Issue #35, opened 2017-04-21
Determine how to build Equella keystore


emcgarrahan:

bjagg:
Emailed Jolse for details on the use of this keystore

ddelblanco:
We have been researching and we propose to use Ant to generate a certificate as explained in:
http://www.asjava.com/ant/how-do-i-create-keystore-with-ant/
for people that don't want or don't have one (for dev environments, for instance), and allow, in some way to use yours if you have one for your institution.  

So, you should be able to configure if you want to use yours or you want a self generated one... 

If using their own certificate, then document where to put your keystore, and ask to configure the password and alias parameters.

If self generated, we will need to ask to configure these parameters (but we can provide some by default for dev purposes, so it would build without need to change this:
  <property name="verisign.key.storepass" value="xxxxx"/>
  <property name="verisign.key.alias" value="yyyyy"/>
  <property name="verisign.key.pass" value="zzzzzz"/>
<param name="CN" value="AsJava.com Group"/>      
<param name="OU" value="Jim"/>     
 <param name="O" value="AsJava.com"/>     
 <param name="C" value="US"/>

doolse:
:+1: The only thing worth also checking is that self signed jars still allow webstart to escape the sandbox.

ddelblanco:
I think I really never been able to make the admin console to be opened by the browser, I needed to launch it with command line based in the jnlp file that is downloaded. In chrome and in firefox. In both of them just downloads the file. Any

doolse:
Yeah that's how you open it, normally jnlp files are associated with javaws
so just opening it rather than going to the command line is enough.

ddelblanco:
I've created this code that allows to select with a build.property if we want to use an existing keystore or we want to create a new one self signed. In the case of the actual keystore, the keypass can be an empty string and it will work. 
https://github.com/equella/Equella/tree/issues/35
Please review it and if you consider good tell me and I will make the PR

ddelblanco:
I've tested the "self-register" keystore. It is true that it doesn't work directly but It is possible to run it if the exception is added in:
${user.home}/.java/deployment/security/exception.sites
or using the java ControlPanel to add the trusted URL (really it does the same writting in that file)

Adding, http://localhost:8080/vanilla/  I can open it. 

So I will update the documentation to explain it. That should happen just for dev environments. Of curse for production people should add a signed keystore.

ddelblanco:
[...]

PR #43, opened 2017-05-03
issue #35 Option to self-generate a keystore when building or use a previously …

ddelblanco:
…created one

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 
 **NOTE: KEYSTORE**
 A keystore with a certificate is required to sign some of the jars or the build will fail.
-In the build.properties.in file you can modify the parameters to configure your own keystore:
+In the build.properties file you can modify the parameters to configure your own keystore:
     
  * The route to the keystore file, by default the root of the user home folder: `tle.signer.certificate=${user.home}/equella.keystore`
  * Set to false to avoid the self-generation of the keystore:  `tle.signer.generateKeystore=false`
--- a/README.md
+++ b/README.md
@@ -8,6 +8,26 @@
 * JDK 8
 * `equella-deps.zip` unzipped into your home directory
 
+**NOTE: KEYSTORE**
+A keystore with a certificate is required to sign some of the jars or the build will fail.
+In the build.properties.in file you can modify the parameters to configure your own keystore. the needed ones are:
+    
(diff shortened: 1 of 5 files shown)


ddelblanco:
To make this work in travis we need to delete the equella.keystore file or... set the build.in properties to the original password.

ddelblanco:
Updated the README.md to explain about the self-generated keystore.

README.md
```diff
@@ -19,17 +19,16 @@ The build artifacts will be produced in the `product` folder.
 
 **NOTE: KEYSTORE**
 A keystore with a certificate is required to sign some of the jars or the build will fail.
-In the build.properties.in file you can modify the parameters to configure your own keystore. the needed ones are:
+In the build.properties.in file you can modify the parameters to configure your own keystore:
```
doolse: build.properties.in shouldn't be modified, it's only for defaults - only build.properties should be modified after: 
`ant generate-build-properties`. However I see the problem with `ant release` is that it always runs the `ant generate-build-properties` - it should probably only do this if `build.properties` doesn't already exist.

ddelblanco: Yes, that was the problem, build.properties is always overwritten by the build.properties.in. I did the check about if the keystore file exists, so I can do the same here checking if the build properties exists.

ddelblanco:
Seems that now it builds with the changes in the travis configuration and only overwrites build.properties if it doesn't exits.

build.xml
```diff
@@ -1,8 +1,16 @@
...
+<project name="EQUELLA-Build" default="build" basedir="." xmlns:if="ant:if"
+    xmlns:unless="ant:unless">
 	<property name="mm" value="6.4" />
 
-	<target name="generate-build-properties">
-		<delete file="build.properties" />
+	<target name="build-properties-exists">
+		<condition property="buildPropExists">     
+	    		<and>
+	    			<available file="build.properties}"/>
```
doolse: There's an extra trailing "}" here, so I don't think this is having any effect. I believe the <copy> task doesn't overwrite if the file exists, so I think just removing the <delete> should be enough

ddelblanco:
Yes, seems that removing the delete command was enough to just overwrite if it doesn't exist. Updated

PR #76, opened 2017-05-22
Allow to configure the no reply email address and remove the hardcode…

ddelblanco:
…d one

### DIFF
--- a/Source/Plugins/Server/com.tle.core.email/src/com/tle/core/email/impl/EmailServiceImpl.java
+++ b/Source/Plugins/Server/com.tle.core.email/src/com/tle/core/email/impl/EmailServiceImpl.java
@@ -176,7 +176,7 @@ public class EmailServiceImpl implements EmailService
 		String smtpUser = systemConfigService.getSmtpUser();
 		String smtpPassword = systemConfigService.getSmtpPassword();
 		mailSettings.setServer(smtp);
-		mailSettings.setSender("donotreply@equella.com");
+		mailSettings.setSender(systemConfigService.getNoReplySender());
 		if( !Check.isEmpty(smtpUser) )
 		{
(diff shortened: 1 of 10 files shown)


Source/Plugins/Server/com.tle.core.email/src/com/tle/core/email/impl/EmailServiceImpl.java
```diff
@@ -176,7 +176,7 @@ private MailSettings getSystemMailSettings()
 		String smtpUser = systemConfigService.getSmtpUser();
 		String smtpPassword = systemConfigService.getSmtpPassword();
 		mailSettings.setServer(smtp);
-		mailSettings.setSender("donotreply@equella.com");
+		mailSettings.setSender(systemConfigService.getNoReplySender());
```
doolse: Existing installs aren't going to have this setting, so you need to use a safe default if it's null

Source/Plugins/System/com.tle.core.system/src/com/tle/core/system/migration/InitialMigration.java
```diff
@@ -92,6 +92,7 @@ protected void executeDataMigration(HibernateMigrationHelper helper, MigrationRe
 			copyProperty(session, "admin.password");
 			copyProperty(session, "admin.emails");
 			copyProperty(session, "smtpserver");
+            copyProperty(session, "noreplysender");
```
doolse: This setting won't exist on older systems, probably won't matter if copyProperty handles nulls but would be safer to remove this line

ddelblanco: I'm almost sure that the code will manage a null. The database allows nulls in the field and so the query will return null and the value will be saved as null.

Issue #78, opened 2017-05-25
METS from edu.harvard.hul has undefined license


bjagg:
To resolve this, we should investigate if this functionality is still needed (METS support). If not, remove due to confusion on license. (One page lists this dependency as LGPL.) If used, we should replace with one of the following:

An alternative is https://sourceforge.net/projects/mets-api/ which is under Apache2 license, and http://sourceforge.net/projects/seventrain which is under BSD.

cbeach47:
Working to pull this code into it's own repo.  I checked with @doolse who didn't conceptually see any major issues, and I've sent a follow up note to a few folks to confirm user stories.

cbeach47:
Code changes and smoke tests complete.  Test binaries with and without METS are built.  Requested some advice on a more complete test suite for METS.  Plan to run the tests using the binaries tomorrow.

We will have to put the hul mets external dep into Maven local for now, and then publish ourselves.

@drewwills - please create an Equella-METS repo similar to #55 .

cbeach47:
METS code published as an 'optional functionality' here:  https://github.com/equella/Equella-METS

doolse:
I've used the way back machine to find the original license - it's LGPL, which means we can publish it yeah?

https://web.archive.org/web/20100413195423/http://hul.harvard.edu:80/mets/download.html

cbeach47:
Confirmed with @drewwills and @bjagg yesterday - LGPL is valid to publish the dependency for Equella in the core base.

cbeach47:
- Removed the 'optional functionality' and 'migration testing' documentation.
- Closed the pull request #91 and deleted the issues/78 branch
- Added this dependency into #79
@drewwills - please remove the Equella-METS repo.  Then we can close this issue.

drewwills:
- [ ] Done

PR #91, opened 2017-05-31
#78 Removed METS code

cbeach47:
Hi @doolse , please merge the changes from issue 78 into master.  The removed code now lives in Equella-METS, and the documentation for integration (including the couple of possible issues clients can run into when running Equella without METS) is linked in that repo's README file.


cbeach47:
As per #78, METS functionality is staying as part of the core Equella code base.

PR #99, opened 2017-06-14
Use Gitignore io template to handle general source ignores

ChristianMurphy:
adds eclipse sbt plugin to automatically generate IDE project configuration with `sbt eclipse` or `sbt gen-idea`


ChristianMurphy:
@doolse @ddelblanco @cbeach47 is there any interest in this PR? Or should this be closed out?

doolse:
I'd prefer the .gitignore to be a small as possible, to me you really need a good reason to add artifacts outside of the "target" directory. I don't think the eclipse and idea plugins should be in a projects sbt, it's more of a user level (~/.sbt) config.

Issue #285, opened 2017-08-07
"resource not visible" warnings in logs


abidingotter:
Noticed by Chris Beach and then Carl Hoffman

There is no functional impact apart from startup time taking a fraction of a second longer (unless you spot missing strings in the app) but it doesn't look good.

You get about 20 or so variations on this message (different plugin IDs)

resource not visible, name=lang/i18n.properties, URL=jar:file:/server/temp/.jpf-shadow/com.tle.core.scripting@1.0.0.jar!/resources/lang/i18n.properties, this={PluginClassLoader: uid=1917498463; {PluginDescriptor: uid=com.tle.core.scripting@1.0.0}}, requestor={PluginClassLoader: uid=1822511256; {PluginDescriptor: uid=com.tle.web.api.search@1.0.0}}

The com.tle.web.api.search plugin needs to include an i18n.properties file, or remove it from the plugin xml.

PR #286, opened 2017-08-08
#285 make resources folder visible

abidingotter:


abidingotter:
Jolse says this section of the XML is auto-generated now, so it wasn't an issue in 6.5, only 6.4

PR #321, opened 2017-09-13
Workflow UI tweaks

doolse:
The new UI is fully functional now. We'll probably want to tweak the look a bit but that can be done in separate issues.


abidingotter:
*thumbsup*

Issue #384, opened 2017-11-02
SRU and SRW endpoints are busted


mrblippy:
I get the following error when accessing the srw or sru endpoints in equella. eg:
http://charliedesk:8092/vanillaext/srw/?operation=searchRetrieve&query=test
seems to happen in 6.4 as well

[...]

PR #386, opened 2017-11-07
#384 static inject again

abidingotter:
@doolse , I know you removed the static injection, but I don't see any other way to get it to work?


doolse:
Make sure you run the `headerCreate` sbt task to create the license headers for you.

```
java.lang.RuntimeException: There are files without headers!
  /home/travis/build/equella/Equella/Source/Plugins/Server/com.tle.web.endpoint.srw/src/com/tle/web/endpoint/srw/guice/SRWEndpointModule.java
```

doolse:
When I last looked at the SRW/SRU, it didn't actually work properly on 6.4 either.. You can access the base url "/sru" (or /srw i forget) but actually doing a search didn't work. Can you confirm that it actually works?

Issue #407, opened 2017-12-14
More detailed Report and Index logging


cbeach47:
Add more detailed logging to Equella for reports and indexing:

Highlights:
- Log @ INFO what report is being run
- Log @ WARN what attachments fail to be indexed
- Log @ TRACE what attachments are indexed

PR #410, opened 2018-01-08
Issue 407:  Added logging for indexing and reporting

cbeach47:
Issue #407 
Provides more detailed logging around attachment indexing and what reports are run.

Request to be merged into master.  Not sure if it's appropriate to be cherry-picked into 65stable...

### DIFF
--- a/Source/Plugins/Generic/com.tle.common.reporting/src/com/tle/beans/entity/report/Report.java
+++ b/Source/Plugins/Generic/com.tle.common.reporting/src/com/tle/beans/entity/report/Report.java
@@ -58,4 +58,8 @@ public class Report extends BaseEntity
 	{
 		this.hideReport = hideReport;
 	}
+	
+	public String getReportSignature() {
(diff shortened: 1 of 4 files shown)


Source/Plugins/Generic/com.tle.common.reporting/src/com/tle/beans/entity/report/Report.java
```diff
@@ -58,4 +58,8 @@ public void setHideReport(Boolean hideReport)
 	{
 		this.hideReport = hideReport;
 	}
+	
+	public String getReportSignature() {
+		return "reportName=[" + this.getName() + "], reportFilename=[" + filename + "]";
```
doolse: Best to use `LangUtils.getString(this.getName())` instead as you'll get prettier output than toString(), which puts some ugly "LB[]" around it

cbeach47:
Thanks for the tip @doolse .  Change committed.  I would find it helpful to have this change cherry-picked into 6.5 Stable.

PR #418, opened 2018-01-17
Issue 229:  Use a wildcard search for the ItemMetadataFilter

cbeach47:
Resolution to https://github.com/equella/Equella/issues/229 : Resources will not be re-indexed when editing Resource Metadata ACLs in the Collection Definition Editor in the Administration Console

I still need to do some final testing on this, but looks to work on initial tests with Postgres so I wanted to get the basic idea of the fix out here for folks to review.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/core/freetext/reindex/ItemMetadataFilter.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/core/freetext/reindex/ItemMetadataFilter.java
@@ -16,11 +16,17 @@
 
 package com.tle.core.freetext.reindex;
 
+import java.util.Arrays;
+
+import org.apache.log4j.Logger;
+
 /**
  * @author Nicholas Read
  */
 public class ItemMetadataFilter extends ReindexFilter
 {
+	private static final Logger LOGGER = Logger.getLogger(ItemMetadataFilter.class);
+
 	private static final long serialVersionUID = 1L;
 
 	private static final String[] NAMES = {"targetId"};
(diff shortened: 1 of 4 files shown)


cbeach47:
Confirmed via an upgrade test on Postgres that the fix now flags the appropriate items for re-indexing.  It's sometimes hard to catch when the RefreshCollectionItemDataFilter.

Source/Plugins/Platform/com.tle.core.hibernate/src/com/tle/core/hibernate/type/HibernateCsvType.java
```diff
@@ -135,43 +138,52 @@ private String escapeString(String str)
 	@Override
 	public void nullSafeSet(PreparedStatement st, Object value, int index) throws SQLException
 	{
+		LOGGER.trace("Checking nullSafeSet for [" + value + "] of type [" + value.getClass().getName() + "] at index [" + index + "].");
```
abidingotter: Presumably "value" could be null, since there is a null check further down (col != null), meaning this line could NPE

doolse: I'd just remove that trace line all together, that's the kind of thing only needed during debugging, in which case you can use breakpoints instead

cbeach47: Thanks for the catch @abidingotter .  

@doolse - I find it useful for applications to have the ability to be more verbose and not rely as much on a debugger, especially when weird issues occur in Production and attaching a remote debugger is not practical.

PR #441, opened 2018-02-21
ci(travis): test open and oracle jdk, test multiple scala versions

ChristianMurphy:
Adding Open JDK and multiple Scala versions to the test matrix allows for gives assurances in terms of language support.
Without adding too much testing burden, Travis CI does the heavy lifting, tests run in parallel 5 at a time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,13 +6,19 @@ cache:
   - "$HOME/.ivy2"
 jdk:
 - oraclejdk8
+- openjdk8
+scala:
+- 2.9.3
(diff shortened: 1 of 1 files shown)


doolse:
EQUELLA is an app, not a library so we don't need to support different versions of Scala, just the one (currently 2.12.4)
Currently we also depend on the Oracle JDK because of the webstart library which is bundled with it. If you can find a replacement on maven that works with OpenJDK, we can relax that requirement.

Issue #492, opened 2018-05-15
Allow a cap on how long the Background Indexer will run


cbeach47:
Certain documents can index for a long time (minutes) and create stability issues for Equella.  This enhancement should allow users via the Settings page to cap the BackgroundIndexer at the item and/or attachment level (in # of seconds) before logging a useful message and canceling the index of the item/attachment.

cbeach47:
There is a check for long running threads [here](https://github.com/equella/Equella/blob/master/Source/Plugins/Core/com.equella.core/src/com/tle/freetext/BackgroundIndexerImpl.java#L394).
- Need a way for a user to configure the `INTERRUPT_MILLIS` - via optional-config would be the cleanest, otherwise we'd have to synchronize access to the config.
- The indexer should be able to terminate long running threads.

I'm looking to implement the following new options via optional-config:
```
# In Seconds
indexer.longRunningThread = 120
# Default is false (original behavior)
indexer.terminateLongRunningThreads = true
```

Still considering what to do with the item currently in the long running thread to be terminated.  If nothing, I believe it'll be picked up in the next batch of items to index, and it'll keep cycling between partially indexed and being killed.

cbeach47:
Ignore the 'new configs' comment above - I initially explored a way to interrupt the indexer thread, but the Tika AutoParser doesn't acknowledge interrupts.  I ended up creating a 'capped' parse handler for the Tika parser so it won't run indefinitely.  Configs to tune this enhancement can be added in `com.tle.core.freetext` > `optional.properties`:

```
# Tuning config for attachment indexing.  Currently implemented for PDFs.
# The attachment indexing duration will go over this amount (in milliseconds), 
# but will be checked each time the parser returns data to the controller, 
# so it will be close
textExtracter.parseDurationCap=30000

# Tuning config for attachment indexing.  Specifies how many times to ignore 
# checking that the duration has not gone over it's cap before checking.  Slight 
# performance boost for higher values
textExtracter.durationCheckFrequency=15
```

PR #493, opened 2018-05-19
#492 Parse duration cap implemented for PDFs

cbeach47:
Initially explored a way to interrupt the indexer thread, but the Tika AutoParser doesn't acknowledge interrupts.  I ended up creating a 'capped' parse handler for the Tika parser so it won't run indefinitely.  

New `com.tle.core.freetext` > `optional.properties`:
```
# Tuning config for attachment indexing.  Currently implemented for PDFs.
# The attachment indexing duration will go over this amount (in milliseconds), 
# but will be checked each time the parser returns data to the controller, 
# so it will be close
textExtracter.parseDurationCap=30000

# Tuning config for attachment indexing.  Specifies how many times to ignore 
# checking that the duration has not gone over it's cap before checking.  Slight 
# performance boost for higher values
textExtracter.durationCheckFrequency=15
```

### DIFF
--- a/Source/Plugins/Core/com.equella.base/src/com/tle/core/TextExtracterExtension.java
+++ b/Source/Plugins/Core/com.equella.base/src/com/tle/core/TextExtracterExtension.java
@@ -26,7 +26,7 @@ import com.tle.beans.mime.MimeEntry;
  */
 public interface TextExtracterExtension
 {
-	void extractText(String mimeType, InputStream input, StringBuilder outputText, int maxSize, long parseDurationCap, int durationCheckFrequency) throws IOException;
+	void extractText(String mimeType, InputStream input, StringBuilder outputText, int maxSize, long parseDurationCap) throws IOException;
 
 	void setEnabledForMimeEntry(MimeEntry mimeType, boolean enabled);
 
--- a/Source/Plugins/Core/com.equella.base/src/com/tle/core/TextExtracterExtension.java
+++ b/Source/Plugins/Core/com.equella.base/src/com/tle/core/TextExtracterExtension.java
@@ -26,7 +26,7 @@ import com.tle.beans.mime.MimeEntry;
  */
 public interface TextExtracterExtension
 {
-	void extractText(String mimeType, InputStream input, StringBuilder outputText, int maxSize) throws IOException;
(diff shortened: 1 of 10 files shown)


cbeach47:
@doolse  - not sure why _codeclimate_ is complaining since I changed the interface `TextExtracterExtension`.   Unless you know of another place _codeclimate_ is keying off of for the method signatures.

Any concerns with merging this pull request?  It'd be nice to get it in before 6.6 is branched.

Source/Plugins/Core/com.equella.base/src/com/tle/core/TextExtracterExtension.java
```diff
@@ -26,7 +26,7 @@
  */
 public interface TextExtracterExtension
 {
-	void extractText(String mimeType, InputStream input, StringBuilder outputText, int maxSize) throws IOException;
+	void extractText(String mimeType, InputStream input, StringBuilder outputText, int maxSize, long parseDurationCap, int durationCheckFrequency) throws IOException;
```
doolse: I don't think "durationCheckFrequency" should be a parameter, it's very specific to the technique your using to check the timeout in the CappedBodyContentHolder.

Source/Plugins/Core/com.equella.core/src/com/tle/core/freetext/extracter/handler/CappedBodyContentHandler.java
```diff
@@ -0,0 +1,80 @@
...
+
+package com.tle.core.freetext.extracter.handler;
+
+import org.apache.tika.sax.BodyContentHandler;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.xml.sax.ContentHandler;
+import org.xml.sax.SAXException;
+
+public class CappedBodyContentHandler extends BodyContentHandler {
```
doolse: Did you try any other approaches to doing a timeout, such as java.util.concurrent.Future? I'll merge with this implementation anyway as I don't think it'll cause too much of a slow down.

cbeach47: Yes, I tried to interrupt the parser, but from the documentation, the AutoParser doesn't check for those exceptions.

cbeach47:
@doolse - thanks for the review.  I've removed the duration check frequency parameter and hard-coded it to 200.

Issue #476, opened 2018-03-28
Script API should allow an XML node to access it's parent and children


cbeach47:
Add methods in the XML Script API to provide access to an XmlScriptType's 'parent' XML node, and list out it's child XML nodes.

cbeach47:
Merged into master.  Closing.

PR #545, opened 2018-08-08
#476 XmlScriptType getParent and getChildren logic in place

cbeach47:
For issue #476 .

Logic is in place to return the parent and children of a given `XmlScriptType`.  I added in a `XmlScriptType.getName()` as well.  The scripting below outputs the expected values.  I chose to go with the call format of `node.getParent()` instead of `node.getParent(node)` or `xml.getParent(node)`.  Both ways functionally worked, so I opted for the simplified way. 

```
xml.set("lom/general/title", "a title");
xml.set("lom/general/description", "a desc");
var titleXml = xml.getSubtree("lom/general/title");
var descXml = xml.getSubtree("lom/general/description");

displayChildren(xml);
displayLineage(xml.getSubtree("lom/general/title"));

var temp = utils.newXmlDocument();
temp.set("/parent1/child1", "p1c1");
temp.set("/parent1/child2", "p1c2");
temp.set("/parent1/child3", "p1c3");
temp.set("/parent2/child1", "p2c1");
temp.set("/parent2/child2", "p2c2");
displayChildren(temp);
displayLineage(temp.getSubtree("/parent2/child1"));

function displayChildren(n) {
	var children = n.getChildren();
	if(children != null) {
		if(children.size() == 0) {
			logger.log("No children found for: " + displayName(n));
		} else {
			logger.log(children.size() + " children found for: " + displayName(n));
			for(var i = 0; i < children.size(); i++) {
				logger.log("Child: " + children.get(i));
				displayChildren(children.get(i));
			}
		}
	} else {
		logger.log("Unable to find children for: " + displayName(n));
	}
}

function displayLineage(n) {
	var parent = n.getParent();
	logger.log("Parent of [" + displayName(n) + "] is [" + parent + "]");
	if(parent != null) {
		displayLineage(parent);
	}
}

function displayName(n) {
	if(n == null) return "NULL";
	else return n.getName();
}
```

### DIFF
--- a/Platform/Plugins/com.tle.platform.common/src/com/dytech/devlib/PropBagEx.java
+++ b/Platform/Plugins/com.tle.platform.common/src/com/dytech/devlib/PropBagEx.java
@@ -1076,10 +1076,43 @@ public class PropBagEx implements Serializable
 	public PropBagEx getParent()
 	{
 		ensureRoot();
-		final Element root = (Element) m_elRoot.getParentNode();
+		final Node rootNode = m_elRoot.getParentNode();
+		if(rootNode instanceof Element) {
+			final Element root = (Element) m_elRoot.getParentNode();
+			if( root != null )
+			{
+				return new PropBagEx(root, true);
+			}
+			else
+			{
+				return null;
+			}
+		} else {
+			return null;
+		}
+	}
+
+	/**
+	 * Creates a List of new PropBags rooted at each of the children of this PropBag, sharing the
+	 * same DOM nodes as the creator.
+	 *
+	 * @return List<PropBagEx> A list of children or null if the node does not exist.
+	 */
+	public List<PropBagEx> getChildren()
+	{
+		ensureRoot();
+		final Element root = (Element) m_elRoot;
 		if( root != null )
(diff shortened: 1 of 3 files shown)


Platform/Plugins/com.tle.platform.common/src/com/dytech/devlib/PropBagEx.java
```diff
@@ -1076,10 +1076,43 @@ public AllNodesWithNameIterator iterateAllNodesWithName(String name)
 	public PropBagEx getParent()
 	{
 		ensureRoot();
-		final Element root = (Element) m_elRoot.getParentNode();
+		final Node rootNode = m_elRoot.getParentNode();
+		if(rootNode instanceof Element) {
```
doolse: Did you change this because the previous getParent() would get classcastexception ?

cbeach47: Yes, as the `displayLineage` function walked up it's parental line, it would eventually hit a classcastexception instead of returning null.

Issue #475, opened 2018-03-28
Script API should allow a control to access it's target XML node


cbeach47:
Add a method into the Control Script API to allow a control to be able to access the control's first target XML node.

cbeach47:
The way the wizard controls integrate with the XML 'model' does not lend itself to cleanly returning the 'backing node'.  I am considering instead to expose the scripting ability for a control to find it's parent control (which looks to only work for controls that are in a group/repeater, and for a control to know what its index is (if it's part of a group / repeater).  Then a script can access the XmlScriptObject via a simple `xml.getSubtree` with `/my/metadata/groups/group[2]/` and inspect the values as needed.  

@doolse / @abidingotter - Does this sound like an appropriate path to take?

abidingotter:
@cbeach47 I could see other legitimate reasons to get a control's index, so I don't have an issue with that.

cbeach47:
Used the following for testing the flows:

getParent:
```
logDetails(ctrl, "Descriptive name of HTMLControl");
return true; 

function logDetails(controlObj, name) {
	if(controlObj == null) {
		logger.log(name + " - No control object available");
	} else {
		var parent = controlObj.getParent();
		var sig = name + "value=[" + getValue(controlObj) + "], formId=[" + controlObj.getFormId() + "], ";
		if(parent == null) {
			logger.log(sig + "parent is null.");
		} else {
			logger.log(sig + "index=[" + parent.getIndex(controlObj) + "], parent-formId=[" + parent.getFormId() + "]");
		}	
	}
}

function getValue(controlObj) {
	try {
		return controlObj.getValue();
	} catch (e) {
		return "NA";
	}
}
```

getIndex:
```
var index = ctrl.getParent().getParent().getIndex(ctrl.getParent());
var xpath = "lom/technical["+(index)+"]/format"; // Setup a list box with 'V1', 'V2', 'V3'.
var value = xml.get(xpath);
logger.log("Nested edit box [" + ctrl.getValue() + "].  Parental xpath: [" + xpath + "].  Parental value: [" + value + "].  Should I be showing? " + (value == "V1"));
return (value == "V1");
```

cbeach47:
Enhancement merged.  Closing.

cbeach47:
Note - the 'getIndex' test flow should be:

```
// Setup a repeater (lom/technical) and have two child widgets:  
//    a list box with 'V1', 'V2', 'V3'
[...]

PR #554, opened 2018-08-16
#475 ControlScriptObject getParent and getIndex(child) implemented

cbeach47:
History in #475.  Allows scripting to be 'aware' of which repeater node the control is in.


cbeach47:
The travis-ci/pr is complaining about a missing header in a scala build file.  This scala file was not part of my changes, so we should be safe merging it.

Issue #555, opened 2018-08-16
Enable SBT building javadoc for the Scripting API


cbeach47:
With additions to the Scripting API ( such as #475 / #476 ), SBT should build the Scripting javadoc as an artifact.

@doolse - Have you done any effort on this?  Otherwise, I'll go ahead and take a crack at it.

cbeach47:
@doolse - thanks for picking this up - confirmed in master this builds the javadoc.  Closing.

PR #597, opened 2018-11-07
#555 Create a scripting api javadoc zip

doolse:
FYI @edalex-ian - EQUELLA used to have a downloads page as part of the server admin which contained documentation downloads but it was removed when it went open source

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,10 +1,10 @@
+
 import java.util.Properties
 
-import complete.DefaultParsers._
 import com.typesafe.sbt.license.LicenseReport
-
-import scala.collection.JavaConverters._
 import sbt.io.Using
+import Path.rebase
+import scala.collection.JavaConverters._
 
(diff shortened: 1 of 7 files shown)


build.sbt
```diff
@@ -145,3 +148,32 @@ mergeJPF := {
     PluginRefactor.mergePlugins(allPluginDirs, basePlugin, newPlugin, args.tail, adminConsole = adminConsole)
   }
 }
+
+writeScriptingJavadoc := {
+  val javadocDir = (doc in Compile).value 
+  val ver = version.value
+  val outZip = target.value / s"scriptingapi-javadoc-$ver.zip"
+  IO.zip((javadocDir ** "*").pair(rebase(javadocDir, "")), outZip)
+  sLog.value.info(s"Writing ${outZip.absolutePath}")
```
edalex-ian: Should that be 'Written' rather than 'Writing'?

Issue #604, opened 2018-11-19
New favicon.ico


edalex-ian:
Current favicon.ico is the old Pearson one, needs to be replaced with an open source specific one.

@mrblippy is currently looking at this as time permits, so assigning accordingly.

SammyIsConfused:
Confirmed working in master.

SammyIsConfused:
Gonna check the admin console too, now I think about it.

SammyIsConfused:
![screenshot from 2018-12-13 12-06-49](https://user-images.githubusercontent.com/24543345/49908674-cadd8280-fecf-11e8-95a9-e5c37113e069.png)
Found one! The admin console window's icon is the old one.

SammyIsConfused:
Confirmed fixed on betadev.

PR #611, opened 2018-11-21
#604 openEquella favicons

mrblippy:


mrblippy:
> Do you also need to replace `./Source/Plugins/Core/com.equella.core/resources/web/favicon.ico`?

That doesn't look like it get used anywhere, but it doesn't hurt to update it anyway

edalex-ian:
hmmm, would it be safe enough to simply remove then to avoid future confusion?

PR #625, opened 2018-11-27
Equella New UI Theme Customiser

SammyIsConfused:
Material UI colour scheme editor and custom logo selector for equella's New UI. Menu is accessible via Settings -> New UI Settings.

Logo uploader tested with invalid file (text file renamed to .png file extension).


Source/Plugins/Core/com.equella.core/js/tsrc/theme/ThemePage.tsx
```diff
@@ -0,0 +1,383 @@
...
+import Divider from "@material-ui/core/Divider/Divider";
+import SnackBar from "@material-ui/core/Snackbar";
+import IconButton from "@material-ui/core/IconButton/IconButton";
+import CloseIcon from "@material-ui/icons/Close";
+import Dialog from "@material-ui/core/Dialog/Dialog";
+import DialogTitle from "@material-ui/core/DialogTitle/DialogTitle";
+import DialogContent from "@material-ui/core/DialogContent/DialogContent";
+import DialogContentText from "@material-ui/core/DialogContentText/DialogContentText";
+import DialogActions from "@material-ui/core/DialogActions/DialogActions";
+import Slide from "@material-ui/core/Slide/Slide";
```
ChristianMurphy: :art: using a single destrutured import for the material ui components would make this easier to follow.
A tree shaker can take care optimizing which files are needed.

```js
import * as React from "react";
import axios from "axios";

import {
  Button,
  Card,
  CardActions,
  CardContent,
  CloseIcon,
  ColorPickerComponent,
  Dialog,
  DialogActions,
  DialogContent,
  DialogContentText,
  DialogTitle,
  Divider,
  FormControl,
  Grid,
  IconButton,
  Slide,
  SnackBar,
  Typography,
  WithStyles,
  withStyles 
} from "@material-ui/core";

import ColorPickerComponent from "./ColorPickerComponent";
import { Bridge } from "../api/bridge";
import { Config } from "../config";
```

Source/Plugins/Core/com.equella.core/js/tsrc/theme/ThemePage.tsx
```diff
@@ -0,0 +1,383 @@
...
+        logoToUpload: file,
+        fileName: file.name
+      });
+    };
+  };
+
+  submitTheme = () => {
+    //TODO: map this in a more elegant way
+    axios.put(`${Config.baseUrl}api/themeresource/update/`,
+      "{\"primaryColor\":\"" + this.state.primary + "\"," +
```
ChristianMurphy: Axios automatically serializes Objects to JSON.

```js
axios.put(
  `${Config.baseUrl}api/themeresource/update/`,
  {
    primaryColor: this.state.primary,
    secondaryColor: this.state.secondary,
    // etc
  }
)
```

Source/Plugins/Core/com.equella.core/js/tsrc/theme/ThemePage.tsx
```diff
@@ -0,0 +1,383 @@
...
+}
+
+function transition(props: any) {
+  return <Slide direction="up" {...props} />;
+}
+
+class ThemePage extends React.Component<ThemePageProps & WithStyles<typeof styles>> {
+
+  state = {
+    primary: themeSettings["primaryColor"],
```
ChristianMurphy: :art: property can be accessed with dot notation `themeSettings.primaryColor`
same applies for properties below.

Source/Plugins/Core/com.equella.core/js/tsrc/theme/ThemePage.tsx
```diff
@@ -0,0 +1,383 @@
...
+import CloseIcon from "@material-ui/icons/Close";
+import Dialog from "@material-ui/core/Dialog/Dialog";
+import DialogTitle from "@material-ui/core/DialogTitle/DialogTitle";
+import DialogContent from "@material-ui/core/DialogContent/DialogContent";
+import DialogContentText from "@material-ui/core/DialogContentText/DialogContentText";
+import DialogActions from "@material-ui/core/DialogActions/DialogActions";
+import Slide from "@material-ui/core/Slide/Slide";
+
+declare var themeSettings: any;
+declare var isCustomLogo: boolean;
```
ChristianMurphy: `themeSettings` and `isCustomLogo` are not and should not be modified in this code, they are `const`s.

Source/Plugins/Core/com.equella.core/js/tsrc/theme/ThemePage.tsx
```diff
@@ -0,0 +1,383 @@
...
+import IconButton from "@material-ui/core/IconButton/IconButton";
+import CloseIcon from "@material-ui/icons/Close";
+import Dialog from "@material-ui/core/Dialog/Dialog";
+import DialogTitle from "@material-ui/core/DialogTitle/DialogTitle";
+import DialogContent from "@material-ui/core/DialogContent/DialogContent";
+import DialogContentText from "@material-ui/core/DialogContentText/DialogContentText";
+import DialogActions from "@material-ui/core/DialogActions/DialogActions";
+import Slide from "@material-ui/core/Slide/Slide";
+
+declare var themeSettings: any;
```
ChristianMurphy: The structure of `themeSettings` is known.
```ts
interface IThemeSettings {
    primaryColor: string;
    secondaryColor: string;
    // etc
}

declare const themeSettings: IThemeSettings;
```

Source/Plugins/Core/com.equella.core/js/tsrc/theme/ThemePage.tsx
```diff
@@ -0,0 +1,383 @@
...
+
+  handleMenuIconChange = (color: string) => {
+    this.setState({menuIcon: color});
+  };
+
+  handleTextChange = (color: string) => {
+    this.setState({text: color});
+  };
+
+  handleImageChange = (e: any) => {
```
ChristianMurphy: Typescript provides DOM typings, including `Event` https://github.com/Microsoft/TypeScript/blob/aa3734c14834afb454c49bc956489980d990a765/src/lib/dom.generated.d.ts#L4943-L5001

Source/Plugins/Core/com.equella.core/src/com/tle/web/api/newuitheme/impl/NewUIThemeResourceImpl.java
```diff
@@ -0,0 +1,193 @@
...
+	@Inject
+	private ConfigurationService configurationService;
+	@Inject
+	private TLEAclManager tleAclManager;
+	@Inject
+	FileSystemService fsService;
+
+	private CustomisationFile customisationFile;
+	private NewUITheme theme;
+	private ObjectMapper objectMapper = new ObjectMapper();
```
edalex-ian: I wonder if there's a global object mapper in openEQUELLA that you should inject. I've not checked, but often a larger system ends up with one with customisations specific to its context. Might just want to check.

Source/Plugins/Core/com.equella.core/src/com/tle/web/api/newuitheme/impl/NewUIThemeResourceImpl.java
```diff
@@ -0,0 +1,193 @@
...
+
+	@Inject
+	private ConfigurationService configurationService;
+	@Inject
+	private TLEAclManager tleAclManager;
+	@Inject
+	FileSystemService fsService;
+
+	private CustomisationFile customisationFile;
+	private NewUITheme theme;
```
edalex-ian: Normally in a singleton you wouldn't store state (or if you do, you'll want to make it thread safe), so I'm wondering if you really need `customisationFile` and `theme`.

Source/Plugins/Core/com.equella.core/src/com/tle/web/api/newuitheme/impl/NewUIThemeResourceImpl.java
```diff
@@ -0,0 +1,193 @@
...
+		}
+		return themeToString;
+	}
+
+	@GET
+	@Path("theme.js")
+	@Produces("application/javascript")
+	public Response retrieveThemeInfo() {
+		//set default theme if none exists in database
+		if (configurationService.getProperty(THEME_KEY) == null) {
```
edalex-ian: Could be good to save the result from `getProperty()` so you don't have to do the same possible data hit below.

Source/Plugins/Core/com.equella.core/js/src/MainUI/Template.js
```diff
@@ -4,6 +4,9 @@ exports.renderData = function()
     return typeof renderData != "undefined" ? renderData : {newUI:false};
 }();
 
+exports.logoPath = function(){
+  return isCustomLogo ? "api/themeresource/newLogo.png" : "p/r/6.7.r88/com.equella.core/images/new-equella-logo.png"
```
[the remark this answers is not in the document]
SammyIsConfused: Yep! Replaced with renderData.baseResources.

Source/Plugins/Core/com.equella.core/js/tsrc/theme/ThemePage.tsx
```diff
@@ -0,0 +1,459 @@
...
+
+  handleUndoButton = () => {
+    this.setState({
+      primary: themeSettings.primaryColor,
+      secondary: themeSettings.secondaryColor,
+      background: themeSettings.backgroundColor,
+      menu: themeSettings.menuItemColor,
+      menuText: themeSettings.menuItemTextColor,
+      menuIcon: themeSettings.menuItemIconColor,
+      text: themeSettings.menuTextColor
```
doolse: This looks the same as the code in the `state` initializer, best to refactor it into a method.

Source/Plugins/Core/com.equella.core/js/tsrc/theme/ThemePage.tsx
```diff
@@ -0,0 +1,459 @@
...
+        secondaryColor: this.state.secondary,
+        backgroundColor: this.state.background,
+        menuItemColor: this.state.menu,
+        menuItemTextColor: this.state.menuText,
+        menuItemIconColor: this.state.menuIcon,
+        menuTextColor: this.state.text,
+        fontSize: 14
+      })
+      .then(() => {
+          window.location.reload();
```
doolse: I'd move the `window.location.reload()` into a method as it's conceivable that reloading might be done in a different way at some point.

Source/Plugins/Core/com.equella.core/js/tsrc/theme/ThemePage.tsx
```diff
@@ -0,0 +1,412 @@
...
+
+  handleInvalidFileErrorClose = () => {
+    this.setState({invalidFileError: false});
+  };
+
+  handleNoFileErrorClose = () => {
+    this.setState({noFileError: false});
+  };
+
+  render() {
```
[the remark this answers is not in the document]
doolse: So for Errors and Notifications there are already components/properties that can be re-used. I will create a document about how to use them.

Source/Plugins/Core/com.equella.core/src/com/tle/web/api/newuitheme/NewUIThemeResource.java
```diff
@@ -0,0 +1,67 @@
...
+	@Path("newLogo.png")
+	@ApiParam("Grabs the current institution's logo from the filestore.")
+	@Produces("image/png")
+	Response retrieveLogo() throws IOException;
+
+	@GET
+	@Path("customlogo.js")
+	@ApiParam("Checks whether or not this institution is using a custom logo.")
+	@Produces("application/javascript")
+	Response customLogoExists();
```
doolse: I would argue that this endpoint doesn't need to exist and the path for the logo should just go in the `NewUITheme` data itself.

Source/Plugins/Core/com.equella.core/src/com/tle/web/api/newuitheme/NewUIThemeResource.java
```diff
@@ -0,0 +1,66 @@
...
+
+	@PUT
+	@Path("update")
+	@ApiParam("Changes the theme settings of the current institution with a JSON string.")
+	Response updateThemeInfo(String themeString);
```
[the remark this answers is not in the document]
doolse: Agreed this should just use the `NewUITheme` as it's parameter type.

Source/Plugins/Core/com.equella.core/src/com/tle/web/api/newuitheme/NewUIThemeResource.java
```diff
@@ -0,0 +1,67 @@
...
+import javax.ws.rs.*;
+import javax.ws.rs.core.Response;
+import java.io.File;
+import java.io.IOException;
+
+/**
+ * @author Samantha Fisher
+ */
+
+@Path("themeresource/")
```
doolse: I'd get rid of `resource` from the path. Resource is really just JAX-RS terminology

Source/Plugins/Core/com.equella.core/src/com/tle/web/api/newuitheme/NewUIThemeResource.java
```diff
@@ -0,0 +1,67 @@
...
+
+	@PUT
+	@Path("updatelogo")
+	@ApiParam("Takes an image file, resizes it to become a logo and saves it to the filestore.")
+	Response updateLogo(File logo) throws IOException;
+
+	@DELETE
+	@Path("resetlogo")
+	@ApiParam("Resets the institution's logo to the default openEQUELLA logo.")
+	Response resetLogo();
```
doolse: The path for uploading and resetting the log should be the same, the HTTP verb decides which action to take. Probably just `logo` for the path would be fine.

Source/Plugins/Core/com.equella.core/src/com/tle/web/api/newuitheme/impl/NewUITheme.java
```diff
@@ -0,0 +1,105 @@
...
+public class NewUITheme {
+	private String primaryColor = "#2196f3";
+	private String secondaryColor = "#ff9800";
+	private String backgroundColor = "#fafafa";
+	private String menuItemColor = "#ffffff";
+	private String menuItemTextColor = "#000000";
+	private String menuItemIconColor = "#000000";
+	private String menuTextColor = "#000000";
+	private int fontSize = 14;
+	@Inject
```
doolse: Using @Inject isn't going to work unless the object is created by the dependency injection framework. I would argue that the serialisation shouldn't be part of the `NewUITheme` class itself but the layer which interacts with the DB properties be responsible.

Source/Plugins/Core/com.equella.core/src/com/tle/web/api/newuitheme/impl/NewUIThemeResourceImpl.java
```diff
@@ -0,0 +1,158 @@
...
+	private static final String LOGO_FILENAME = "newLogo.png";
+	private static final String PERMISSION_KEY = "EDIT_SYSTEM_SETTINGS";
+
+	private void setTheme(String themeString) {
+		configurationService.setProperty(THEME_KEY, themeString);
+	}
+
+	private void setTheme(NewUITheme theme) throws JsonProcessingException {
+		setTheme(theme.themeToString(theme));
+	}
```
doolse: I would create a `ThemeService` class which had getTheme/setTheme methods which hid the serialization/deserialization from the client. 
Also it could have the code for getting/setting the logo, the idea being that this class shouldn't need to know about ConfigrationService, FilesystemService or CustomisationFile.

Source/Plugins/Core/com.equella.core/src/com/tle/web/api/newuitheme/impl/NewUIThemeResourceImpl.java
```diff
@@ -0,0 +1,158 @@
...
+
+	@GET
+	@Path("theme.js")
+	@Produces("application/javascript")
+	public Response retrieveThemeInfo() throws IOException {
+		String themeString = configurationService.getProperty(THEME_KEY);
+		//set default theme if none exists in database
+		if (themeString == null) {
+			setTheme(new NewUITheme());
```
doolse: Although it probably won't cause any harm to save a default, I don't think a GET method should really modify the database if it doesn't have to.

Source/Plugins/Core/com.equella.core/src/com/tle/web/api/newuitheme/NewUIThemeResource.java
```diff
@@ -0,0 +1,67 @@
...
+	Response retrieveLogo() throws IOException;
+
+	@GET
+	@Path("customlogo.js")
+	@ApiParam("Checks whether or not this institution is using a custom logo.")
+	@Produces("application/javascript")
+	Response customLogoExists();
+
+	@PUT
+	@Path("update")
```
doolse: I'd make this `settings` and also have a corresponding @GET method which returns the `NewUITheme`. The `theme.js` endpoint is really just there to make sure the theme is loaded before the app javascript is run, it's shouldn't be used as the source for editing the theme.

Source/Plugins/Core/com.equella.core/src/com/tle/web/api/newuitheme/impl/NewUIThemeResourceImpl.java
```diff
@@ -0,0 +1,158 @@
...
+		setTheme(themeString);
+		return Response.accepted().build();
+	}
+
+	@PUT
+	@Path("/updatelogo")
+	public Response updateLogo(File logoFile) throws IOException {
+		if (tleAclManager.filterNonGrantedPrivileges(Collections.singleton(PERMISSION_KEY), false).isEmpty()) {
+			throw new PrivilegeRequiredException(PERMISSION_KEY);
+		}
```
doolse: The permission check could be refactored into a method. Could also go in the `ThemeService` class.

Source/Plugins/Core/com.equella.core/js/tsrc/theme/ThemePage.tsx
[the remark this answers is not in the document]
SammyIsConfused: Is the way I changed it in [this commit](https://github.com/equella/Equella/pull/625/commits/e988159d4ab78ec07e8a751899c0fe315868305c) acceptable?
I could separate the page into separate component files outside of the ThemePage class, but that would require passing around some data in different ways, and it isn't like the components are going to be used elsewhere.

Source/Plugins/Core/com.equella.core/js/tsrc/theme/ThemePage.tsx
SammyIsConfused: I don't think I can use a method call within the state initializer, and this is the only other place it is. I can refactor it but it won't reduce anything, it will just make handleUndoButton be a method empty except another method call.

Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/service/ThemeSettingsService.java
```diff
@@ -0,0 +1,40 @@
...
+import com.tle.web.api.newuitheme.impl.NewUITheme;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+
+public interface ThemeSettingsService {
+
+	//GETTERS//
+	String getTheme() throws JsonProcessingException;
```
doolse: `getTheme()` should return `NewUITheme`

Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/service/ThemeSettingsService.java
```diff
@@ -0,0 +1,40 @@
...
+	String getLogoURL();
+	InputStream getCustomLogo() throws IOException;
+	boolean getCustomLogoStatus();
+
+	//SETTERS (returns true if successful)//
+	boolean setTheme(NewUITheme theme) throws JsonProcessingException;
```
doolse: No need for `boolean` return values

Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/service/ThemeSettingsService.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+
+public interface ThemeSettingsService {
+
+	//GETTERS//
+	String getTheme() throws JsonProcessingException;
+	String getLogoURL();
```
doolse: URLs that are part of the Resource should be generated by the Resource

Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/service/ThemeSettingsService.java
```diff
@@ -0,0 +1,40 @@
...
+	InputStream getCustomLogo() throws IOException;
+	boolean getCustomLogoStatus();
```
doolse: The convention is usually to use `is` instead of `get` for booleans

Source/Plugins/Core/com.equella.core/src/com/tle/web/api/newuitheme/NewUIThemeResource.java
```diff
@@ -0,0 +1,75 @@
...
+
+@Path("theme/")
+@Api("New UI Theme Resource")
+public interface NewUIThemeResource {
+
+	@GET
+	@Path("settings")
+	@ApiParam("Retrieves the current institution's theme settings from the database.")
+	@Produces("application/javascript")
+	Response retrieveThemeInfo() throws IOException;
```
doolse: This method should return `NewUITheme` instead of javascript.
Need a new method which returns javascript for both the theme and logo url at the path `theme.js`

Source/Plugins/Core/com.equella.core/src/com/tle/web/api/newuitheme/NewUIThemeResource.java
```diff
@@ -0,0 +1,75 @@
...
+	@Path("newLogo.png")
+	@ApiParam("Retrieves the current institution's logo from the filestore.")
+	@Produces("image/png")
+	Response retrieveLogo() throws IOException;
+
+	@GET
+	@Path("customlogo.js")
+	@ApiParam("Checks whether or not this institution is using a custom logo and returns the URL.")
+	@Produces("application/javascript")
+	Response retrieveCustomLogoURL();
```
doolse: This should be removed in favour of returning the logo url in `theme.js`

Source/Plugins/Core/com.equella.core/src/com/tle/web/api/newuitheme/NewUIThemeResource.java
```diff
@@ -0,0 +1,75 @@
...
+	@Path("settings")
+	@ApiParam("Retrieves the current institution's theme settings from the database.")
+	@Produces("application/javascript")
+	Response retrieveThemeInfo() throws IOException;
+
+	@GET
+	@Path("logo")
+	@ApiParam("Retrieves the path to the logo used by this institution.")
+	@Produces("text/plain")
+	Response retrieveLogoPath();
```
doolse: I don't think this should return the uri to the logo, if it needs to exist at all it should return the actual image stream of the custom logo

Source/Plugins/Core/com.equella.core/js/tsrc/theme/ColorPickerComponent.tsx
```diff
@@ -0,0 +1,97 @@
...
+    popover: {
+      position: "absolute",
+      zIndex: 2
+    },
+    cover: {
+      position: "fixed",
+      top: "0px",
+      right: "0px",
+      bottom: "0px",
+      left: "0px"
```
ChristianMurphy: :art: `0` is unitless

SammyIsConfused: It is, but I've used px everywhere else so I'd prefer to keep it consistent.

Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/service/impl/ThemeSettingsServiceImpl.java
```diff
@@ -0,0 +1,133 @@
...
+			bImage.getWidth() - 1, bImage.getHeight() - 1, null);
+		g2d.dispose();
+		RenderedImage rImage = resizedImage;
+
+		//write resized image to image file in the institution's filestore
+		ByteArrayOutputStream os = new ByteArrayOutputStream();
+		ImageIO.write(rImage, "png", os);
+		InputStream fis = new ByteArrayInputStream(os.toByteArray());
+		fileSystemService.write(customisationFile, LOGO_FILENAME, fis, false);
+	}
```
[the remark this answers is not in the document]
doolse: @cbeach47 The custom links icon isn't related to this, best make an issue for that.

SammyIsConfused:
@cbeach47 
Totally agree with the IO streams. That has been done.

Menu background refers to all Material UI menu panes - not just the sidebar. The background refers to the colour behind that. These are two different colours.

Custom link icons is out of the scope of this PR. Best to have your own issue for that.

The buttons are different styles based upon [Material Design guidelines.](https://material.io/design/components/buttons.html#hierarchy-placement) I can easily make them uniform, but the differences in emphasis were chosen deliberately.

Moving the link to the UI section will be done today.

PR #629, opened 2018-11-29
build: add dependency graph plugin

ChristianMurphy:
allows Equella's dependency graph to be visualized.
allows metadata to be extracted by dependency analysis tools.


edalex-ian:
Heya @ChristianMurphy ,

Just having a read of the [plug-in](https://github.com/jrudolph/sbt-dependency-graph) and it advises people doing a global install if they need it.

> sbt-dependency-graph is an informational tool rather than one that changes your build, so you will more than likely wish to install it as a global plugin so that you can use it in any SBT project without the need to explicitly add it to each one.

I think that'd be great, as our SBT already talks awhiles to start, so we probably want to be cautious adding more.

Ian.

Issue #608, opened 2018-11-20
Various drop down's are incorrectly sized


doolse:
Due to #528 some of the drop downs aren't sized correctly.

- [x] Manage Resources / Bulk action / Edit metadata.. / "Select schema nodes"

mrblippy:
Also look borked in the navigation builder (probably other places too)
![capture](https://user-images.githubusercontent.com/4625498/49201690-053a2080-f3f6-11e8-9181-d09d93be23e0.PNG)

edalex-ian:
Heya @SammyIsConfused 

With 2018.2 approaching, could you please have a look at this?

Cheers!

edalex-ian:
**UPDATE:** @SammyIsConfused  has found that there are many cases of this broken. It seems to be because the change for #528 was to *all* drop downs rather than just those for the course selector. And there are different explicit CSS rules for different exact pixel widths on different `select` elements. Further to that, the select2 library (which was used for #528) is then also trying to specify the width inline on the elements [based on it's own rules](https://select2.org/appearance#container-width).

It would seem the best way forward is to only do the #528 change to the course drop downs, and then some different CSS can be applied which will be supported by select2.

mrblippy:
Only a minor thing, but the bulk actions drop-downs aren't quite lined up with the content below it.
![capture](https://user-images.githubusercontent.com/4625498/49838429-c09f8380-fdfe-11e8-88c6-0aa490866c22.PNG)

edalex-ian:
Is that any different from 6.6?

mrblippy:
Oops, it's also present in 6.6

edalex-ian:
Okay, so I'll say the initial issue this was raise for is fixed, but there is a further older issue. So, safe to close this and create a new issue @mrblippy ?

mrblippy:
Yep, will do. I will log that as a separate issue. Verified in 2018.2

PR #642, opened 2018-12-11
#608 Revert to defaulting to stylish select

doolse:


edalex-ian:
Cool, looks fine to me - I see the main change is the support for specifying the render type, and then rest is a revert.

And I understand @doolse is going to update the auto tests and check locally, then once that is working a merge will be done.

Issue #648, opened 2018-12-12
Course editor: Server error when saving changes more than once


SammyIsConfused:
**Describe the bug**
In the new Course editor, if you go to create a new course and hit save, then edit again without leaving the page and resave, a 400 error labelled "Save failed due to server error" appears and the new changes are not saved.

**To Reproduce**
1. Go to '[institution URL]/vanilla/page/course'.
2. Click on the '+' button in the bottom right of the page to add a new course. You should now be on ''[institution URL]/vanilla/page/course/new'.
3. Type any string into at least the mandatory fields 'Name*' and 'Code*'.
4. Click save. The message "Successfully saved" should appear at the bottom of the page.
5. Without leaving the page, edit either of these fields.
6. Click save again.
ERROR: An error message pops up labelled "Save failed due to server error" appears and changes are not saved. This will trigger as many times as you click save.

**Expected behavior**
When clicking save, the course should be updated to whatever the form has been modified to, no matter how many times it is edited.

**Screenshots**
![screenshot from 2018-12-12 11-35-18](https://user-images.githubusercontent.com/24543345/49839282-846e2200-fe02-11e8-90bd-2d1aeb06861c.png)

**Stacktrace**
```
ERROR [CurrentLocale] Missing language string for 
'com.equella.core.course.edit.validation.codeinuse'
java.lang.Error
	at com.tle.common.i18n.CurrentLocale$AbstractCurrentLocale.get(CurrentLocale.java:148)
        ....
```

**Platform:**
 - OpenEquella Version: 6.6.r18, 6.7.r174
 - OS: Ubuntu 18.04.1
 - Browser: Firefox, Chrome

**Additional context**
None.

edalex-ian:
No need to backport any more, and looks like we have a PR just waiting for @abidingotter to merge.

PR #668, opened 2018-12-21
Bugfix/courseeditor

abidingotter:
#648


edalex-ian:
@abidingotter this looks like it's good to go other than the conflict. Any chance you could fix the conflict this Friday and merge?

PR #671, opened 2018-12-21
chore: update tomcat to version 8.5.37

ChristianMurphy:


edalex-ian:
Heya @ChristianMurphy 

Being a revision update only, we're hopefully that this should be painless.

But we'd like to make sure we can run the test suite over this for more confidence it doesn't break anything. We're currently making some updates to that, so we'll just hold this until that's work done before merging.

ChristianMurphy:
rebased on latest `develop`

ChristianMurphy:
closing in favor of renovate

PR #674, opened 2018-12-21
chore: update axis2 to version 1.7.9

ChristianMurphy:


ChristianMurphy:
This one does mostly builds, but does not pass tests yet.
Planning to dig into the issues sometime in January.

Issue #680, opened 2018-12-27
SOAP "fastSearch" fails with exception


doolse:
**Describe the bug**
Using the Soap51 endpoint method "searchItemsFast" fails with 

`org.apache.cxf.binding.soap.SoapFault: an implementation is missing`

**To Reproduce**
1. Run the Autotest `Soap51Test`

**Platform:**
 - OpenEquella Version: 2018.2

SammyIsConfused:
Tested on 6.6-stable, 2018.2-stable and 2019.1-alpha. Soap51Test checks out.

PR #681, opened 2018-12-27
#680 Replace the ??? with a no-op

doolse:


edalex-ian:
Link PR to #680

edalex-ian:
Hey @doolse 

Can you give a run down of how this fixes the issue please?

doolse:
`???` is the equivalent of `throw new UnsupportedOperationException()` in java, generally it's generated by the IDE when you implement something as a reminder to actually implement the method. In this case the correct implementation is a no-op.

edalex-ian:
Gotcha, just found this too: https://alvinalexander.com/scala/what-does-three-question-marks-in-scala-mean

Kind of hard to search for `???`.

PR #698, opened 2019-01-03
Fix deserialisation of File API folder renames 

doolse:
We accept `GenericFileBean` as our JSON representation, which is meant to allow us to pass in the `FolderBean` which we retrieved from the GET, however Jackson has a bug which is preventing that from working. 

We don't need to deserialise different types of `GenericFileBean` so we can just use `@JsonTypeName` instead of a mapper extension.


Source/Plugins/Core/com.equella.base/src/com/tle/web/api/item/interfaces/beans/GenericFileBean.java
```diff
@@ -18,19 +18,17 @@
 
 import javax.xml.bind.annotation.XmlRootElement;
 
+import com.fasterxml.jackson.annotation.JsonSubTypes;
```
edalex-ian: Is that redundant? I see it added, but not used. :thinking:

PR #694, opened 2019-01-02
docs: update copyright year to 2019

ChristianMurphy:

PR #699, opened 2019-01-03
docs: apply Apache license header

ChristianMurphy:
license text from https://apache.org/legal/src-headers.html
closes #694


Installer/src/com/dytech/edge/installer/CacheConfig.java
```diff
@@ -1,17 +1,20 @@
 /*
- * Copyright 2017 Apereo
+ * Licensed to the Apache Software Foundation (ASF) under one
```
ChristianMurphy: Should this be updated to read
```
Licensed to the Apereo Foundation under one
```
?

ChristianMurphy:
@cbeach47 @edalex-ian good idea, updated notice file with Apache license text.

NOTICE.md
```diff
@@ -1,4 +1,24 @@
-# equella-licenses
+# Open Equella
+
+Copyright 2017-2019 The Apereo Foundation
```
ChristianMurphy: Is 2017-2019 accurate?

cbeach47:
I've sent a note to the licensing group at Apereo to get their take on the wording in the source code headers and NOTICE file.

ChristianMurphy:
@cbeach47 Any update from the licensing group?

cbeach47:
I am still waiting on a confirmation for the exact date (or range) for the copyright in the NOTICE file, but the rest of the change was confirmed (example below).  I suggest we leave the copyright as-is (2017 - 2019), and we can open a small PR to change the date if needed.

Our friends at Opencast provided an example that was confirmed as an appropriate path to take:

We include a common licensing header [1]. We also automatically check that it is included in exactly this format in our Java and JavaScript files so that people cannot forget to put it there or use another header [2,3].  A formal copyright claim is included in the NOTICES file.
[1]:
https://github.com/opencast/opencast/blob/develop/modules/animate-api/src/main/java/org/opencastproject/animate/api/AnimateService.java#L1-L20
[2]:
https://github.com/opencast/opencast/blob/develop/pom.xml#L352
[3]:
https://github.com/opencast/opencast/blob/develop/docs/checkstyle/eslintrc.js#L14-L18

~

Let's open a ticket to check the source files for the common header, but I don't think it needs to be included with this PR (I'd like to keep this one very clean).

Can you please expand 'ASF' to be 'Apereo Software Foundation' - it could get confused with 'Apache' vs 'Apereo'.

Otherwise looks good to merge :+1:

cbeach47:
As per the last CDM, I'm closing this PR and will create a new one off of the latest `develop` branch.

PR #704, opened 2019-01-08
Make icon colours consistent with each other

SammyIsConfused:


Source/Plugins/Core/com.equella.core/js/src/MainUI/Template.purs
```diff
@@ -325,7 +325,7 @@ templateClass = withStyles ourStyles $ R.component "Template" $ \this -> do
       badgedLink iconName count uri tip = 
         let iconOnly = icon_ [ D.text iconName ] 
             buttonLink :: forall a. IsTSEq a ButtonColors => a -> ReactElement -> ReactElement
-            buttonLink col linkContent = iconButton {"aria-label": tip, href: uri, color: asTS col :: ButtonColors} [ linkContent ]
```
doolse: If `buttonLink` isn't going to use the col parameter anymore it should be removed. 
So that means removing from the type definition: `a ->` (the actual parameter) and `forall a. IsTSEq a ButtonColors =>` (the constraint which checks that a is one of the color types). Also the "col" parameter from the definition.

SammyIsConfused: check c6dea84. Is this what you meant?

PR #708, opened 2019-01-10
Bugfix/649 search breadcrumb fix

SammyIsConfused:


Source/Plugins/Core/com.equella.core/src/com/tle/web/searching/navigation/SearchCollectionBreadcrumbProvider.java
```diff
@@ -84,25 +86,39 @@ public TagState getBreadcrumb(SectionInfo info, Map<String, ?> params)
...
 		collectionLink.setRel("parent");
 		collectionLink.setTitle(BREADCRUMB_TITLE);
 		return collectionLink;
 	}
+
+	private boolean isNewSearchActive()
+	{
+		UISettings uis = UISettingsJava.getUISettings();
+		return uis.newUI().enabled() && uis.newUI().newSearch();
+	}
```
edalex-ian: I wonder if this method should be up in `UISettings.scala`. Thoughts?

doolse: Actually yeah it should just be in `UISettings` itself.

Source/Plugins/Core/com.equella.core/src/com/tle/web/searching/navigation/SearchCollectionBreadcrumbProvider.java
```diff
@@ -87,7 +87,7 @@ public TagState getBreadcrumb(SectionInfo info, Map<String, ?> params)
 		collectionLink.setLabel(label);
 
 
-		if (isNewSearchActive())
+		if (UISettingsJava.isNewSearchActive())
```
doolse: Sorry yeah UISettings itself would be a better place for that check: So this would end up being:
`if (UISettingsJava.getUISettings().isNewSearchActive())`

Source/Plugins/Core/com.equella.core/src/com/tle/web/searching/navigation/SearchCollectionBreadcrumbProvider.java
SammyIsConfused: 0311bc3

Source/Plugins/Core/com.equella.core/src/com/tle/web/searching/navigation/SearchCollectionBreadcrumbProvider.java
SammyIsConfused: 0311bc3

Issue #709, opened 2019-01-11
Store IntelliJ IDEA Project Settings


edalex-ian:
We want to store the IntelliJ project settings in the repo, primarily to lever full code style settings, but also aid with other developers starting up fresh.

So we need to add the `.idea` folder, however we should make sure out `.gitignore` has:

```
### Intellij ###
## https://intellij-support.jetbrains.com/hc/en-us/articles/206544839-How-to-manage-projects-under-Version-Control-Systems

.idea/artifacts/
.idea/httpRequests/

# User-specific stuff:
.idea/workspace.xml
.idea/tasks.xml
.idea/dictionaries
.idea/vcs.xml
.idea/jsLibraryMappings.xml

# Sensitive or high-churn files:
.idea/dataSources.ids
.idea/dataSources.xml
.idea/dataSources.local.xml
.idea/dataSources/
.idea/sqlDataSources.xml
.idea/dynamic.xml
.idea/uiDesigner.xml

## And seeing SBT projects also regenerates .iml module files
## lets ignore them all:
.idea/modules/

### Intellij - END ###
```

We can start there, and see how we go.

But key is to make sure we include the `codeStyles/` folder.

ChristianMurphy:
Just a thought, rather than commiting the intellij files, locking contributors into a specific config.
Maybe offer a way to generate the .idea folder if wanted, while still allowing customization.
Something like: https://github.com/mpeltonen/sbt-idea

edalex-ian:
I think this can now be closed.

PR #710, opened 2019-01-11
Change .gitignore to include parts of .idea folder

SammyIsConfused:
#709


edalex-ian:
Heya @ChristianMurphy ,

The main reason for these are:

* To ease start-up for contributors who are using IntelliJ; and
* To leverage the IDEs codestyle configuration for those which are - a common reason for storing these folders in projects

With regards to locking people to specific config, it should be noted that this is not too dissimilar to build definitions (such as `build.sbt`). And just like those, you can still have your local machine specific configuration sitting alongside it - hence a number of those items placed into the `.gitignore`.

As for generating these kind of configs (from SBT, Gradle and all the others) my experience has often been mixed/poor - especially with plugins not staying up to date.

I'd be happy to see this proceed.

ChristianMurphy:
> To ease start-up for contributors who are using IntelliJ

Running `sbt gen-idea` once seems pretty easy :slightly_smiling_face: 

> To leverage the IDEs codestyle configuration for those which are - a common reason for storing these folders in projects

If possible it would be preferable to keep code style checking/formatting out of the IDE and in the build system.
It would be less than ideal for developers to _need_ to use Intellij to submit a merge-able PR. :-1: 

> As for generating these kind of configs (from SBT, Gradle and all the others) my experience has often been mixed/poor - especially with plugins not staying up to date.

And from my experience using the idea plugin for Gradle, the experience is straight forward and pleasant, and that plugin always stays up-to-date as it is a part of Gradle core. :man_shrugging:

edalex-ian:
Thanks for your feedback Christian. For now we'll proceed, and if it adds issues later we can back it out.

It also addresses our work right now with the number of Devs in Edalex office attempting to ensure we have consistent code format. When the contributors grow, we can re-asses.

Good news is, for now no one is locked into anything - even with it there.

Cheers.

Issue #660, opened 2018-12-19
Login Screen Notice


edalex-ian:
Ability to use the unused space on the login screen to display notices and information - also with the ability to include links and images - without having to resort to using Language Packs.

**Analysis**

Seems straight-forward. Would need some type of editor and would need to be accessed by admins only.

**Additional context**

Would be cool if this could include an editor which allowed formatting and images.

cbeach47:
:+1:

PR #723, opened 2019-01-29
Login HTML Notice: delete notice

SammyIsConfused:
Add delete functionality, as well as feedback snackbars for the creation and deletion methods.
#660

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/loginnotice/LoginNoticeConfigPage.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/loginnotice/LoginNoticeConfigPage.tsx
@@ -22,7 +22,7 @@ export const strings = prepLangStrings("loginnoticepage",
   {
     title: "Login Notice Editor",
     label: "Login Notice",
-    notifications:{
+    notifications: {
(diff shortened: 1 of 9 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/loginnotice/LoginNoticeConfigPage.tsx
```diff
@@ -42,39 +59,93 @@ class LoginNoticeConfigPage extends React.Component<LoginNoticeConfigPageProps,
...
+  };
+
+  handleDeletedDescriptionClose = () => {
+    this.setState({deleted: false});
+  };
+
+  Notifications = () => {
+    return (
+      <div>
+        <Snackbar anchorOrigin={{vertical: "bottom", horizontal: "right"}}
```
doolse: Best to use '&lt;MessageInfo/>' instead direct '&lt;SnackBar/>'

SammyIsConfused: 532e699

Issue #689, opened 2018-12-28
Search page layout editor


edalex-ian:
Creation of a new WYSIWYG editor for controlling exactly what filtering/sorting controls show on your search page.

Will utilise the API released in 2018.2: #587

PR #738, opened 2019-02-01
Listing of search configurations using EntityList react component

doolse:
Part of #689

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/.editorconfig
+++ b/Source/Plugins/Core/com.equella.core/js/.editorconfig
@@ -1,0 +1,8 @@
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 2
+trim_trailing_whitespace = true
+
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/components/EntityList.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/components/EntityList.tsx
@@ -14,8 +14,6 @@ const styles = (theme: Theme) => ({
     padding: theme.spacing.unit * 2, 
     position: "relative"
   }, 
-  resultList: {
-  },
   resultHeader: {
     display: "flex",
     justifyContent: "flex-end"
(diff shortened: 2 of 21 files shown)


Source/Plugins/Core/com.equella.core/js/.editorconfig
```diff
@@ -0,0 +1,8 @@
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 2
+trim_trailing_whitespace = true
```
edalex-ian: What the additional `.editorconfig`? Is there any concern/consideration required for having multiple around the repo?

Source/Plugins/Core/com.equella.core/js/tsrc/components/EntityList.tsx
```diff
@@ -0,0 +1,69 @@
...
+  resultText: {
+      flexGrow: 1
+  }, 
+  progress: {
+      display: "flex", 
+      justifyContent: "center"
+  },
+  fab: {
+    zIndex: 1000,
+    position: 'fixed',
```
SammyIsConfused: just a nitpick, but on lines 11, 15, 20, 21, 27 and 28 you use double quotes. Best to be consistent one way or the other.

Source/Plugins/Core/com.equella.core/js/tsrc/components/EntityList.tsx
```diff
@@ -0,0 +1,69 @@
...
+  overall: {
+    padding: theme.spacing.unit * 2, 
+    height: "100%"
+  }, 
+  results: {
+    padding: theme.spacing.unit * 2, 
+    position: "relative"
+  }, 
+  resultList: {
+  },
```
SammyIsConfused: `resultList` is unecessary

Source/Plugins/Core/com.equella.core/js/.editorconfig
doolse: The only reason for this is that my VSCode workspace is rooted from the "Source/Plugins/Core/com.equella.core/js" folder so it doesn't pick up the root .editconfig. I'm fairly sure that it had to be like that because of the purescript IDE support didn't work properly otherwise.

edalex-ian: :+1:

Issue #718, opened 2019-01-22
Stop users creating new item versions when moderation is in progress


edalex-ian:
**Is your feature request related to a problem? Please describe.**

Different end users can create new versions of an item and submit it for moderation, allowing for multiple versions of the same resource in to be under moderation simultaneously. But once one is accepted, the others will be left in moderation of attempts are made to finalise will trigger an error.

**Describe the solution you'd like**

If an item exists in moderation (whether it has a status of moderating or live), a new version cannot be created from any other version of that item (current live version or archived version).
A suitable message to the user will be displayed when the new version option is selected when there is an existing version in moderation.

**Additional context**

Seen on site for a user, who are currently using the manual work-around of requesting users to ensure they don't start a new moderation workflow if one is already in progress.

edalex-ian:
All tested and backported. Closing.

PR #740, opened 2019-02-04
Feature/718 display warning for new version

PenghaiZhang:
#718

### DIFF
--- a/Source/Plugins/Core/com.equella.core/resources/lang/i18n-resource-centre.properties
+++ b/Source/Plugins/Core/com.equella.core/resources/lang/i18n-resource-centre.properties
@@ -4020,6 +4020,7 @@ summary.restricted=You do not have permission to view this attachment
 summary.sidebar.actions.archive.receipt=Successfully archived
 summary.sidebar.actions.archive.title=Archive this version
 summary.sidebar.actions.changeownership.title=Change ownership
+summary.sidebar.actions.newversion.moderationconflict=A new version of this item cannot be created because there is a current version awaiting moderation.
 summary.sidebar.actions.delete.confirm=Are you sure you wish to delete this item?
 summary.sidebar.actions.delete.receipt=Successfully marked as deleted
 summary.sidebar.actions.delete.title=Delete this version
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/viewitem/actions/NewVersionSection.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/viewitem/actions/NewVersionSection.java
@@ -65,15 +65,13 @@ public class NewVersionSection extends AbstractWizardViewItemActionSection {
     // An item may have one or more versions.
     String selectedItemUuid = selectedItem.getUuid();
     List<Item> items = itemService.getVersionDetails(selectedItemUuid);
-    items.forEach(
-        item -> {
-          if (item.isModerating()) {
-            // Display a warning when creating a new version of this item because there is a current
-            // version awaiting moderation.
-            getComponent()
-                .setClickHandler(context, new OverrideHandler(Js.alert_s(NO_NEW_VERSION_LABEL)));
-          }
-        });
+    Boolean isModerating = items.stream().anyMatch(item -> item.isModerating());
+    if (isModerating) {
+      // Display a warning when creating a new version of this item because there is a current
+      // version awaiting moderation.
+      getComponent()
+          .setClickHandler(context, new OverrideHandler(Js.alert_s(NO_NEW_VERSION_LABEL)));
+    }
     return super.renderHtml(context);
   }
 
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/viewitem/actions/NewVersionSection.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/viewitem/actions/NewVersionSection.java
@@ -16,12 +16,20 @@
 
 package com.tle.web.wizard.viewitem.actions;
 
+import com.tle.beans.item.Item;
 import com.tle.beans.workflow.WorkflowStatus;
 import com.tle.core.guice.Bind;
+import com.tle.core.item.service.ItemService;
 import com.tle.web.sections.SectionInfo;
+import com.tle.web.sections.SectionResult;
 import com.tle.web.sections.equella.annotation.PlugKey;
+import com.tle.web.sections.events.RenderEventContext;
+import com.tle.web.sections.js.generic.Js;
+import com.tle.web.sections.js.generic.OverrideHandler;
 import com.tle.web.sections.render.Label;
 import com.tle.web.viewurl.ItemSectionInfo;
+import java.util.List;
+import javax.inject.Inject;
 
 @Bind
 @SuppressWarnings("nls")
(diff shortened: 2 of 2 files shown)


Source/Plugins/Core/com.equella.core/resources/lang/i18n-resource-centre.properties
```diff
@@ -4020,6 +4020,7 @@ summary.restricted=You do not have permission to view this attachment
 summary.sidebar.actions.archive.receipt=Successfully archived
 summary.sidebar.actions.archive.title=Archive this version
 summary.sidebar.actions.changeownership.title=Change ownership
+summary.sidebar.actions.newversion.no=Sorry, you cannot create a new version because the status of this item is moderating or live.
```
SammyIsConfused: The name `summary.sidebar.actions.newversion.no` is a little strange. I'd consider renaming to  `summary.sidebar.actions.newversion.moderationconflict` or something similar.

Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/viewitem/actions/NewVersionSection.java
```diff
@@ -29,7 +30,8 @@
 {
 	@PlugKey("viewitem.actions.newversion.title")
 	private static Label LABEL;
-
+	@PlugKey("summary.sidebar.actions.newversion.no")
+	private static Label No_NEW_VERSION_LABEL;
 	@Override
```
SammyIsConfused: I would reccommend adding a whitespace line between line 34 and 35.

Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/viewitem/actions/NewVersionSection.java
```diff
@@ -45,7 +47,24 @@ protected boolean canView(SectionInfo info, ItemSectionInfo itemInfo, WorkflowSt
...
+		/*
+		* Modified by Penghai Zhang ( 04/02/2019)
+		*
+		* Github number : #718
+		*
+		* Display a warning when creating a new version of an item  if this item is in moderation or the status of this item is LIVE.
+		*
+		*/
+		Item item = getItemInfo(info).getItem();
+		System.err.println("Moderation:" +item.getStatus().toString());
```
SammyIsConfused: While good for debugging, `System.err.println` calls should not be used in production code. I would omit these two lines, or use a logger.

Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/viewitem/actions/NewVersionSection.java
```diff
@@ -45,7 +47,24 @@ protected boolean canView(SectionInfo info, ItemSectionInfo itemInfo, WorkflowSt
 	@Override
 	protected void execute(SectionInfo info) throws Exception
 	{
-		forwardToWizard(info, false, false, true);
+		/*
```
SammyIsConfused: It is unnecessary to write commit and timestamp information in your code, as the git history logs all of this information already.

Source/Plugins/Core/com.equella.core/resources/lang/i18n-resource-centre.properties
edalex-ian: I see this message is literal from the GH issue, but after review ideally it should read:

> A new version of this item cannot be created because there is a current version awaiting moderation

This highlights that we actually need to revisit the implementation. See comment below.

Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/viewitem/actions/NewVersionSection.java
```diff
@@ -45,7 +47,24 @@ protected boolean canView(SectionInfo info, ItemSectionInfo itemInfo, WorkflowSt
...
+		*
+		* Github number : #718
+		*
+		* Display a warning when creating a new version of an item  if this item is in moderation or the status of this item is LIVE.
+		*
+		*/
+		Item item = getItemInfo(info).getItem();
+		System.err.println("Moderation:" +item.getStatus().toString());
+		System.err.println("id:"+item.getItemId());
+		if ("live".equals(item.getStatus().toString()) || "moderating".equals(item.getStatus().toString()))
```
edalex-ian: Unfortunately if we do this simple status check, then that would mean we can never create a new version - due to the 'is live' check.

What we actually need is something a bit more complicated to go and see if there are any moderation records active for this item. (Let's discuss in person more.)

Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/viewitem/actions/NewVersionSection.java
```diff
@@ -48,6 +56,18 @@ protected void execute(SectionInfo info) throws Exception
 		forwardToWizard(info, false, false, true);
 	}
 
+	@Override
+	public SectionResult renderHtml(RenderEventContext context)
+	{
+		Item item = getItemInfo(context).getItem();
+		if (item.isModerating())
+		{
+//		Display a warning when creating a new version of this item because there is a current version awaiting moderation.
```
edalex-ian: The comment marker (`//`) should align with the following line (i.e. `getComponent()...`) and then we only need one space after it.

```suggestion
			// Display a warning when creating a new version of this item because there is a current version awaiting moderation.
```

Source/Plugins/Core/com.equella.core/resources/lang/i18n-resource-centre.properties
```diff
@@ -4020,6 +4020,7 @@ summary.restricted=You do not have permission to view this attachment
 summary.sidebar.actions.archive.receipt=Successfully archived
 summary.sidebar.actions.archive.title=Archive this version
 summary.sidebar.actions.changeownership.title=Change ownership
+summary.sidebar.actions.newversion.moderationConflict=A new version of this item cannot be created because there is a current version awaiting moderation.
```
SammyIsConfused: Just noticed one more thing; the naming convention in this file seems to be to be all lowercase. So `moderationConflict` should be `moderationconflict`.

Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/viewitem/actions/NewVersionSection.java
```diff
@@ -48,6 +62,23 @@ protected void execute(SectionInfo info) throws Exception
...
 	}
 
+	@Override
+	public SectionResult renderHtml(RenderEventContext context)
+	{
+		Item selectedItem = getItemInfo(context).getItem();
+		// An item may have one or more versions.
+		String selectedItemUuid = selectedItem.getUuid();
+		List<Item> items = itemService.getVersionDetails(selectedItemUuid);
+		items.forEach(item -> {
```
edalex-ian: This will unnecessarily iterate over all instances/versions of an item. However, it really only needs to find one or none.

I suggest you re-write this to use a [java stream](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html) and then the `anyMatch( predicate )` method. You can then branch based on the boolean result from that.

PR #746, opened 2019-02-12
Login Notice: Add scheduling system to login notice

SammyIsConfused:
Login notices can now be set to "Always on", "Scheduled" or "Disabled". Scheduled login notices have a validated start and end date. 

#660 

image
image


edalex-ian:
@SammyIsConfused the travis checks are failing due to missing headers. Wont to fix that up and then we can have a look see?

Source/Plugins/Core/com.equella.core/js/tsrc/loginnotice/LoginNoticeModule.ts
```diff
@@ -9,7 +9,17 @@ export enum NotificationType {
...
+  OFF = "OFF",
+  ON = "ON",
+  SCHEDULED = "SCHEDULED"
+}
+export interface PreLoginNotice {
+  notice?: String;
+  scheduleSettings: ScheduleSettings;
+  startDate?: Date;
+  endDate?: Date;
+}
```
edalex-ian: This seems slightly curious. I would've though more along the lines of:

```
export interface PreLoginNotice {
  notice?: String;
  enabled: boolean;
  schedule?: ScheduleSettings;
}
```

And then because start and end dates are purely about scheduling, that's all you should need in `ScheduleSettings`. And then you can also argue that if you do have a schedule, then you must have a start and end date - so they're no longer optional (and don't need to be checked).

Source/Plugins/Core/com.equella.core/js/tsrc/loginnotice/PreLoginNoticeConfigurator.tsx
```diff
@@ -1,35 +1,50 @@
...
 interface PreLoginNoticeConfiguratorProps {
   handleError: (axiosError: AxiosError) => void;
   notify: (notificationType: NotificationType) => void;
 }
 
 interface PreLoginNoticeConfiguratorState {
-  preNotice?: string; //what is currently in the textfield
-  dbPreNotice?: string; //what is currently in the database
+  preNotice?: String; //what is currently in the textfield
+  dbPreNotice?: String; //what is currently in the database
```
edalex-ian: What's going on here. Why do we have both `string` and `String`?

Source/Plugins/Core/com.equella.core/js/tsrc/loginnotice/PreLoginNoticeConfigurator.tsx
```diff
@@ -41,16 +56,41 @@ class PreLoginNoticeConfigurator extends React.Component<
     this.state = {
       preNotice: "",
       dbPreNotice: "",
-      clearStaged: false
+      clearStaged: false,
+      checked: "ON",
+      dbChecked: "ON",
```
edalex-ian: Hard-coded strings are error prone. Be good to use an `enum`.

Source/Plugins/Core/com.equella.core/js/tsrc/loginnotice/PreLoginNoticeConfigurator.tsx
```diff
@@ -41,16 +56,41 @@ class PreLoginNoticeConfigurator extends React.Component<
...
+      dbEndDate: new Date()
     };
   }
 
   handleSubmitPreNotice = () => {
     if (this.state.preNotice != undefined) {
-      submitPreLoginNotice(this.state.preNotice)
+      let scheduleToSend: ScheduleSettings;
+      switch (this.state.checked) {
+        case "ON":
```
edalex-ian: I wonder... If Typescript allows you to use a string enum in this case statement (i.e. so you could do `case ScheduleSettings.ON:') like some languages. :thinking: 

But I guess my earlier comment means a change to this code anyway.

Source/Plugins/Core/com.equella.core/js/tsrc/loginnotice/PreLoginNoticeConfigurator.tsx
```diff
@@ -81,10 +121,47 @@ class PreLoginNoticeConfigurator extends React.Component<
     this.setState({ preNotice: e.value });
   };
 
+  mapScheduleSetting = (scheduleSetting: ScheduleSettings): string => {
```
edalex-ian: I wonder if you need this. :thinking: 

https://www.typescriptlang.org/docs/handbook/enums.html#enums-at-runtime

But again, this will change.

Source/Plugins/Core/com.equella.core/js/tsrc/loginnotice/PreLoginNoticeConfigurator.tsx
```diff
@@ -1,35 +1,50 @@
...
 }
 
 interface PreLoginNoticeConfiguratorState {
-  preNotice?: string; //what is currently in the textfield
-  dbPreNotice?: string; //what is currently in the database
+  preNotice?: String; //what is currently in the textfield
+  dbPreNotice?: String; //what is currently in the database
   clearStaged: boolean;
+  checked: string;
+  dbChecked: string;
```
edalex-ian: Might need a better name here. What does `checked` represent? What has been checked?

Source/Plugins/Core/com.equella.core/js/tsrc/loginnotice/PreLoginNoticeConfigurator.tsx
```diff
@@ -122,9 +199,96 @@ class PreLoginNoticeConfigurator extends React.Component<
...
+    //state matches database?
+    return (
+      this.state.checked == this.state.dbChecked &&
+      this.state.preNotice == this.state.dbPreNotice &&
+      this.state.startDate == this.state.dbStartDate &&
+      this.state.endDate == this.state.dbEndDate
+    );
+  };
+
+  handleRadioButtons = (event: ChangeEvent, value: string) => {
```
edalex-ian: You have good names for your below handlers, but just 'RadioButtons' here is unclear. Which radio buttons for what? (This probably also covers my query around what is `checked` - seems it's just the state of an unnamed radio button.)

Source/Plugins/Core/com.equella.core/js/tsrc/loginnotice/PreLoginNoticeConfigurator.tsx
```diff
@@ -142,14 +306,17 @@ class PreLoginNoticeConfigurator extends React.Component<
               inputProps={{ length: 12 }}
               placeholder={strings.prelogin.description}
               onChange={e => this.handlePreTextFieldChange(e.target)}
-              value={preNotice}
+              value={preNotice + ""}
```
edalex-ian: This (`""`) looks like a code smell. What's being attempted here?

Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/loginnotice/impl/LoginNoticeServiceImpl.java
```diff
@@ -78,6 +102,29 @@ public void deletePostLoginNotice() {
...
+  @Override
+  public boolean isActive(PreLoginNotice preLoginNotice) {
+    Date now = new Date();
+
+    Date startDateMidnight = preLoginNotice.getStartDate(); // 12am at the start the first day
+    DateUtils.truncate(startDateMidnight, Calendar.DAY_OF_MONTH);
+
+    Date endDateMidnight = preLoginNotice.getEndDate(); // 12am at the end of the last day
+    DateUtils.truncate(endDateMidnight, Calendar.DAY_OF_MONTH);
+    endDateMidnight = DateUtils.addDays(endDateMidnight, 1);
```
edalex-ian: Seems redundant to do all the above (and I wonder if it triggers any interesting NPEs) if `getScheduleSettings()` returns a value other than `SCHEDULED`. I'd probably change to extract the setting into a `final`, and then do something like:

```
boolean isActive = false;
if( scheduleSetting == ON ) {
  isAcitve = false;
}
else if( scheduleSetting == OFF ) {
  // Calculate the date points
  // ...

  // Set isActive to the comparison you already have
  isActive = ...
}

return isActive;
```

Source/Plugins/Core/com.equella.core/src/com/tle/web/login/LogonSection.java
```diff
@@ -221,7 +222,15 @@ public SectionResult renderHtml(RenderEventContext context) {
     Decorations decorations = Decorations.getDecorations(context);
     decorations.setTitle(TITLE_LABEL);
     decorations.setMenuMode(MenuMode.HIDDEN);
-    model.setLoginNotice(loginNoticeService.getPreLoginNotice());
+    PreLoginNotice preLoginNotice = new PreLoginNotice();
+    try {
+      preLoginNotice = loginNoticeService.getPreLoginNotice();
+    } catch (IOException e) {
+      e.printStackTrace();
```
edalex-ian: No `printStackTrace()`. Let's log it.

PR #748, opened 2019-02-15
Feature/689 New entity table for Search Configs

doolse:
Related to #689 

Best compared with ignoring of space changes.

This pull request contains:

* The new table definition - `OEQEntity`
* A migration which converts the old search configurations / creates the new table and indexes - `NewEntityTable`
* An importer/exporter for the new table which will work for new entities based on this table - `OEQEntityConverter`
* A library API class for working with the new table - `EntityDB`


Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/institution/OEQEntityConverter.scala
```diff
@@ -0,0 +1,172 @@
...
+      j.typeid,
+      j.name,
+      LocaleStrings(j.nameStrings),
+      j.description,
+      LocaleStrings(j.descriptionStrings),
+      UserId(j.owner),
+      Instant.ofEpochMilli(j.created),
+      Instant.ofEpochMilli(j.modified),
+      j.data
+    )
```
edalex-ian: This would benefit from named arguments. Originally I asked this question on `toEntityJson()`, but continued on. However here the readability starts to really suffer.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/searches/SearchDefaults.scala
```diff
@@ -15,13 +15,20 @@
...
     "search" -> SearchConfig(
-      None,
+      UUID.randomUUID(),
       "item",
+      "Default search",
+      None,
+      None,
+      None,
+      new Date,
+      new Date,
```
edalex-ian: The above parameters would greatly benefit from named parameters, and would also better fit with the idea we're generating a JSON block here. The parameters below work due to the `Map` and then the key names.

Source/Plugins/Core/com.equella.serverbase/scalasrc/com/tle/core/db/dao/EntityDB.scala
```diff
@@ -0,0 +1,90 @@
...
+          typeid = ee.typeId,
+          "",
+          LocaleStrings.empty,
+          None,
+          LocaleStrings.empty,
+          UserId(uc.user.getUserBean.getUniqueID),
+          Instant.now(),
+          Instant.now(),
+          Json.obj()
+      )
```
edalex-ian: Needs named arguments. Indeed, lines 52 and 53 are the stand examples of when named arguments are needed.

Source/Plugins/Core/com.equella.serverbase/scalasrc/com/tle/core/validation/EntityValidation.scala
```diff
@@ -0,0 +1,71 @@
...
+  def standardValidation(
+      edits: OEQEntityEdits,
+      oeq: OEQEntity,
+      locale: Locale
+  ): ValidatedNec[EntityValidation, OEQEntity] = {
+    EntityValidation
+      .nonBlank("name", edits.name, edits.nameStrings, locale)
+      .map { n =>
+        val desc = LocaleStrings.fromStrings(
+          edits.description.getOrElse(""),
```
edalex-ian: I wonder why below on line 67 you used `LocalStrings.empty`, however here you went with `""`?

Source/Plugins/Core/com.equella.core/plugin-jpf.xml
```diff
@@ -5535,11 +5525,6 @@
     <parameter id="class" value="bean:com.tle.core.institution.convert.MigrationConverter" />
     <parameter id="order" value="100" />
   </extension>
-  <extension plugin-id="com.tle.core.institution.convert" point-id="converter" id="zippingConverter">
-    <parameter id="id" value="ZIPFILES" />
-    <parameter id="class" value="bean:com.tle.core.institution.convert.ZippingConverter" />
-    <parameter id="order" value="20000" />
-  </extension>
```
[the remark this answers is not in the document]
doolse: > A lot of code here. Ideally this should have been split up and then spread across a few PRs. This may have been one reason it has taken so longer for anyone to review. But also, I'm not confident this was the best review, due to retention by the end.
> 
> My general comments are:
> 
> 1. There are several places here due to classes with many properties where named arguments would greatly benefit readability. In the end though, I stopped pointing them out;

I will name them :+1:

> 2. We had mentioned my concern with continued use of SimpleDBA, and now seeing the proliferation here I'm even more concerned. I think after seeing the quantity it's really highlighted a need to discuss further.
As I mentioned the only thing that would be a suitable replacement would be Doobie. SimpleDBA was designed with oEQ use cases in mind, so a lot of that would have to be re-done from scratch with another library.

> 3. Before starting I note there is now a minor conflict that needs to be resolved.
I think I fixed that before you did your review.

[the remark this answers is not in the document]
doolse: BTW yeah they were removed and put into code

PR #750, opened 2019-02-22
Formatters with pre commit hooks and checks

doolse:
Related to #711, #712 and #713. 

The idea is to create a single commit which contains the tooling first which can then be cherry-picked onto current feature branches and the formatters can be run to ease merge conflicts.


CONTRIBUTING.md
```diff
@@ -124,18 +124,42 @@ EclipseKeys.createSrc := EclipseCreateSrc.Default + EclipseCreateSrc.ManagedClas
 EclipseKeys.eclipseOutput := Some("target/scala-2.11/classes/")
 ```
 
+#### Code Formatters
+
+All code should be formatted using the following formatting tools:
+
+- Scala - [scalafmt](https://scalameta.org/scalafmt/)
+- Java - [Google Java Format](https://github.com/google/google-java-format)
+- Typescript/JS - [Prettier](https://prettier.io/)
```
edalex-ian: Should we either:

* Note that this does more than TS/JS; or
* Should we setup a prettier-ignore file and block all the things other than JS/JSX/TS/TSX

I'm inclined to bite the bullet and just do the first.

doolse: I've noted that it does some more filetypes

PR #752, opened 2019-02-27
Modify the new item REST API (#718)

PenghaiZhang:
Since the logic of creating a new version of an item has been changed, the new item REST API needs changes as well.

#718


edalex-ian:
@PenghaiZhang : I'm concerned that this logic is now in two places. Ideally it should be somewhere central, like the `ItemService`. Is it possible to simply have a `newVersion` method in the item service that handles all this, and then we use it in both view layers (i.e. the UI and the REST)?

abidingotter:
@edalex-ian Normally I'd agree, but the REST and "normal UI" methods of saving an item are very different codebases.  But perhaps the check should be somewhere within itemDeserializerService.newItem instead of the REST endpoint itself though.

PenghaiZhang:
> @PenghaiZhang : I'm concerned that this logic is now in two places. Ideally it should be somewhere central, like the `ItemService`. Is it possible to simply have a `newVersion` method in the item service that handles all this, and then we use it in both view layers (i.e. the UI and the REST)?

Thanks Ian. A new method called "isItemInModeration" has been added to ItemService.

PenghaiZhang:
> @edalex-ian Normally I'd agree, but the REST and "normal UI" methods of saving an item are very different codebases. But perhaps the check should be somewhere within itemDeserializerService.newItem instead of the REST endpoint itself though.

Thank you! 
the check now is in ItemEditorServiceImpl.newItemEditor

edalex-ian:
Can we please tweak the message text as part of this.

'A new version of this item cannot be created because there is already a version in moderation.'

PR #765, opened 2019-03-07
Merging the autotests into a subfolder of oEQ

doolse:


edalex-ian:
So to be sure, to be sure... Is the goal here to also have passing checks @doolse ? Seems like it'd make sense to - especially the AWS CodeBuild one. :thinking:

doolse:
Yeah the idea is to not merge until the builders can build it

doolse:
I did this on my own fork to not pollute the official one if it wasn't going to work but the history appears to have come over so I will close this and do it in a branch instead

Issue #749, opened 2019-02-21
Contribution - "Allow non-sequential navigation" checked allows saving without filling in mandatory fields of unvisited pages


mrblippy:
**Describe the bug**
Currently you are able to save and publish an item that has incomplete mandatory fields, provided you haven't visited that page yet. This was changed in 6.6 to satisfy a particular client https://github.com/equella/Equella/issues/406

But it doesn't really make sense that you can save a wizard without having completing mandatory fields.

**To Reproduce**
1. Import the scheme and collections attached to this issue
[Schema - Test of wizard schema.zip](https://github.com/equella/Equella/files/2887749/Schema.-.Test.of.wizard.schema.zip)
[Collection Definition - TEST of wizard.zip](https://github.com/equella/Equella/files/2887752/Collection.Definition.-.TEST.of.wizard.zip)
2. Open up a contribution and save it without visiting page 2 (which contains mandatory fields)
3. Click the save button
4. The item will be published and made live

**Expected behavior**
User should be presented with the following dialog upon clicking save
![capture](https://user-images.githubusercontent.com/4625498/53454378-10473000-3a7b-11e9-83f8-574b0bcc90d2.PNG)

**Platform:**
 - OpenEquella Version: 6.6, 2018.2
 - OS: Windows
 - Browser Chrome

abidingotter:
I don't believe that you should be able to submit, only save as draft.  I think that may have been the original intention.

edalex-ian:
Ahhh, so can you submit/publish @mrblippy , or just save and then come back and edit (like a draft)?

mrblippy:
Yeah, you can submit/publish the item

mrblippy:
Found another case that needs to be handled:
Below the save button is a 'Save and continue' button, which saves the item as a draft, but keeps the user in the contribution wizard. 

When save and continue is clicked the user can save and publish without completing the mandatory field on page 2

PR #770, opened 2019-03-12
Fix another case where mandatory field bug occurs (#749)

PenghaiZhang:
This PR fixes another case where the mandatory field bug happens.

"When save and continue is clicked the user can save and publish without completing the mandatory field on page 2"


PenghaiZhang:
> This PR fixes another case where the mandatory field bug happens.
> 
> After Charlie's testing, the last possible case exposed : "When save and continue is clicked the user can save and publish without completing the mandatory field on page 2"

PenghaiZhang:
Fix two cases where the mandatory field bugs happens.
1. When editing an item, clicking 'save and continue' button will enable users to publish an item without filling in mandatory fields displayed in pages other than page 1.
2. When moderator is moderating an item, this item can be edited and saved without filling in mandatory fields displayed in pages other than page 1.

PR #774, opened 2019-03-13
Cloud provider registration callback

doolse:


doolse:
As discussed we will do the testing as part of the add (or at least in another story)

PR #775, opened 2019-03-13
Provide a registration API

doolse:
Add provider API using a replicate cache to store registration tokens. Depends on #774


doolse:
> Looks good, but just one thing:
> 
> * What is the `register/init` endpoint for? Could we make this clearer in the doco? If it needs more than the one liner, should we add more top level doco for this API?

The `register/init` call is actually the one the openEQUELLA UI will call before it forwards to the given provider URL. So it's actually not something that cloud providers themselves will be using. However it does bring up that at some point we need a "how do i implement a cloud provider guide"

PR #779, opened 2019-03-14
Build and ide cleanup

doolse:
- .sbt files formatted
- google java intellij style added
- Use npm instead of yarn because it's not prone to race conditions on install
- Try to keep yarn/npm refs out of support scripts


Source/Plugins/Core/com.equella.core/build.sbt
```diff
@@ -114,7 +114,8 @@ resourceGenerators in Compile += Def.task {
 
 resourceGenerators in Compile += Def.task {
   val baseSwagger = baseDirectory.value / "swaggerui"
-  Common.runYarn("browserify", baseSwagger)
+  Common.nodeInstall(baseSwagger)
+  Common.nodeScript("browserify", baseSwagger)
```
edalex-ian: Obviously not directly related to your work here, but I thought we weren't browserfying any more. Should this parcel?

Source/Plugins/Core/com.equella.core/swaggerui/package.json
```diff
@@ -1,12 +1,13 @@
 {
   "scripts": {
-    "browserify": "yarn install && browserify index.js -o target/bundle.js"
+    "browserify": "browserify index.js -o target/bundle.js"
```
edalex-ian: As mentioned above, I thought we were Parcel.js these days. Can we change this to be the same? If so, we can split it out to another ticket for another day.

buildspec.yml
```diff
@@ -22,9 +22,9 @@ phases:
   post_build:
     commands:
       - publishBuildResults
-      - echo "Finished"
 cache:
   paths:
     - "/root/.ivy2/**/*"
     - "/root/.npm/**/*"
     - "/root/.sbt/**/*"
+    - "/root/.cache/yarn/**/*"
```
edalex-ian: Why add  the yarn dir to the cache if we're no longer using yarn?

Source/Plugins/Core/com.equella.core/build.sbt
doolse: Swagger doesn't have any equella code, so we don't get that much benefit from using parcel, but I will try and convert for consistency.

buildspec.yml
doolse: True, the autotests haven't been changed yet but I will change them to npm too

PR #785, opened 2019-03-20
Add a page to register cloud providers

PenghaiZhang:
Add a page for registering cloud providers

Screenshot from 2019-03-20 12-13-18
Screenshot from 2019-03-20 12-13-37

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/cloudprovider/CloudProviderModule.ts
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/cloudprovider/CloudProviderModule.ts
@@ -4,14 +4,22 @@ import { Config } from "../config";
 import { CloudProviderEntity } from "./CloudProviderEntity";
 import { prepLangStrings } from "../util/langstrings";
 
-export const GET_CLOUD_PROVIDER_URL = `${Config.baseUrl}api/cloudprovider`;
-
+export const GET_CLOUD_PROVIDER_LIST_URL = `${Config.baseUrl}api/cloudprovider`;
+export const POST_CLOUD_PROVIDER_REGISTER_INIT_URL = `${
+  Config.baseUrl
+}api/cloudprovider/register/init`;
 export const langStrings = prepLangStrings("cp", {
-  title: "cloud providers",
+  title: "Cloud providers",
   cloudProviderAvailable: {
     zero: "No cloud providers available",
     one: "%d cloud provider",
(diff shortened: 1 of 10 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/cloudprovider/CloudProviderModule.ts
```diff
@@ -4,21 +4,35 @@ import { Config } from "../config";
...
 export function getCloudProviders(): Promise<
   PagingResults<CloudProviderEntity>
 > {
   return Axios.get<PagingResults<CloudProviderEntity>>(
-    GET_CLOUD_PROVIDER_URL
+    GET_CLOUD_PROVIDER_LIST_URL
   ).then(res => res.data);
 }
+
+export function registerCloudProviderInit(params: object) {
```
doolse: Instead of `params: object` the function should simply take the url and construct the parameters inside the function itself.

You should also put the type of the promise return value here instead of having to cast it later

Source/Plugins/Core/com.equella.core/js/tsrc/cloudprovider/NewCloudProvider.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+interface NewCloudProviderProps extends WithStyles<typeof styles> {
+  bridge: Bridge;
+}
+interface NewCloudProviderState {
+  error?: ErrorResponse;
+  cloudProviderUrl: string;
+  canRegister: boolean;
+}
+interface registerCloudProviderInitRes {
```
doolse: The convention is to use upper case as the first letter for interfaces.
Also I would change this to `CloudProviderInitResponse`

Source/Plugins/Core/com.equella.core/js/tsrc/cloudprovider/NewCloudProvider.tsx
```diff
@@ -0,0 +1,141 @@
...
+    }
+  });
+
+interface NewCloudProviderProps extends WithStyles<typeof styles> {
+  bridge: Bridge;
+}
+interface NewCloudProviderState {
+  error?: ErrorResponse;
+  cloudProviderUrl: string;
+  canRegister: boolean;
```
doolse: No need to have `canRegister` as the cloudProviderUrl can be used to calculate that

PR #790, opened 2019-03-22
Login Notice: migrate to TinyMCE

SammyIsConfused:
#660


Source/Plugins/Core/com.equella.core/resources/web/reactjs/tinymce/skins/content/default/content.css
```diff
@@ -0,0 +1,45 @@
+/**
+ * Copyright (c) Tiny Technologies, Inc. All rights reserved.
+ * Licensed under the LGPL or a commercial license.
+ * For LGPL see License.txt in the project root for license information.
```
edalex-ian: Hmmm, I wonder if we need to make sure we still have that License.txt, and or whether we need to include something in the oEQ NOTICES file. :thinking:

autotest/Tests/src/test/scala/equellatests/pages/LoginNoticePage.scala
```diff
@@ -13,23 +13,23 @@ case class LoginNoticePage(ctx: PageContext)
...
-    preNoticeAddImagePopup.findElement(
-      By.xpath("//span[contains(@class, 'rdw-image-modal-header-option') and contains(.,'URL')]"))
+  private def preNoticeAddImageField: WebElement =
+    preNoticeAddImageForm.findElement(By.xpath("//input"))
 
-  private def preNoticeAddImageURLSection: WebElement =
-    findElement(By.className("rdw-image-modal-url-section"))
   private def preNoticeAddImageOK: WebElement =
-    preNoticeAddImagePopup.findElement(By.xpath("//button[text()='Add']"))
+    preNoticeAddImagePopup.findElement(By.xpath("//button[text()='Save']"))
```
edalex-ian: I note a lot of findElement()s above with rather low specificity. These can significantly slow down execution of selenium tests (as it has to walk the whole DOM). Can they be improved at all? At least with a higher level `id`?

(I've put this comment here, but please browse the lines above.)

Source/Plugins/Core/com.equella.core/resources/web/reactjs/tinymce/skins/content/default/content.css
SammyIsConfused: We will discuss this more in the CDM.

edalex-ian: Yup, looks like maybe good enough for now. Have added to CDM agenda.

autotest/Tests/src/test/scala/equellatests/pages/LoginNoticePage.scala
SammyIsConfused: The issue with some of these is that TinyMCE generates a unique number that gets appended to IDs of elements every time the editor loads, so a specific ID isn't available for all the elements. Another issue is that in the IDs that _are_ persistent, often there are multiple elements with the same ID and so I must find another way.
The example on line 32 doesn't walk through the whole DOM, just the children of `preNoticeAddImagePopup`.

doolse: Generally you would specify the id of something else in the page, perhaps the footer bar where the buttons exist. I don't think the overhead of doing a full DOM search is really a problem but what if the page happens to have another button with "Save" on it? Unlikely but it's a good practice to be a bit more specific

SammyIsConfused: @doolse I was under the impression that calling `findElement` on `preNoticeAddImagePopup` only searches the children of `preNoticeAddImagePopup`?

doolse:
@ChristianMurphy Are you sure you aren't getting confused with GPL? We already use LGPL libraries in openEQUELLA (JPF for example).

Source/Plugins/Core/com.equella.core/js/tsrc/components/RichTextEditor.tsx
```diff
@@ -1,120 +1,106 @@
...
+require("tinymce/plugins/toc");
+require("tinymce/plugins/visualblocks");
+require("tinymce/plugins/visualchars");
+require("tinymce/plugins/wordcount");
 
 interface RichTextEditorProps {
   htmlInput: string;
   onStateChange(html: string): void;
-  imageUploadCallBack?(file: object): Promise<any>;
+  imageUploadCallBack?(file: any): AxiosPromise<imageReturnType>;
```
doolse: The `file` param should be typed.

Source/Plugins/Core/com.equella.core/js/tsrc/components/RichTextEditor.tsx
```diff
@@ -1,120 +1,106 @@
...
 }
 
-interface RichTextEditorState {
-  editorState: EditorState;
+interface imageReturnType {
```
doolse: The naming convention for interface's says this should start with a capital

autotest/Tests/src/test/scala/equellatests/pages/LoginNoticePage.scala
doolse: Sorry @SammyIsConfused you're absolutely correct. @edalex-ian the specificity comes from the element that it's being called on

ChristianMurphy:
> Are you sure you aren't getting confused with GPL?

@doolse not confused, LGPL is on the Category X list of banned licenses.
https://apache.org/legal/resolved.html#category-x
Screenshot_2019-03-21 ASF 3rd Party License Policy

> We already use LGPL libraries in openEQUELLA (JPF for example).

That's something that should be discussed and remediated.

autotest/Tests/src/test/scala/equellatests/pages/LoginNoticePage.scala
edalex-ian: Yup, but as I said, review the ones higher up where it's not called on an element. This was just the last findElement() call.

I raise this, as I've been on projects that have had tests runs suddenly jump from an hour to two or more because of lacking specificity. So yes, like this line, if the above can at least be limited to a parent (like @doolse mentioned) that'd do the trick. If we can't even do that, then oh well, at least we tried.

SammyIsConfused:
@ChristianMurphy Are you sure that Apache projects in this case means all Apache licensed products, and not official Apache projects?

edalex-ian:
@ChristianMurphy wrote:

> That's something that should be discussed and remediated.

Totally agree. It's on the CDM agenda for next week. So how's about we continue for now and discuss there? Maybe in prep for the CDM @cbeach47 can also check in with the usual legal people. As apparently back at the start of the project a green light was given for things like JPF, Hibernate, etc.

Sorry all for opening a can of worms.

SammyIsConfused:
Merging as discussed with @doolse and @edalex-ian in the office.

PR #793, opened 2019-03-25
Login Notice: Add schedule settings 

SammyIsConfused:
#660 

(PR redo after migration to TinyMCE)


Source/Plugins/Core/com.equella.core/js/tsrc/loginnotice/PreLoginNoticeConfigurator.tsx
```diff
@@ -22,6 +34,12 @@ interface PreLoginNoticeConfiguratorProps {
 interface PreLoginNoticeConfiguratorState {
   html: string;
   dbHtml: string;
+  scheduleType: ScheduleTypeSelection;
+  dbScheduleType: ScheduleTypeSelection;
+  startDate?: Date;
+  endDate?: Date;
+  dbStartDate?: Date;
+  dbEndDate?: Date;
```
doolse: I can vaguely remember discussing this before not sure if it was this pull request or another but instead of having a state which contains two copies of every field it would be better to simply use two instances of `PreLoginNotice`

PR #794, opened 2019-03-26
Login Notice: Preserve content of image files on upload

SammyIsConfused:
#660 
Uploaded images were getting converted to PNG. Now, they retain their original file type.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/loginnotice/LoginNoticeModule.ts
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/loginnotice/LoginNoticeModule.ts
@@ -77,9 +77,10 @@ export function clearPostLoginNotice(): AxiosPromise {
 }
 
 export function uploadPreLoginNoticeImage(file: any): AxiosPromise {
+  let imageBlob: Blob = file.blob();
   return axios.put(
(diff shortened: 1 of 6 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/loginnotice/LoginNoticeModule.ts
```diff
@@ -77,5 +77,9 @@ export function clearPostLoginNotice(): AxiosPromise {
 }
 
 export function uploadPreLoginNoticeImage(file: any): AxiosPromise {
-  return axios.put(PRE_LOGIN_NOTICE_IMAGE_API_URL + file.name, file);
+  return axios.put(
+    PRE_LOGIN_NOTICE_IMAGE_API_URL + file.filename(),
+    file.blob(),
+    { headers: { "content-type": "application/octet-stream" } }
```
doolse: Did you try using the blob's `type` property instead of "application/octet-stream"?

SammyIsConfused: Just tried it, it works. Once that commit is in I'll merge.

PR #801, opened 2019-03-29
Login notice: Finishing touches

SammyIsConfused:
#660 
- Reword language strings
- Fix padding for titles with a backroute
- Fix special character uploading for image file names
- Refactor PreLoginNotice state into objects using the spread operator
- Make Save button float and add a backroute
- Change display format of date picker


Source/Plugins/Core/com.equella.core/js/tsrc/loginnotice/LoginNoticeConfigPage.tsx
```diff
@@ -101,25 +106,39 @@ class LoginNoticeConfigPage extends React.Component<
           <PreLoginNoticeConfigurator
             handleError={this.handleError}
             notify={this.notify}
+            submit={this.state.submitPreNotice}
```
doolse: Instead of passing in a flag like this, there should be a method on each configurator for `save`ing and you can save the ref for the child component.

Source/Plugins/Core/com.equella.core/js/tsrc/loginnotice/LoginNoticeConfigPage.tsx
```diff
@@ -132,6 +151,19 @@ class LoginNoticeConfigPage extends React.Component<
           </Tabs>
         }
         errorResponse={this.state.error}
+        footer={
+          <Paper>
```
doolse: This isn't going to floating if it's on top of `Paper`

doolse: Also use a class instead of inline styles

Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/loginnotice/impl/LoginNoticeServiceImpl.java
```diff
@@ -124,15 +125,16 @@ public boolean isActive(PreLoginNotice preLoginNotice) {
   public String uploadPreLoginNoticeImage(InputStream imageFile, String name) throws IOException {
     checkPermissions();
     CustomisationFile customisationFile = new CustomisationFile();
-    String nameToUse = iterateImageNameIfDuplicateExists(name);
+    String nameToUse = iterateImageNameIfDuplicateExists(URLEncoder.encode(name, "UTF-8"));
```
doolse: Only encode the resulting URI, the filename can stay unencoded until then.

Source/Plugins/Core/com.equella.core/js/tsrc/loginnotice/LoginNoticeConfigPage.tsx
SammyIsConfused: 382eb22 Yep, thats a much neater way of doing it, thanks.

Source/Plugins/Core/com.equella.core/js/tsrc/loginnotice/LoginNoticeConfigPage.tsx
SammyIsConfused: ebfaea3
25ecdb2

Source/Plugins/Core/com.equella.core/js/tsrc/loginnotice/LoginNoticeConfigPage.tsx
```diff
@@ -25,8 +32,17 @@ interface LoginNoticeConfigPageState {
...
+    floatingButton: {
+      right: theme.spacing.unit * 2,
+      bottom: theme.spacing.unit * 2,
+      position: "fixed"
+    }
+  });
+
 class LoginNoticeConfigPage extends React.Component<
-  LoginNoticeConfigPageProps,
+  LoginNoticeConfigPageProps & WithStyles<typeof styles>,
```
PenghaiZhang: When declaring LoginNoticeConfigPageProps, we can extend it with 'WithStyles<typeof styles>', then we don't need to do a And Operation here and in constructor.

Source/Plugins/Core/com.equella.core/js/tsrc/loginnotice/LoginNoticeConfigPage.tsx
```diff
@@ -123,10 +132,14 @@ class LoginNoticeConfigPage extends React.Component<
   handleSubmitButton = () => {
     switch (this.state.selectedTab) {
       case 0:
-        this.setState({ submitPreNotice: !this.state.submitPreNotice });
+        if (this.preLoginNoticeConfigurator.current != null) {
```
PenghaiZhang: if (this.preLoginNoticeConfigurator.current) should work

SammyIsConfused: 3276bb2

Source/Plugins/Core/com.equella.core/js/tsrc/loginnotice/LoginNoticeConfigPage.tsx
SammyIsConfused: f4a63b9

Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/loginnotice/impl/LoginNoticeServiceImpl.java
doolse: We have now tested the encoding with % # (space) and they all worked :+1:

Issue #797, opened 2019-03-27
The New UI Theme Editor is only setting the secondary text colour, not the primary


edalex-ian:
**Describe the bug**

Currently the New UI theme editor only provides on setting for text colours: 'Text Colour'. Unfortunately, it turns out this is currently controlling the secondary colours for Material UI.

![Screenshot_2019-03-27 Theme Settings openEQUELLA](https://user-images.githubusercontent.com/43919233/55045448-eab94080-5091-11e9-8ed4-2643f80e36df.png)

**Expected behavior**

If we only have one control that should do the primary colour. However, it would be better to see this fixed with an additional control - i.e. one for primary and one for secondary.

**Platform:**

- OpenEquella Version: 2019.1 maybe 2018.2

**Additional context**

Not sure if this in 2018.2, as there's a chance it's come about only in our 2019.1 work while updating Material UI. Either way, we need to fix.

mrblippy:
Verified in 2019.1

PR #804, opened 2019-04-01
New UI Theme Editor: Fix color palette settings

SammyIsConfused:
#797


Source/Plugins/Core/com.equella.core/src/com/tle/web/api/newuitheme/impl/NewUITheme.java
```diff
@@ -21,9 +21,9 @@
   private String secondaryColor = "#ff9800";
   private String backgroundColor = "#fafafa";
   private String menuItemColor = "#ffffff";
-  private String menuItemTextColor = "#000000";
   private String menuItemIconColor = "#000000";
-  private String menuTextColor = "#000000";
+  private String primaryTextColor = "#000000";
+  private String secondaryTextColor = "#000000";
```
edalex-ian: I think the default is that normally the secondary text colour is a lighter colour than the primary. Indeed, some of the components only look right if this is the case.

edalex-ian:
Heya @SammyIsConfused , could we please have a screenshot of how the it looks with an additional option?

edalex-ian:
Also, what will this mean for someone who had the settings set in 2018.2?

SammyIsConfused:
image

Source/Plugins/Core/com.equella.core/src/com/tle/web/api/newuitheme/impl/NewUITheme.java
```diff
@@ -24,7 +24,7 @@
   private String menuItemIconColor = "#000000";
   private String menuItemTextColor = "#000000";
   private String primaryTextColor = "#000000";
-  private String menuTextColor = "#000000";
+  private String menuTextColor = "#444444";
```
PenghaiZhang: Instead of using a color hex code, I wonder if using Color palette of MUI would be more easily to understand?

SammyIsConfused: It would be more readable, however these values are used in the Color Picker components - which requires hex codes.

Issue #810, opened 2019-04-03
OAuth token servlet requires redirect_uri even when using client_credentials grant


doolse:
**Describe the bug**
The oEQ OAuth token servlet (`oauth/access_token`) makes the `redirect_uri` parameter mandatory even when being used with `client_credentials` grant. Also the error messages are broken language strings.

**To Reproduce**
An OAuth client using `client_credentials` grant doesn't supply the `redirect_uri`.

**Expected behavior**
Using the `client_credentials` grant should work without the `redirect_uri` parameter being specified. The error messages should work properly.

**Platform:**
- OpenEquella Version: [6.5 - 2019.1]

edalex-ian:
@doolse could you please advise where this issue's at?

Thanks.

doolse:
This is ready to test in 2018.2  and 2019.1. It probably not worth backporting as it was only an issue because of the Cloud provider work

edalex-ian:
For some reason I can't assign this to @mrblippy , but could you please test it? (Maybe self assign.)

mrblippy:
This is working in 2019.1, but not in 2018.2

SammyIsConfused:
Modified milestone to 2019.2, since it's working in 2019.1.

PR #811, opened 2019-04-03
Fix oauth redirect param and language strings

doolse:
#810


Source/Plugins/Core/com.equella.core/resources/lang/i18n-resource-centre.properties
```diff
@@ -2919,6 +2919,7 @@ oauth.clients.refuse.delete=This OAuth client application is referenced by a {0}
 oauth.error.accessdenied=User denied access
 oauth.error.clientcodemismatch=The client_id ({0}) does not match the client_id assigned to this code
 oauth.error.clientnotfound=No OAuth client can be found with the supplied client_id ({0}) and redirect_uri ({1})
+oauth.error.clientnotfoundnoredir=No OAuth client can be found with the supplied client_id ({0})
```
SammyIsConfused: There are plenty of long variable names in here, might as well call this `oath.error.clientnotfoundnoredirect` rather than abbreviating.

PR #812, opened 2019-04-03
Component/760 oauthtoken service

doolse:
In this pull request

* Modifications to existing `Cache` service to allow you to control the in-memory invalidation semantics
* An OAuth token service (`OAuthTokenService`) which requests and caches tokens (using the cached_value table)
* A Uri templating service which can be used to generate URLs to cloud provider services


Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/cloudproviders/CloudProviderService.scala
```diff
@@ -0,0 +1,84 @@
...
+      }
+      .getOrElse(dbLiftIO.liftIO(IO.raiseError(new Throwable("No OAuth service URL"))))
+  }
+
+  def proxyRequest(uuid: UUID, serviceId: String): DB[String] = {
+    CloudProviderDB
+      .get(uuid)
+      .flatMap { cp =>
+        val tempCP = cp.copy(
+          baseUrl = "http://doolse-sabre:8080/my/",
```
SammyIsConfused: is this hardcoded baseURL going to be an issue?

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/cloudproviders/package.scala
```diff
@@ -28,7 +28,7 @@ package object cloudproviders {
 
   case class Viewer(name: String, serviceId: String)
 
-  case class ServiceUri(authenticated: Boolean, uri: String)
+  case class ServiceUri(uri: String, authenticated: Boolean)
```
SammyIsConfused: Any reason why these parameters swapped places?

doolse: Yeah the uri is the more important thing

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/cloudproviders/CloudProviderService.scala
doolse: Yeah this is test data. This testing code is only here to keep pull request size down instead of waiting till the end when everything is done.

SammyIsConfused: Gotcha. Makes sense, thanks.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/cloudproviders/package.scala
SammyIsConfused: okay, cool!

Issue #758, opened 2019-03-05
Configure a cloud control


cathfitz:
* Display new cloud controls in the wizard control list in the Collection Definition Editor
* Add ability for an administrator to add a cloud control to a contribution wizard page
* Add ability for an administrator to edit a cloud control's configuration options

edalex-ian:
All work on this is now complete, ready for final testing. Over to you @mrblippy

PR #816, opened 2019-04-05
Configure cloud control

PenghaiZhang:
#758 
* Create dynamically generated UI to configure cloud controls.
* The configuration is supported by five basic UI components at this stage. They are MultiTargetChooser(XPath), textfield, combobox , checkbox and radiobutton.
* A warning is given to users if  UI components other than above five types are found.
* Saving configuration without filling in all mandatory fields results in a warning message and stopping saving.


PenghaiZhang:
Screenshot from 2019-04-05 14-08-45
Screenshot from 2019-04-05 14-09-01

Source/Plugins/Core/com.equella.admin/resources/lang/i18n-admin-console.properties
```diff
@@ -41,12 +41,15 @@
...
 /wizard.controls.timerange=This control should select a time range
 /wizard.controls.title=Title
+/wizard.controls.isrequired= *
 /wizard.controls.toplevel=Automatically select all top-level files
 /wizard.controls.uniqueness.dontshowagain=Do not show again 
 /wizard.controls.uniqueness.message=For duplication and uniqueness checking to perform correctly,\nplease ensure that each metadata target is "Indexed for Advanced\nSearching" on the chosen schema for this collection.
 /wizard.controls.uniqueness.ok=OK
 /wizard.controls.uniqueness.title=Note
 /wizard.controls.webdav=Allow WebDAV file management
+/wizard.cloudcontrol.validation.message= {0} is mandatory
```
doolse: Strings that start with '/' are global strings and shouldn't really be used as it's only for legacy code.

Source/Plugins/Core/com.equella.admin/src/com/tle/admin/controls/cloudcontrol/CloudControlEditor.java
```diff
@@ -0,0 +1,310 @@
...
+      String labelPosition = (configType.equals("XPath") ? BorderLayout.NORTH : BorderLayout.WEST);
+      configPanel.add(
+          new JLabel(
+              cloudControlConfig.name()
+                  + (cloudControlConfig.isConfigMandatory()
+                      ? CurrentLocale.get("wizard.controls.isrequired")
+                      : "")),
+          labelPosition);
+      switch (configType) {
+        case "Textfield":
```
doolse: I'd rather see each of the control types have their own classes instead of having it all in this one method.

Source/Plugins/Core/com.equella.admin/src/com/tle/admin/controls/cloudcontrol/CloudControlEditor.java
```diff
@@ -0,0 +1,310 @@
...
+
+          break;
+        default:
+          try {
+            configPanel.removeAll();
+            cloudControlModel.setErrorMessage(
+                CurrentLocale.get("wizard.cloudcontrol.unknownconfig.message", configType));
+            throw new EditorException(
+                CurrentLocale.get("wizard.cloudcontrol.unknownconfig.message"));
+          } catch (EditorException e) {
```
doolse: Why is this try catch here? I'm guessing you copied and pasted it from some other admin console code?

Source/Plugins/Core/com.equella.admin/src/com/tle/admin/controls/cloudcontrol/CloudControlEditor.java
```diff
@@ -0,0 +1,310 @@
...
+            e.printStackTrace();
```
SammyIsConfused: We should be using a logger here.

Source/Plugins/Core/com.equella.admin/src/com/tle/admin/controls/cloudcontrol/CloudControlEditor.java
```diff
@@ -0,0 +1,310 @@
...
+    subPanelLayout.setVgap(10);
+    configPanel.setLayout(subPanelLayout);
+    return configPanel;
+  }
+
+  private void renderTitle() {
+    JPanel titlePanel = createSubPanel();
+    JLabel titleLabel =
+        new JLabel(
+            CurrentLocale.get("wizard.controls.title")
```
doolse: We don't really want to use `CurrentLocale.get` directly, better to have a class which can look up these common strings

Source/Plugins/Core/com.equella.admin/resources/lang/i18n-admin-console.properties
```diff
@@ -41,12 +41,15 @@
 /wizard.controls.scanmetadata=Extract name and description from the web page
 /wizard.controls.timerange=This control should select a time range
 /wizard.controls.title=Title
+/wizard.controls.isrequired= *
```
SammyIsConfused: I'd probably recommend "required" as opposed to "isrequired".

Source/Plugins/Core/com.equella.admin/src/com/tle/admin/controls/cloudcontrol/CloudControlEditor.java
```diff
@@ -0,0 +1,310 @@
...
+                + CurrentLocale.get("wizard.controls.isrequired"));
+    title = new I18nTextField(BundleCache.getLanguages());
+    changeDetector.watch(title);
+    titlePanel.add(titleLabel, BorderLayout.WEST);
+    titlePanel.add(title, BorderLayout.CENTER);
+    mainPanel.add(titlePanel, gridBagConstraints);
+  }
+
+  private void renderBody() {
```
SammyIsConfused: This looks like it needs a refactor, there is a lot going on in each switch case. Perhaps a different method for each?

Source/Plugins/Core/com.equella.admin/src/com/tle/admin/controls/cloudcontrol/CloudControlEditor.java
PenghaiZhang: original idea was to popup a dialog that gives user a warning by raising this exception, but this seems to be redundant so I removed it.

PR #822, opened 2019-04-08
Add MIME-type checker to FileSystemService

SammyIsConfused:
#660 
MIME type as a wildcard was sufficient for modern browsers, but not for IE11. Added a method to get the MIME type from login notice images to fix in IE11.
image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/web/api/loginnotice/impl/PreLoginNoticeResourceImpl.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/web/api/loginnotice/impl/PreLoginNoticeResourceImpl.java
@@ -29,7 +29,6 @@ import java.io.InputStream;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 import javax.ws.rs.core.Context;
-import javax.ws.rs.core.MediaType;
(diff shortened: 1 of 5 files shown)


Source/Plugins/Core/com.equella.core/src/com/tle/web/api/loginnotice/impl/PreLoginNoticeResourceImpl.java
```diff
@@ -66,7 +65,7 @@ public Response deletePreLoginNotice() {
 
   @Override
   public Response getPreLoginNoticeImage(String name) throws IOException {
-    return Response.ok(noticeService.getPreLoginNoticeImage(name), MediaType.MEDIA_TYPE_WILDCARD)
+    return Response.ok(noticeService.getPreLoginNoticeImage(name), noticeService.getMimeType(name))
```
PenghaiZhang: just wondering what would happen if getMimeType returns null due to a unrecognised file

SammyIsConfused: Good catch. I'll replace it with a MimeType wildcard.

Issue #781, opened 2019-03-15
Failing to edit an item as a moderator


PenghaiZhang:
**Describe the bug**
This bug only occurs in new UI.
Moderators cannot edit an item when they are moderating

**To Reproduce**

To test the second bug,

1. Login in as a moderator,
2. Select a task to moderate,
3. Click Edit resource button,

Expected behavior
An error message pops up.

**Screenshots**
![Screenshot from 2019-03-15 16-35-02](https://user-images.githubusercontent.com/47203811/54411760-51f6fc80-4744-11e9-956c-efdc1ff40b3a.png)

**Platform:**

OpenEquella Version: 6.6, 2018.2
OS: Ubuntu
Browser Chrome

edalex-ian:
There is no strict requirement for this to be backported, and seeing the fix in 2019.1 was in an area of the Purescript that has been significant reworked it would require undue to prepare a different fix.

But it's all tested in `develop`, so closing now.

PR #828, opened 2019-04-09
Bugfix/781 failing to edit an item

PenghaiZhang:
#781 
This bugs happened when moderators click "edit resource", "cancel" and "save" when moderating an item.
All fixed now.

"info.forceredirect" doesn't work perfectly this time because the redirecting URL is as same as the current one, so REACT won't re-render UI. The solution is to do an identity comparison, to tell REACT that the redirecting URL is a different object though it has same value.


Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/command/EditInWizard.java
```diff
@@ -59,6 +61,13 @@ public void execute(SectionInfo info, WizardSectionInfo winfo, String data) thro
     for (WebWizardPage page : winfo.getWizardState().getPages()) {
       wizardService.ensureInitialisedPage(info, page, ps.getReloadFunction(), true);
     }
+    info.forceRedirect();
+  }
+
+  @Override
```
doolse: I think you may have left this in by accident?

Issue #757, opened 2019-03-05
Register a cloud provider


cathfitz:
* Provide the ability for an administrator to create (register) new and delete existing cloud providers
* Add ability to comment OAuth authentication to the service
* Add cloud provider detail storage
* Set access permission by creating a new MANAGE_CLOUD_PROVIDERS ACL

edalex-ian:
All work on this is now complete, ready for final testing. Over to you @mrblippy

PR #831, opened 2019-04-10
Validate the cloud provider registration URL

PenghaiZhang:
#757 
Validate the cloud provider registration URL.
After discussing with Charlie, the validation will only check if a URL starts with "http(s)://".

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/cloudprovider/CloudProviderAddDialog.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/cloudprovider/CloudProviderAddDialog.tsx
@@ -18,7 +18,6 @@ interface CloudProviderAddDialogProps {
 }
 interface CloudProviderAddDialogState {
   cloudProviderUrl: string;
-  isUrlValid: boolean;
 }
 
 export default class CloudProviderAddDialog extends React.Component<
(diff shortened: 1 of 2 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/cloudprovider/CloudProviderAddDialog.tsx
```diff
@@ -27,19 +28,22 @@ export default class CloudProviderAddDialog extends React.Component<
   constructor(props: CloudProviderAddDialogProps) {
     super(props);
     this.state = {
-      cloudProviderUrl: ""
+      cloudProviderUrl: "",
+      isUrlValid: true
```
doolse: isUrlValid doesn't need to be in the state, just a pure function

Source/Plugins/Core/com.equella.core/js/tsrc/cloudprovider/CloudProviderAddDialog.tsx
```diff
@@ -27,19 +28,22 @@ export default class CloudProviderAddDialog extends React.Component<
...
     };
   }
 
   handleTextChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    let url = e.target.value.toLowerCase();
```
doolse: Naughty lowercasing

PR #837, opened 2019-04-16
Feature/757 set permission for cloud provider

PenghaiZhang:
#757 
1. Add a new privilege called MANAGE_CLOUD_PROVIDER
2. Add a new node under the privilege tree to grant/revoke this privillege 
3. Accessing the page of cloud provider without this privilege will result in the list of cloud providers not displayed.
4. The add/delete buttons on cloud provider page will not work if users do not have this privilege.
Screenshot from 2019-04-16 16-14-45

### DIFF
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/cloudproviders/package.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/cloudproviders/package.scala
@@ -21,6 +21,9 @@ import java.util.UUID
 import com.tle.beans.Institution
 import com.tle.common.usermanagement.user.AbstractUserState
 import com.tle.common.usermanagement.user.valuebean.DefaultUserBean
+import com.tle.exceptions.PrivilegeRequiredException
+import com.tle.legacy.LegacyGuice
+import com.tle.web.cloudprovider.CloudProviderConstants
 
 package object cloudproviders {
 
@@ -68,5 +71,13 @@ package object cloudproviders {
 
     override def isSystem: Boolean = true
     override def getSessionID      = "PROVIDER"
+  }
+
+  def checkPermissions(): Unit = {
+    if (LegacyGuice.aclManager
+          .filterNonGrantedPrivileges(CloudProviderConstants.PRI_MANAGE_CLOUD_PROVIDER)
+          .isEmpty) {
+      throw new PrivilegeRequiredException(CloudProviderConstants.PRI_MANAGE_CLOUD_PROVIDER)
+    }
   }
 }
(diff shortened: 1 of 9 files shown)


Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/cloudproviders/package.scala
```diff
@@ -69,4 +72,12 @@ package object cloudproviders {
     override def isSystem: Boolean = true
     override def getSessionID      = "PROVIDER"
   }
+
+  def checkPermissions(): Unit = {
+    if (LegacyGuice.aclManager
```
SammyIsConfused: Since you're already using LegacyGuice here, you can inject your `CloudProviderSettingPrivilegeTreeProvider.java` to LegacyGuice and then the permissions checks can be simplified to
`if (CloudProviderSettingPrivilegeTreeProvider.isAuthorised())`
This one isn't a big issue. Both ways work, I just think it's slightly more concise that way.

Source/Plugins/Core/com.equella.core/src/com/tle/web/cloudprovider/CloudProviderSettingPrivilegeTreeProvider.java
```diff
@@ -0,0 +1,43 @@
...
+import com.tle.core.security.PrivilegeTreeProvider;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+
+@Bind
+public class CloudProviderSettingPrivilegeTreeProvider implements PrivilegeTreeProvider {
+
+  @Override
+  public void mapTargetIdsToNames(Collection<TargetId> targetIds, Map<TargetId, String> results) {}
```
SammyIsConfused: Is this empty override necessary?

PenghaiZhang: Yes, because this class implements PrivilegeTreeProvider.

SammyIsConfused: Okay! That's fine.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/cloudproviders/package.scala
PenghaiZhang: CloudProviderSettingPrivilegeTreeProvider actually cannot call isAuthorised  because it doesn't extend AbstractSettingsPrivilegeTreeProvider.

SammyIsConfused: Gotcha, I didn't realise that was System Setting specific. In that case, good to go :)

PR #842, opened 2019-04-17
Fix cloud provider ACL label text

PenghaiZhang:
Fix the label text of cloud provider ACL


PenghaiZhang:
Sam did codebuild again and passed this it is OK to merge into Develop.

Issue #744, opened 2019-02-07
Add static code analysis to build tooling


edalex-ian:
As part of the [2019/2 CDM](https://github.com/equella/Equella/wiki/CDM-2019-Feb) it was agreed to incorporate static code analysis tools into the oEQ build tooling - in place of the previous Code Climate.

To that end, this 'issue' acts as the starting point for that. The intention is that we add the following tools into the SBT process. And have it such that the various CI setups can call it (arguably very early on - fail quick) and return false/non-zero if checks fail.

The tools currently to integrate are:

- [ ] [checkstyle](http://checkstyle.sourceforge.net/) for Java
- [ ] [Scalastyle](http://www.scalastyle.org/) for Scala
- [ ] One of the linters for JS
- [ ] And if needs be, another linter for the Typescript (including TSX)
- [ ] For bonus points, something for Purescript

It would be great if we ended up with a simple `sbt codeCheck` at the end of it all.

This is open to contributions for one and all. Probably best if we do one PR per checker...

(And we also have a current discussion going on around Code Styles. So it was interesting to note that out of the box checkstyle has "example configuration files [...] supporting the Sun Code Conventions, Google Java Style".)

edalex-ian:
@edalex-joe is going to have a look at adding checkstyle.

edalex-ian:
As can be seen above, @PenghaiZhang is currently progressing with Java static analysis.

edalex-ian:
So we've got checkstyle in there nowadays. We still need to look at the others.

PR #863, opened 2019-05-01
Feature/744 add checkstyle to sbt

PenghaiZhang:
#744 
Add the plugin - sbt checkstyle into our project
* Create a new sbt command checkJavaCodeStyle, which is automatically executed after sbt compile
* A HTML report will be produced if checkJavaCodeStyle runs successfully. The report lives in the root target folder
* Currently, Java codes are checked based on Sun's coding conventions, and we can tailor what checks we need later on
* The HTML report template is provided by the plugin itself
* The version of checkstyle is updated to 8.19 in order to support checks used in Sun

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,6 @@ before_install:
   - npm install --global purescript psc-package
 install:
   - sbt checkJavaCodeStyle
-  - cat /home/travis/build/equella/Equella/checkstyle-config.xml
-  - cat /home/travis/build/equella/Equella/target/checkstyle-report.html
-  - cat /home/travis/build/equella/Equella/target/checkstyle-report.xml
   - git fetch --unshallow
   - npm install && npm run check
   - sbt headerCheck
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,11 +12,11 @@ before_install:
   - sudo apt-get install -y awscli
   - npm install --global purescript psc-package
 install:
-  - sbt checkJavaCodeStyle
   - git fetch --unshallow
   - npm install && npm run check
   - sbt headerCheck
 script:
+  - sbt checkJavaCodeStyle
   - sbt test installerZip writeLanguagePack writeScriptingJavadoc
 after_failure:
   - aws s3 --region us-east-1 cp target/checkstyle-report.html s3://edalexdev/equella_artifacts/$TRAVIS_BRANCH/$TRAVIS_BUILD_NUMBER/
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ before_install:
   - npm install --global purescript psc-package
 install:
   - sbt checkJavaCodeStyle
+  - cat /home/travis/build/equella/Equella/target/checkstyle-report.html
   - git fetch --unshallow
   - npm install && npm run check
   - sbt headerCheck
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,11 +12,11 @@ before_install:
   - sudo apt-get install -y awscli
   - npm install --global purescript psc-package
 install:
+  - sbt checkJavaCodeStyle
   - git fetch --unshallow
   - npm install && npm run check
   - sbt headerCheck
 script:
-  - sbt checkJavaCodeStyle || travis_terminate 1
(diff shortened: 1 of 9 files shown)


build.sbt
```diff
@@ -63,6 +64,35 @@ lazy val equella = (project in file("."))
...
+    Some(Set(CheckstyleXSLTSettings(htmlTemplateFile, outputHtmlFile)))
+  // We don't want to exit SBT if checkstyle finds any issue, so severityLevel should be None
+  val severityLevel = None
+  Checkstyle.checkstyle(sourceFile,
+                        resourceFile,
+                        outputXmlFile,
+                        configFile,
+                        xsltTransformationSetting,
+                        severityLevel,
+                        streams.value)
```
edalex-ian: Seeing all the above are single use, probably would be better to go with named arguments.

build.sbt
```diff
@@ -63,6 +64,35 @@ lazy val equella = (project in file("."))
...
+}
+
+// Make checkJavaCodeStyle executed after compile is done
+(checkJavaCodeStyle in Compile) := (checkJavaCodeStyle in Compile)
+  .triggeredBy(compile in Compile)
+  .value
```
edalex-ian: Normally when I've used static analysis tools I've run them pre-compile - as no point going through the slow compile if there are issues. Can we have it before compile?

PenghaiZhang:
Finally work out the threshold is 870, which means if checkstyle finds more than 870 errors it will fail the Travis build and send the report to S3.

Screenshot from 2019-05-08 15-45-30
Screenshot from 2019-05-08 15-45-54

.travis.yml
```diff
@@ -16,7 +16,10 @@ install:
   - npm install && npm run check
   - sbt headerCheck
 script:
+  - sbt checkJavaCodeStyle
   - sbt test installerZip writeLanguagePack writeScriptingJavadoc
+after_failure:
+  - aws s3 --region us-east-1 cp target/checkstyle-report.html s3://edalexdev/equella_artifacts/$TRAVIS_BRANCH/$TRAVIS_BUILD_NUMBER/
```
edalex-ian: There's more than one failure cases here, so sometimes we'd be saving a successful checkStyle report to S3 for now real reason. We should use shell logic handling instead. So rather than having this `after_failure` block, just have this up when we execute checkstyle:

    sbt checkJavaCodeStyle || aws s3 etc etc

.travis.yml
```diff
@@ -16,7 +16,10 @@ install:
   - npm install && npm run check
   - sbt headerCheck
 script:
+  - sbt checkJavaCodeStyle
```
edalex-ian: Let's move this into the `install` section so that if it fails it stops the build. Place it after the `headerCheck`. Also note the below command for handling if it fails.

build.sbt
```diff
@@ -63,6 +64,47 @@ lazy val equella = (project in file("."))
...
+    val outputFile = new File("target/checkstyle-report.xml")
+    if (outputFile.exists()) {
+      val report = scala.xml.XML.loadFile(outputFile)
+      (report \ "file").flatMap { file =>
+        (file \ "error").map { _ =>
+          {
+            1
+          }
+        }
+      }.sum
```
edalex-ian: This `map` then `sum` seems overly convoluted. Could just be a simple `reduce`.

build.sbt
```diff
@@ -63,6 +64,47 @@ lazy val equella = (project in file("."))
...
+    } else {
+      Int.MaxValue
```
edalex-ian: This should probably throw an exception (e.g. `FileNotFoundException`). Just returning a really high value is going to be misleading down below.

edalex-ian:
Sorry @PenghaiZhang , my bad there, you now also need to return false after uploading to s3. e.g.:

    sbt checkJavaStyle || (aws s3 cp ... ; false)

PenghaiZhang:
> Sorry @PenghaiZhang , my bad there, you now also need to return false after uploading to s3. e.g.:
> 
> ```
> sbt checkJavaStyle || (aws s3 cp ... ; false)
> ```

No worries, Ian. I also realised this before lunch time.

Issue #873, opened 2019-05-09
REST API for easier editing of attachments / metadata


doolse:
The new cloud provider control functionality added an API for doing the following while the wizard is open:
* Edit the XML metadata
* Add/Edit/Delete attachment(s) and manipulate their associated XML metadata node

It would be good if the existing Item REST APIs also had access to this same functionality. It should be located at:
`PUT /api/item/<UUID>/<version>/edit`

edalex-ian:
This is one of the items listed in #760

PR #874, opened 2019-05-09
Abstract the ItemEdits for use with existing API

doolse:
#873

### DIFF
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/item/ItemEdits.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/item/ItemEdits.scala
@@ -1,0 +1,99 @@
+/*
+ * Licensed to The Apereo Foundation under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * The Apereo Foundation licenses this file to you under the Apache License,
+ * Version 2.0, (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.tle.web.api.item
+
+import com.dytech.devlib.PropBagEx
+import com.fasterxml.jackson.annotation.JsonSubTypes.Type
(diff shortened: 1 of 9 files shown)


Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/item/ItemEdits.scala
```diff
@@ -0,0 +1,99 @@
...
+case class AddAttachmentResponse(attachment: EquellaAttachmentBean)  extends ItemEditResponse
+case class EditAttachmentResponse(attachment: EquellaAttachmentBean) extends ItemEditResponse
+case class DeleteAttachmentResponse(uuid: String)                    extends ItemEditResponse
+
+case class ItemEdits(xml: Option[String], edits: Iterable[ItemEditCommand])
+case class ItemEditResponses(xml: String, results: Iterable[ItemEditResponse])
+
+object ItemEdits {
+  val editorMap            = LegacyGuice.attachmentSerializerProvider.getAttachmentSerializers.asScala
+  val attachmentSerializer = LegacyGuice.attachmentSerializerProvider
```
SammyIsConfused: Couldn't you set `attachmentSerializer` first and then derive `editorMap` from that rather than referring to `LegacyGuice.attachmentSerializerProvider` twice?

Source/Plugins/Core/com.equella.core/src/com/tle/web/api/item/resource/EquellaItemResource.java
```diff
@@ -93,4 +95,29 @@ Response deleteComment(
       @ApiParam(APIDOC_ITEMUUID) @PathParam("uuid") String uuid,
       @ApiParam(APIDOC_ITEMVERSION) @PathParam("version") int version,
       @ApiParam(APIDOC_ITEMVERSION) @QueryParam("commentuuid") String commentUuid);
+
+  @PUT
+  @Path("/{uuid}/{version}/edit")
+  @Consumes("application/json")
+  @ApiOperation(value = "Edit attachments and metadata")
+  public ItemEditResponses editCommands(
+      // @formatter:off
```
SammyIsConfused: What is the purpose of the `// @formatter:off/on` lines? Is it to avoid the precommit hooks for this section? If so, why?

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/item/ItemEdits.scala
doolse: Yep :+1:

Source/Plugins/Core/com.equella.core/src/com/tle/web/api/item/resource/EquellaItemResource.java
doolse: `// @formatter:off` was meant to tell ide's / formatters not to reformat the code inside it but I just copied it from another API call it wasn't intentional, so i've just removed it.

Issue #88, opened 2017-05-26
import/export tool SBT conversion


emcgarrahan:

cbeach47:
Source is https://github.com/equella/Equella/tree/master/Source/Tools/ImportExportUtils

edalex-ian:
This has currently come to the attention of Edalex and so some preliminary work has been done on the branch: `bugfix/build-import-export`

We plan to continue work on it, as we have at least one client who needs it.

PR #879, opened 2019-05-10
Bugfix/build import export

PenghaiZhang:
#88 

* Move this project to a new gradle-based project which is a sub-project of Equella
* Create a new gradle command createImportExportTools to create two jars of the import and export tools
* Find these two jars from /Equella/import-export-tool/build/libs

* changes from build.sbt were made by Jolse before. Not sure if we want to remove them, or even the whole old project of import/export.

Screenshot from 2019-05-09 15-21-47
Screenshot from 2019-05-09 15-25-25


Source/Tools/ImportExportUtils/build.sbt
```diff
@@ -0,0 +1,33 @@
+import org.apache.axis2.wsdl.WSDL2Java
```
edalex-ian: So we probably don't need this file any more.

build.sbt
```diff
@@ -207,3 +207,6 @@ sources in (Compile, doc) := {
     +++ javadocSources((baseDirectory in LocalProject("com_equella_core")).value)).get
 }
 javacOptions in (Compile, doc) := Seq()
+
+lazy val importExport = (project in file("Source/Tools/ImportExportUtils"))
```
edalex-ian: So seeing we wont have the SBT project there any more, should we revisit this? Can we make it trigger gradle? Actually, I also wonder if we even need it in the root project. :thinking:

import-export-tool/build.gradle
```diff
@@ -0,0 +1,103 @@
...
+      srcDir 'build/codegen'
+      srcDir 'src/main/java'
+    }
+  }
+}
+
+allprojects {
+  apply plugin: "java"
+  sourceCompatibility = 1.8
+  version = '1.0.0'
```
edalex-ian: Is that the right version? What was the version of the previous one - e.g. the one Charlie had access to, or the public have access to.

import-export-tool/build.gradle
```diff
@@ -0,0 +1,103 @@
...
+  from {
+    configurations.runtimeClasspath.collect { it.isDirectory() ? it : zipTree(it) }
+  }
+  with jar
+}
+
+task createImportExportTools {
+  dependsOn(importToolJar).dependsOn(exportToolJar).dependsOn(compileJava)
+  doLast {
+    println "import and export tools are created under $projectDir/build/libs"
```
edalex-ian:
```suggestion
    println "Import and export tools are created under $projectDir/build/libs"
```

import-export-tool/build.gradle
```diff
@@ -0,0 +1,103 @@
...
+      'Implementation-Version': version,
+    )
+  }
+  from {
+    configurations.runtimeClasspath.collect { it.isDirectory() ? it : zipTree(it) }
+  }
+  with jar
+}
+
+task createImportExportTools {
```
edalex-ian: I wonder if we really need this task. Should we modify the standard `assemble` or `build` task to run the `importToolJar` and `exportToolJar`? Or, at least have `assemble` perhaps `dependOn` this task.

Either way, that'd give users the standard approach of coming into this directory and simply doing:

    ./gradlew assemble

import-export-tool/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.10-bin.zip
```
edalex-ian: This is a bit out of date, we're already up to Gradle 5.4.1.

import-export-tool/build.gradle
PenghaiZhang: Yes we can make this task executed after build, which means uses just type ./gradlew build

import-export-tool/build.gradle
PenghaiZhang: Just asked Charlie and he is also uncertain about this. He suggests we start from 1.0.0 and we keep maintaining from now.

import-export-tool/build.gradle
PenghaiZhang: ./gradlew assemble now can generate these two jars

import-export-tool/build.gradle
```diff
@@ -80,9 +81,9 @@ task exportToolJar(type: Jar) {
 }
 
 task createImportExportTools {
-  dependsOn(importToolJar).dependsOn(exportToolJar).dependsOn(compileJava)
+  dependsOn(importToolJar).dependsOn(exportToolJar)
```
edalex-ian: Let's move these two dependencies into the task definition, e.g.

    task createImportExportTools( dependsOn: [ importToolJar, exportToolJar ] ) 

or otherwise have a single line of:

    dependsOn( [ importToolJar, exportToolJar ] )

Issue #829, opened 2019-04-10
Rewrite New UI Layout in Typescript


edalex-ian:
**Is your feature request related to a problem? Please describe.**

Currently the top level container/layout for the New UI and the supporting components are done in Purescript. It has been agreed that this poses an impediment to future maintenance of openEQUELLA due to the step required learning curve and minimal adoption of the language.
 
**Describe the solution you'd like**

It has been agreed to refactor this to Typescript, leaving Purescript to only be used for other existing pages/components - and possibly also later refactored.

**Additional context**

This aligns also with the new direction of no new code being done in Purescript.

Creation of this ticket (and ideally actioning it) came out of the recent Community Developer Meeting action items.

**Testing**

* Nav links change when you login/logout
* Menu links work (logout, user prefs if you have them)
* Tasks and notifications work
* Leaving a page which has changes forces you to confirm leaving (course editor, contribution)

edalex-ian:
@doolse has highlighted that the effort to do the router is also significant. So first the layout/template will be done and bedded in. Then part two will be doing the router.

PR #920, opened 2019-05-16
Typescript template

doolse:
#829 
Full conversion of template which should have zero difference from the purescript one


edalex-ian:
Looks like it broke all the login notice tests. Please fix those first.

Following that, then at least we can also see it run the rest of the test suite.

edalex-ian:
Being this is a bit of a key area (as is the routing rework that's coming too) be great if we can have a few review it.

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/Template.tsx
```diff
@@ -0,0 +1,563 @@
...
+  preventNavigation?: boolean;
+  /* An optional Route for showing a back icon button */
+  backRoute?: Route;
+  /* Markup to show at the bottom of the main area. E.g. save/cancel options */
+  footer?: React.ReactNode;
+  /* Unexpected errors can be displayed by setting this property */
+  errorResponse?: ErrorResponse;
+  fullscreenMode?: string;
+  hideAppBar?: boolean;
+  menuMode?: string;
```
abidingotter: Could also use an enum string type.

PR #930, opened 2019-05-22
Allow new UI in IE11 to change CSS

SammyIsConfused:
Resolves an issue in which switching to "Videos" or "Images" in the search tab would break the styles upon moving back to "Standard" ( which looked like this)
image
Now working as intended.
image
Also fixed an issue in which the menu icons would not render in IE11.
Also fixed an issue which would prevent the login notice editor from rendering in IE11.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/src/UI/LegacyContent.js
+++ b/Source/Plugins/Core/com.equella.core/js/src/UI/LegacyContent.js
@@ -1,6 +1,7 @@
-exports.resolveUrl = function(url) {
+function resolveUrl(url) {
   return new URL(url, $("base").attr("href")).href;
-};
+}
+exports.resolveUrl = resolveUrl;
 
 exports.setInnerHtml = function(props) {
   return function() {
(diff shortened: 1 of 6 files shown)


Source/Plugins/Core/com.equella.core/js/src/UI/LegacyContent.js
```diff
@@ -105,3 +106,44 @@ exports.setupLegacyHooks_ = function(ps) {
...
+          delete existingSheets[cssUrl];
+        } else {
+          var newCss = doc.createElement("link");
+          newCss.rel = "stylesheet";
+          newCss.href = cssUrl;
+          head.insertBefore(newCss, insertPoint);
+          lastLink = newCss;
+        }
+        return lastLink;
+      }, null);
```
PenghaiZhang: I think a blank line here can help readability such as quickly locating the opening bracket

Source/Plugins/Core/com.equella.core/js/src/UI/LegacyContent.js
```diff
@@ -105,3 +106,44 @@ exports.setupLegacyHooks_ = function(ps) {
...
+      const deleteSheets = function() {
+        for (key in existingSheets) {
+          head.removeChild(existingSheets[key]);
+        }
+      };
+      if (!lastSheet) onSuccess(deleteSheets);
+      else {
```
PenghaiZhang: Maybe I am wrong, but I don't think this else block can be executed, even doubt it can pass compile?? The reason is the if statement in above line has no curly braces, which means nothing will be done if the condition is not satisfied, and this 'else' will be believed that it has no corresponding 'if'.

Source/Plugins/Core/com.equella.core/js/src/UI/LegacyContent.js
doolse: The formatting is handled by `prettier`, so if you don't like it you'll have to take it up with them :)

Source/Plugins/Core/com.equella.core/js/src/UI/LegacyContent.js
doolse: An if statement doesn't require braces for either of the statements, it's the same as java.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/template/RenderNewTemplate.scala
```diff
@@ -57,6 +57,7 @@ object RenderNewTemplate {
         s"api/language/bundle/${LocaleLookup.selectLocale.getLocale.toLanguageTag}/bundle.js")
         .preRender(info)
       new IncludeFile(s"api/theme/theme.js").preRender(info)
+      RenderTemplate.STYLES_CSS.preRender(info)
```
doolse: This actually breaks the new search UI querybar because the default old stylesheet has to be available to be removed, if  it's included here it can't be removed.

SammyIsConfused: 6f076a9

PR #932, opened 2019-05-22
Fix Import/Export tools not running with JAVA 10

PenghaiZhang:
One dependency required in the two tools is the package javax.xml.soap, which is no longer available from Java 9. So we need to add this dependency through Gradle, in order for users who use Java 9 or higher version to get them running properly.


edalex-ian:
Go for merge! :+1:

PR #936, opened 2019-05-23
Add IE11 xpath support

SammyIsConfused:
Prevents IE11 crash on contribution wizard when cloudcontrol is present.


Source/Plugins/Core/com.equella.core/js/tsrc/cloudcontrol/index.ts
```diff
@@ -98,6 +99,7 @@ function encodeFilepath(filepath: string): string {
 }
 
 const observer = new MutationObserver(function() {
+  wgxpath.install(window);
```
doolse: I'm fairly sure you shouldn't be `install()`ing this every single mutation, it should only be done once straight after the `require()`

SammyIsConfused: e7ac0ff

PR #948, opened 2019-05-24
Feature/783 remove jnlp

abidingotter:
Remove JNLP download servlet and anything else JNLP+console related.  The link in the settings page now goes to the GitHub releases page for the admin console.


edalex-ian:
That's a good question @ChristianMurphy .

@cathfitz do you know of any doco? If there is none, I wonder if we should add some publicly available?

PenghaiZhang:
I got a 404 error when clicking the downloading link.  I wonder if this is caused by the new template or route stuff. I am fixing this.

cathfitz:
It would be good to inform the user of the change in access to the Admin Console on clicking the Admin Console link from the Settings page. Something like:

The Administration Console is no longer accessed from this link. The Administration Console Package must be downloaded and configured on your system. Once installed, the launcher file is then used to open the openEQUELLA Administration Console Launcher dialog to open the Admin Console.

Can we make the work 'downloaded' the link to the download page?

Issue #983, opened 2019-06-11
The Navigation Builder does not save edits in new UI


cathfitz:
**Describe the bug**
1. When using the Navigation Builder wizard control to set up a structure for displaying a group of attachments, the changes are not saved when the new UI template is switched on (in 2018.2 and 2019.1). 

**To Reproduce the first part (Nav Builder structure not save with new UI switched on)**
Add a Navigation Builder page to a Contribution Wizard via the Admin Console. Once set up:

1. Make sure the new UI template is switched on
2. Contribute an item to the collection with the Nav Builder page - make sure to add multiple attachments.
2. Go to the Nav Builder page and set up a structure
3. Save the item
4. Edit the item and go to the Nav Builder page - the structure has not been saved.

cathfitz:
Tested and fixed.

PR #995, opened 2019-06-14
Fix the Navigation builder in the new UI

doolse:
#983

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ cache:
     - "$HOME/.cache"
     - "$HOME/.sbt"
 jdk:
-  - oraclejdk8
+  - openjdk8
 before_install:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -5,7 +5,7 @@ cache:
     - "$HOME/.cache"
     - "$HOME/.sbt"
 jdk:
-  - oraclejdk8
+  - openjdk8
```
abidingotter: Is this meant to be in the PR?

abidingotter:
Looks fine except for the one issue.

.travis.yml
doolse: yeah it was, not exactly related but we needed to test openjdk8 on travis and this branch was the sacrificial lamb

PR #1009, opened 2019-06-18
Add German stemming

PenghaiZhang:
* Add a new property which enables users to specify what language stemming to use

* Add German stemming search if the value of above property is German

### DIFF
--- a/Installer/data/server/learningedge-config/optional-config.properties
+++ b/Installer/data/server/learningedge-config/optional-config.properties
@@ -65,4 +65,4 @@ configurationService.proxyPassword = ${proxy/pass}
 libav.path = ${libav/path#t\/}
 
 # Uncomment and specify the stemming language
-#freetext.analyzer.language = English
+#freetext.analyzer.language = en
--- a/Installer/data/server/learningedge-config/optional-config.properties
+++ b/Installer/data/server/learningedge-config/optional-config.properties
@@ -63,3 +63,6 @@ configurationService.proxyPassword = ${proxy/pass}
 
 # Libav path. For example C:/Libav/usr/bin  
 libav.path = ${libav/path#t\/}
+
+# Specify what language stemming to use
+freetext.stemming.language = english
--- a/Installer/data/server/learningedge-config/optional-config.properties
+++ b/Installer/data/server/learningedge-config/optional-config.properties
@@ -65,4 +65,4 @@ configurationService.proxyPassword = ${proxy/pass}
 libav.path = ${libav/path#t\/}
 
 # Uncomment and specify the stemming language
-#freetext.stemming.language = English
+#freetext.analyzer.language = English
--- a/Installer/data/server/learningedge-config/optional-config.properties
+++ b/Installer/data/server/learningedge-config/optional-config.properties
@@ -64,5 +64,5 @@ configurationService.proxyPassword = ${proxy/pass}
 # Libav path. For example C:/Libav/usr/bin  
 libav.path = ${libav/path#t\/}
 
-# Specify what language stemming to use
-freetext.stemming.language = english
+# Uncomment and specify the stemming language
+#freetext.stemming.language = English
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/core/freetext/index/AbstractIndexEngine.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/core/freetext/index/AbstractIndexEngine.java
@@ -26,7 +26,11 @@ import com.tle.freetext.TLEAnalyzer;
 import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
 import java.util.Timer;
 import java.util.TimerTask;
 import javax.annotation.PostConstruct;
@@ -34,6 +38,7 @@ import org.apache.log4j.Logger;
 import org.apache.lucene.analysis.Analyzer;
 import org.apache.lucene.analysis.CharArraySet;
 import org.apache.lucene.analysis.PerFieldAnalyzerWrapper;
+import org.apache.lucene.analysis.ReusableAnalyzerBase;
 import org.apache.lucene.analysis.WordlistLoader;
 import org.apache.lucene.index.IndexWriter;
 import org.apache.lucene.index.IndexWriterConfig;
@@ -42,6 +47,8 @@ import org.apache.lucene.search.NRTManager;
 import org.apache.lucene.search.NRTManager.TrackingIndexWriter;
 import org.apache.lucene.search.NRTManagerReopenThread;
 import org.apache.lucene.store.FSDirectory;
+import org.apache.lucene.util.Version;
+import org.reflections.Reflections;
 
 /**
  * When Lucene creates an IndexSearcher it keeps a snap-shot of the index at that point in time, and
@@ -76,7 +83,6 @@ import org.apache.lucene.store.FSDirectory;
 @SuppressWarnings("nls")
 public abstract class AbstractIndexEngine {
   protected final Logger LOGGER = Logger.getLogger(getClass());
-
   private File indexPath;
   private PerFieldAnalyzerWrapper analyzer = null;
   private File stopWordsFile;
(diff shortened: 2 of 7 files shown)


Installer/data/server/learningedge-config/optional-config.properties
```diff
@@ -63,3 +63,6 @@ configurationService.proxyPassword = ${proxy/pass}
 
 # Libav path. For example C:/Libav/usr/bin  
 libav.path = ${libav/path#t\/}
+
+# Specify what language stemming to use
+freetext.stemming.language = english
```
abidingotter: Instead of "english" and "german" could we use the codes listed here:
https://www.oracle.com/technetwork/java/javase/javase7locales-334809.html
You could treat anything starting with de_ as generic German.
It would open up the possibility of including different dialects later on, and it's a list of known code. "german" and "english"   are kind of magic values at the moment.

PenghaiZhang: Yeah this is a good point! Let me discuss with @edalex-ian . Thanks Aaron.

PenghaiZhang:
image
A testing screenshot.

PenghaiZhang:
For French 
image


For Chinese 
image

Installer/data/server/learningedge-config/optional-config.properties
edalex-ian: hmmm, I see the idea and it would kind of align with other things... However, it'd mean we need to maintain a map to look up the Language in the Lucene Analyzers, and currently they only support languages not dialects.

Actually, in some ways I think we even go more raw here and specify a FQN for a class so that people could even place their own analyzers in the class path. But... That's a bit OTT.

So to start with, I'm happy to stick with English, German, French, etc.

Installer/data/server/learningedge-config/optional-config.properties
```diff
@@ -63,3 +63,6 @@ configurationService.proxyPassword = ${proxy/pass}
 
 # Libav path. For example C:/Libav/usr/bin  
 libav.path = ${libav/path#t\/}
+
+# Uncomment and specify the stemming language
+#freetext.stemming.language = English
```
edalex-ian:
```suggestion
#freetext.analyzer.language = English
```

Source/Plugins/Core/com.equella.core/src/com/tle/core/freetext/index/AbstractIndexEngine.java
```diff
@@ -199,9 +210,48 @@ protected Analyzer getAnalyser() {
...
+                languageAnalyzer.get().getDeclaredMethod("getDefaultStopSet");
+            nonStemmedAnalyzer =
+                new TLEAnalyzer(
+                    new CharArraySet(Version.LUCENE_36, (Set) getDefaultStopSet.invoke(null), true),
+                    false);
+          } catch (InstantiationException
+              | NoSuchMethodException
+              | InvocationTargetException
+              | IllegalAccessException e) {
+            // For analyzers that don't have constructors or the getDefaultStopSet method
```
edalex-ian: We should log a WARNing here.

Source/Plugins/Core/com.equella.core/src/com/tle/core/freetext/index/AbstractIndexEngine.java
```diff
@@ -199,9 +210,48 @@ protected Analyzer getAnalyser() {
...
+                .findFirst();
+
+        if (languageAnalyzer.isPresent()) {
+          try {
+            normalAnalyzer =
+                languageAnalyzer
+                    .get()
+                    .getDeclaredConstructor(Version.class)
+                    .newInstance(Version.LUCENE_36);
+            // For the non-stemmed one, we can reuse TLEAnalyzer
```
edalex-ian: I'm not sure I'm groking the below, but assuming I am, should it read:

    // For the non-stemmed analyzer we still use the TLEAnalyzer, however we use the language specific stop words
    // by loading them from the language specific analyzer.

???

PenghaiZhang:
> Looks good, thanks @PenghaiZhang !

My pleasure Christian !

Installer/data/server/learningedge-config/optional-config.properties
ChristianMurphy: > hmmm, I see the idea and it would kind of align with other things... However, it'd mean we need to maintain a map to look up the Language in the Lucene Analyzers, and currently they only support languages not dialects

The Java locales are BCP-47 based.
https://www.ietf.org/rfc/bcp/bcp47.txt

BCP-47 describes 3 a language in 3 part.
A language tag (E.G. `en`, `de`, `es`)
A script (character set) tag (E.G. `ja-Kana`).
And a region tag (E.G. `es-ES`)

For Equella's use case just the language tag could be a good fit, and it aligns with how Lucene scopes the stemmers.

edalex-ian: Yup, totes get it, but the issue is then the application of these. That is, the class we're trying to dynamically load is in the form of `<Language>Analyzer`. E.g. `EnglishAnalyzer`, `GermanAnalyzer`, `FrenchAnalyzer` etc.

But that said, you mention that "Lucene scopes the stemmers" using the language tag. I guess by this you mean they group their stemmers and analyzers in packages of the form `org.apache.lucene.analysis.<tag>`. And from our stroll through these packages it does seem to only have one analyzer for package. So arguably we could introspect the package for classes ending with `Analyzer` and load the first. (And then maybe at least log the name of the one we're using for if we have to debug things.)

But it all just feels less explicit. :thinking: 

The other option is to maintain a map between tags and Analyzers like I mentioned above. And we don't want to do that.

(P.S. And yes, you'll note I've focused here on Analyzers, that's because using the stemmers alone would've been insufficient (the issue title/request was too specific) - as then what would users do for their stop words. But interestingly note that for some language tag packages there's  several stemmers, and for some none.)

ChristianMurphy: Gotcha, makes sense :+1:

Source/Plugins/Core/com.equella.core/src/com/tle/core/freetext/index/AbstractIndexEngine.java
```diff
@@ -216,15 +216,17 @@ protected Analyzer getAnalyser() {
...
+        Reflections reflections = new Reflections("org.apache.lucene.analysis." + analyzerLanguage);
         Set<Class<? extends ReusableAnalyzerBase>> languageAnalyzers =
             reflections.getSubTypesOf(ReusableAnalyzerBase.class);
         // Lucene3.6.2 provides one analyzer for each language supported so we can use findFirst
         Optional<Class<? extends ReusableAnalyzerBase>> languageAnalyzer =
             languageAnalyzers.stream()
-                .filter(stemFilter -> stemFilter.getName().contains(analyzerLanguage))
+                .filter(
+                    analyzerClassFullName ->
+                        analyzerClassFullName.getName().contains(analyzerLanguage))
```
edalex-ian: ummmm, so if we're searching in the package `org.apache.lucene.analysis.en` then all classes are going to contain `en` in their full name... So this seems kind of redundant.

I guess seeing that we believe there is only one Analyzer in each package, and above we're doing `getSubTypesOf()` could we then just:

1. Confirm that there is one (and only one) result; and then
2. Use that one result?

No need for streams and filtering.

Source/Plugins/Core/com.equella.core/src/com/tle/core/freetext/index/AbstractIndexEngine.java
```diff
@@ -199,9 +210,54 @@ protected Analyzer getAnalyser() {
...
+                    .getDeclaredConstructor(Version.class)
+                    .newInstance(Version.LUCENE_36);
+            // For the non-stemmed analyzer we still use the TLEAnalyzer, however we use the
+            // language specific stop words by loading them from the language specific analyzer.
+            Method getDefaultStopSet =
+                languageAnalyzer.get().getDeclaredMethod("getDefaultStopSet");
+            nonStemmedAnalyzer =
+                new TLEAnalyzer(
+                    new CharArraySet(Version.LUCENE_36, (Set) getDefaultStopSet.invoke(null), true),
+                    false);
```
edalex-ian: If we've got down here, we're now using a dynamically loaded analyzer. So could we finish this `try` block with a:

    LOGGER.info("Using Lucene analyzer: " + languageAnalyzer.get().getClass().getName());

PR #1029, opened 2019-06-21
Change the event that causes the Flickr search tree to be registered

SammyIsConfused:
Since this handler has it's own tree, there is a race condition for
removal of the tree between the Flickr control and InfoEventListener. So
a method was added that ensures it is added from the Flickr control.


PenghaiZhang:
Travis build failed due to oraclejdk 8 is not supported in Ubuntu Xenial.

edalex-ian:
@PenghaiZhang do you want to fix the 2018.2 branch directly then, and then we'll get @SammyIsConfused to rebase this branch.

PR #1040, opened 2019-06-25
Fix scroptions so they actually work

doolse:
* Needed to make sure they were inside the `<form>` element
* Also make sure they don't get unmounted so they retain their changes


Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyForm.tsx
```diff
@@ -0,0 +1,24 @@
+import * as React from "react";
```
ChristianMurphy:
```suggestion
import * as React, { ReactNode, Fragment } from "react";
```

SammyIsConfused:
> What is a `scoption`? Can we add this term to the oEQ glossary or rename it?

@ChristianMurphy 
`scroptions` is a playful in-house abbreviation for `screen options`, that has kinda stuck in the dev office as of late :) Definitely not an official term.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/LegacyContentApi.scala
```diff
@@ -558,6 +555,14 @@ class LegacyContentApi {
     }
   }
 
+  def renderScroptions(context: RenderContext): Option[String] = {
```
SammyIsConfused: Should we at least capitalise the O of Options?

ChristianMurphy:
> `scroptions` is a playful in-house abbreviation for `screen options`, that has kinda stuck in the dev office as of late :) Definitely not an official term.

It is fun to name a variable to sound like a Scorpion-Tarantula hybrid. :scorpion: :spider: :stuck_out_tongue_winking_eye: 
It may be helpful to camel case the term for clarity, `scrOptions`. :slightly_smiling_face:

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyForm.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export function LegacyForm(props: {
+  state: { [key: string]: string[] };
+  children: React.ReactNode;
+}) {
+  const { state, children } = props;
+  return (
+    <form name="eqForm" id="eqpageForm" onSubmit={e => e.preventDefault()}>
+      <div style={{ display: "none" }} className="_hiddenstate">
```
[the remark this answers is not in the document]
doolse: It's only done like this because this markup is what the old UI has and I dare not change it in case something in the old UI depends on it (unlikely but not worth the risk).

Issue #1022, opened 2019-06-20
Remove use of Joda Time


edalex-ian:
Java 8+ projects are recommended by the Joda project to convert to using the matching features no in `java.time`:

> The standard date and time classes prior to Java SE 8 are poor. By tackling this problem head-on, Joda-Time became the de facto standard date and time library for Java prior to Java SE 8. **Note that from Java SE 8 onwards, users are asked to migrate to java.time (JSR-310) - a core part of the JDK which replaces this project.**

From: https://www.joda.org/joda-time/

**Describe the solution you'd like**

Remove the joda-time dependency, and then update code to use the new `java.time` classes.

PenghaiZhang:
We currently have a few files using Joda Time. 
Links to these files:
https://github.com/apereo/openEQUELLA/blob/develop/Source/Plugins/Core/com.equella.core/src/com/tle/web/controls/youtube/YoutubeUtils.java#L50

https://github.com/apereo/openEQUELLA/blob/develop/Source/Plugins/Core/com.equella.core/src/com/tle/web/controls/youtube/YoutubeAttachmentSerializer.java#L59

https://github.com/apereo/openEQUELLA/blob/develop/Source/Plugins/Core/com.equella.core/src/com/tle/web/echo/EchoUtils.java#L55

PR #1041, opened 2019-06-25
Remove Joda time

PenghaiZhang:
PR #1022 
Joda Time was used in the wizard control of Youtube and Echo360.

Removing Joda from the Youtube one is simple, but doing that for  Echo360 is harder because we are unable to test it at this moment.


Source/Plugins/Core/com.equella.core/src/com/tle/web/echo/EchoUtils.java
```diff
@@ -51,29 +47,10 @@
...
-            .appendSeparator(ampersand)
-            .appendSeconds()
-            .appendSuffix(ds("second"), ds("seconds"))
-            .toFormatter();
-
-    return pf.print(period).trim();
+    LocalDateTime now = LocalDateTime.now();
+    LocalDateTime plus = now.plus(Duration.ofMillis(duration * 1000));
+    Duration interval = Duration.between(now, plus);
+    return YoutubeUtils.formatDuration(interval.toString());
```
PenghaiZhang: Current change just simply reuses the format used in the Youtube control for two reasons: 
1. It takes more efforts to convert the PeriodFormatter which Java 8 has no similar API provided.
2. According to Charlie, we can't test Echo360 at this moment. Without seeing how this really works, I wonder if we can rework the formatting later on.

edalex-ian: Seems fine. I actually wonder if what you've done now address the TODO up on line 48.

edalex-ian:
Aha! The fix you made for the autotest was indeed what I was alluding to when I said:

> I assume you've tested and ensured the resultant format is as it was before.

As that exact bit of code gave me pause. Lucky the autotest picked it up, but something to keep in mind next time. ;)

Go for merge.

PenghaiZhang:
> Aha! The fix you made for the autotest was indeed what I was alluding to when I said:
> 
> > I assume you've tested and ensured the resultant format is as it was before.
> 
> As that exact bit of code gave me pause. Lucky the autotest picked it up, but something to keep in mind next time. ;)
> 
> Go for merge.

Sorry about that. Now I know why I did not test it.  This bit of code gets executed when using Youtube API v2 which has been deprecated. So in my tests I never got this bit running. 
Apologise!

PR #1061, opened 2019-06-27
Do proper synchronization for wizard state updates

doolse:


SammyIsConfused:
Getting lots of TestNG fails on this one. Once that's fixed, good to go.

PR #991, opened 2019-06-13
Feature/specify travis dist

PenghaiZhang:
Specify travis dist as Ubuntu Trusty

PR #1064, opened 2019-06-27
Make sure directory is correct

PenghaiZhang:
We actually fixed this for Develop in PR #991, but did not cherry pick it to Release.


PenghaiZhang:
Although the Travis check for push failed due to Travis dependency issues, the other check passed and building results were copied to S3. So it's safe to merge.

PR #1079, opened 2019-07-01
Reset Docker image hash after Renovate updated it

edalex-joe:
Renovate automatically updated the pinned Docker hash on the openEQUELLA docker build, resulting in a broken build (couldn't find apt-get) - this reverses it (I haven't looked at disabling that behaviour in Renovate yet).


edalex-ian:
As discussed, we can close this one and address the other issue.

PR #1080, opened 2019-07-01
Add one config option to enable attachment duplicate check

PenghaiZhang:
* Add a checkbox to the file/link attachment config panel, to enable or disable the feature of attachment duplicate check
image


Source/Plugins/Core/com.equella.base/src/com/tle/common/wizard/controls/universal/handlers/LinkSettings.java
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ */
+
+package com.tle.common.wizard.controls.universal.handlers;
+
+import com.dytech.edge.wizard.beans.control.CustomControl;
+import com.tle.common.wizard.controls.universal.UniversalSettings;
+
+public class LinkSettings extends UniversalSettings {
+  private static final String DUPLICATIONCHECK = "LINK_DUPLICATIONCHECK";
```
SammyIsConfused: Maybe there should be an underscore between DUPLICATION and CHECK, since then it would be MACRO_CASE. Though, I see there are other variables formatted like this in FileUploadSettings.java, so perhaps its not worth changing them all. :thinking:

PenghaiZhang: Thanks Sam. I think we should change . It won't take too long.

PR #1082, opened 2019-07-02
Remove commons-compress from sbt

PenghaiZhang:
According to the discussion in PR #1036, the commons-compress library used in Equella is not the one declared in sbt. So remove the declaration.


edalex-ian:
Heya @PenghaiZhang 

Fine to just stick with your work over on the #1036 PR and branch. No need to create another. ;)

I'll close and delete this for now.

PR #1098, opened 2019-07-11
Fix textfield duplciates not allow saving item

PenghaiZhang:
A bit background here: when the text field duplicate check is enabled, a warning will be given if duplicates are found. Users must tick the checkbox in 'Duplicate data'  page in order to publish.  However, if the item includes attachments and one or more attachments are of type 'Web Pages', then users will not be able to save even though they tick the checkbox. 
Screenshot from 2019-07-11 16-27-08
image
image

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -39,6 +39,7 @@
     </JavaCodeStyleSettings>
     <ScalaCodeStyleSettings>
       <option name="FORMATTER" value="1" />
+      <option name="MULTILINE_STRING_SUPPORT" value="-1" />
     </ScalaCodeStyleSettings>
     <codeStyleSettings language="JAVA">
       <option name="KEEP_CONTROL_STATEMENT_IN_ONE_LINE" value="false" />
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -39,7 +39,6 @@
     </JavaCodeStyleSettings>
     <ScalaCodeStyleSettings>
       <option name="FORMATTER" value="1" />
-      <option name="MULTILINE_STRING_SUPPORT" value="-1" />
     </ScalaCodeStyleSettings>
     <codeStyleSettings language="JAVA">
       <option name="KEEP_CONTROL_STATEMENT_IN_ONE_LINE" value="false" />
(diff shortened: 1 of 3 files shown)


PenghaiZhang:
Normally, the Wizard validates controls first and then checks if the checkbox is ticked on or not. With 'Web Page' being selected, it checks the checkbox first and then validates. So the fix tries to ensure validating controls gets working first.

.idea/codeStyles/Project.xml
```diff
@@ -39,7 +39,6 @@
     </JavaCodeStyleSettings>
     <ScalaCodeStyleSettings>
       <option name="FORMATTER" value="1" />
-      <option name="MULTILINE_STRING_SUPPORT" value="-1" />
```
PenghaiZhang: I found my Intellij keeps removing this line no matter I modified Intellij Scala style setting,  reset this file or checkout it again...Can someone give me any idea?

PenghaiZhang:
Just feel I should add a bit more details here. 
Based on what I said above,  the reason why duplicate information will be overwritten is Wizard uses a hashmap to store duplicate information. Because the hashmap key is the value of a text field and value is an instance of DuplicateData, Wizard will overwrite this key-value pair if it validates fields again. 

DuplicateData has an important boolean property called "accepted" which is false by default and used to decide if a wizard page is valid or not.  Ticking the checkbox will turn  "accepted" to be true.

Hope the above explanation helps a bit!

Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/section/DuplicateDataSection.java
```diff
@@ -69,7 +69,7 @@ public String getDefaultPropertyName() {
     return PROP_NAME;
   }
 
-  @DirectEvent(priority = SectionEvent.PRIORITY_AFTER_EVENTS)
+  @DirectEvent(priority = SectionEvent.PRIORITY_NORMAL)
```
abidingotter: Maybe change to:
```
// Must run after wizard validation (The event queued by PagesSection handleWizardState)
@DirectEvent(priority = SectionEvent.PRIORITY_AFTER_EVENTS - 1)
```

This would ensure it still runs before NORMAL (if it really needs to), but always after the duplicate validation

Issue #1118, opened 2019-07-17
Using nginx as proxy fails for uploading due to oEQ custom headers


doolse:
**Describe the bug**
Some front-end proxies don't like passing on headers with '_' in their name, in particular nginx requires a flag to enable them. See: https://stackoverflow.com/questions/22856136/why-underscores-are-forbidden-in-http-header-names

openEQUELLA uses two custom headers for file uploading: X_UUID and X_FILENAME. I propose we switch them to X-UUID and X-FILENAME to prevent any problems.

**To Reproduce**

1. Use nginx without `underscores_in_headers on;`
2. Importing an institution file
3. You will only get "Please select an institution import file"

**Platform:**

- OpenEquella Version: [e.g. 6.6, 2018.2, 2019.1]

PR #1119, opened 2019-07-17
Rename custom upload headers for better compatibility with frontend proxies

doolse:
#1118 
Targeted towards `develop` as there are workarounds available.

99.9% of the javascript changes are due to the prettier formatting them for the first time.


abidingotter:
@doolse Given that this is just nginx blocking perfectly valid headers, and that this will break any existing scripts reading the underscore value, should this be enabled by (dare I say it?) another config value?

doolse:
> @doolse Given that this is just nginx blocking perfectly valid headers, and that this will break any existing scripts reading the underscore value, should this be enabled by (dare I say it?) another config value?

Where would a script be using these values?

abidingotter:
> Where would a script be using these values?

They probably aren't, given it's undocumented.

abidingotter:
Let's squerge after 2019.2

PR #1127, opened 2019-07-22
Fix saving uuid as item title

PenghaiZhang:
This bug happens in some cases when hitting Save button the XML of an item is still not available .

### DIFF
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/section/PagesSection.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/section/PagesSection.java
@@ -122,7 +122,7 @@ public class PagesSection extends WizardSection<PagesSection.PagesModel>
     if (model.isSubmit()) {
       String xmlDoc = model.getXmldoc();
       if (xmlDoc != null) {
-        PropBagEx itemXml = state.getItemPack().getXml();
+        PropBagEx itemXml = state.getItemxml();
         PropBagEx cloudControlXml = new PropBagEx(xmlDoc);
         itemXml.mergeTree(cloudControlXml);
         state.setItemXml(itemXml.toString());
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/section/PagesSection.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/section/PagesSection.java
@@ -18,6 +18,7 @@
 
 package com.tle.web.wizard.section;
 
+import com.dytech.devlib.PropBagEx;
 import com.google.common.collect.Lists;
 import com.tle.annotation.NonNullByDefault;
 import com.tle.beans.entity.itemdef.Wizard;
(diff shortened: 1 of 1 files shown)


abidingotter:
Not sure this is the correct fix.  I mean, it may work, but I think the problem comes from the XML doc being out of date because the client script getState promise only runs once.
There is also still the problem of validation hooks not being unloaded, but that's a different issue :)

PenghaiZhang:
> Not sure this is the correct fix. I mean, it may work, but I think the problem comes from the XML doc being out of date because the client script getState promise only runs once.
> There is also still the problem of validation hooks not being unloaded, but that's a different issue :)

Yes, getState promise only runs once. So I try to get it running again inside the "presubmit". And then I found the real reason is the item's XML is not really ready when calling getState.
If you go to line 106 of WizardApi, you would see 'itemPack.getXml ' actually does not have the XML we need yet.

abidingotter:
Something I would test is:
1. Put a repeater on the first page and add some repeater items. 
2. Now remove a repeater item.  
3. Now save + publish. 
4. Check the item you removed is not in the XML

And:
1. Put some optional controls on page 2
2. Go to page 2 and fill in some values
3. Go back to page 1 
4. Go back to page 2 and put in some new values
5. Now save + publish. 
6. Check that the first set of values you put on page 2 are not in the XML.

Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/section/PagesSection.java
```diff
@@ -121,7 +122,10 @@ public void handleWizardState(SectionInfo info, final WizardState state) {
     if (model.isSubmit()) {
       String xmlDoc = model.getXmldoc();
       if (xmlDoc != null) {
-        state.setItemXml(xmlDoc);
+        PropBagEx itemXml = state.getItemPack().getXml();
```
abidingotter: state.getItemxml() should be sufficient

PenghaiZhang:
> Something I would test is:
> 
> 1. Put a repeater on the first page and add some repeater items.
> 2. Now remove a repeater item.
> 3. Now save + publish.
> 4. Check the item you removed is not in the XML
> 
> And:
> 
> 1. Put some optional controls on page 2
> 2. Go to page 2 and fill in some values
> 3. Go back to page 1
> 4. Go back to page 2 and put in some new values
> 5. Now save + publish.
> 6. Check that the first set of values you put on page 2 are not in the XML.

Thanks Aaron. Let me test these two cases now.

PenghaiZhang:
For the second test, looks good. 
Screenshot from 2019-07-23 10-05-24
Screenshot from 2019-07-23 10-05-36
Screenshot from 2019-07-23 10-05-59

PenghaiZhang:
For the Repeater test : 
Screenshot from 2019-07-23 10-21-13
Screenshot from 2019-07-23 10-18-46
Screenshot from 2019-07-23 10-18-33

PR #1134, opened 2019-07-24
Fix newUI same URL redirect bug

doolse:
So the problem with some actions not being triggered first go is because when the Purescript / Typescript conversion happened, the trigger for calling the legacy page submit API moved into React effect hook dependency list but it didn't take into account redirecting to the same URL. [LegacyPage.tsx](https://github.com/apereo/openEQUELLA/blob/develop/Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContent.tsx#L215)

In order to fix it properly we can use the "key" property of the react router location object, which will have a different value for each time a history location is pushed, thus it can trigger even with the same path and query.

The original fix is just a workaround which is checking for values very specific to the moderation page, however this problem will happen anywhere in oeq, so it's best to fix it properly.

I have run the new ui autotests on this and it fixed a number of tests.


doolse:
The page content update code now has it's own function.

PR #1136, opened 2019-07-25
link attachment duplicate check

PenghaiZhang:
* add the feature of link attachment duplicate check
* update duplicate information when editing, replacing or deleting a link attachment

Please note that the duplicate warning message and duplicate data page now can be affected by both file and link attachments.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/controls/universal/UniversalWebControlNew.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/controls/universal/UniversalWebControlNew.scala
@@ -258,33 +258,21 @@ class UniversalWebControlNew extends AbstractWebControl[UniversalWebControlModel
       val attachments         = dialog.getAttachments.asScala
       val state               = repository.getState
       val uploadedAttachments = dialog.getAttachments.size
+      val fileDuplicateCheckEnabled =
+        dialog.getControlConfiguration.getBooleanAttribute("FILE_DUPLICATION_CHECK")
+      val linkDuplicateCheckEnabled =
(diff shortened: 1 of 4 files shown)


Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/controls/universal/UniversalWebControlNew.scala
```diff
@@ -255,7 +255,47 @@ class UniversalWebControlNew extends AbstractWebControl[UniversalWebControlModel
...
     def validate(): Unit = {
+      val attachments         = dialog.getAttachments.asScala
+      val state               = repository.getState
       val uploadedAttachments = dialog.getAttachments.size
+
+      def fileDuplicateCheck(): Unit = {
+        val DuplicateCheckEnabled =
+          dialog.getControlConfiguration.getBooleanAttribute("FILE_DUPLICATION_CHECK")
+        if (DuplicateCheckEnabled) {
+          val fileAttachments =
```
doolse: The idiomatic way to do this kind of thing would be use scala's [pattern matching](https://docs.scala-lang.org/tour/pattern-matching.html), so you can replace the `filter`, `map` and `for` with:
```scala
attachments.collect {
  case fileAttachment : FileAttachment => wizardService.checkFileAttachmentDuplicate(state,
                                          fileAttachment.getFilename, fileAttachment.getUuid)
}
```

In fact in general there's never a time you would want to use `isInstanceOf` in scala.

PenghaiZhang: Thanks Jolse!! I will rework.

PR #1139, opened 2019-07-26
Fix the uuid title bug without mergeTree

PenghaiZhang:
The previous fix still breaks things due to the mergeTree not working properly.
This fix will just update part of the xml rather than do a merge.
Locally tested in singe/multiple pages with cloud control, I believe this fix is a better workaround.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/section/PagesSection.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/section/PagesSection.java
@@ -124,8 +124,24 @@ public class PagesSection extends WizardSection<PagesSection.PagesModel>
       if (xmlDoc != null) {
         PropBagEx itemXml = state.getItemxml();
         PropBagEx cloudControlXml = new PropBagEx(xmlDoc);
-        itemXml.mergeTree(cloudControlXml);
-        state.setItemXml(itemXml.toString());
+
+        /*
+         * Cloud control writes its metadata based on the item's xml.
+         * However in some cases (e.g. single page collection), at the moment when it is writing,
+         * the item's xml doesn't include the latest values of other controls in its node called <item>.
+         * So we must update this node, and don't change anything returned from cloud control.
+         */
+        final String ITEM_XML_NODE = "item";
+        PropBagEx oldItemNode = cloudControlXml.getSubtree(ITEM_XML_NODE);
+        PropBagEx newItemNode = itemXml.getSubtree(ITEM_XML_NODE);
+
+        if (oldItemNode != null) {
+          cloudControlXml.deleteSubtree(oldItemNode);
+        }
+        if (newItemNode != null) {
+          cloudControlXml.appendChildren(ITEM_XML_NODE, newItemNode);
+        }
+        state.setItemXml(cloudControlXml.toString());
       }
       info.queueEvent(
           new AbstractDirectEvent(SectionEvent.PRIORITY_AFTER_EVENTS, getSectionId()) {


PenghaiZhang:
The solution ideally should be provides cloud control with the latest xml when it is being validated. 
However, this probably requires a lot changes.

Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/section/PagesSection.java
```diff
@@ -124,8 +124,24 @@ public void handleWizardState(SectionInfo info, final WizardState state) {
...
-        itemXml.mergeTree(cloudControlXml);
-        state.setItemXml(itemXml.toString());
+
+        /*
+         * Cloud control writes its metadata based on the item's xml.
+         * However in some cases (e.g. single page collection), at the moment when it is writing,
+         * the item's xml doesn't include the latest values of other controls in its node called <item>.
+         * So we must update this node, and don't change anything returned from cloud control.
+         */
+        final String ITEM_XML_NODE = "item";
```
abidingotter: "item" is just part of the schema.  The node immediately under "xml" can be called anything.

SammyIsConfused: Ahh, hmm. So it should just be set to whatever the direct child of xml is, rather than item?

abidingotter: I don't even think it's the best solution.  :)  Would be better to have @doolse look at the client side script that is passing in stale XML

PenghaiZhang: No this is not a good solution. The best one should be providing the correct xml to cloud control when it is being validated. But anyway I will keep working on it..

doolse: @PenghaiZhang @abidingotter Can somebody please explain the problem trying to be fixed here? I'm happy to have a call about it..

PenghaiZhang: @doolse Hello Jolse. The issue is when you are contributing an item which includes cloud controls, the title of this item will be a uuid.  
To reproduce:
1. create a collection that has a cloud control
2. start contributing an item of this collection.
3. put any string as the value of title
4. save and publish

doolse: Ok seems straightforward, i'll take a look

PenghaiZhang: @doolse I guess there are two issues causing this problem.
1. When clicking Save, front-end doesn't get the new state (line 175-180 in cloudcontrol/index.ts)
2. the xml provided to cloud control doesn't include the latest values of other controls (line 94 of WizardApi.scala)

PenghaiZhang: if you just give a value to the title and do nothing else, and then save. You would see at this moment the itemXml is just </xml>. And then you cancel saving, and change the value of the title, and save again. Now you would see the xml includes the old title. 
If my explanation is not clear please grab me anytime. THank you!!

abidingotter: @doolse The problem stems from the cloud control validator.  If they edit the XML, they don't always get the current state first, and so the xmldoc passed to the server is stale and overwrites the session.  

Also, a side issue is that the validators are not unloaded.  Finish a contribution and go to contribute a new item by clicking on a collection name -> your validator is run even before the wizard loads up.  (My validator was looking for an element my control adds to the page, so it blew up and the contribution never started)

abidingotter: I believe it's because the getState promise is already resolved, and won't fire again.

doolse: > @doolse The problem stems from the cloud control validator. If they edit the XML, they don't always get the current state first, and so the xmldoc passed to the server is stale and overwrites the session.

The current xml should already be up to date, what circumstances are you seeing it not being up to date?

> Also, a side issue is that the validators are not unloaded. Finish a contribution and go to contribute a new item by clicking on a collection name -> your validator is run even before the wizard loads up. (My validator was looking for an element my control adds to the page, so it blew up and the contribution never started)

Are you calling `deregisterValidator(validator)` when your control is unmounted? If not the validator will run again.

I'm going to need someone to show me the problem as I can't reproduce it or see the xml not being up to date correctly. (After reverting the mergeTree stuff anyway)

doolse: > @doolse I guess there are two issues causing this problem.
> 
> 1. When clicking Save, front-end doesn't get the new state (line 175-180 in cloudcontrol/index.ts)

> 2. the xml provided to cloud control doesn't include the latest values of other controls (line 94 of WizardApi.scala)

This is true, you won't see the values from server side controls because they haven't actually been saved to the xml yet. You won't be able to rely on other controls on the page unless you have they do a reload (drop downs have this option).

abidingotter: > Are you calling deregisterValidator(validator) when your control is unmounted? If not the validator will run again.

I would have assumed that would be done by the cloud control API.  I could easily add code to remove the validator, but I think the API should be very forgiving of third party controls not cleaning up.

As for the original bug:
1. Make a 1 page wizard, containing a name field and a cloud control (the cloud control *must* have a validator that edits the XML)
2. Start a contribution in this collection.
3. Fill in the name field.
4. Click Save
5. Click Publish

In step 5, the cloud control API adds an xmldoc post parameter that contains essentially blank XML, and the server uses that as gospel.

abidingotter: Can't jump on a call today to show it to you though sorry.  I am working at 3TT today.

doolse: > > Are you calling deregisterValidator(validator) when your control is unmounted? If not the validator will run again.
> 
> I would have assumed that would be done by the cloud control API. I could easily add code to remove the validator, but I think the API should be very forgiving of third party controls not cleaning up.

If you're using react hooks you will get used to cleaning up properly :)

> As for the original bug:
> 
> 1. Make a 1 page wizard, containing a name field and a cloud control (the cloud control _must_ have a validator that edits the XML)
> 2. Start a contribution in this collection.
> 3. Fill in the name field.
> 4. Click Save
> 5. Click Publish
> 
> In step 5, the cloud control API adds an xmldoc post parameter that contains essentially blank XML, and the server uses that as gospel.

That is correct, the xml that is posted is what the server side controls will update with their changes. E.g. it's the xml BEFORE the title has been saved into it. After the server side controls have written into the xml any cloud controls will get told of the latest value of the xml (which should now have the title in it).

abidingotter: That's the problem.  The controls don't overwrite it, at least in the build I was testing.  The xmldoc is set into the wizard state, but the control writing stuff never happens.  I assume a forward happens that prevents it.

PenghaiZhang: Hi @doolse , are you able to reproduce this bug now ? 
I don't know the purpose of posting xml that is BEFORE saving, but what I found is xmlDoc will overwrite xml even though xmlDoc does not have the latest values, and then in the doSave method, old values will be saved.

For example,  the docXml may look like 
<xml><validated>true</validated><validated>true</validated></xml>
and item xml  may look like <xml><title>title</title></xml>.
After calling state.setItemXml, the item's xml is incorrect any more.

doolse: @PenghaiZhang Nope I was unable to reproduce it. The integration tester has a cloud control with a validator that edits the xml and it works fine with changes to the controls on the same page.

PenghaiZhang: I am also using integtester and its validator.  
image
image

doolse: @PenghaiZhang Can we have a call? I am doing the same thing and it works for me..

PenghaiZhang: @doolse ok

doolse: @PenghaiZhang Actually I just reproduced it :) I was clicking save, not publish.

PR #1140, opened 2019-07-26
Fix cloud provider / server side xml sync problems

doolse:
There are two problems which combined to cause the "uuid" as title problem:

* The cloud provider xml is only sent to the server when the "validate" global jquery event is called - the save button doesn't trigger this when opening the "publish" dialog, so the form fields need to be setup in the "presubmit" handler
* The cloud provider xml is being used when it shouldn't be - it should only be used inside the 
`SectionEvent.PRIORITY_AFTER_EVENTS` queued event, as this event get cancelled by the publish buttons.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/section/PagesSection.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/section/PagesSection.java
@@ -18,7 +18,6 @@
 
 package com.tle.web.wizard.section;
 
-import com.dytech.devlib.PropBagEx;
 import com.google.common.collect.Lists;
 import com.tle.annotation.NonNullByDefault;
 import com.tle.beans.entity.itemdef.Wizard;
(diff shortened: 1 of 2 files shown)


Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/section/PagesSection.java
```diff
@@ -120,17 +119,14 @@ public void handleWizardState(SectionInfo info, final WizardState state) {
...
-        state.setItemXml(itemXml.toString());
-      }
       info.queueEvent(
           new AbstractDirectEvent(SectionEvent.PRIORITY_AFTER_EVENTS, getSectionId()) {
             @Override
             public void fireDirect(SectionId sectionId, SectionInfo info) throws Exception {
+              String xmlDoc = model.getXmldoc();
+              if (xmlDoc != null) {
+                state.setItemXml(xmlDoc);
+              }
```
PenghaiZhang: I am not quite sure about this.  Before#1038, this bit of code was here in a synchronized block, and then it was moved to where it is now. And Now we are moving it back without the synchronized block. 

Please let me do some testing.

Thanks a lot @doolse

PenghaiZhang:
> There are two problems which combined to cause the "uuid" as title problem:
> 
> * The cloud provider xml is only sent to the server when the "validate" global jquery event is called - the save button doesn't trigger this when opening the "publish" dialog, so the form fields need to be setup in the "presubmit" handler
> * The cloud provider xml is being used when it shouldn't be - it should only be used inside the
>   `SectionEvent.PRIORITY_AFTER_EVENTS` queued event, as this event get cancelled by the publish buttons.

I think this PR works well.  I can't reproduce errors we already know.

My understanding is "validating cloud control in "presubmit" event, which gets fired when clicking Save or changing pages. This event firstly set xmlDoc item's xml, and then validating other fields and adding their latest values to item's xml.

PR #1141, opened 2019-07-30
Minimise docker image size and optimise layer order

edalex-ian:
There were two issues:

1. The docker image was weighing in at 2.06G (now 1.28G)
2. A new docker image for a new build resulted in almost all the layers
   having to be pulled - as they'd all change.

Large Docker Image

This was primarily caused by a double up of the layer which contained
the installed equella which was caused by the RUN statement after it had
been copied into the final equella image. But seeing this is a
multi-stage build with one stage dedicated to building / installing, it
made sense for it to completely finalise the installed equella directory
ready to copy into the final image.

To do this various logic was moved into the install 'stage' and effort
was also undertaken to minimise RUN statements.

Excessive layer churn

In docker each statement in a Dockerfile creates a layer. When
rebuilding an image, if a statement results in a changed layer then all
following layers will also change.

In the case of this build, the only statement which should really result
in a change to a layer is that with the new equella install - i.e. where
it copies from the `installer` image. So it is best to have this as late
as possible in the Dockerfile to ensure all previous layers can be
re-used as much as possible.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -6,42 +6,34 @@ FROM openjdk@sha256:0e25c8428a56e32861fe996b528a107933155c98fb2a9998a4a4e9423aad
 ENV LANG C.UTF-8
 ENV LC_ALL C.UTF-8
 
-RUN \
-  apt-get update \
-  && apt-get install -y \
+RUN useradd -ms /bin/bash equella \
+  && apt-get update \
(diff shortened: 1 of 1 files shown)


edalex-ian:
Codebuild failed due to flakey test on one build, but the other was successful.

PR #1143, opened 2019-08-01
Fix invalid <a/> tag in harvestlist.ftl, add robots meta tag to new UI

SammyIsConfused:
Allows DiscoverabilityTest - Harvest test to work in the new UI, and
also makes the harvest.do links appear as links in the new UI. Up 
until now, they were just text.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/resources/view/layouts/outer/react.ftl
+++ b/Source/Plugins/Core/com.equella.core/resources/view/layouts/outer/react.ftl
@@ -5,6 +5,7 @@
     <head>
       <meta http-equiv="X-UA-Compatible" content="IE=edge" >
     	<meta http-equiv="Content-Type" content="text/html;charset=UTF-8">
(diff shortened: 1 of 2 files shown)


PenghaiZhang:
Please hold on for a moment.

Source/Plugins/Core/com.equella.core/resources/view/layouts/outer/react.ftl
```diff
@@ -5,6 +5,7 @@
     <head>
       <meta http-equiv="X-UA-Compatible" content="IE=edge" >
     	<meta http-equiv="Content-Type" content="text/html;charset=UTF-8">
+      <meta name="robots" content="noindex, follow">
```
PenghaiZhang: This tag should be added into head as long as the page is harvest. However, it is everywhere now if new UI is turned on.

PenghaiZhang:
I think we can keep all changes at this stage as we need all test passed. We can come back later to fix these bugs.

PR #1145, opened 2019-08-02
Use the newly published 0.1.8 simpledba

doolse:


SammyIsConfused:
@doolse 
Any idea why the "circeJsonUnsafe" function isn't accessible from this one?

edalex-ian:
Thanks @doolse . CodeBuild is tripping on flakey tests now, but once I see that green we'll go for merge. :+1:

PR #1154, opened 2019-08-07
Make DRMPrivilegeTest pass in the new UI

SammyIsConfused:
There is a bug in the new UI which prevents navigating directly (via URL or fullscreen button) to an attachment that has DRM. In the old UI, the DRM agreement prompt would pop up, in the new UI it simply says that the item does not exist.


SammyIsConfused:
TestNG results confirm that this now passes in the new UI.

PR #1156, opened 2019-08-07
Make the getToken method wait for a token

SammyIsConfused:
Common issue with many of the rest tests in the new UI is the check for a token was called before the token existed.

### DIFF
--- a/autotest/OldTests/src/test/java/com/tle/webtests/test/webservices/rest/AbstractRestApiTest.java
+++ b/autotest/OldTests/src/test/java/com/tle/webtests/test/webservices/rest/AbstractRestApiTest.java
@@ -159,7 +159,11 @@ public abstract class AbstractRestApiTest extends AbstractSessionTest {
 
   protected String getToken() throws IOException {
     if (token == null) {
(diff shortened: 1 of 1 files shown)


autotest/OldTests/src/test/java/com/tle/webtests/test/webservices/rest/AbstractRestApiTest.java
```diff
@@ -158,6 +159,8 @@ protected void cleanupAfterClass() throws Exception {
   }
 
   protected String getToken() throws IOException {
+    LoginPage page = new LoginPage(context);
+    page.getWaiter().until((webDriver -> clients.size() >= 1));
```
edalex-ian: You shouldn't need to bring in the login page. Could you just use `WebDriverWait`?

SammyIsConfused: I didn't realise it was possible to use the WebDriverWait without a page, but it turns out it is! Fixed.

PR #1158, opened 2019-08-07
Fix the slow typing issue in new search UI

PenghaiZhang:
Debounce the search  functionality in new search UI to improve user experience.
A search is triggered after user stops typing for half second.

Successfully tested with throttling in local environment.


PenghaiZhang:
> Nice! I like it. 
> 
> Happy to proceed, however you're leveraging a transient dependency on `lodash`. As well as the type dependency, should we now specify an explicit dependency as well?

Yep :)

Issue #1162, opened 2019-08-08
The search result is not consistent with the configuration of screen option


PenghaiZhang:
**Describe the bug**
The configuration of the new UI screen option is not consistent with the search results.

**To Reproduce**

1. Enable the checkbox of showing non-live items from Settings > Searching and content indexing
2. Go to the search page, open the screen option, and tick the option "Include results that are not live"
3. The search result list should refresh now and non-live items are displayed.
4. Reload the page.
5. The search result list keeps the same
6. Search a non-live item.
7. The result is "No results found"
8. Open the screen option and the option of showing non-live items is still ticked.
9. Tick this option twice again and the non-live item is displayed now.
10. After step 7,  we can also remove the search term the hit the search button
11. The search result list does not include non-live items.
12. Open the screen option and see the option of showing non-live items is still ticked.

**Expected behavior**
If the option of showing non-live items is ticked, then the result list should really include those items even though the search page is reloaded. Or, reloading the page should result in that option deselected.

**Screenshots**
![Screenshot from 2019-08-08 15-42-26](https://user-images.githubusercontent.com/47203811/62678049-acdf0800-b9f3-11e9-9e57-9cf69c5b71d5.png)

**After reloading the search page**, the result list does not change. Then search a draft item.

![Screenshot from 2019-08-08 15-43-47](https://user-images.githubusercontent.com/47203811/62678057-b5cfd980-b9f3-11e9-97f6-121b5eee170d.png)

**Stacktrace**
None

**Platform:**

- OpenEquella Version: 2019.2
- OS: Ubuntu18.04
- Browser: Chrome
**Additional context**
None

cathfitz:
Further investigation shows that if you select the 'Include results that are not live' checkbox, it will work until unless you use the Refresh browser button. Once you've refreshed the browser, it still includes the non-live items in the count at the top of the page, but as soon as you move away from the page (e.g. view a summary page, go to another page of results), the count changes to only Live items.
Interestingly, you don't need to uncheck then check the option again, you just need to open page options then shut it again to re-enable the non-live items displaying.

abidingotter:
Ok, finally got to the bottom of this one, and the fix is quite simple.
The problem is that the cloud search sends an AJAX request to the server before various React components have been rendered on the page. This means that the screen options div is not present when the cloud search collects all the form elements on the page to submit to the server.
The fix is to do a setTimeout(..., 0) so that the AJAX request does not happen until all the rendering scripts have occurred.

PR #1163, opened 2019-08-08
make SearchSettingTest pass in new UI

PenghaiZhang:
This test failed due to the Screen Option. More details are logged here #1162  .

In addition, I think we can keep this fix even though #1162 is fixed.

### DIFF
--- a/autotest/OldTests/src/test/java/com/tle/webtests/test/searching/SearchSettingsTest.java
+++ b/autotest/OldTests/src/test/java/com/tle/webtests/test/searching/SearchSettingsTest.java
@@ -119,12 +119,13 @@ public class SearchSettingsTest extends AbstractCleanupAutoTest {
     ssp.setOrder(SearchSettingsPage.Order.RANK);
     ssp.includeNonLive(true).save();
 
-    SearchScreenOptions sso = new SearchPage(context).load().openScreenOptions();
-
+    SearchPage searchPage = new SearchPage(context).load();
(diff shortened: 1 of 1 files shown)


autotest/OldTests/src/test/java/com/tle/webtests/test/searching/SearchSettingsTest.java
```diff
@@ -119,12 +119,13 @@ public void testShowNonLive() {
     ssp.setOrder(SearchSettingsPage.Order.RANK);
     ssp.includeNonLive(true).save();
 
-    SearchScreenOptions sso = new SearchPage(context).load().openScreenOptions();
-
+    SearchPage searchPage = new SearchPage(context).load();
+    SearchScreenOptions sso = searchPage.openScreenOptions();
     assertTrue(sso.hasNonLiveOption());
     sso.setNonLiveOption(true);
-    ItemListPage results = SearchPage.searchExact(context, "SearchSettings - Draft Item");
```
PenghaiZhang: searchExact will reload the page and then search. But I can't get the point of reloading the search page. So just call search can fix.

PenghaiZhang:
> Make sure to test it in the old UI too, since this is changing the test functionality on both sides.
> Looks good to me, though.

Yes works fine in old UI.
What searchExact does is : 
return new SearchPage(context).load().search('"' + itemFullName + '"');

Not sure if we really need  "new SearchPage(context).load()"

PR #1170, opened 2019-08-13
Testfix/more rest tests

PenghaiZhang:
The first thing these rest tests do is to login. However, in the new UI, these tests do not wait until the home page is completed loaded. As a result, they fail to open the oauthadmin page, and then fail to get oauth token.

### DIFF
--- a/autotest/OldTests/src/test/java/com/tle/webtests/test/AbstractSessionTest.java
+++ b/autotest/OldTests/src/test/java/com/tle/webtests/test/AbstractSessionTest.java
@@ -3,6 +3,8 @@ package com.tle.webtests.test;
 import com.tle.webtests.framework.PageContext;
 import com.tle.webtests.pageobject.HomePage;
 import com.tle.webtests.pageobject.LoginPage;
+import org.openqa.selenium.By;
+import org.openqa.selenium.support.ui.ExpectedConditions;
 
 public class AbstractSessionTest extends AbstractTest {
   public static String GENERIC_TESTING_COLLECTION = "Generic Testing Collection";
@@ -28,7 +30,11 @@ public class AbstractSessionTest extends AbstractTest {
   }
 
   public void logonOnly(String username, String password) {
-    new LoginPage(context).load().loginWithRedirect(username, password);
+    LoginPage loginPage = new LoginPage(context).load();
+    loginPage.loginWithRedirect(username, password);
+    if (loginPage.usingNewUI()) {
+      loginPage.getWaiter().until(ExpectedConditions.presenceOfElementLocated(By.id("eqpageForm")));
+    }
   }
 
   public void logonWithNotice(String username, String password) {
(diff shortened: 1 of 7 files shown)


autotest/OldTests/src/test/java/com/tle/webtests/test/webservices/rest/AbstractRestApiTest.java
```diff
@@ -159,7 +158,6 @@ protected void cleanupAfterClass() throws Exception {
   }
 
   protected String getToken() throws IOException {
-    new WebDriverWait(context.getDriver(), 30).until((webDriver -> clients.size() >= 1));
```
edalex-ian: Interesting to change this, as it's the bit that @SammyIsConfused just recently added. But I guess I see why, as I was wondering before who `clients` was meant to increase in size, and you're highlighting the root cause here was the delay with the login page.

autotest/OldTests/src/test/java/com/tle/webtests/test/AbstractSessionTest.java
```diff
@@ -28,7 +30,12 @@ public HomePage logon(String username, String password, PageContext context) {
   }
 
   public void logonOnly(String username, String password) {
-    new LoginPage(context).load().loginWithRedirect(username, password);
+    LoginPage loginPage = new LoginPage(context).load();
+    loginPage.loginWithRedirect(username, password);
+    // If using new UI, ensure the home page is completed loaded.
+    if (loginPage.usingNewUI()) {
```
edalex-ian: I wonder if we really need this `if` check. Potentially the comment with the line is sufficient, and in theory the line below will just pass straight away when in the Legacy UI mode.

Thoughts?

autotest/OldTests/src/test/java/com/tle/webtests/test/webservices/rest/AbstractRestApiTest.java
PenghaiZhang: > Interesting to change this, as it's the bit that @SammyIsConfused just recently added. But I guess I see why, as I was wondering before who `clients` was meant to increase in size, and you're highlighting the root cause here was the delay with the login page.

If these tests fail to open oauthadmin page due to login delay, then `clients` will always be empty regardless of how long the tests wait here.

I got some key code here:

      for (OAuthClient clientInfo : oauthClients) {
        OAuthUtils.createClient(context, clientInfo);   ---this line fails due to oauthadmin not open
        clients.add(clientInfo);  --this line won't get executed
      }

autotest/OldTests/src/test/java/com/tle/webtests/test/AbstractSessionTest.java
PenghaiZhang: > I wonder if we really need this `if` check. Potentially the comment with the line is sufficient, and in theory the line below will just pass straight away when in the Legacy UI mode.
> 
> Thoughts?

Yes, this `if` is redundant. I was thinking `eqpageForm` only exists in new UI but now I confirm it also exists in the Legacy UI. So let me remove this if.

PR #1171, opened 2019-08-13
Add end to end / functional testing to Travis builds

edalex-ian:
This PR makes it so that the same set of Selenium tests which were previously only run on the private AWS Codebuild is now run on the public Travis CI builds. This way full visibility can be provided for all the community.

Further, splitting up the build into a multi-stage/multi-job build means that once the work to get the same tests running with the New UI turned on is complete, we can readily incorporate them being run for every build in parallel. Hopefully this will occur by the end of this week - if not the next.

Other than the significant refactoring to support the above, items of note are:

* Build image has been updated all the way from Trusty (14) to Bionic (18) so that we can get access to a version of exiftools which is compliant with the functional tests (specifically, an ASC test). We actually have been using this on Codebuild for awhile now.
  * Doing so means I had to coerce it to allow me to utilise openjdk8 - Travis only supports 9+ on Bionic.
* Chromium is used rather than Chrome (as it is on the Edalex Codebuild). This was for convenience mainly, and it seems to work. (This does actually concern me a little, as being installed from APT it will change overtime and we'll need to update the chrome web driver. This can mean that when we go back to build future stable branches we may find things not immediately passing. To pin this version would require significant additional work - I think...)
* FFmpeg is used with symlinks to emulate libav until we get that sorted.
* When you run `sbt` on Travis you're actually using the [sbt-extras tool](https://github.com/paulp/sbt-extras/). (This is actually the same as the `sbt` in the project's root.) Of specific note is that Paul P decided to remove support for the `-mem` option of normal `sbt` so instead we use the `-jvmopts` to specify memory needs as per the use of `-mem` on Codebuild.
* The creation of all the TestNG suite YAML files was done by grabbing the one big YAML file used on codebuild and splitting it into logical groups. As a result, in theory we're running the exact same suite of tests as there.
* AWS S3 (provided by Edalex) is still used to host the final artefacts, but also to share files between stages;
  * This also includes all test results, as well as a final Jacoco coverage report (with the URL being displayed at the end of the 'Generate coverage report' job).
  * It could be useful to generate a HTML landing page in the future to access all the bits.
* There is further work which could be done to further improve the DRYness of this, however I'm sure you'll agree this change is big enough as is.
* In the future we could reduce reliance on the use of S3 through the use of the new [Travis CI Workspaces](https://docs.travis-ci.com/user/using-workspaces/) feature. However, that's still in beta, so I held off for now. Further, as we move towards bedding down our use of Git Flow we could utilise the [Travis CI Releases](https://docs.travis-ci.com/user/deployment/releases/).

(Once approved, I plan to merge this branch in rather than doing a squash and merge so as to preserve the history of the work done here. I believe the quality of the commit messages support this, plus the contain knowledge for future.)

P.S. And just as I've finished this I note [GitHub's Actions in beta](https://github.com/features/actions). However, that'd not be enough for us.


.travis.yml
```diff
@@ -1,32 +1,167 @@
...
+    - JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64
+    # include $HOME/.local/bin for `aws`, and $JAVA_HOME/bin to force use of APT installed Java 8
+    - PATH=$JAVA_HOME/bin:$HOME/.local/bin:$PATH
+services:
+  - postgresql
+addons:
+  postgresql: "9.6"
+  apt:
+    packages:
+      - chromium-browser
```
ChristianMurphy: Out of general curiosity, not something that needs to be changed.
Why chromium over [Chrome](https://docs.travis-ci.com/user/chrome) or [Firefox](https://docs.travis-ci.com/user/firefox/) for test suite?

.travis.yml
```diff
@@ -1,32 +1,167 @@
...
+      - ffmpeg
+      - libimage-exiftool-perl
+      - openjdk-8-jdk
+      - python3-pip
+      - python3-setuptools
 cache:
   directories:
     - "$HOME/.ivy2"
     - "$HOME/.cache"
     - "$HOME/.sbt"
```
ChristianMurphy: maybe we could also cache `$HOME/.cache/pip`?

.travis.yml
```diff
@@ -1,32 +1,167 @@
...
+      name: Build primary artefacts
+      after_success:
+        - aws s3 $S3_REGION_OPT cp Source/Server/equellaserver/target/tle-upgrade*.zip $S3_DEST_BUILD
+        - aws s3 $S3_REGION_OPT cp Installer/target/equella-installer*.zip             $S3_DEST_BUILD
+        - aws s3 $S3_REGION_OPT cp target/reference-language-pack.zip                  $S3_DEST_BUILD
+        - aws s3 $S3_REGION_OPT cp target/scriptingapi-javadoc-*.zip                   $S3_DEST_BUILD
+    - stage: build and check
+      script: sbt test
+      name: Unit test
+    - stage: build and check
```
PenghaiZhang: We have three "build and check" here. So I guess what is after `stage:` is a description of this stage, and different stages are identified by their `name:` ?

PenghaiZhang: oh I understand now. They are 3 jobs in the stage "build and check"

.travis.yml
edalex-ian: Heya @ChristianMurphy 

First, the easy one - Firefox. Currently the test suite does not pass with them. :( Ideally one day we'll go back and do that.

As for Chrome, it was easier to go with Chromium and seeing that's the same engine as Chrome I decided that made sense. Then when the tests all passed, I was happy.

I note though that your links are to the Travis CI addons. So at first I _was_ using the Chrome addon and it was really nice. But alas, you'll note at that link it only works with Trusty (or macOS). Seeing we needed to go Bionic, I then had to rework it to use APT.

But that said previously you mentioned there is possibly a provider out there who gives OSS projects access to being able to run against multiple browsers. I'm hoping in the future we get to do that.

autotest/build.sbt
```diff
@@ -108,14 +108,22 @@ coverageLoader := {
   val l   = new ExecFileLoader()
   optPath(cc, "file").filter(_.canRead).foreach { f =>
     log.info(s"Loading coverage data from ${f.absolutePath}")
-    l.load(f)
+    if (f.isDirectory)
+      f.listFiles().foreach(ef => l.load(ef))
```
PenghaiZhang: Can I ask in what cases f would be a directory?

.travis.yml
```diff
@@ -1,32 +1,167 @@
...
+      script:
+        # Copy down previous stored Jacoco coverage exec files
+        - aws s3 $S3_REGION_OPT cp ${S3_DEST_BUILD}coverage/ autotest/coverage $S3_CP_OPTS
+        # First we need to installEquella so that it extracts the install zip with all the JARs and
+        # as a result all the class files which Jacoco needs to map against to generate the
+        # coverageReport
+        - sbt -jvm-opts autotest/.jvmopts "project autotest" installEquella coverageReport
+      name: Generate coverage report
+      after_success:
+        - aws s3 $S3_REGION_OPT cp autotest/target/coverage-report ${S3_DEST_BUILD}coverage-report/ $S3_CP_OPTS
```
PenghaiZhang: I am a bit confused here. 
In the end of a successful build a full coverage report will be copied to s3.
Then why we also use `coverage-save-results.sh`  in each job's after_script?

autotest/build.sbt
edalex-ian: Good question @PenghaiZhang 

So this code was initially by @doolse but it looked like it suited my purpose and just needed a little tweaking.

The way it works is that it loads Jacoco Exec files from a location specified in your autotest config file by:

```
coverage {
  file = <location>
}
```

You'll see in this PR that such a change has been made to `travis.conf`. The value I specify there is a directory with which I populate all the Jacoco Exec files from each of the jobs in the 'functional test' stage. So I needed to modify this config value to support either a single file specification, or a directory.

Now when it is a directory, the effect is they're all loaded in and represented as sessions in the Jacoco report. e.g. https://edalexdev.s3.amazonaws.com/equella_artifacts/travis-end2end/3377/coverage-report/jacoco-sessions.html

.travis.yml
edalex-ian: Another good question.

The process for arriving at the 'Coverage Report' (i.e. the set of HTML files detailing the coverage) is:

1. Start oEQ with the Jacoco java assist agent running in it;
2. Run tests against oEQ which results in the Jacoco agent accumulating metrics;
3. Before stopping oEQ request the metrics (via TCP) from the agent and save into a Jacoco Exec file
4. Save that exec file off to S3 (`coverage-save-results.sh`) as every job starts fresh, so if you don't save it you lose it
5. At the end (in this job) pull down all the Exec files from S3
6. Load them up with the Jacoco API (with matching classes - found in the installer zip) and request the HTML report be generated (see `coverageReport` in `autotest/build.sbt`)
7. On successful generation of this report, copy to S3.

You don't want to generate the HTML at the end of each test step, as you'll only have partial coverage results.

Ideally, it'd be good if we generated it for the unit tests too. But that's a wee bit tricker.

.travis.yml
```diff
@@ -1,32 +1,167 @@
...
+        - aws s3 $S3_REGION_OPT cp --recursive $S3_DEST_LATEST $S3_DEST_BUILD
+      name: Update S3
+    - stage: finalise
+      script:
+        # Copy down previous stored Jacoco coverage exec files
+        - aws s3 $S3_REGION_OPT cp ${S3_DEST_BUILD}coverage/ autotest/coverage $S3_CP_OPTS
+        # First we need to installEquella so that it extracts the install zip with all the JARs and
+        # as a result all the class files which Jacoco needs to map against to generate the
+        # coverageReport
+        - sbt -jvm-opts autotest/.jvmopts "project autotest" installEquella coverageReport
```
cbeach47: It's just a trademark thing, but could you switch `installEquella` to `installOpenequella`?

edalex-ian: Interesting thought there @cbeach47 . These are the task and variable definitions in the SBT scripts, and there's a whole swag which simply use Equella.

I think possibly if we want to address this we raise another GitHub issue against the SBT tasks and then update them there. Once that is done, then we'd have to update our builds and our doco on building.

So for now, I'll proceed with this PR as the above would be out of scope for this one.

PR #1185, opened 2019-08-22
Example docker compose of clustered oEQ environment

PenghaiZhang:
* Use docker-compose to setup the oEQ clustering environment

* Update Readme.md


PenghaiZhang:
As I don't have docker swarm working locally, I use `docker-compose up -d --scale oeq=3`  instead of `docker stack deploy`

PR #1186, opened 2019-08-22
Setup oEQ clustering environment

PenghaiZhang:
* Use docker-compose to setup the oEQ clustering environment
 * Update docker/Readme.md

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -104,3 +104,26 @@ See https://github.com/equella/Equella#keystore for more details.
 ## Future
 
 For ideas on how to enhance docker with openEQUELLA, please review the (GitHub issues)[https://github.com/apereo/openEQUELLA/issues?q=is%3Aissue+docker+label%3ADocker] with the `docker` label.
+
+## Setup a local clustering environment of oEQ
+
+- From the root use SBT to build the installer;
+
+```
+sbt installerZip
+```
+
+- Copy the installer zip from `Equella/Installer/target` to the docker folder `Equella/docker`; then
+- Open Dockerfile and change values of oEQ environment variables according to how you configure the docker-compose.yml
+
+```
+For instance, ARG EQ_ADMIN_URL=http://oeq.localhost/admin/
+
+Here database is the service name of Postgres defined in docker-compose.yml
+```
+
+- Run `docker-compose up -d` from `Equella/docker`
+- Check logs of docker containers which run oEQ instances, you should expect to see `[ClusterMessagingServiceImpl] Successful connection from NODE: xxxx (a string in UUID format)`
+- Open `oeq.localhost/admin/` from a browser, go the oEQ Administer server page() and then open Health check, you should expect to see a table which lists all node IDs
+- If you have a new oEQ installer, you can run `docker-compose up -d --force-recreate --build`
+- You can also specify the number of oEQ instance by running 'docker-compose up -d --scale oeq=3', here `oeq` is the service name defined in the yml file
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,0 +1,63 @@
+version: "3.6"
+
+services:
+  oeq:
+    build: .
+    networks:
+      - oeq-cluster
+    links:
+      - database
+    labels:
+      - "traefik.enable=true"
+      - "traefik.backend=oeq"
+      - "traefik.docker.network=oeq-cluster"
+      - "traefik.frontend.rule=Host:oeq.localhost"
+      - "traefik.port=8080"
+    environment:
+      - EQ_MESSAGING_BINDPORT=8999
+      - EQ_MESSAGING_USEHOSTNAME=true
+      - EQ_USERSERVICE_USEXFORWARDEDFOR=true
+      - EQ_ZOOKEEPER_INSTANCES=zookeeper
+      - EQ_ZOOKEEPER_PREFIX=eqdev
+    depends_on:
+      - zookeeper
+    volumes:
+      - /home/penghai/Edalex/Equella/Dev/learningedge-config/data/filestore:/home/equella/equella/filestore
+      - /home/penghai/Edalex/Equella/Dev/learningedge-config/hikari.properties:/home/equella/equella/learningedge-config/hikari.properties
+      - /home/penghai/Edalex/Equella/Dev/learningedge-config/learningedge-log4j.properties:/home/equella/equella/learningedge-config/learningedge-log4j.properties
+
+  zookeeper:
+    image: zookeeper:latest
+    environment:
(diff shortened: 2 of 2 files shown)


PenghaiZhang:
I have login/logout issues sometimes with this docker-compose file. I will investigate after fixing the ECS issue.

docker/README.md
```diff
@@ -104,3 +104,24 @@ See https://github.com/equella/Equella#keystore for more details.
...
+## Setup a local clustering environment of oEQ
+
+- From the root use SBT to build the installer;
+
+```
+sbt installerZip
+```
+
+- Copy the installer zip from `Equella/Installer/target` to the docker folder `Equella/docker`; then
+- Open Dockerfile and change values of oEQ environment variables according to how you configure the docker-compose.yml
```
edalex-ian: Shouldn't need to edit `Dockerfile` at all. Should just set the environment variable either in your shell (say, with `export VAR=value`) and/or via the `docker-compose.yml`. Keep in mind, if a docker-compose YAML file has a an environment variable listed in the environment key with no value set it pulls it from the calling environment.

docker/README.md
```diff
@@ -104,3 +104,24 @@ See https://github.com/equella/Equella#keystore for more details.
...
+
+```
+For instance, ARG EQ_ADMIN_URL=http://oeq.localhost/admin/
+```
+
+- Run `docker-compose up -d` from `Equella/docker`
+- Check logs of docker containers which run oEQ instances, you should expect to see `[ClusterMessagingServiceImpl] Successful connection from NODE: xxxx (a string in UUID format)`
```
edalex-ian: Wonder if it's worth noting this can be done via `docker-compose logs` from the directory.... :thinking:

docker/README.md
```diff
@@ -104,3 +104,24 @@ See https://github.com/equella/Equella#keystore for more details.
...
+- Open `oeq.localhost/admin/` from a browser, go the oEQ Administer server page() and then open Health check, you should expect to see a table which lists all node IDs
```
edalex-ian:
```suggestion
- Open `oeq.localhost/admin/` from a browser, go the oEQ Administer server page and then open Health check, you should expect to see a table which lists all node IDs
```

docker/docker-compose.yml
```diff
@@ -0,0 +1,66 @@
...
+    build: .
+    networks:
+      - oeq-cluster
+    links:
+      - database
+    labels:
+      - "traefik.enable=true"
+      - "traefik.backend=oeq"
+      - "traefik.docker.network=oeq-cluster"
+      - "traefik.frontend.rule=Host:oeq.localhost"
```
edalex-ian: Does one need to put anything in their `/etc/hosts` file to have that hostname work, or is it a docker compose special?

docker/docker-compose.yml
```diff
@@ -0,0 +1,66 @@
...
+      - EQ_MESSAGING_USEHOSTNAME=true
+      - EQ_USERSERVICE_USEXFORWARDEDFOR=true
+      - EQ_ZOOKEEPER_INSTANCES=zookeeper
+      - EQ_ZOOKEEPER_PREFIX=eqdev
+    depends_on:
+      - zookeeper
+    volumes:
+      - /home/penghai/Edalex/Equella/Dev/learningedge-config/data/filestore:/home/equella/equella/filestore
+      - /home/penghai/Edalex/Equella/Dev/learningedge-config/hikari.properties:/home/equella/equella/learningedge-config/hikari.properties
+      - /home/penghai/Edalex/Equella/Dev/learningedge-config/learningedge-log4j.properties:/home/equella/equella/learningedge-config/learningedge-log4j.properties
```
edalex-ian: Hold on a second, these are explicitly mapped to paths on your machine. We need to do something about that.

Probably best to go with relative paths here and provide the basic set of for `hikari.properties` and `learningedge-log4j.properties`.

docker/docker-compose.yml
```diff
@@ -0,0 +1,66 @@
...
+  zookeeper:
+    image: zookeeper:latest
+    environment:
+      ZOO_MY_ID: 1
+      ZOO_SERVERS: server.1=0.0.0.0:2888:3888;2181
+      ZOO_LOG4J_PROP: INFO, CONSOLE, ROLLINGFILE
+      ZOO_AUTOPURGE_PURGEINTERVAL: 1
+    volumes:
+      - /zookeeper/zoo1/data:/data
+      - /zookeeper/zoo1/datalog:/datalog
```
edalex-ian: I wonder if you need to explicitly define volumes for this, as it's only single node Zk setup...

docker/docker-compose.yml
```diff
@@ -0,0 +1,66 @@
...
+      - EQ_ZOOKEEPER_PREFIX=eqdev
+    depends_on:
+      - zookeeper
+    volumes:
+      - /home/penghai/Edalex/Equella/Dev/learningedge-config/data/filestore:/home/equella/equella/filestore
+      - /home/penghai/Edalex/Equella/Dev/learningedge-config/hikari.properties:/home/equella/equella/learningedge-config/hikari.properties
+      - /home/penghai/Edalex/Equella/Dev/learningedge-config/learningedge-log4j.properties:/home/equella/equella/learningedge-config/learningedge-log4j.properties
+
+  zookeeper:
+    image: zookeeper:latest
```
edalex-ian: Probably best not to use `latest` here, should pin to a version.

docker/docker-compose.yml
```diff
@@ -0,0 +1,66 @@
...
+      ZOO_LOG4J_PROP: INFO, CONSOLE, ROLLINGFILE
+      ZOO_AUTOPURGE_PURGEINTERVAL: 1
+    volumes:
+      - /zookeeper/zoo1/data:/data
+      - /zookeeper/zoo1/datalog:/datalog
+    networks:
+      - oeq-cluster
+
+  database:
+    image: postgres
```
edalex-ian: Again, probably best to pin to an exact version.

```suggestion
    image: postgres:9.6
```

docker/docker-compose.yml
```diff
@@ -0,0 +1,66 @@
...
+    restart: always
+    networks:
+      - oeq-cluster
+
+  traefik:
+    image: traefik
```
edalex-ian: Again, best to pin version here.

docker/docker-compose.yml
PenghaiZhang: Hi Ian, this is a docker compose special hostname.

docker/docker-compose.yml
PenghaiZhang: `learningedge-log4j.properties` has been provided here for a long time, so I add `hikari.properties`.

PR #1189, opened 2019-08-22
Fix ECS registration token issue

PenghaiZhang:
1. In clustering environment, the ECS registration token is stored in the database, and the token will be deleted from the database once it is used.
2. Say node1 generates a token, then the only way that node 2 and 3 can know this token is to grab the token from the database, because oEQ uses Guava cache which does not support being shared among instances.
3. According to [the guava cache api](https://guava.dev/releases/19.0/api/docs/com/google/common/cache/LoadingCache.html#refresh(K)), refreshing the cache will call the CacheLoader's load method, and this method is where reading the token from the database was implemented before. Therefore, each node is able to know the token is the token is available

### DIFF
--- a/Source/Plugins/Core/com.equella.serverbase/src/com/tle/core/replicatedcache/impl/ReplicatedCacheServiceImpl.java
+++ b/Source/Plugins/Core/com.equella.serverbase/src/com/tle/core/replicatedcache/impl/ReplicatedCacheServiceImpl.java
@@ -51,9 +51,12 @@ import com.tle.core.zookeeper.ZookeeperService;
 import java.io.Serializable;
 import java.util.Date;
 import java.util.Iterator;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import javax.inject.Inject;
 import javax.inject.Singleton;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 
 @Singleton
 @NonNullByDefault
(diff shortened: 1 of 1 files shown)


PenghaiZhang:
Travis failed because this branch was created before we merged the ASC fix branch into Develop.

edalex-ian:
Heya @PenghaiZhang 

I'm not sure this is the right approach. As it means whenever we use the the cache in a clustered environment first it's going to read from the DB again? That then voids the value of the cache.

I thought oEQ has it's own messaging protocol, so that ideally if new data is added then the other nodes can be signalled to refresh their cache at the point. So, here when a new registration token is created a message should be trigged. And once a token is removed, then another message should trigger to evict.

Let's discuss.

doolse:
I just had a look at `ReplicatedCache` and I think the problem might be in the `get` - 
```java
    public synchronized Optional<V> get(@NonNull String key) {
      checkNotNull(key);

      LoadingCache<String, Optional<ExpiringValue<V>>> c =
          cache.getIfPresent(CurrentInstitution.get());
      if (c != null) {
         ... 
      }
```

The `getIfPresent` means that the per institution cache is never actually checked at all if it hasn't been created yet, thus never even going to the db at all. I reckon it would work if it was changed to plain `get` and remove the null check.

PenghaiZhang:
> I just had a look at `ReplicatedCache` and I think the problem might be in the `get` -
> 
> ```java
>     public synchronized Optional<V> get(@NonNull String key) {
>       checkNotNull(key);
> 
>       LoadingCache<String, Optional<ExpiringValue<V>>> c =
>           cache.getIfPresent(CurrentInstitution.get());
>       if (c != null) {
>          ... 
>       }
> ```
> 
> The `getIfPresent` means that the per institution cache is never actually checked at all if it hasn't been created yet, thus never even going to the db at all. I reckon it would work if it was changed to plain `get` and remove the null check.

Thanks Jolse! I will try your suggestion now.

PR #1190, opened 2019-08-26
Fix incorrectly prerendering styles.css in new layout

SammyIsConfused:
Fixed an issue with https://github.com/openequella/openEQUELLA/pull/878 in which the styles.css was wrongly moved out of the oldLayout, causing graphical issues. Tested that it still works in IE11.


edalex-ian:
So, what's the deal with the failing build?

SammyIsConfused:
image
Here's the screenshot for @PenghaiZhang!
@edalex-ian - the build was failing because the branch was incorrectly pulling down the 'master' autotest branch rather than the 'stable-6.6' autotest branch. This meant that it simply couldn't even install any institutions, so all tests failed. Fixed that now, with an environment var.

PR #1191, opened 2019-08-26
Bugfix/drag and drop update duplicate msg

PenghaiZhang:
Modify Drag and Drop to update attachment duplicate warning message
* If a collection has multiple attachment controls, each attachment control has a warning message
* Only update the warning message of an attachment control which has attachments changed

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/src/Uploads/UploadList.js
+++ b/Source/Plugins/Core/com.equella.core/js/src/Uploads/UploadList.js
@@ -5,15 +5,16 @@ exports.updateCtrlErrorText = function(ctrlId, text) {
   contElem.querySelector("P.ctrlinvalidmessage").textContent = text;
 };
 
-exports.updateDuplicateMessage = function(display) {
+exports.updateDuplicateMessage = function(id, display) {
   var duplicateMessageDiv = document.querySelector(
-    ".attachment-duplicate-message"
+    "#" + id + "_duplicateWarningMessage"
   );
-  console.log(display);
-  if (display) {
-    duplicateMessageDiv.setAttribute("style", "color:red; display:inline");
-  } else {
-    duplicateMessageDiv.setAttribute("style", "color:red; display:none");
+  if (duplicateMessageDiv != null) {
+    if (display) {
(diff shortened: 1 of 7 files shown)


PenghaiZhang:
> From what I understand the approach here is to add `AttachmentDuplicateInfo` status changes to `UpdateEntry` and `RemoveEntries` and update the warning link with DOM manipulation. As far as I can tell this approach will work .
> 
> It's not really worth changing if it works, although I'll just throw my 2c in and suggest a slightly different approach:
> 
> * Add a flag to `FileElement` for whether or not it is a duplicate or not
> * Render the warning based on any entry having the flag set in the `UploadList` react component itself.
> 
> You would need to pass through another language string in the `ControlStrings` + the onClick handler for the warning link but that's trivial.
> 
> Anyway nice work on editing some purescript @PenghaiZhang

Thanks Jolse!! Editing purescript is really fun as I got a lot type match errors at the beginning lol.

Source/Plugins/Core/com.equella.core/js/src/Uploads/UploadList.js
```diff
@@ -5,6 +5,21 @@ exports.updateCtrlErrorText = function(ctrlId, text) {
   contElem.querySelector("P.ctrlinvalidmessage").textContent = text;
 };
 
+exports.updateDuplicateMessage = function(id, display) {
+  // The div id of all duplicate warning message automatically follow
```
SammyIsConfused: Potentially could do with a comment rephrase.
` // the div id of all duplicate warning messages automatically follows `
`  // this format: its parent div id concatenated with "_duplicateWarningMessage"`

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/controls/universal/UniversalWebControlNew.scala
```diff
@@ -343,9 +343,15 @@ class UniversalWebControlNew extends AbstractWebControl[UniversalWebControlModel
                     children)
     }
 
+    // Return if any attachment in this control has duplicate information found
```
SammyIsConfused: `// Returns true if any attachment in this control has duplicate information found`

PR #1195, opened 2019-08-27
Change Travis branch build targets

PenghaiZhang:
Make Travis do branch build only for master, develop, release and stable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,11 @@
 language: scala
 dist: bionic
+branches:
+  only:
+    - develop
+    - master
+    - /^release\/.+/
+    - /^stable-.+/
 env:
   global:
     ### AWS START ###


edalex-ian:
Heya @ChristianMurphy 

That's what'll happen. Because we have the Travis option to be ON for PRs they will get built regardless of branch.

But before this change, both a PR build would trigger, and so would a branch build (which is now almost 7 hours worth of travis time). So now _only_ the PR build will trigger, which is the main one we need - as it confirms that all should work after merge. (And then branch builds will only occur on key branches - ones that also typically don't have a PR.)

Also, we're still going to see what further optimisations we can do to improve build time - such as a possibly warm up step to do all the TS/JS/Purescript bits which get done again and again every test job - and each take around 7mins.

Cheers.

Issue #1147, opened 2019-08-02
GET taxonomy REST endpoint doesn't work


abidingotter:
**Describe the bug**
Error when listing all taxonomies for an institution

**To Reproduce**
Go into apidocs.do and then "Try it out" + "Execute" on the list taxonomies endpoint. You will receive the following error:
```
{
  "code": 415,
  "error": "Unsupported Media Type",
  "error_description": "RESTEASY003200: Could not find message body reader for type: interface javax.ws.rs.core.UriInfo of content type: */*"
}
```

Other entity endpoints don't seem to have the same problem.

abidingotter:
This is fixed as a result of other changes in 2019.2

PR #1202, opened 2019-08-30
Taxonomy term sorting

abidingotter:
Includes changes to be able to create / edit a taxonomy via the REST API, and a fix to the List taxonomies endpoint as logged under #1147


abidingotter:
Closing to use forking model

PR #1205, opened 2019-09-02
Add duplicate check to Editbox

PenghaiZhang:
* this check is triggered when the value of an Editbox is changed and its focus is lost.
* The warning message links to Duplicate Data page.

Screenshot from 2019-09-02 14-07-15
Screenshot from 2019-09-02 14-07-01


Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/controls/CEditBox.java
```diff
@@ -60,20 +79,29 @@ public void resetToDefaults() {
...
+
+  private void checkDuplicate(ImmutableCollection<String> list) {
+    // We need to inform the wizard to check for uniqueness every time,
+    // no matter what
+    isUnique =
+        getRepository().checkDataUniqueness(getFirstTarget().getXoqlPath(), list, !forceUnique);
+
+    setInvalid(
+        forceUnique && !isUnique && !isInvalid(),
+        new KeyLabel("wizard.controls.editbox.uniqueerror")); // $NON-NLS-1$
```
edalex-ian: I see you've essentially refactored this code into its own method... But even then, we don't need these `$NON-NLS-1$` comments any-more, so might as well remove when refactoring.

Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/controls/CEditBox.java
```diff
@@ -51,6 +56,20 @@ public CEditBox(WizardPage page, int controlNumber, int nestingLevel, WizardCont
...
+    return checkDuplication;
+  }
+
+  public boolean isForceUnique() {
+    return forceUnique;
+  }
+
+  // An instace of CEditBox is wrapped by an EditBox Section which controls
+  // the UI and event handling
+  public void setEditBoxSection(com.tle.web.wizard.standard.controls.EditBox editBoxSection) {
```
edalex-ian: Do you need the fully qualified class name here?

PenghaiZhang:
Just an update : new warning message
image

PR #1206, opened 2019-09-02
Hotfix/2019.1.1

PenghaiZhang:
The intention of this PR is to make review easier.


autotest/build.sbt
```diff
@@ -69,9 +69,18 @@ installOptions := {
 def optPath(bc: Config, p: String) = if (bc.hasPath(p)) Some(file(bc.getString(p))) else None
 
 autotestInstallerZip := {
-  val bc = autotestBuildConfig.value
-  optPath(bc, "install.zip").orElse(optPath(bc, "install.dir").map(d =>
-    (d * "equella-installer-*.zip").get.head))
+  val bc                       = autotestBuildConfig.value
+  val currentFullVersion       = equellaVersion.value
```
edalex-ian:
```suggestion
  val equellaFullVersion       = equellaVersion.value
```

autotest/build.sbt
```diff
@@ -69,9 +69,18 @@ installOptions := {
...
+  val installerFileName        = s"equella-installer-${currentFullVersion.majorMinor}.zip"
+  val installerTargetDirectory = (target in LocalProject("Installer")).value
```
edalex-ian:
```suggestion
  val installerDirectory       = (target in LocalProject("Installer")).value
```

autotest/build.sbt
```diff
@@ -69,9 +69,18 @@ installOptions := {
 def optPath(bc: Config, p: String) = if (bc.hasPath(p)) Some(file(bc.getString(p))) else None
 
 autotestInstallerZip := {
-  val bc = autotestBuildConfig.value
-  optPath(bc, "install.zip").orElse(optPath(bc, "install.dir").map(d =>
-    (d * "equella-installer-*.zip").get.head))
+  val bc                       = autotestBuildConfig.value
+  val currentFullVersion       = equellaVersion.value
+  val installerFileName        = s"equella-installer-${currentFullVersion.majorMinor}.zip"
```
edalex-ian:
```suggestion
  val installerFileName        = s"equella-installer-${equellaFullVersion.majorMinor}.zip"
```

autotest/build.sbt
```diff
@@ -69,9 +69,18 @@ installOptions := {
...
+  val installerTargetDirectory = (target in LocalProject("Installer")).value
+  val installerFile            = installerTargetDirectory / installerFileName
```
edalex-ian:
```suggestion
  val installerAbsoloutePath   = installerTargetDirectory / installerFileName
```

autotest/build.sbt
```diff
@@ -69,9 +69,18 @@ installOptions := {
...
+  // If the Installer named as installerFileName exists then return it, otherwise returns the default Installer
+  if (installerFile.exists) {
```
edalex-ian:
```suggestion
  if (installerAbsoloutePath.exists) {
```

autotest/build.sbt
```diff
@@ -69,9 +69,18 @@ installOptions := {
...
+    Some(installerFile)
```
edalex-ian:
```suggestion
    Some(installerAbsoloutePath)
```

Source/Tools/UpgradeManager/src/com/tle/upgrademanager/Utils.java
```diff
@@ -24,8 +24,9 @@
 import java.util.regex.Pattern;
 
 public final class Utils {
+  // Modify the pattern to support artifacts that have no 'mmr'
```
edalex-ian: Maybe a comment more along the lines of:

```
// The below pattern supports both the old automatic versioning scheme (`<maj>.<min>.r<commits>` - `2019.1.r123`)
// as well as the newer semver hotfix scheme (<maj>.<min>.<rev> - `2019.1.1`). Hence the optional `r?`.

File: autotest/build.sbt (hunk 4 above)
Comment (SammyIsConfused) [reply autotest/build.sbt]: Make sure that's `installerAbsolutePath` - you've got a rogue 'O' in there.

**Comment by edalex-ian** [comment]
Actually, I just had a bit more thought around the upgrader bits...

1. Modifying the regex in the upgrader to support the two formats is really only setting the stage for the future when we can remove the `rnn` from the filename; but
2. For now we should keep a hardcoded `r0` in the filename so that people with older versions can still readily upgrade - without us needing to provide instructions on how to do it manually; then
3. At some point in the future, we'd finally remove that hardcoded `r0` - maybe once we reach 2021.1.0 (i.e. all three supported versions can support non `rnnn` syntax filenames).

To support that, we'll need to further modify the regex though, as the filename will have `2019.1.1.r0` (I think)....

**Comment by PenghaiZhang** [comment]
> Actually, I just had a bit more thought around the upgrader bits...
> 
> 1. Modifying the regex in the upgrader to support the two formats is really only setting the stage for the future when we can remove the `rnn` from the filename; but
> 2. For now we should keep a hardcoded `r0` in the filename so that people with older versions can still readily upgrade - without us needing to provide instructions on how to do it manually; then
> 3. At some point in the future, we'd finally remove that hardcoded `r0` - maybe once we reach 2021.1.0 (i.e. all three supported versions can support non `rnnn` syntax filenames).
> 
> To support that, we'll need to further modify the regex though, as the filename will have `2019.1.1.r0` (I think)....

I just bring `r0` back for

**Comment by PenghaiZhang** [comment]
> > Actually, I just had a bit more thought around the upgrader bits...
> > 
> > 1. Modifying the regex in the upgrader to support the two formats is really only setting the stage for the future when we can remove the `rnn` from the filename; but
> > 2. For now we should keep a hardcoded `r0` in the filename so that people with older versions can still readily upgrade - without us needing to provide instructions on how to do it manually; then
> > 3. At some point in the future, we'd finally remove that hardcoded `r0` - maybe once we reach 2021.1.0 (i.e. all three supported versions can support non `rnnn` syntax filenames).
> > 
> > To support that, we'll need to further modify the regex though, as the filename will have `2019.1.1.r0` (I think)....
> 
> I just bring `r0` back for 2019.1.1

**Comment by PenghaiZhang** [comment]
`mmr` is used to help Manager correctly recognise newer and older versions, however it is removed. So we replace it with `mm`.

File: project/EquellaVersion.scala [hunk 8]
```diff
@@ -1,11 +1,13 @@
-case class EquellaVersion(majorMinor: String, releaseType: String, sha: String) {
-  def fullVersion = s"$majorMinor-$releaseType-$sha"
+case class EquellaVersion(major: Int, minor: Int, patch: Int, releaseType: String, sha: String) {
+  def semanticVersion = s"$major.$minor.$patch"
+  def fullVersion     = s"$semanticVersion-$releaseType-$sha"
 }
 
 object EquellaVersion {
-  val RegEx = """(.*)-(.*)-(.*)""".r
+  val RegEx = """(\d+).(\d+).(\d+)-(.*)-(.*)""".r
```
edalex-ian: Use of `.*` is probably a bit gratuitous, should it be `.+`?

autotest/build.sbt
```diff
@@ -69,9 +69,18 @@ installOptions := {
...
 
 autotestInstallerZip := {
-  val bc = autotestBuildConfig.value
-  optPath(bc, "install.zip").orElse(optPath(bc, "install.dir").map(d =>
-    (d * "equella-installer-*.zip").get.head))
+  val bc                     = autotestBuildConfig.value
+  val equellaFullVersion     = equellaVersion.value
+  val installerFileName      = s"equella-installer-${equellaFullVersion.semanticVersion}.zip"
+  val installerDirectory     = (target in LocalProject("Installer")).value
+  val installerAbsoloutePath = installerDirectory / installerFileName
```
edalex-ian: Don't forget to fix the typo @SammyIsConfused mentioned:

```suggestion
  val installerAbsolutePath = installerDirectory / installerFileName
```

Source/Tools/UpgradeManager/src/com/tle/upgrademanager/helpers/Version.java
```diff
@@ -57,6 +57,11 @@ public Version(ManagerConfig config) {
     return versions;
   }
 
+  private static String getSemanticVersion(String displayName) {
+    // semanticVersion is part of displayName, e.g. 2019.1.1-Stable.OSE
+    return displayName.substring(0, displayName.indexOf("-"));
```
edalex-ian: Has the (does it need to be) displayName been validated at any point. Otherwise you could end up with a substring of (0, -1) which will throw and array out of bounds exception.

Update: Actually, I see there is code below to do this, but for some reason it's external... I wonder if it should actually be internal to this method - as well as the optionally returning null. (i.e. the whole apply() method below, but then what's the point of the apply method.)

Source/Tools/UpgradeManager/src/com/tle/upgrademanager/helpers/Version.java
```diff
@@ -101,12 +107,15 @@ public WebVersion getDeployedVersion() {
     try (FileInputStream in = new FileInputStream(versionFile)) {
       Properties p = new Properties();
       p.load(in);
-      version.setDisplayName(p.getProperty("version.display"));
-      // As mmr is removed from `version.properties`, replace mmr with mm
-      String mmr = p.getProperty("version.mm");
-      version.setMmr(mmr);
+
+      String displayName = p.getProperty("version.display");
+      String semanticVersion = getSemanticVersion(displayName);
```
edalex-ian: Ah see, then display name here is not validated - but if the above comment was actioned it would've been.

Source/Server/equellaserver/build.sbt
```diff
@@ -327,11 +327,14 @@ additionalPlugins := {
   }
 }
 
+import java.time.LocalDate
+import java.time.format.DateTimeFormatter
```
edalex-ian: I think although scala supports the ideas of imports anywhere, I'm not sure this is common in the oEQ scala code. So let's move them back to the top.

Issue #1210, opened 2019-09-04
Support preventing "dangerous" scripts from being run


doolse:
**Is your feature request related to a problem? Please describe.**

While being extremely powerful, openEQUELLA's scripts are unsuitable to be used in a shared hosting environment because you must trust that a script won't use too many resources as you have a turing complete language at your disposal. It would be nice is we could support a safe subset that is suitable for shared hosting.

**Describe the solution you'd like**

The expressions you can generate in the "Basic" tab of the script editing panel (visibility/workflow scripts etc) is a safe subset which still allows for some useful configurations. So it would be good to limit scripts to this subset.

I propose an optional flag (in optional-config.properties) which: 
* Doesn't change anything at all if not set
* Checks scripts conform to the restrictions set in the "Basic" tab of the scripting dialog(s)
* Disables 
  * "Advanced" scripting tab
  * "Expert Scripting" tab on collection editor
  * "Advanced scripting" wizard control
  * "Scripted section" on display wizard
  * "Scripted" portlet
  * "Execute Script" bulk item action

It is not intended that an installation which already contains "unsafe" scripts should function in a friendly manner (exceptions will be displayed upon use of non-conforming scripts). This is a flag which would be set once upon installation and not changed.

abidingotter:
Let's leave this issue open since there is already a PR for it

PR #1212, opened 2019-09-04
Safe scripting config parameter

doolse:
#1210 
* Moved the `BasicParser` from the `com.equella.admin` so that it could be used by the `ScriptingService` to validate the scripts.
* Check the "scripting.safe" flag to disable functionality in all the places mentioned in the issue.


edalex-ian:
Heya @doolse 

Looks like the travis build needs a fix so that fork PRs can still build without erroring on encrypted creds. I'll look into it.

Also, I note there are no tests included with this change/feature. Any chance you could please add some?

Cheers.

doolse:
> Heya @doolse
> 
> Looks like the travis build needs a fix so that fork PRs can still build without erroring on encrypted creds. I'll look into it.
> 
> Also, I note there are no tests included with this change/feature. Any chance you could please add some?
> 
> Cheers.

I believe we have enough test coverage on the "nothing has changed if the flag is not set" side of things, and unfortunately our test setup can't test config parameter changes atm, so I've added a unit test which tests that a couple of scenarios are picked up by the parser.

edalex-ian:
> so I've added a unit test

Good stuff. Anything we can test with a unit test is great! Aligns with the new direction of any new feature or bug fix we make sure we've got testing coverage for.

And I've just restarted one of the failed test suites for you so then we should have a nice green PR ready for review. :+1:

Source/Plugins/Core/com.equella.core/test/java/com/tle/core/scripting/UnsafeScriptParserTest.java
```diff
@@ -0,0 +1,42 @@
...
+  public void testUnsafeScript() {
+    try {
+      parseScript(
+          "var bRet = false; \n"
+              + "if( xml.contains('/item/name', console.log(\"Dangerous\") ) \n"
+              + "{ \n"
+              + "    bRet = true; \n"
+              + "} \n"
+              + "return bRet; \n");
+      throw new Error("This script is dangerous and should not be allowed to run");
```
cbeach47: Seems odd - shouldn't the call the `BasicParser` invocation throw the error/exception?  Currently it reads that if `parseScript` _succeeds_, we throw an error in the test, and then catch the error in the test.

Source/Plugins/Core/com.equella.core/src/com/tle/core/scripting/guice/ScriptingGuiceModule.java
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+package com.tle.core.scripting.guice;
+
+import com.tle.core.config.guice.OptionalConfigModule;
+
+public class ScriptingGuiceModule extends OptionalConfigModule {
+
+  @Override
```
cbeach47: Please add a comment explaining what setting this config will affect in the system - at least directing users to the description in https://github.com/openequella/openEQUELLA/issues/1210 .  It would be good to add this documentation into the openEQUELLA.github.io repo as well.

PR #1216, opened 2019-09-05
Add support for forks to the Travis CI Build

edalex-ian:
As a security measure [Travis CI removes encrypted variables for builds on forks][1]. To support this, here we now wrap calls to `aws s3 cp` to check that required env vars are present. If they're not, copying is skipped.

However, S3 was also used to share the installer across all the test jobs. To still have the possibility of running the tests against fork PRs I've added code to use the new(ish) [Travis Workspaces][2]. But it should be noted this feature is still Beta, so I hope it works. :crossed_fingers: 

Assuming this build all works fine on this PR. The next test will have to be after it's merged to develop to see if the build still copies all bits to S3 nicely.

[1]: https://docs.travis-ci.com/user/pull-requests#pull-requests-and-security-restrictions
[2]: https://docs.travis-ci.com/user/using-workspaces/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,8 +52,6 @@ before_install:
 install:
   - curl -OL https://chromedriver.storage.googleapis.com/76.0.3809.68/chromedriver_linux64.zip
   - unzip chromedriver_linux64.zip
-  # Copy down the installer from the build stage - if available
-  - ci/s3cp.sh ${S3_DEST_BUILD} Installer/target/ --recursive --exclude "*" --include "equella-installer*.zip"
   # retrieve full git history for automatic versioning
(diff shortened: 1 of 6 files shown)


edalex-ian:
Early days, but promising to see the build primary artefacts work with both creating the workspace and skipping S3 commands:

image

If the workspaces do indeed work, in the future we can then look to use them for speeding up some of the repetitive JS and Purescript building for all the tests. That in turn will probably save 1hr build time.

edalex-ian:
Cool, well the build all passed. Any chance you can quickly cast your eye over it @PenghaiZhang ?

Then I'll merge it in and then #1212 can be updated with it to confirm building.

Issue #1221, opened 2019-09-06
Admin console banner does not fit inside window in the new launcher


abidingotter:
Due to some difference with Oracle Java vs OpenJDK java, the window size in the new launcher is smaller than it should be (I couldn't work out why) and so the banner no longer fits.

![image](https://user-images.githubusercontent.com/1694589/64393000-a4c9c500-d092-11e9-81f7-49450d349ddd.png)

abidingotter:
The actual whole window is narrower in OpenJDK, despite the same call to setSize to set the width to 450pixels.  I have “fixed” the issue by making the window wider and resizing the header and footer images. (I also de-giffed them 🙂 )

![image](https://user-images.githubusercontent.com/1694589/64393028-e195bc00-d092-11e9-99c1-bd9d16f124ac.png)

cathfitz:
Tested and looks good

PR #1222, opened 2019-09-06
#1221 Increase the console width

abidingotter:
Screenshot is in related issue


abidingotter:
> Looks good to me. Just curious if Swing supports adjusting the size of this dialog. (dialog.setResizable(true);)

If we did something like that we'd have to make it look ok in all sizes...  (and if the banner image isn't large enough it leaves a weird looking space to the right)

PR #1223, opened 2019-09-06
Refactor Duplicate Data page

PenghaiZhang:
* Remove checkboxes
* Duplicate data are displayed in two sections( one for Editbox and the other one for attachments)
* Display the title of an Editbox which has duplicates

Screenshot from 2019-09-06 14-32-07


PenghaiZhang:
Before, if an editbox is set to check uniqueness and duplicates found, in order to publish items users must click Save, click Complete, and then go to Duplicate data page to tick checkboxes.   
However, the logic has been changed. Now, Users can publish items even though duplicates are found.
Therefore two tests failed because the Complete button cannot be found.

edalex-ian:
I see this PR has failed due to the coverage report not having the workspace. So i think that's fine and can be ignored.

edalex-ian:
Is there a GitHub Issue to link this to?

Issue #725, opened 2019-01-29
Ability to sort Taxonomy terms in the Taxonomy Editor


cathfitz:
Add ability to sort Taxonomy terms in the Taxonomy Editor alphabetically and at selected levels.

edalex-ian:
See PR #1225

abidingotter:
Still waiting on the merging the whole tree PR..

PR #1225, opened 2019-09-06
Feature/taxonomy term sorting

abidingotter:
#725, #1310 and #1311


edalex-ian:
Cool, replaces #1202

Source/Plugins/Core/com.equella.core/src/com/tle/web/api/taxonomy/TaxonomyResourceImpl.java
```diff
@@ -86,108 +75,131 @@
...
+  //					"Taxonomy is locked in a different session.  Call unlockTaxonomy with a force parameter
+  // value of true.",
+  //					Status.UNAUTHORIZED);
+  //			} else {
+  //				throw new WebApplicationException(
+  //					"You do not own the lock on this taxonomy.  It is held by user ID " + ex.getUserID(),
+  //					Status.UNAUTHORIZED);
+  //			}
+  //		}
+  //		return Response.ok().build();
```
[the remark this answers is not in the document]
abidingotter: > One of the key changes in this PR is the use of termUuid. Can I ask why is termUuid needed ?

It allows you to specify your own UUID for the term instead letting the system generate one.  This is useful so that you know how to retrieve the term later on.  We already allow it for other types of objects.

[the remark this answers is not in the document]
abidingotter: > Are there any particular reasons for keeping these commented code ?

No, I will delete the commented out code

PenghaiZhang:
TaxonomyApiTest and UserGroupManagementApiTest failed in both Old and New UI. Maybe the issue comes from `create` in TaxonomyResource.java

PenghaiZhang:
Travis test still failed `TaxonomyApiTest` due to the 406 error, but AWS passed...how strange

PenghaiZhang:
Finally found that Travis builds failed because Travis disk spaces were running out.

Source/Plugins/Core/com.equella.core/src/com/tle/web/api/taxonomy/interfaces/TaxonomyResource.java
```diff
@@ -0,0 +1,303 @@
...
+
+  /**
+   * Search terms
+   *
+   * @param uuid
+   * @param query
+   * @param restriction
+   * @param limit
+   * @param searchfullterm
+   * @return Response
```
edalex-ian: I wont mention this all the way through this (it applies to all the following methods), but if you're not going to define the `@param` and `@return` then please remove. Arguably in this case they're sufficiently documented by the annotations.

Source/Plugins/Core/com.equella.core/src/com/tle/web/api/taxonomy/TaxonomyResourceImpl.java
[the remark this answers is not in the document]
edalex-ian: Seems this code is still here.

Source/Plugins/Core/com.equella.core/src/com/tle/web/core/filter/LoggingContextFilter.java
```diff
@@ -109,7 +109,11 @@ public void afterServlet(HttpServletRequest request, HttpServletResponse respons
         Logger.getLogger(LoggingContextFilter.class.getName() + ".RequestLogger");
 
     public void logRequest(HttpServletRequest request) {
-      final StringBuilder params = new StringBuilder(request.getRequestURI() + "\n");
+      final StringBuilder params =
+          new StringBuilder(request.getMethod())
+              .append(" ")
+              .append(request.getRequestURI())
+              .append("\n");
```
edalex-ian: This is a nice change, but potentially people may be already processing this log. So should we raise a GitHub issue to include against 2019.2 to capture this change.

autotest/OldTests/src/test/java/com/tle/webtests/test/webservices/rest/TaxonomyApiTest.java
```diff
@@ -120,13 +127,201 @@ public void testSearchTerms() throws IOException {
...
+            true);
+    assertResponse(sortRootsResponse, 200, "failed to sort root terms");
+    // Sort test terms children
+    final HttpResponse sortTestTermChildrenResponse =
+        postEntity(
+            null,
+            PathUtils.urlPath(
+                context.getBaseUrl(), API_TAXONOMY_PATH, taxonomyUuid, "sortchildren"),
+            getToken(),
+            true,
```
edalex-ian: This part of this postEntity call is the same now in three places. The only difference here is the subsequent two varargs. So it'd make sense to create a method of `sortTaxonomyChildren(...)` and use it in these three places.

autotest/OldTests/src/test/java/com/tle/webtests/test/webservices/rest/TaxonomyApiTest.java
```diff
@@ -120,13 +127,201 @@ public void testSearchTerms() throws IOException {
...
+    }
+    assertEquals(rootTerms, returnedSortedRootTerms);
+
+    // Check test terms children sorted
+    final List<String> returnedSortedChildTerms = new ArrayList<>(10);
+    final ArrayNode testTermSortedChildren =
+        (ArrayNode)
+            getEntity(
+                PathUtils.urlPath(context.getBaseUrl(), API_TAXONOMY_PATH, taxonomyUuid, "term"),
+                getToken(),
```
edalex-ian: It'd be good to keep this DRY too like mentioned above for `sortchildren`.

autotest/OldTests/src/test/java/com/tle/webtests/test/webservices/rest/TaxonomyApiTest.java
```diff
@@ -120,13 +127,201 @@ public void testSearchTerms() throws IOException {
...
+        createTerm(taxonomyUuid, null, subTermName, termUuid);
+      }
+    }
+
+    // These are the orders we will expect
+    // Sort using same mechanism as server
+    final Collator collator = Collator.getInstance(Locale.getDefault());
+    collator.setStrength(Collator.PRIMARY);
+    Collections.sort(rootTerms, collator);
+    Collections.sort(testTermChildren, collator);
```
edalex-ian: I've pondered over this block of code. As it's very close to simply repeating the code from the server, and so of course you'd expect it to achieve the same results. Something you typically avoid - but admittedly in this case it's not identical.

However it got me wondering if you'd be better not using a random input source set, and then you could also have an explicit expected result.

But if you did still want to stay with the random approach, I'd suggest you do not use `Collator` at all so as to ensure you use a different implementation for validation.

autotest/OldTests/src/test/java/com/tle/webtests/test/webservices/rest/TaxonomyApiTest.java
```diff
@@ -120,13 +127,201 @@ public void testSearchTerms() throws IOException {
...
+    final Collator collator = Collator.getInstance(Locale.getDefault());
+    collator.setStrength(Collator.PRIMARY);
+    Collections.sort(rootTerms, collator);
+
+    // Sort root
+    final HttpResponse sortRootsResponse =
+        postEntity(
+            null,
+            PathUtils.urlPath(
+                context.getBaseUrl(), API_TAXONOMY_PATH, taxonomyUuid, "sortchildren"),
```
edalex-ian: `sortchildren` is repeated several times and as a result should be in a string constant. (Although I do mention below refactoring into a method, maybe then this wouldn't be duplicated.)

autotest/OldTests/src/test/java/com/tle/webtests/test/webservices/rest/TaxonomyApiTest.java
```diff
@@ -120,13 +127,201 @@ public void testSearchTerms() throws IOException {
...
+            getToken(),
+            true);
+    assertResponse(sortRootsResponse, 200, "failed to sort root terms");
+
+    final List<String> returnedSortedRootTerms = new ArrayList<>(10);
+    final ArrayNode arrayNode =
+        (ArrayNode)
+            getEntity(
+                PathUtils.urlPath(context.getBaseUrl(), API_TAXONOMY_PATH, taxonomyUuid, "term"),
```
edalex-ian: `term` is repeated several times and as a result should be in a string constant. (Although I do mention below refactoring into a method, maybe then this wouldn't be duplicated.)

Source/Plugins/Core/com.equella.core/src/com/tle/web/api/taxonomy/interfaces/TaxonomyResource.java
```diff
@@ -0,0 +1,303 @@
...
+  Response delete(
+      @Context UriInfo uriInfo, @ApiParam(value = "Taxonomy UUID") @PathParam("uuid") String uuid);
+
+  @POST
+  @ApiOperation(value = "Create a new taxonomy", notes = "Create new or list existing taxonomies")
+  @ApiResponses({@ApiResponse(code = 201, message = "Location: {newtaxonomy uri}")})
+  Response create(
+      @Context UriInfo uriInfo,
+      @ApiParam(value = "Taxonomy") TaxonomyBean bean,
+      @ApiParam(required = false) @QueryParam(value = "file") String stagingUuid);
```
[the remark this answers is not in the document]
abidingotter: stagingUuid is actually more descriptive than the "file" API parameter.  I think it's probably exposed as "file" because it's meant to abstract away the fact that it's really a staging UUID.

Source/Plugins/Core/com.equella.core/src/com/tle/web/api/taxonomy/interfaces/TaxonomyResource.java
```diff
@@ -0,0 +1,303 @@
...
+
+  @PUT
+  @Path("/{uuid}")
+  @ApiOperation("Edit a taxonomy")
+  @ApiResponses({@ApiResponse(code = 200, message = "Location: {taxonomy uri}")})
+  Response edit(
+      @Context UriInfo uriInfo,
+      @ApiParam(value = "Taxonomy UUID") @PathParam("uuid") String uuid,
+      @ApiParam TaxonomyBean bean,
+      @ApiParam(required = false, value = "Staging area UUID") @QueryParam("file")
```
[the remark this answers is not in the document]
abidingotter: As above

Source/Plugins/Core/com.equella.core/src/com/tle/web/api/taxonomy/interfaces/TaxonomyResource.java
abidingotter: Removed

Source/Plugins/Core/com.equella.core/src/com/tle/web/core/filter/LoggingContextFilter.java
abidingotter: No-one should be processing this log, it's not documented anywhere.  I only added it some time ago for my own purposes (and so we could tell clients to turn it on if we needed to diagnose an issue)

abidingotter: Note my comment in the code :)
```
// This is for DEV ONLY.  Do not use in a test / staging / prod environment.
// There is no guarantee that all sensitive information will be blanked!
```

autotest/OldTests/src/test/java/com/tle/webtests/test/webservices/rest/TaxonomyApiTest.java
abidingotter: It would have to use the Collator, otherwise it does an ASCII comparison: BANANA < apple
I like the random approach as it is more likely to catch incorrect sorting than a handful of unchanging values.

It's not really testing that the actual sorting is working (it really should work, and perhaps we could add an actual unit test for that), it's more about how the term index changes have been captured correctly, are stored in the database correctly, and are fetched back from the database in the right order.  This was the part that was quite challenging to get right, and is the most likely part of the implementation to behave incorrectly.

autotest/OldTests/src/test/java/com/tle/webtests/test/webservices/rest/TaxonomyApiTest.java
abidingotter: Have significantly refactored the test class

autotest/OldTests/src/test/java/com/tle/webtests/test/webservices/rest/TaxonomyApiTest.java
abidingotter: Have significantly refactored the test class

autotest/OldTests/src/test/java/com/tle/webtests/test/webservices/rest/TaxonomyApiTest.java
abidingotter: Have significantly refactored the test class

autotest/OldTests/src/test/java/com/tle/webtests/test/webservices/rest/TaxonomyApiTest.java
abidingotter: Have significantly refactored the test class

PenghaiZhang:
Hi @abidingotter, in order to pass the Travis CI build checks we need to reduce the amount of terms to be created from 1000 to 10.

autotest/project/build.properties
```diff
@@ -0,0 +1 @@
+sbt.version=1.3.0-RC1
```
edalex-ian: @PenghaiZhang can you please tweak this to make sure we match the same version as the root build.sbt. Otherwise this would be a third version being used in the one code base. (At some-point we should update the one in IntegTester.)

PR #1226, opened 2019-09-06
Configure Stale Bot

edalex-ian:
Following the Community Developer Meeting it was agreed to use Stale Bot
to try and keep our Issue List under control. After enabling it on our
organisation, I'm now following the instructions as provided in the
Installed GitHub Apps for Stale.

These are essentially the recommended defaults.

As per https://github.com/organizations/equella/settings/installations/1802878


.gitlab/stale.yml
```diff
@@ -0,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
```
ChristianMurphy: the docco for stale bot mentions `.github/stale` as the folder.
https://github.com/probot/stale#usage

will `.gitlab/stale` also work?

.gitlab/stale.yml
```diff
@@ -0,0 +1,17 @@
...
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
```
ChristianMurphy: some other labels that might be exempt-able.
- tech debt
- roadmap
- edelex
- unicon se

thoughts? 💭

.gitlab/stale.yml
edalex-ian: Ah bugga. Sorry, my bad. Jumping between too many different systems. Will fix.

.gitlab/stale.yml
edalex-ian: Good idea. I'll add them. 👍

PR #1229, opened 2019-09-07
fix: correct typo in theme options

ChristianMurphy:


edalex-ian:
This looks fine. But I wonder where `autotestMode` is turned on. :thinking: 

Could be worth checking (and maybe documenting).. Or whether this is dead code.

ChristianMurphy:
I was curious about that too.
This is the conditional:
https://github.com/openequella/openEQUELLA/blob/57f8452686e274f9d293fd1bca2b09ea1daaa7c7/Source/Plugins/Core/com.equella.core/js/tsrc/theme/index.ts#L52

It appears this may be able to trigger the conditional
https://github.com/openequella/openEQUELLA/blob/e55409d824f3caba1439e4bd7b29d24a7c06381a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/template/RenderNewTemplate.scala#L128-L134

edalex-ian:
Ah yes, I just eventually found that too. :)

Cool, I reckon it's good to go then. :+1:

PR #1230, opened 2019-09-07
build: upgrade react visibility sensor

ChristianMurphy:
remove custom typing, the latest version provides a typing.
remove require import, es6 style import is now suported.


Source/Plugins/Core/com.equella.core/js/tsrc/course/SearchCourse.tsx
```diff
@@ -8,7 +8,7 @@ import ConfirmDialog from "../components/ConfirmDialog";
 import SearchResult from "../components/SearchResult";
 import { courseService } from "../services";
 import { formatSize, languageStrings } from "../util/langstrings";
-import VisibilitySensor = require("react-visibility-sensor");
+import VisibilitySensor from "react-visibility-sensor";
```
PenghaiZhang: Hi @ChristianMurphy , just one more place needs this change :  `tsrc/api/errors.ts`

ChristianMurphy:
test failure looks like a timeout (probably a false positive).
I can't see the codebuild logs, any issues there?

ChristianMurphy:
@edalex-ian any insights?

edalex-ian:
Interesting @ChristianMurphy 

Both failed the `com.tle.webtests.test.contribute.controls.asc.AdvancedScriptControlTests` tests under New UI. And considering this is a change that would impact New UI, I'd suggest its possibly breaking something.

But to be sure, to be sure, I've retries the Travis CI one. If it fails again though, then looks like it'll need to be addressed.

edalex-ian:
Hmm, bugga. Looks like ASC still not happy.

At least since when you first submitted this, you can now see the whole run details in the Travis build.

PR #1231, opened 2019-09-07
refactor: update prop types to use interfaces and destructuring

ChristianMurphy:
This is a pattern used through a majority of the code, bringing the few places that are different, over to the standard.


ChristianMurphy:
>  Previously, I can easily see what parameters are required. With the change I must open `LegacyForm.tsx` to know what `isLegacyFormProps`.

It's true that that TypeScript in VSCode currently shows the interface name rather than the properties of that interface.
That is a developer experience choice of VSCode, which I agree, at times, is not ideal.

The trade off here is that typings for properties declared directly in the function cannot be reused, interfaces can.

PR #1232, opened 2019-09-08
refactor: add missing typings

ChristianMurphy:
Typescript's `any` feature disables type checking for a variable, this change goes through areas that have been disabled and adds appropriate types.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/components/index.ts
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/components/index.ts
@@ -7,7 +7,8 @@ export { default as Loader } from "./Loader";
 export { default as AppBarQuery } from "./AppBarQuery";
 export { default as SearchResult } from "./SearchResult";
 
-export const handleUnexpectedApiError = <P extends TemplateUpdateProps>({
+export function handleUnexpectedApiError<P extends TemplateUpdateProps>({
   props: { updateTemplate }
-}: React.Component<P>) => (err: Error) =>
-  updateTemplate(templateError(generateFromError(err)));
+}: React.Component<P>) {
+  return (err: Error) => updateTemplate(templateError(generateFromError(err)));
+}
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/components/index.ts
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/components/index.ts
@@ -7,10 +7,7 @@ export { default as Loader } from "./Loader";
 export { default as AppBarQuery } from "./AppBarQuery";
 export { default as SearchResult } from "./SearchResult";
 
-export function handleUnexpectedApiError<P extends TemplateUpdateProps>(
-  t: React.Component<P>
-): (err: any) => void {
-  return function(err) {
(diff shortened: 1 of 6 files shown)


edalex-ian:
Looks good, but unfortunately we don't have test coverage on some of these areas I fear. :thinking: 

At least one bit of this code is yours @SammyIsConfused , so please review.

The the question is how to we test it. Seems the areas of interest would be:

* Course editor
* General display of errors
* And places validating integers
* Plus some search results - which is curious, as I thought the current New Search UI was all PS (but maybe this is something else)

Did you have a chance to test any of this @ChristianMurphy ?

ChristianMurphy:
I tested this by running `npm run build`.
The changes updates the typescript typings, which exist only during compile time and do not affect the run time.

The one change that could affect the run time is `validation.ts`, which I checked with a truth table.
But have not done extensive testing of course service.

ChristianMurphy:
ping @edalex-ian :bell:

Source/Plugins/Core/com.equella.core/js/tsrc/util/validation.ts
```diff
@@ -1,16 +1,12 @@
 export function isInteger(
-  val: any,
+  val: number | undefined,
```
abidingotter: How about `val?: number`

Source/Plugins/Core/com.equella.core/js/tsrc/components/index.ts
```diff
@@ -7,10 +7,7 @@ export { default as Loader } from "./Loader";
...
 export { default as SearchResult } from "./SearchResult";
 
-export function handleUnexpectedApiError<P extends TemplateUpdateProps>(
-  t: React.Component<P>
-): (err: any) => void {
-  return function(err) {
-    t.props.updateTemplate(templateError(generateFromError(err)));
-  };
-}
+export const handleUnexpectedApiError = <P extends TemplateUpdateProps>({
```
abidingotter: I would argue that `export const` is less readable than `export function`

ChristianMurphy: Fair, my tendency is to use arrow functions in any situation where I don't need to [`bind`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind) the `function`.
I'd be fine walking that back to a `function` if you feel strongly that `function` is more readable.

abidingotter: I'm not that concerned about it :)  Just thought that `export function` tells you right up what is being exported, without you having to decide what it is.  I agree about use of arrow functions elsewhere.

Source/Plugins/Core/com.equella.core/js/tsrc/util/validation.ts
ChristianMurphy: Updated

Source/Plugins/Core/com.equella.core/js/tsrc/components/index.ts
ChristianMurphy: switched to `function`

Issue #1199, opened 2019-08-28
Item and attachment view counts not working properly


cathfitz:
**Describe the bug**
The view count functionality at both an attachment and item level is flaky - item level doesn't seem to work at, and attachment level works only sometimes

**To Reproduce**

1. Make sure that the VIEW_VIEWCOUNT ACL is granted to some users
2. Log in as a user and go to Search
3. Open an item with attachments, then click on the attachment links within that item to view them.
4. Note the Views: numbers in the Details part of the summary page (stays at 0 for items)
5. Use the attachment drop-down to view the Views: for each attachment (sometimes they increase by one, sometimes they don't).
6. Log in as a different user and go to the same item and check the Views areas. The item level count still says 0. View each attachment then check the counts. The view number may or may not have increased.
Repeat the above for a few different items.

**Expected behavior**
When an item is first opened in a session, the View count should increase by 1. (Note: if smart logging is set in the mandatory.properties file, this count will only go up once per user session, if not, it should go up each time you open that item).
Each time an attachment is viewed within a session, the View count should increase by 1.

NOTE: the xml for the items only shows a <views>x<views> section for each attachment if the system has registered a count >0. It doesn't show at all at an item level (presumably because it's not registering a view at all at an item level).

**Screenshots**

![image](https://user-images.githubusercontent.com/32081370/63826421-386c0900-c9a2-11e9-87bb-f651f662ce02.png)

**Platform:**

- OpenEquella Version: [e.g. 2019.1,]
- OS: [e.g. Windows 10, MacOS 10.13]
- Browser [e.g. Chrome, Firefox]

abidingotter:
I suspect that the attachment views (sometimes) not increasing is due to browser caching.  I.e. if you view an image attachment a second time, it will use an image already downloaded by the browser into a temp folder, so Equella doesn't know about the viewing.

The item summary views not increasing is definitely a bug, and introduced by the fix to #856

abidingotter:
This should not be an issue in 2018.2.  Can someone please confirm?

abidingotter:
The large image viewer doesn't trigger an attachment view

PR #1248, opened 2019-09-13
Fix item summary view count

abidingotter:
(No longer a huge commit)


edalex-ian:
The test appear broken.

abidingotter:
#1199 tweaked the PR so that is only a small edit of institution.  The PR now contains a mod to the installer that allows you to select your auditing options.  This was done so that autotests can run with auditing enabled.

abidingotter:
Note: turn off whitespace changes when viewing the PR!  :)

edalex-ian:
Unfortunately @abidingotter , this is still failing (on all three cases) in test `com.tle.webtests.test.viewing.VersionShowAllTest`.

abidingotter:
@edalex-ian Yeah I saw that, but didn't get time to see what the issue was.

Dev/learningedge-config/build.sbt
```diff
@@ -44,6 +45,11 @@ prepareDevConfig := {
   imc.setProperty("imageMagick.path", imPath)
   imc.save(baseDir / "plugins/com.tle.core.imagemagick/config.properties")
 
+  val vic = new PropertiesConfiguration()
```
PenghaiZhang: Does vic stand for 'view item count'?

Installer/resources/script/app-script.xml
```diff
@@ -1,364 +1,390 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
PenghaiZhang: I have never played these two big xml files. Are they used when the installer gets running?

autotest/codebuild.conf
```diff
@@ -16,4 +16,5 @@ tests {
 
 install {
   dir = ${HOME}
+  auditLevel = "NORMAL"
```
PenghaiZhang: We already have 'auditLevel' configured in `autotest-defaults.conf` so that tests can run properly. 
I wonder if we still need `auditLevel` set in both codebuild and travis configure files?

Dev/learningedge-config/build.sbt
abidingotter: It stands for "view item configuration" :)  I was copying the existing bad variable names.

Installer/resources/script/app-script.xml
abidingotter: Yes.  All the pages of the installer are defined in these XML files.

autotest/codebuild.conf
abidingotter: True.  It should be NORMAL by default, not NONE, so that autotests will work in any environment.  I will revert these files.

Dev/learningedge-config/build.sbt
abidingotter: Renamed

autotest/codebuild.conf
abidingotter: Reverted

Issue #1198, opened 2019-08-28
Package Viewer: Error messages getting rendered with Side menu, header and footer


mrblippy:
**Describe the bug**
When an error occurs inside the Package Viewer (like a 404 error) the full oEQ interface is displayed, with Side menu, header and footer
![current](https://user-images.githubusercontent.com/4625498/63826482-97318280-c9a2-11e9-92bd-6a1845cb106a.PNG)

That's different to how it used to work in 6.5 and earlier. Only the error was displayed (this screenshot from a client install, so it has different styling applied 
![old](https://user-images.githubusercontent.com/4625498/63825959-1b363b00-c9a0-11e9-9ee8-3f477d5343ac.jpg)

I would say something happened with this during the great plugin merge when the product went open source?

A good place to start looking to fix this is 
https://github.com/openequella/openEQUELLA/blob/786e43d70d65e453dbb8c25b3d57808520bad624/Source/Plugins/Core/com.equella.core/src/com/tle/web/template/RenderTemplate.java#L168

Happens in old and new ui

**To Reproduce**

1. Contribute/publish an item with the following package 
[ContentPackagingOneFilePerSCO_SCORM12.zip](https://github.com/openequella/openEQUELLA/files/3549122/ContentPackagingOneFilePerSCO_SCORM12.zip)
2. Click the scorm attachment in the item summary to open the Package Viewer
3. Click on the 'How to Have Fun Playing Golf' link

**Platform:**

- OpenEquella Version: 2018.2, 2019.1

PR #1279, opened 2019-09-27
Dirtiness to prevent navigation bars on error screen

abidingotter:
when the original request is already hiding them

#1198

### DIFF
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/web/viewitem/section/RootItemFileSection.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/web/viewitem/section/RootItemFileSection.java
@@ -25,10 +25,12 @@ import com.tle.beans.item.Item;
 import com.tle.beans.item.attachments.Attachment;
 import com.tle.beans.item.attachments.IAttachment;
 import com.tle.common.Check;
+import com.tle.common.URLUtils;
 import com.tle.common.usermanagement.user.CurrentUser;
 import com.tle.core.institution.InstitutionService;
 import com.tle.core.mimetypes.MimeTypeService;
 import com.tle.core.services.FileSystemService;
+import com.tle.core.services.UrlService;
 import com.tle.exceptions.AccessDeniedException;
 import com.tle.web.integration.Integration;
 import com.tle.web.integration.IntegrationSessionData;
@@ -43,6 +45,7 @@ import com.tle.web.sections.SectionUtils;
 import com.tle.web.sections.annotations.Bookmarked;
 import com.tle.web.sections.annotations.DirectEvent;
 import com.tle.web.sections.events.BeforeEventsListener;
+import com.tle.web.sections.events.BookmarkEvent;
 import com.tle.web.sections.events.ForwardEventListener;
 import com.tle.web.sections.events.RenderEventContext;
 import com.tle.web.sections.events.SectionEvent;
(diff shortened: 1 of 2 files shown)


Source/Plugins/Core/com.equella.core/src/com/tle/web/viewitem/section/RootItemFileSection.java
```diff
@@ -145,7 +149,7 @@ public ViewItemResource getViewItemResource(SectionInfo info) {
       mimeType = mimeService.getMimeTypeForFilename(filename);
     }
     resource =
-        new BaseViewItemResource(getViewableItem(info), filename, mimeType, model.getViewer());
+        getBaseViewItemResource(getViewableItem(info), filename, mimeType, model.getViewer());
```
PenghaiZhang: I just looked into the method `getBaseViewItemResource` and it  returns a new `BaseViewItemResource`.  
So it seems that changing `new BaseViewItemResource` to `getBaseViewItemResource`  does not make any differences.

Source/Plugins/Core/com.equella.core/src/com/tle/web/viewitem/section/RootItemFileSection.java
```diff
@@ -260,6 +265,20 @@ public SectionResult renderHtml(RenderEventContext info) throws IOException {
     }
   }
 
+  // Appalling hack to hide navigation when an error screen shows up...
+  private String modifyHref(SectionInfo info, String href) {
+    if (urlService.isRelativeUrl(href)) {
+      BookmarkEvent ev = new BookmarkEvent(info.lookupSection(RenderTemplate.class), false, info);
```
PenghaiZhang: I am not quite familiar with this event, but from the comments written in this file, I guess we want to fire a BookmarkEvent to stop `RenderTemplate` rendering not only the navigation bar but also the header. 
The important thing is the second parameter `isRendering` is set to false.

edalex-ian:
This appears to break `com.tle.webtests.test.drm.DRMTest`. I've rerun the Admin tests (old UI) several times now, and it keeps hitting this.

```
[TestNG] INVOKING: "Admin" - com.tle.webtests.test.drm.DRMTest.attachmentTest()
java.lang.AssertionError: expected [http://localhost:8080/fiveo/file/677a4bbc-defc-4dc1-b68e-4e2473b66a6a/1/Biscuit factory complex ratios.zip/index.html] but found [http://localhost:8080/fiveo/file/677a4bbc-defc-4dc1-b68e-4e2473b66a6a/1/Biscuit factory complex ratios.zip/index.html?]
	at org.testng.Assert.fail(Assert.java:93)
	at org.testng.Assert.failNotEquals(Assert.java:512)
	at org.testng.Assert.assertEqualsImpl(Assert.java:134)
	at org.testng.Assert.assertEquals(Assert.java:115)
	at org.testng.Assert.assertEquals(Assert.java:189)
	at org.testng.Assert.assertEquals(Assert.java:199)
	at com.tle.webtests.test.drm.DRMTest.attachmentTest(DRMTest.java:235)
```

Source/Plugins/Core/com.equella.core/src/com/tle/web/viewitem/section/RootItemFileSection.java
abidingotter: Yeah, it doesn't make a difference, but the function was already there, and if there is extra logic to create one in the future it will be in a single place.

PR #1287, opened 2019-10-07
Fix the bug that the file attachment check treats the file itself as a duplicate

PenghaiZhang:
At the moment, if a file attachment has been published and then when editing this file, the duplicate warning appears despite no duplicates being really found. This is because duplicates are found by the file MD5 and the file itself is in the result list. So the duplicate check treats the file itself as a duplicate. 

So this PR aims to fix this issue by excluding the file itself if it has been published.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/impl/WizardServiceImpl.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/impl/WizardServiceImpl.java
@@ -766,11 +766,10 @@ public class WizardServiceImpl
   @Override
   public void checkFileAttachmentDuplicate(WizardState state, String fileName, String fileUuid) {
     try {
-      String md5 = fileSystemService.getMD5Checksum(state.getFileHandle(), fileName);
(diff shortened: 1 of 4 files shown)


Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/impl/WizardServiceImpl.java
```diff
@@ -769,6 +769,8 @@ public void checkFileAttachmentDuplicate(WizardState state, String fileName, Str
       String md5 = fileSystemService.getMD5Checksum(state.getFileHandle(), fileName);
       List<Attachment> duplicateFileAttachments =
           attachmentDao.findByMd5Sum(md5, state.getItemDefinition(), true);
+      // Exclude the file itself if it has been published.
+      duplicateFileAttachments.removeIf(attachment -> attachment.getUuid().equals(fileUuid));
```
edalex-ian: Instead of doing it post DB call, can we update `findByMd5Sum` to consider the `fileUuid` in the DB query?

PR #1288, opened 2019-10-07
view a deleted owner's name with a deleted indicator

PenghaiZhang:
View a deleted owner's name with a deleted indicator on the Item Summary and Moderation History pages.

Here are a few more explanations for how this work is done.
1. oEQ uses a UI component called `UserLinkState` to render user names.
2. Before this component gets rendered, it calls `getTitle`  and `getLabel` to prepare what will be rendered. Then, these two methods call `ensureLookup` to collect user details.
3. Change the signature of a few methods to include one more boolean parameter `useLastOwnerDetails`, making sure changes only affect  Item Summary and Moderation History pages.
4. When useLastOwnerDetails is true and no details of current user are found, the new method `findUserInfoBackup` created in the UserService is called to retrieve user details from the user_info_backup table.

Attached is a screenshot for illustration.  For example,  'Aaron Holland' is a user who has been deleted externally. So instead of `unknown user`,  'Aaron Holland' is displayed, and the deleted indicator ' (removed)' is also appended.

Screenshot from 2019-10-07 14-28-42


Source/Plugins/Core/com.equella.core/src/com/tle/web/sections/equella/utils/UserLinkSection.java
```diff
@@ -65,6 +66,9 @@
   @PlugKey("userlink.systemuser")
   private static Label LABEL_SYSTEMUSER;
 
+  @PlugKey("userlabel.lastownersuffix")
```
edalex-ian: Maybe we should stick with the existing namespacing:

```suggestion
  @PlugKey("userlink.lastownersuffix")
```

PenghaiZhang:
In order to have a test to cover this feature, a few things are done.
1. Create a new `converter` to enable importing data to 'user_info_backup',  so that  `setupForTest` can prepare necessary testing data.
2. `UserInfoBackupDao` has two more methods to support point 1.
3. A new Selenium test will check if the details of last owner is displayed.

PR #1290, opened 2019-10-09
Sass->css compilation of legacy css

mrblippy:
Functionality included:

- Server-side sass->css compilation, using values from NewUiTheme. 
- Basic caching has been implemented to ensure that compilation only happens on first visit to any new ui page, or there is a newer version of the source sass file,or the theme editor is changed
- Client rendering of sass compiled css for legacy content

Only commit c15136f needs to be reviewed

Note: The majority of the contents of legacy.scss is placeholder (copied from the original legacy css file).  So there's not much point in reviewing that part yet. It will be updated in subsequent tickets.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/package.json
+++ b/Source/Plugins/Core/com.equella.core/js/package.json
@@ -21,8 +21,7 @@
     "bundleps:main": "cross-env-shell \"purs bundle \"${npm_package_config_psglob}\" -m Bridge --main Bridge -o entrybuild/target/main.js\"",
     "bundleps:selection": "cross-env-shell \"purs bundle \"${npm_package_config_psglob}\" -m OEQ.SelectionUI.Main --main OEQ.SelectionUI.Main -o entrybuild/target/selection.js\"",
     "bundleps:uploadlist": "cross-env-shell \"purs bundle \"${npm_package_config_psglob}\" -m Uploads.UploadList --main Uploads.UploadList -o entrybuild/target/uploadlist.js\"",
-    "bundleps:langbundle": "cross-env-shell \"purs bundle \"${npm_package_config_psglob}\" -m Tools.GenLangStrings --main Tools.GenLangStrings -o target/genlang.js\"",
-    "scss": "sass ../resources/web/css/saas/legacy.scss ../resources/web/css/saas/output/legacy.css"
+    "bundleps:langbundle": "cross-env-shell \"purs bundle \"${npm_package_config_psglob}\" -m Tools.GenLangStrings --main Tools.GenLangStrings -o target/genlang.js\""
   },
   "dependencies": {
     "@date-io/luxon": "1.3.9",
--- a/Source/Plugins/Core/com.equella.core/js/package.json
+++ b/Source/Plugins/Core/com.equella.core/js/package.json
@@ -21,7 +21,8 @@
     "bundleps:main": "cross-env-shell \"purs bundle \"${npm_package_config_psglob}\" -m Bridge --main Bridge -o entrybuild/target/main.js\"",
     "bundleps:selection": "cross-env-shell \"purs bundle \"${npm_package_config_psglob}\" -m OEQ.SelectionUI.Main --main OEQ.SelectionUI.Main -o entrybuild/target/selection.js\"",
     "bundleps:uploadlist": "cross-env-shell \"purs bundle \"${npm_package_config_psglob}\" -m Uploads.UploadList --main Uploads.UploadList -o entrybuild/target/uploadlist.js\"",
-    "bundleps:langbundle": "cross-env-shell \"purs bundle \"${npm_package_config_psglob}\" -m Tools.GenLangStrings --main Tools.GenLangStrings -o target/genlang.js\""
+    "bundleps:langbundle": "cross-env-shell \"purs bundle \"${npm_package_config_psglob}\" -m Tools.GenLangStrings --main Tools.GenLangStrings -o target/genlang.js\"",
+    "scss": "sass ../resources/web/css/saas/legacy.scss ../resources/web/css/saas/output/legacy.css"
   },
   "dependencies": {
     "@date-io/luxon": "1.3.9",
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContentRenderer.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContentRenderer.tsx
@@ -2,6 +2,7 @@ import * as React from "react";
 import JQueryDiv from "./JQueryDiv";
 import { makeStyles } from "@material-ui/styles";
 import { PageContent } from "./LegacyContent";
+import "./legacy.scss";
 
 const useStyles = makeStyles(t => ({
   withPadding: {
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContentRenderer.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContentRenderer.tsx
@@ -2,7 +2,6 @@ import * as React from "react";
 import JQueryDiv from "./JQueryDiv";
 import { makeStyles } from "@material-ui/styles";
 import { PageContent } from "./LegacyContent";
-import "./legacy.scss";
 
 const useStyles = makeStyles(t => ({
   withPadding: {
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/LegacyContentApi.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/LegacyContentApi.scala
@@ -480,7 +480,7 @@ class LegacyContentApi {
 
       val scripts  = preRenderPageScripts(context, context).map(_.getStatements(context))
       val jsFiles  = context.getJsFiles.asScala
-      val cssFiles = Iterable(RenderTemplate.LEGACY_CSS.getHref(context))
+      val cssFiles = Iterable()
       val title =
         Option(decs.getBannerTitle).orElse(Option(decs.getTitle)).map(_.getText).getOrElse("")
       val menuMode       = decs.getMenuMode.toString
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/LegacyContentApi.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/LegacyContentApi.scala
@@ -478,11 +478,9 @@ class LegacyContentApi {
             JQueryCore.JQUERY,
             new AnonymousFunction(new StatementBlock(ready).setSeperate(true))))
 
-      val scripts = preRenderPageScripts(context, context).map(_.getStatements(context))
-      val jsFiles = context.getJsFiles.asScala
-      val cssFiles = context.getCssFiles.asScala.collect {
-        case css: CssInclude => css.getHref(context)
-      }
+      val scripts  = preRenderPageScripts(context, context).map(_.getStatements(context))
+      val jsFiles  = context.getJsFiles.asScala
+      val cssFiles = Iterable(RenderTemplate.LEGACY_CSS.getHref(context))
       val title =
         Option(decs.getBannerTitle).orElse(Option(decs.getTitle)).map(_.getText).getOrElse("")
       val menuMode       = decs.getMenuMode.toString
@@ -608,6 +606,7 @@ class LegacyContentApi {
     renderAjaxBody(renderedBody)
     val responseCallback = arc.getJSONResponseCallback
     info.setRendered()
+    arc.clearCss()
     Response.ok(responseCallback.getResponseObject(arc))
   }
 }
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/LegacyContentApi.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/LegacyContentApi.scala
@@ -80,7 +80,6 @@ case class MenuItem(title: String,
                     newWindow: Boolean)
 
 case class LegacyContent(html: Map[String, String],
-                         css: Iterable[String],
                          js: Iterable[String],
                          script: String,
                          state: Map[String, Array[String]],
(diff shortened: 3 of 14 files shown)


Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/service/ThemeSettingsService.java
```diff
@@ -23,18 +23,23 @@
...
-  void setTheme(NewUITheme theme) throws JsonProcessingException;
+  void setTheme(NewUITheme theme) throws JsonProcessingException, IOException;
 
   void setLogo(File logoFile) throws IOException;
 
   boolean isCustomLogo();
 
   void deleteLogo();
+
+  InputStream compileSass() throws IOException, URISyntaxException;
```
edalex-ian: I'm not convinced this needs to be part of the interface. Should just be internal and private.

Probably remove from here.

Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/service/impl/ThemeSettingsServiceImpl.java
```diff
@@ -146,4 +157,63 @@ private String themeToJSONString(NewUITheme theme) throws JsonProcessingExceptio
     themeToString = objectMapper.writeValueAsString(theme);
     return themeToString;
   }
+
+  public String newUIToSass(NewUITheme theme) throws IOException {
```
edalex-ian: Should be private...

```suggestion
  private String newUIToSass(NewUITheme theme) throws IOException {
```

I also wonder of the name of this. Hmmm, needs more than just `newUI`... Maybe `newUIThemToSassVars`... Longish, but maybe warranted...

Actually, why not move this functionality to a method on the `NewUITheme` class. Maybe `toSassVars()`. Then done on line 201 you'd have:

```
fileSystemService.write(
        staging, SASS_VARS_FILENAME, new StringReader(getTheme().toSassVars()), false);
```

Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/service/impl/ThemeSettingsServiceImpl.java
```diff
@@ -146,4 +157,63 @@ private String themeToJSONString(NewUITheme theme) throws JsonProcessingExceptio
...
+          > fileSystemService.lastModified(customisationFile, LEGACY_CSS_FILENAME)) {
+        compileSass();
+      }
+    } else {
+      compileSass();
+    }
+    return fileSystemService.read(customisationFile, LEGACY_CSS_FILENAME);
+  }
+
+  public InputStream compileSass() throws IOException {
```
edalex-ian: As mentioned in the `interface`, seems this only needs to be private to this class:

```suggestion
  private InputStream compileSass() throws IOException {
```

Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/service/impl/ThemeSettingsServiceImpl.java
```diff
@@ -146,4 +157,63 @@ private String themeToJSONString(NewUITheme theme) throws JsonProcessingExceptio
...
+    CustomisationFile customisationFile = new CustomisationFile();
+    StagingFile staging = stagingService.createStagingArea();
```
edalex-ian: Why using the staging area for this rather than Temp?

Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/service/impl/ThemeSettingsServiceImpl.java
```diff
@@ -146,4 +157,63 @@ private String themeToJSONString(NewUITheme theme) throws JsonProcessingExceptio
...
+  public InputStream getLegacyCss() throws IOException {
+    File baseLegacySass =
+        new File(getClass().getResource("/web/sass/" + SASS_LEGACY_CSS_FILENAME).getFile());
+    File baseLegacyOutput;
+    CustomisationFile customisationFile = new CustomisationFile();
+
+    if (fileSystemService.fileExists(customisationFile, LEGACY_CSS_FILENAME)) {
+      if (baseLegacySass.lastModified()
+          > fileSystemService.lastModified(customisationFile, LEGACY_CSS_FILENAME)) {
+        compileSass();
```
PenghaiZhang: Maybe it would be better if change to :
boolean fileExist = fileSystemService.fileExists(customisationFile, LEGACY_CSS_FILENAME);
boolean lastModified = baseLegacySass.lastModified()
          > fileSystemService.lastModified(customisationFile, LEGACY_CSS_FILENAME);

And then one `if` should work.

Source/Plugins/Core/com.equella.core/js/package.json
```diff
@@ -21,7 +21,8 @@
     "bundleps:main": "cross-env-shell \"purs bundle \"${npm_package_config_psglob}\" -m Bridge --main Bridge -o entrybuild/target/main.js\"",
     "bundleps:selection": "cross-env-shell \"purs bundle \"${npm_package_config_psglob}\" -m OEQ.SelectionUI.Main --main OEQ.SelectionUI.Main -o entrybuild/target/selection.js\"",
     "bundleps:uploadlist": "cross-env-shell \"purs bundle \"${npm_package_config_psglob}\" -m Uploads.UploadList --main Uploads.UploadList -o entrybuild/target/uploadlist.js\"",
-    "bundleps:langbundle": "cross-env-shell \"purs bundle \"${npm_package_config_psglob}\" -m Tools.GenLangStrings --main Tools.GenLangStrings -o target/genlang.js\""
+    "bundleps:langbundle": "cross-env-shell \"purs bundle \"${npm_package_config_psglob}\" -m Tools.GenLangStrings --main Tools.GenLangStrings -o target/genlang.js\"",
+    "scss": "sass ../resources/web/css/saas/legacy.scss ../resources/web/css/saas/output/legacy.css"
```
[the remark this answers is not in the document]
mrblippy: Yeah, this was leftover from previous pull request. I will remove

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContentRenderer.tsx
```diff
@@ -2,6 +2,7 @@ import * as React from "react";
 import JQueryDiv from "./JQueryDiv";
 import { makeStyles } from "@material-ui/styles";
 import { PageContent } from "./LegacyContent";
+import "./legacy.scss";
```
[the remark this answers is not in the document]
mrblippy: Yeppers

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/LegacyContentApi.scala
```diff
@@ -608,6 +606,7 @@ class LegacyContentApi {
     renderAjaxBody(renderedBody)
     val responseCallback = arc.getJSONResponseCallback
     info.setRendered()
+    arc.clearCss()
```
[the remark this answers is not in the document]
mrblippy: Everytime an ajax request gets made using sections it was pulling in the old css files. 
I added that clearCss call in to ensure old ui css files didn't get loaded while using the newui, while leaving the old ui functionality untouched

Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/service/impl/ThemeSettingsServiceImpl.java
```diff
@@ -146,4 +157,63 @@ private String themeToJSONString(NewUITheme theme) throws JsonProcessingExceptio
...
+    try {
+      Output output =
+          compiler.compileFile(
+              fileSystemService.getExternalFile(staging, SASS_LEGACY_CSS_FILENAME).toURI(),
+              dstFile.toURI(),
+              options);
+      fileSystemService.write(
+          customisationFile, LEGACY_CSS_FILENAME, new StringReader(output.getCss()), false);
+    } catch (Exception e) {
+      e.printStackTrace();
```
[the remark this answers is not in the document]
mrblippy: Agreed, I just had that in for testing. But forgot to replace it with a logger before submitting the pull request

Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/service/ThemeSettingsService.java
mrblippy: Agreed

Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/service/impl/ThemeSettingsServiceImpl.java
mrblippy: Yeah, it makes more sense for it to be in the NewUITheme class, since it's more closely related to that

Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/service/impl/ThemeSettingsServiceImpl.java
mrblippy: I can't see much code inside core oEQ that uses temp files (apart from upgrader/installer code). So I thought I should be consistent

That being said, this is something that could be happening often (anytime the user alters their Themes). So maybe I should actually be using temp files?

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/LegacyContentApi.scala
```diff
@@ -478,11 +478,9 @@ class LegacyContentApi {
...
             new AnonymousFunction(new StatementBlock(ready).setSeperate(true))))
 
-      val scripts = preRenderPageScripts(context, context).map(_.getStatements(context))
-      val jsFiles = context.getJsFiles.asScala
-      val cssFiles = context.getCssFiles.asScala.collect {
-        case css: CssInclude => css.getHref(context)
-      }
+      val scripts  = preRenderPageScripts(context, context).map(_.getStatements(context))
+      val jsFiles  = context.getJsFiles.asScala
+      val cssFiles = Iterable()
```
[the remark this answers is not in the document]
mrblippy: Turns out this still needs to be there. The LegacyContentApi collects up all included css files that are referenced in Sections code. I don't want that old css to come over.

I guess I could refactor the case class definition on line 82 or LegacyContentApi. But I didn't know if it would break other parts of the code that I wasn't super familiar with

Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/service/impl/ThemeSettingsServiceImpl.java
mrblippy: Good idea Penghai

Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/service/impl/ThemeSettingsServiceImpl.java
edalex-ian: Hmmm, interesting... Could be a desire to ensure files are only within the confines of oEQ filestore. :thinking: 

Well, you've already got this code and it's consistent. So I say keep it as you've got it.

What'd be even better is if you could simply do it in mem, but after a quick read of the library you're using it looks like that's not an option. :'(

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/LegacyContentApi.scala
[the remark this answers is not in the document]
edalex-ian: Okay, cool. Could you put that comment in the code please above this call?

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/LegacyContentApi.scala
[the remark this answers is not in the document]
edalex-ian: So by my reading it's simply used to pass the response back to calls from within `LegacyContent.tsx` and then it's only used in two places and passed into a common method. So a refactor to remove this may not be too difficult, but probably warranted - otherwise we're leaving dead code around the place. Want to give it a go?

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/LegacyContentApi.scala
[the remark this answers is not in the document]
mrblippy: sure thing, it's a bit ambiguous otherwise

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/LegacyContentApi.scala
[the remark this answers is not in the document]
mrblippy: Yep, i think I initially got scared off by scala code when i first did it. But yeah, it now looks like not much effort to refactor, especially based on the knowledge I have now

Source/Plugins/Core/com.equella.core/src/com/tle/web/api/newuitheme/impl/NewUITheme.java
```diff
@@ -18,7 +18,13 @@
 
 package com.tle.web.api.newuitheme.impl;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.util.Map;
+
 public class NewUITheme {
+
+  private ObjectMapper objectMapper = new ObjectMapper();
+
```
edalex-ian: This is not actually a real member of this class, and it's only ever needed if they choose to use `toSassVars()`. As a result, best to just instantiate in `toSassVars()` and have as a local.

Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/service/impl/ThemeSettingsServiceImpl.java
```diff
@@ -212,7 +203,7 @@ public InputStream compileSass() throws IOException {
       fileSystemService.write(
           customisationFile, LEGACY_CSS_FILENAME, new StringReader(output.getCss()), false);
     } catch (Exception e) {
-      e.printStackTrace();
+      LOGGER.error("Failed to compile Sass to css " + e);
```
edalex-ian: If you want the stacktrace still optionally logged, you'd want to do:

    LOGGER.error("Failed to compile Sass to CSS.", e);

Source/Plugins/Core/com.equella.core/src/com/tle/web/api/newuitheme/impl/NewUITheme.java
mrblippy: Yep, that's a good idea. Will do

Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/service/impl/ThemeSettingsServiceImpl.java
mrblippy: Oops, thought that's what I was doing there. I will change it

edalex-ian: All good, that's a common one. ;)

mrblippy:
The autotests are failing because some elements are hidden (because they don't have any styling). 
Jsass isn't liking importing css files from inside jar files. Apparently you have to write your own custom importer to do that.

I'm just going to merge this into the feature branch as-is, and revisit the importing issue down the track, before the feature gets merged to develop

PR #1322, opened 2019-10-25
Log external link views as attachment views

abidingotter:
Also show view count on link attachments that was never shown before
(but should have been)

Resolves: #1199

### DIFF
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/web/cloud/search/CloudSearchListEntry.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/web/cloud/search/CloudSearchListEntry.java
@@ -19,10 +19,7 @@
 package com.tle.web.cloud.search;
 
 import com.dytech.devlib.PropBagEx;
-import com.google.common.collect.Maps;
 import com.tle.annotation.NonNullByDefault;
-import com.tle.beans.item.attachments.AttachmentType;
-import com.tle.beans.item.attachments.IAttachment;
 import com.tle.beans.item.attachments.UnmodifiableAttachments;
 import com.tle.core.cloud.beans.converted.CloudItem;
 import com.tle.core.guice.Bind;
@@ -43,8 +40,6 @@ import com.tle.web.sections.standard.model.HtmlLinkState;
 import com.tle.web.sections.standard.model.SimpleBookmark;
 import com.tle.web.sections.standard.renderers.LinkRenderer;
 import com.tle.web.viewable.ViewableItem;
-import java.util.Collections;
-import java.util.Map;
 import javax.inject.Inject;
 
 /** @author Aaron */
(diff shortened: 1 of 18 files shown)


PenghaiZhang:
Hi @abidingotter , before merging this PR I did some tests, and I found the issue still exists. 

Firstly, the audit level is NORMAL.
Then, I viewed an item which has a file attachment six times so the view count of item is 6, which is right. However, the view count of attachment is 1 all the time. I opened the attachment every time I viewed this item.

image

abidingotter:
Yes, it is because it's an image attachment that is being cached by the
browser. You will need to try a combination of deleting your browser cache,
force refreshing and/or closing the browser completely and starting it
again. Whatever works :)

PenghaiZhang:
Sorry @abidingotter , I accidentally deleted your comment...but I got your ideas. Thanks !

PenghaiZhang:
Hi @abidingotter , may I ask why when I click `Full screen` the view number increase by 1 if this behaviour is caused by the browser cache ?  Either way, the browser uses the cache. 
image

abidingotter:
It would be because it has a different URL.  Seeing that, I am wondering if
the change to do with error screens in content packages is likely to break
things (that change would be adding the temp.hb=true etc parameters to the
image URL)

Source/Plugins/Core/com.equella.core/src/com/tle/web/cloud/search/CloudSearchListEntry.java
```diff
@@ -124,127 +119,4 @@ protected void setupMetadata(RenderContext context) {
           new StdMetadataEntry(LABEL_MODIFIED, JQueryTimeAgo.timeAgoTag(item.getDateModified())));
     }
   }
-
-  @NonNullByDefault(false)
-  public static class CloudAttachmentDisplay implements IAttachment {
```
edalex-ian: Was this just removed because it was unused?

autotest/Tests/src/main/java/com/tle/webtests/pageobject/viewitem/AttachmentsPage.java
```diff
@@ -123,7 +136,15 @@ private By folderBy(String description) {
...
-      row.click();
+      // You cannot just call click on the row, as it may be clicking on the actual attachment link
+      // instead.
+      // Note the yOffset is negative, as despite their javadoc saying the coord are top-left based,
+      // it's actually a bald-faced lie, and the position is relative to the center of the element.
+      new Actions(driver)
+          .moveToElement(((WrapsElement) row).getWrappedElement(), 0, -13)
+          .click()
+          .build()
+          .perform();
```
edalex-ian: This concerns me for future maintainability... However looking at the page nothing really jumps out as to what to do if simple `row.click()` is not working. But that said, how come `row.click()` used to work - i.e. why'd it only need to be changed now?

abidingotter: row.click works when your attachment name is short enough that it doesn't reach the middle of the row

edalex-ian: Gotcha.

But not to worry, if the page does significantly change (as it will in the new UI support) then arguably we wont even have the row, so... All good.

Source/Plugins/Core/com.equella.core/src/com/tle/web/cloud/search/CloudSearchListEntry.java
abidingotter: Yep, I removed it because it unused, and it was related to the changes I was making.  (i.e.: I would have had to have added a new method to this class to implement the new method on the interface, which would have been pointless as the whole class is unused)

abidingotter:
> Generally fine with this, just a few queries.
> But a bigger concern is the discussions around various issues still popping up. Let's discuss in person tomorrow with @PenghaiZhang .

I don't believe there are any issues.  For attachments like images, the browser is caching them as it should, hence not hitting the server.  The max-age=0 modification made recently may change that however.

abidingotter:
@edalex-ian @PenghaiZhang I can confirm that the latest max-age changes in develop have caused this branch to register a view each time an image is viewed, regardless of caching.

PenghaiZhang:
> @edalex-ian @PenghaiZhang I can confirm that the latest max-age changes in develop have caused this branch to register a view each time an image is viewed, regardless of caching.

Thanks @abidingotter . Actually I did test with the latest max-age change and the issue I found is the view count increases by 2 not 1.

abidingotter:
It only increased by 1 for me...

> Another thing I am concerned with is you also mentioned that the change to do with error screens in content packages is likely to break things.

Yes, but that is a different issue that has already been merged, so not relevant to this PR :)
When I said "break things", I meant a bunch of issues could crop up that we hadn't thought about, I didn't mean anything specific to the view count issue.

SammyIsConfused:
@abidingotter This is what @PenghaiZhang is talking about. The item view count indeed only goes up by one, but the _attachment_ view count goes up by two each time. Here's a clip showing that happening (zipped, because apparently GitHub doesn't natively support WebM files in comments like Gitlab does).
[Peek 2019-11-18 14-30.zip](https://github.com/openequella/openEQUELLA/files/3856695/Peek.2019-11-18.14-30.zip)

abidingotter:
@SammyIsConfused Does this only happen with the full screen link?  Ie., if you click on File Viewer and go back (and probably have to refresh the page) does the view count only go up by one?  It did for me (attachment, not item)

If it's only when using the Full screen link, then I suspect it's this ridiculous browser bug:
https://bugzilla.mozilla.org/show_bug.cgi?id=254144
But to be sure, open up the dev tools and see if there is a duplicate request for the same image.

SammyIsConfused:
> @SammyIsConfused Does this only happen with the full screen link? Ie., if you click on File Viewer and go back (and probably have to refresh the page) does the view count only go up by one? It did for me (attachment, not item)
> 
> If it's only when using the Full screen link, then I suspect it's this ridiculous browser bug:
> https://bugzilla.mozilla.org/show_bug.cgi?id=254144
> But to be sure, open up the dev tools and see if there is a duplicate request for the same image.

Just checked it and yes, this only with the full screen link. With the file viewer we only get one.

SammyIsConfused:
@abidingotter And there is no duplicate call for the same image with the fullscreen. Sounds like you may be right about the browser iframe bug.

abidingotter:
It's also possibly that the full screen viewer is logging a "view" just to get the link for the left hand side (which is invisible when there is only one attachment) and one view to actually view the image.
Could you add another attachment to the item, then click Full screen, and see if the attachment that you *don't* view gains a view count or not?

SammyIsConfused:
> It's also possibly that the full screen viewer is logging a "view" just to get the link for the left hand side (which is invisible when there is only one attachment) and one view to actually view the image.
> Could you add another attachment to the item, then click Full screen, and see if the attachment that you _don't_ view gains a view count or not?

Just tried that,  only the one viewed goes up at all and that is by 2. Also, it goes up by two every time you switch back and forth in the fullscreen viewer.
Just open fullscreen, just +2 for attachment 1. +0 for attachment 2.

Open fullscreen, +2 for attachment 1
Click on the left pane for attachment 2, +2 for attachment 2

abidingotter:
@SammyIsConfused will look at it as soon as I get time.  I don't feel like it should be a blocker however.

SammyIsConfused:
@abidingotter That sounds good :) Yeah I don't think it's a major issue, its quite a narrow use case.

PenghaiZhang:
Thanks @abidingotter and @SammyIsConfused. I reckon we can merge this PR, but let's wait for Ian's approval.

edalex-ian:
Alright, had a chat with @cathfitz and we're thinking let's get this moving so merge this for now and then an issue will be raised for the full screen viewer.

Issue #1181, opened 2019-08-20
Attachment caching not functioning as expected


mrblippy:
**Describe the bug**
Each request that gets made to an attachment in the filestore gets a Cache-Control max-age=86400, s-maxage=0, must-revalidate header on the response. 
Here is the code 
https://github.com/openequella/openEQUELLA/blob/c9941d45ef6f8e598521993b408ea7ffd843a36b/Source/Plugins/Core/com.equella.core/src/com/tle/web/viewitem/ItemFilestoreServlet.java#L198

The cache doesn’t get invalidated when the content of an attachment gets changed (eg. in-place editing, or replacing the file attachment with a file with the same name.

It’s good that we have caching on attachments. But the issue  is that the cache isn’t getting invalidated when the contents of an attachment is changed (eg, when the attachment is replaced by a file with the same filename)

Maybe we could look at implementing something like this?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag

**To Reproduce**

1. Contribute an item to oEQ with a file attachment
2. On the item summary click the attachment to view it (your browser will cache the file for 24hrs)
3. Edit the item
4. Replace the file attachment with a different file that contains the same filename
5. Save the item
6. On the item summary click the attachment to view it
7. Notice that the original file is still displayed

**Expected behavior**
Cache should be invalidated when attachment content is changed

**Platform:**

- OpenEquella Version: All oEQ versions
- OS: Windows 10

PR #1329, opened 2019-10-30
Replace max-age with Etag checksum check to fix attachment preview caching

SammyIsConfused:
##### Checklist

- [x] documentation is changed or added (in-code documentation)

##### Description of change

Solves #1181. The original method for file attachment cache validation was to use a max age parameter, which meant that modifying a file but maintaining its filename would cause stale preview images. Using a checksum instead forces the invalidation should the file content change. 

### DIFF
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/web/viewitem/ItemFilestoreServlet.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/web/viewitem/ItemFilestoreServlet.java
@@ -141,7 +141,7 @@ public class ItemFilestoreServlet extends HttpServlet {
       if (itemId != null) {
         auditor.audit(request, new ViewAuditEntry("file:" + mimeType, path), itemId, attachment);
       }
-      contentStreamWriter.outputStream(request, response, filteredStream);
+      contentStreamWriter.outputStream(request, response, filteredStream, true);
     } catch (AccessDeniedException ade) {
       if (CurrentUser.isGuest()) {
         LogonSection.forwardToLogon(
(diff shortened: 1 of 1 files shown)


Source/Plugins/Core/com.equella.core/src/com/tle/web/viewitem/ItemFilestoreServlet.java
```diff
@@ -185,17 +185,8 @@ public InitialFilestoreStream(
...
-      // a copy of the resource - the age values only specify how long
-      // before the cache must *revalidate* their copy. With these
-      // settings, browser caches can serve their copy for a whole day
-      // without needing to talk to the server. Shared caches (eg, a Squid
-      // proxy cache) must always revalidate before releasing their copy
-      // to ensure we're not returning a resource to a user who doesn't
-      // have access.
-      return "max-age=86400, s-maxage=0, must-revalidate";
+      // this is to ensure that the etag is checked when attempting to pull down attachments
+      return "must-revalidate";
```
PenghaiZhang: Considering ETag is enabled, I wonder if we still need Cache-Control HTTP header.

SammyIsConfused: We need the must-revalidate to enforce that the etag check is always made. The max-age and s-maxage is no longer necessary.

abidingotter:
This is probably going to kill the server. It reads the entire file into memory to generate the hash every time. That's why I didn't enable it in general. The real fix would be to generate the hash once upon upload and store it... Somewhere

abidingotter:
I wonder if changing to max-age=0 will make the browser check if the file has changed or not?

PenghaiZhang:
> I wonder if changing to max-age=0 will make the browser check if the file has changed or not?

I may be wrong, but as far as I know, max-age=0 means the cache expires immediately so it enforces the browser to re-validate the resource every time.  Then the server will have to check if the requested resource is latest for every request by either Etag or Last_Modified_date.

abidingotter:
Yep, that's what I mean. The check for last modified date should be super quick.

abidingotter:
We don't calculate etag for anything except the admin console files.

Issue #1313, opened 2019-10-18
New UI Contribution wizard: alert dialog appearing upon clicking Publish (on one page wizards)


mrblippy:
**Describe the bug**
Seems specific to one page wizards in the new UI

Please watch this fantastic video I recorded to see the issue in action
![broked](https://user-images.githubusercontent.com/4625498/67067807-66f8aa00-f1c3-11e9-8d84-00222dd819aa.gif)

![Capture](https://user-images.githubusercontent.com/4625498/67067828-7e379780-f1c3-11e9-97bc-97fc03ec1188.PNG)
![borked 2](https://user-images.githubusercontent.com/4625498/67067833-82fc4b80-f1c3-11e9-9e0c-6edf000ee100.PNG)

**Expected behavior**
Should be able to save and publish an item without an alert appearing

**Platform:**

- OpenEquella Version: 2019.2
- OS: Windows 10
- Browser Chrome

PR #1330, opened 2019-10-30
Fix the alert dialog bug when publishing items

PenghaiZhang:
##### Checklist

- [x] the [contributor license agreement is signed

##### Description of change
#1313 
According to my investigation, this bug is caused because the the route to the Item Summary page is considered as an external route due to the new function `isClientPath` in LegacyContentApi. 

This function  only marks Urls which end with '.do' as internal routes, excluding 'logon.do'. However, routes to the Item Summary page (e.g. http://localhost:8080/charlie/items/802798b4-5bf8-4b70-b7d5-4c91d366386a/1/)  should also be internal ones.

So the proposed (and maybe the safest) fix is to let `isClientPath` mark the Urls of  Item Summary page as internal routes.

Lastly, in terms of why this bug only happens in one page collection, the answer could be found from line 88 - 99 of `index.tsx` .

### DIFF
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/LegacyContentApi.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/LegacyContentApi.scala
@@ -111,11 +111,21 @@ object LegacyContentController extends AbstractSectionsController with SectionFi
 
   def isClientPath(relUrl: RelativeUrl): Boolean = relUrl.path.parts match {
     case Vector("logon.do") => false
-    case p                  => p.last.endsWith(".do")
+    case p                  => p.last.endsWith(".do") || isItemSummaryUrl(relUrl.toString())
   }
 
   def internalRoute(uri: String): Option[String] = {
     relativeURI(uri).filter(isClientPath).map(r => "/" + r.toString())
+  }
+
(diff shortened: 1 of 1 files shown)


PenghaiZhang:
> treeExistsForUrlPath

Had a short discussion with Aaron about this issue. I tried `treeExistsForUrlPath` and it did not fix the issue because there is no tree registered for a particular Url of the Item Summary page.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/LegacyContentApi.scala
```diff
@@ -111,13 +111,23 @@ object LegacyContentController extends AbstractSectionsController with SectionFi
...
     case Vector("logon.do") => false
-    case p                  => p.last.endsWith(".do")
+    case p                  => p.last.endsWith(".do") || isItemSummaryUrl(relUrl.toString())
   }
 
   def internalRoute(uri: String): Option[String] = {
     relativeURI(uri).filter(isClientPath).map(r => "/" + r.toString())
   }
 
+  def isItemSummaryUrl(path: String): Boolean = {
```
PenghaiZhang: If we want to keep this method, I feel  `LegacyContentApi.scala` isn't the best place. It should be in somewhere in the `UrlService` or something like that.

Issue #1194, opened 2019-08-26
Adding a new datum to a taxonomy term fails


cbeach47:
**Describe the bug**
Unable to PUT to `/api/taxonomy/{tax-uuid}/term/{term-uuild}/data/{key-1}/{value-1}`.  oEQ 2019.1 yields the error:
```
{
  "code": 500,
  "error": "Internal Server Error",
  "error_description": "No existing transaction found for transaction marked with propagation 'mandatory'"
}
```

**Platform:**
- OpenEquella Version: 2019.1

phette23:
I think this is a dupe of #1192 but we _really_ need this feature fixed, it basically invalidated an entire week of work I did trying to convert one of our processes from SOAP to the REST API.

edalex-ian:
As part of #1225 the taxonomy REST API implementation has been significantly reworked to match some of the other oEQ REST APIs. This has been merged into `develop` so it could be interesting to see if this has improved things for you.

abidingotter:
This is a separate issue to #1192 , but I have fixed both in one pull request: https://github.com/openequella/openEQUELLA/pull/1333

SammyIsConfused:
Tested and verified in develop. This is now fixed. Closing.

Issue #1192, opened 2019-08-26
Taxonomy PUT requests for data is broken


cbeach47:
**Describe the bug**
For the PUT request `/{uuid}/term/{termUuid}/data/{datakey}/{datavalue}` in `TaxonomyResource`, the `datakey` parameter is listed twice instead of once and then the `datavalue`parameter.

This messes up the Swagger UI apidocs.do page as well.

**Platform:**

- OpenEquella Version: 2019.1

edalex-ian:
As part of #1225 the taxonomy REST API implementation has been significantly reworked to match some of the other oEQ REST APIs. This has been merged into `develop` so it could be interesting to see if this has improved things for you.

SammyIsConfused:
Verified and tested in current develop, this now works. Closing.

PR #1321, opened 2019-10-24
Ability to sort all taxonomy nodes in the entire tree at once

abidingotter:
##### Checklist

- [x] the [contributor license agreement][] is signed
- [x] commit message follows [commit guidelines][]
- [x] tests are included

##### Description of change

Ability to sort all taxonomy nodes in the entire tree at once. #725 

[contributor license agreement]: https://www.apereo.org/node/676
[commit guidelines]: https://chris.beams.io/posts/git-commit/

PR #1333, opened 2019-10-31
Fix issues adding term data via the taxonomy api

abidingotter:
##### Checklist

- [x] tests are included

##### Description of change

Add missing transactional annotations, fix name of params in API, add
test for adding data

#1192 and #1194 


Source/Plugins/Core/com.equella.core/src/com/tle/core/taxonomy/impl/TermServiceImpl.java
```diff
@@ -386,19 +388,25 @@ public void setAllData(Taxonomy taxonomy, String termFullPath, Map<String, Strin
   }
 
   @Override
-  public void setDataByTermUuid(
+  @SecureOnCall(priv = "EDIT_TAXONOMY")
```
SammyIsConfused: We should have a constant string for "EDIT_TAXONOMY".

autotest/OldTests/src/test/java/com/tle/webtests/test/webservices/rest/AbstractRestApiTest.java
```diff
@@ -233,9 +235,22 @@ protected JsonNode getEntity(String uri, String token, Object... params) throws
   }
 
   protected HttpResponse postEntity(
-      String json, String uri, String token, boolean consume, Object... paramNameValues)
+      @Nullable String json, String uri, String token, boolean consume, Object... paramNameValues)
```
PenghaiZhang: As we have a new method `putOrPostEntity`, do we still need `@Nullable` here and also in the signature of `putEntity` ?

Source/Plugins/Core/com.equella.base/src/com/tle/common/taxonomy/terms/Term.java
```diff
@@ -282,9 +282,12 @@ public String getAttribute(String key) {
     return null;
   }
 
-  public void setAttribute(String key, String value) {
+  /** @return true if the attribute did not previously exist */
+  public boolean setAttribute(String key, String value) {
     ensureAttributes();
+    final boolean created = (attributes.get(key) == null);
```
PenghaiZhang: If the key does not exist in the attributes, then `created` will be true. Shouldn't it be false?

abidingotter: No, this is before the new attribute is added.  
If there was no value before, then the attribute is getting created, otherwise it's updating an existing attribute.

autotest/OldTests/src/test/java/com/tle/webtests/test/webservices/rest/AbstractRestApiTest.java
abidingotter: Well, technically we don't *need* it anywhere :)  But I like them, and it helps catch errors.
To answer your question though, yes, we do need it on both, but only because we allow null to be passed in in both.  It basically stops you doing something like this:

```
protected void aMethod(@Nullable String thisIsNullable)
{
    int length = thisIsNullable.length();  << this is an error, you have to check is thisIsNullable is null first.
}
```
Someone could still stick code into putEntity and use the json as if it was definitely not null.

Source/Plugins/Core/com.equella.core/src/com/tle/core/taxonomy/impl/TermServiceImpl.java
abidingotter: Yeah possibly

abidingotter: Added constant

PenghaiZhang:
Unfortunately due to merging #1321 into Develop, some code conflicts come up. I will resolve them later.

PR #1338, opened 2019-11-04
Disable the browser's cache for attachments

PenghaiZhang:
##### Checklist

##### Description of change

Set the value of `max-age` to 0 so that the browser will send a request to server all the time. Then the server will check `Last-Modified` of the attachment, which does not cost too many resources.

#1181


Source/Plugins/Core/com.equella.core/src/com/tle/web/viewitem/ItemFilestoreServlet.java
```diff
@@ -195,7 +195,7 @@ public String getCacheControl() {
       // proxy cache) must always revalidate before releasing their copy
       // to ensure we're not returning a resource to a user who doesn't
       // have access.
```
edalex-ian: Need to update the comment. ;)

Issue #1184, opened 2019-08-22
Create a new default css that displays legacy content within the new UI template so that the legacy openEQUELLA content has a new look and feel.


cathfitz:
Create new styles that can be applied to legacy content within the new ui template using CSS

edalex-ian:
All done and tested.

PR #1350, opened 2019-11-20
search page styling mega commit/monster

mrblippy:
##### Description of change

https://github.com/openequella/openEQUELLA/issues/1184

sass code has been written to style the search page in a material-like way. the code is a mixture of existing legacy css and new sass code written by me 

For testing all functionality on the search page will need to be tested, to ensure everything has been styled in a pleasing way. I have been looking at it all for too long, so there are likely still issues that i haven't yet found/glossed over. Any feedback welcome
beautification search filtering
beautification search user selector
beautification search 1
beautification search add to favs message
beautification search attachments

### DIFF
--- a/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
+++ b/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
@@ -162,6 +162,21 @@
   text-transform: uppercase;
 }
 
+/*****************************************************************************
+Animations
+*****************************************************************************/
+@keyframes snackbar {
+  0%,
+  100% {
+    bottom: -80px;
+  }
+
+  10%,
+  90% {
+    bottom: 0px;
+  }
+}
+
 //**sass functions**//
 @function set-text-color($color) {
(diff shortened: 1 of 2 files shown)


SammyIsConfused:
Interesting issue with the snackbars. They seem to be fixed in position, so if you scroll they don't move with you.

SammyIsConfused:
Tested in Chromium, Firefox, Edge and IE11.
Chromium, Firefox and Edge all seem to display in an identical fashion. I wasn't able to make them behave differently.
Firefox:
image
Edge:
image
Chromium:
image

However, IE11 support has some minor issues. Icons seem to turn up twice in slightly different forms (almost as if code that is hiding them doesn't work) and there are some colouring and button oddities. 
 IE11:
Screenshot from 2019-11-25 11-57-35
Screenshot from 2019-11-25 11-56-26
Screenshot from 2019-11-25 11-52-16

So I'd say once IE11 compatibility is matched, the main outstanding issue with the search page EBP is the static snackbars. They should stay where they are when scrolling, and also they should probably disappear after a short time.

Issue #1344, opened 2019-11-14
Config to disable task Remove Staging Areas


cbeach47:
**Is your feature request related to a problem? Please describe.**
Parent issue: #1137 .   Some adopters have temporarily disabled this on their installs in efforts to address #1137 , but it makes it difficult for them to upgrade while #1137 is being worked.

**Describe the solution you'd like**
A config in `optional-config.properties` to turn off running the scheduled task 'Remove Staging Areas' as a background task.  Done this way, it should still allow the task to show up on the admin page `scheduledtasksdebug.do`.  If possible, it'd be useful to allow a manual run of the task from that admin page.

cbeach47:
Tested on the latest develop build (2019.2.r20191206).  Looks good.

PR #1359, opened 2019-11-26
Added config to disable RemoveStagingAreas

cbeach47:
##### Checklist
- [ ] documentation is changed or added - once approved / merged, I'll update the docs repo.

##### Description of change
Allows a system admin to turn off the `RemoveStagingAreas` via the boolean system-level config `com.tle.core.tasks.RemoveStagingAreas.enable` in `optional-config.properties`.  Also adding more file audit logging.  

This will affect both the background task and the manual invocation via `/access/scheduledtasksdebug.do`.  Didn't seem like a clean way to allow the manual and block the background method, and this _should_ be a rarely used configuration.

Issue: #1344 

### DIFF
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/core/filesystem/staging/service/impl/StagingServiceImpl.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/core/filesystem/staging/service/impl/StagingServiceImpl.java
@@ -67,10 +67,12 @@ public class StagingServiceImpl implements StagingService {
     if (s == null) {
       LOGGER.error("Staging area does not exist");
     } else {
+      LOGGER.info("Deleting Staging entry in DB [" + s.getStagingID() + "]");
       stagingDao.delete(s);
     }
 
     if (removeFiles) {
+      LOGGER.info("Removing Staging area in the filestore [" + s.getStagingID() + "]");
       fileSystemService.removeFile(staging);
     }
   }
@@ -100,9 +102,10 @@ public class StagingServiceImpl implements StagingService {
             String uuid = file.getFileName().toString();
             if (!stagingExists(uuid)) {
               try {
+                LOGGER.debug("Deleting staging area [" + uuid + "]");
                 FileUtils.delete(file, null, true);
               } catch (IOException ex) {
-                LOGGER.warn("Could not delete staging area: " + uuid, ex);
+                LOGGER.warn("Could not delete staging area [" + uuid + "]", ex);
               }
             }
           }
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/core/filesystem/staging/service/impl/StagingServiceImpl.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/core/filesystem/staging/service/impl/StagingServiceImpl.java
@@ -67,12 +67,12 @@ public class StagingServiceImpl implements StagingService {
     if (s == null) {
       LOGGER.error("Staging area does not exist");
     } else {
-      LOGGER.info("Deleting Staging entry in DB [" + s.getStagingID() + "]");
+      LOGGER.debug("Deleting Staging entry in DB [" + s.getStagingID() + "]");
       stagingDao.delete(s);
     }
 
     if (removeFiles) {
-      LOGGER.info("Removing Staging area in the filestore [" + s.getStagingID() + "]");
+      LOGGER.debug("Removing Staging area in the filestore [" + s.getStagingID() + "]");
       fileSystemService.removeFile(staging);
     }
   }
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/core/scheduler/impl/SchedulerModule.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/core/scheduler/impl/SchedulerModule.java
@@ -26,6 +26,6 @@ public class SchedulerModule extends OptionalConfigModule {
   protected void configure() {
     bindInt("com.tle.core.tasks.RemoveDeletedItems.daysBeforeRemoval");
     bindInt("com.tle.core.tasks.RemoveOldAuditLogs.daysBeforeRemoval");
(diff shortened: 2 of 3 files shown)


Source/Plugins/Core/com.equella.core/src/com/tle/core/scheduler/impl/SchedulerModule.java
```diff
@@ -26,5 +26,6 @@
   protected void configure() {
     bindInt("com.tle.core.tasks.RemoveDeletedItems.daysBeforeRemoval");
     bindInt("com.tle.core.tasks.RemoveOldAuditLogs.daysBeforeRemoval");
+    bindBoolean("com.tle.core.tasks.RemoveStagingAreas.enable");
```
PenghaiZhang: The default value is null. What about `bindBoolean("com.tle.core.tasks.RemoveStagingAreas.enable", true);` ?

Source/Plugins/Core/com.equella.core/src/com/tle/core/scheduler/standard/task/RemoveStagingAreas.java
```diff
@@ -22,16 +22,29 @@
...
 @Singleton
 public class RemoveStagingAreas implements ScheduledTask {
+  private static final Logger LOGGER = Logger.getLogger(RemoveStagingAreas.class);
+
   @Inject private StagingService stagingService;
 
+  @com.google.inject.Inject(optional = true)
+  @Named("com.tle.core.tasks.RemoveStagingAreas.enable")
+  // Can be overrode by the optional-config.properties
+  private boolean enableTask = true;
```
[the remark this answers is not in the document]
cbeach47: Noted in documentation thread on #1360 . Seems to be inconsistent in how the configs are documented.  I added this to the next CDM agenda.  For now, I'm leaning towards just merging as-is.

Source/Plugins/Core/com.equella.core/src/com/tle/core/scheduler/impl/SchedulerModule.java
cbeach47: Good point - the default is now set to `true`.

edalex-ian:
Heya @cbeach47 

Just a heads up on a little housekeeping I did on this ticket. I moved the issue reference number from the PR's title to the body. This way GitHub processes it and establishes the links between the issue and this PR.

If you could keep that in mind for other ones, that'd be great.

I did try to include this instruction in the PR template, but you're not the first to miss it, so I wonder if I need to improve the template. :thinking:

Source/Plugins/Core/com.equella.core/src/com/tle/core/scheduler/standard/task/RemoveStagingAreas.java
```diff
@@ -22,16 +22,29 @@
...
+
   @Override
   public void execute() {
-    stagingService.removeUnusedStagingAreas();
+    if (enableTask) {
+      stagingService.removeUnusedStagingAreas();
+    } else {
+      LOGGER.info("RemoveStagingAreas is disabled.  Not running task.");
```
edalex-ian: I think at most this should only be at `debug` as it's going to get logged every-time this job is scheduled to run which is a bit verbose. Which raises two things:

1. If you just want to know what the read in state was, then perhaps that should be done `SchedulerModule`; but maybe
2. It could be better if this flag was detected in the config, then it simply cancels the schedule for the task.

Source/Plugins/Core/com.equella.core/src/com/tle/core/filesystem/staging/service/impl/StagingServiceImpl.java
```diff
@@ -67,10 +67,12 @@ public void removeStagingArea(StagingFile staging, boolean removeFiles) {
     if (s == null) {
       LOGGER.error("Staging area does not exist");
     } else {
+      LOGGER.info("Deleting Staging entry in DB [" + s.getStagingID() + "]");
       stagingDao.delete(s);
     }
 
     if (removeFiles) {
+      LOGGER.info("Removing Staging area in the filestore [" + s.getStagingID() + "]");
```
edalex-ian: Could this and the one above simply be `debug`? I feel `info` is too verbose, and could end up spamming resource centre logs when attempting to analyse support issues. (And especially in a case of any bulk item creation.)

Source/Plugins/Core/com.equella.core/src/com/tle/core/scheduler/standard/task/RemoveStagingAreas.java
cbeach47: If this becomes a 'thing' were we want to turn off scheduled tasks, then I can see delving into the logic more.  For now, since this is really to root-cause the Attachment Health issue and generally no one should be setting this flag, I am cautious about messing with the scheduler / task generation logic too much at this time.  I'd like to leave this check for the flag as-is, and revisit if a need arises to toggle other tasks.  I'll downgrade the log messages though - if folks are setting the flag to disable the task, they can modify their log4j as well.

Source/Plugins/Core/com.equella.core/src/com/tle/core/filesystem/staging/service/impl/StagingServiceImpl.java
cbeach47: ok - that's downgraded now

Source/Plugins/Core/com.equella.core/src/com/tle/core/scheduler/standard/task/RemoveStagingAreas.java
edalex-ian: Note a problem.

> I am cautious about messing with the scheduler / task generation logic too much at this time.

I'm hoping (like standard schedulers) there's simply a method that we can call to remove it from the schedules. However, I also don't have time to confirm that, and as you said maybe this isn't a 'thing' yet.... So, cool.

PR #1395, opened 2019-12-16
Fix flakiness for PortalsTest

SammyIsConfused:
##### Checklist

- [x] tests are included

##### Description of change

The PortalsTest keeps failing in Travis when looking for the options button in the Recent Portal. Added a couple of waits to ensure that the header itself exists in the tree and then the options button. 
During local testing the PortalsTest would break roughly once in 5 times, since making this change  I have run the test locally 25 times without a fail. 


PenghaiZhang:
This test still fails due to another reason: it can't load the HomePage after waiting for 30 seconds.
Therefore I feel it's worth to have a look at line 56 of `AbstractPortalEditPage`

PenghaiZhang:
We need to fix the Chrome issue first in order to have those functional tests work again. 
Will confirm with @SammyIsConfused  when she is back.

Issue #1356, opened 2019-11-26
Form data is not being proxied through using Cloud Provider SDK


edalex-tom:
**Describe the bug**
Using the openEQUELLA Cloud Provider SDK and an extension control that submits data via a HTTP POST, the form data (have tried a file and string) does not appear to be proxied through to the cloud provider endpoint.

**To Reproduce**

1. Register a cloud provider
2. Add an cloud provider control to a contribution item that supports submitting form data (e.g. file) through a HTTP POST
3. Submit the form data through the control
4. Request is proxied through openEQUELLA but the form data does not appear to be available in the cloud provider endpoint

**Expected behavior**
The request including form data is proxied through openEQUELLA and sent to the cloud provider endpoint

**Screenshots**
N/A

**Stacktrace**
Not available

**Platform:**

- OpenEquella Version: 2019.2
- OS: Ubuntu 18.04
- Browser: Chrome

cathfitz:
Not going to be done

PR #1402, opened 2019-12-19
Ensure form data are properly sent to ECS services

PenghaiZhang:
* Copy headers to the requests sent to ECS services by the Proxy
* Add requests' content inputstream to 'RemappedRequest'
* Add streaming upload check to IntegTester
#1356 

##### Checklist

- [x] tests are included

##### Description of change

The 'inputstream' passed to the Cloud Provider Proxy endpoint has been consumed by Servlet filters so the requests sent to ECS do not have the original body information. Also the requests do not have correct headers.

The proposed fix is to copy the origin inputstream to the wrapped request (RemappedRequest) so we can read the inputstream at the Proxy endpoint. Also add headers to the requests. 

In order to test if files are uploaded in streaming, I slightly modify the Integtester. It is expected that the response message should include the multiparts information and  large files are saved locally in the root folder of oEQ without the 'OutOfMemory' error.

Screenshot from 2019-12-19 16-28-42

### DIFF
--- a/Source/Plugins/Core/com.equella.base/src/com/tle/web/dispatcher/RemappedRequest.java
+++ b/Source/Plugins/Core/com.equella.base/src/com/tle/web/dispatcher/RemappedRequest.java
@@ -19,8 +19,10 @@
 package com.tle.web.dispatcher;
 
 import com.tle.common.URLUtils;
+import java.io.IOException;
 import java.io.UnsupportedEncodingException;
 import java.net.URLDecoder;
+import javax.servlet.ServletInputStream;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletRequestWrapper;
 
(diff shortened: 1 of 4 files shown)


Source/Plugins/Core/com.equella.base/src/com/tle/web/dispatcher/RemappedRequest.java
```diff
@@ -70,6 +78,11 @@ public String getRequestURI() {
     return requestURI;
   }
 
+  @Override
+  public ServletInputStream getInputStream() {
+    return servletInputStream;
```
edalex-ian: We shouldn't need another property on the class, this method could just do `request.getInputStream()`.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/wizard/WizardApi.scala
```diff
@@ -205,10 +205,13 @@ class WizardApi {
                 content: InputStream): Response = {
     val streamedBody =
       readInputStream(IO(content), 4096, Implicits.global).chunks.map(_.toByteBuffer)
+
+    val headers = for {
+      headerName <- req.getHeaderNames.asScala
+    } yield (headerName, req.getHeader(headerName))
```
edalex-ian: A key thing is to include all headers - as part of this fix - however there are some that we'd probably want to drop (such as security tokens from client - jsession etc - and maybe any HTTPS related headers).

Source/Plugins/Core/com.equella.base/src/com/tle/web/dispatcher/RemappedRequest.java
PenghaiZhang: The code here does not make much sense so I will try another way

edalex-ian:
After further discussions, it's become apparent that oEQ is not really setup to proxy data through it due to its various servlet filters - especially those for OAuth. Support would require significant changes which we're prefer to avoid for now.

The approach instead is that cloud providers would go with store and forward. That is, first do a 'standard' upload into the staging area as part of contribution, and then use from there.

As a result, closing this PR.

PR #1408, opened 2020-01-07
Do not use 'cd -' after building import-export-tool

PenghaiZhang:
##### Checklist

##### Description of change
Recently, Codebuild has failed several times due to the directory for publishing build results being wrong. We do not know why this is happening yet, but we can avoid it by removing `cd -` after building the import/export tools


PenghaiZhang:
This PR only affects Codebuild so I cancelled Travis builds.

PenghaiZhang:
Retried two times and all builds are green. But today Mr Build stop complaining without any changes. I wonder if we still want this change.

PR #1429, opened 2020-01-15
fix functional tests

PenghaiZhang:
##### Checklist

##### Description of change

This PR aims to make the Selenium tests more stable. 

PackageAndNavigationTest is less likely to fail. This test was very stable before the version 79 of Chrome . Although we use the beta version Chrome and make some changes, this test can still fail intermittently. So this test should become stable again once the next stable version of Chrome is available.

PortalsTest and AdvancedScriptingControlTest are less likely to fail.

For the oEQ installing issue (only failed on Travis), just need more builds to verify the fix.

For the Scala tests which sometimes fail on Codebuild, it should not happen any more.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,6 +57,9 @@ install:
   - unzip chromedriver_linux64.zip
   - whereis google-chrome
   - whereis google-chrome-beta
+  - apt-get purge google-chrome-stable
+  - whereis google-chrome
+  - whereis google-chrome-beta
   - echo "google chrome version"
   - echo `google-chrome --version`
   - echo `google-chrome-beta --version`
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ addons:
       - sourceline: "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main"
         key_url: "https://dl-ssl.google.com/linux/linux_signing_key.pub"
     packages:
-      - google-chrome
+      - google-chrome-beta
       - ffmpeg
       - libimage-exiftool-perl
       - openjdk-8-jdk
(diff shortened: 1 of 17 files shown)


.travis.yml
```diff
@@ -6,6 +6,7 @@ branches:
     - master
     - /^release\/.+/
     - /^stable-.+/
+    - /^travis\/.+/
```
PenghaiZhang: I feel that we can keep Travis builds for this branch at the moment because when we can install new stable version of Chrome and Chromedriver via APT we may need this branch for testing.

.travis.yml
```diff
@@ -51,6 +51,12 @@ before_install:
   - sudo ln -s /usr/bin/ffplay /usr/bin/avplay
   - sudo ln -s /usr/bin/ffprobe /usr/bin/avprobe
 install:
+  #  - wget https://dl.google.com/linux/direct/google-chrome-beta_current_amd64.deb
```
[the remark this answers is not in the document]
PenghaiZhang: Yes, I will delete these comments.

.travis.yml
```diff
@@ -101,33 +115,46 @@ jobs:
...
 
     - stage: functional test
-      script: sbt -jvm-opts autotest/.jvmopts "project autotest" \
-        installEquella startEquella configureInstall setupForTests \
-        Tests/test dumpCoverage
+      script:
+        # DEBUG: Confirm PS directory is being properly cached
+        - ls -l autotest/IntegTester/ps/
+        # DEBUG: Confirm Target directory is being properly cached
+        - ls -l autotest/IntegTester/target/
```
edalex-ian: I think we can dump these now - as I think this is now working?

PR #1445, opened 2020-01-24
Accessibility: Add alt tags where they are missing

SammyIsConfused:
##### Checklist

##### Description of change

- Add alt tags to the logo (including in the theme editor page)
- Change the ImageRenderer such that if the alt tag is null or empty, the image src becomes the alt tag rather than just an empty string.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/web/sections/standard/renderers/ImageRenderer.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/web/sections/standard/renderers/ImageRenderer.java
@@ -100,9 +100,13 @@ public class ImageRenderer extends TagRenderer {
     attrs.put("src", source); // $NON-NLS-1$
     attrs.put("width", width); // $NON-NLS-1$
     attrs.put("height", height); // $NON-NLS-1$
-    if (alt != null) {
+    if (alt != null && alt.getText().length() > 0) {
       attrs.put("alt", alt.getText()); // $NON-NLS-1$
       attrs.put("title", alt.getText()); // $NON-NLS-1$
+    } else {
+      // If the alt tag is not set, just use the source as the alt
+      attrs.put("alt", source); // $NON-NLS-1$
+      attrs.put("title", source); // $NON-NLS-1$
     }
   }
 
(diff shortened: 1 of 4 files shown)


Source/Plugins/Core/com.equella.core/src/com/tle/web/sections/standard/renderers/ImageRenderer.java
```diff
@@ -100,9 +100,13 @@ protected void prepareFirstAttributes(SectionWriter writer, Map<String, String>
     attrs.put("src", source); // $NON-NLS-1$
     attrs.put("width", width); // $NON-NLS-1$
     attrs.put("height", height); // $NON-NLS-1$
-    if (alt != null) {
+    if (alt != null && alt.getText().length() > 0) {
       attrs.put("alt", alt.getText()); // $NON-NLS-1$
       attrs.put("title", alt.getText()); // $NON-NLS-1$
+    } else {
+      // If the alt tag is not set, just use the source as the alt
+      attrs.put("alt", source); // $NON-NLS-1$
```
abidingotter: No need for new // $NON-NLS-1$ comments.  In fact, can you please bulk delete them when you see them?  :)

Source/Plugins/Core/com.equella.core/src/com/tle/web/sections/standard/renderers/ImageRenderer.java
```diff
@@ -100,9 +100,13 @@ protected void prepareFirstAttributes(SectionWriter writer, Map<String, String>
     attrs.put("src", source); // $NON-NLS-1$
     attrs.put("width", width); // $NON-NLS-1$
     attrs.put("height", height); // $NON-NLS-1$
-    if (alt != null) {
+    if (alt != null && alt.getText().length() > 0) {
```
abidingotter: Just an FYI, the original intention of ImageRenderer was that alt was never optional.  (See the comment on the constructor :) )  This change seems fine however.

edalex-ian: I'm thinking there's nothing (at least in the doco for the Label interface) that says `getText()` will return a non-null value. That combined with convention elsewhere in the code base could I suggest using `StringUtils`:

```suggestion
    if (alt != null && StringUtils.isNotBlank(alt.getText())) {
```

PR #1455, opened 2020-01-29
Accessibility: Set width of the save button to remove horizontal scroll bar

SammyIsConfused:
##### Checklist

##### Description of change

image
The Save button changes in the EBP caused the save button to be slightly too wide for the page, causing an unnecessary horizontal scrollbar. 
 


edalex-ian:
hmmm, this is an interesting one. As isn't the button already in some kind of container/element and shouldn't the width of it be controlled?

SammyIsConfused:
> hmmm, this is an interesting one. As isn't the button already in some kind of container/element and shouldn't the width of it be controlled?

Fixed, we shouldn't have been resizing the button at all, rather settings it's direct parent's display type. It was defaulting to "block", when it should have been "flex".

Issue #1432, opened 2020-01-21
Accessibility enhancements


cathfitz:
**Describe the bug**

There are a number of areas in oEQ (new UI) that require work to conform to WCAG 2.0 AA.

**Platform:**

- OpenEquella Version: 2019.2, 2020.1

PR #1477, opened 2020-02-12
show options when enter key is pressed, instead of tab

mrblippy:
##### Checklist

##### Description of change

#1432
 
After revisiting this issue the dropdown doesn’t need to announce its collapsed/uncollapsed state.

**BUT…..**

The fact that the within dropdown automatically expands as soon as it receives focus (pressing the tab key) violates https://www.w3.org/TR/UNDERSTANDING-WCAG20/consistent-behavior-receive-focus.html

So i made the dropdown only activate when the user:

Focuses on the element and presses the enter key (and the enter or esc key will close it)

The behaviour of the mouse if unchanged


SammyIsConfused:
Looks good to me, though there is some strange merging going on. A lot of these commits are already in the component branch.

Issue #1479, opened 2020-02-12
Chore:  remove jetty-utils dependency


cbeach47:
`jetty-utils` is referenced in `Source/Server/equellaserver/build.sbt` and could use an upgrade, but doesn't appear to be used anywhere - so instead, let's remove the dependency altogether.

PR #1480, opened 2020-02-12
#1479:  Remove jetty-utils dep

cbeach47:
##### Checklist

##### Description of change
#1479 - Removed the `jetty-utils` dependency.  It was needing an upgrade, but didn't look to be used anywhere.

### DIFF
--- a/Source/Server/equellaserver/build.sbt
+++ b/Source/Server/equellaserver/build.sbt
@@ -183,7 +183,6 @@ libraryDependencies ++= Seq(
   "org.dspace"             % "cql-java"          % "1.0",
   //  "org.dspace.oclc" % "oclc-srw" % "1.0.20080328",
   "org.omegat"                           % "jmyspell-core"                  % "1.0.0-beta-2",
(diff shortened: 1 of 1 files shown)


cbeach47:
@edalex-ian - do you see any need to keep these `jetty-utils` around?

edalex-ian:
It's an interesting one to be sure. I can't find any compile time usage, and nothing is jumping out at me as far as a runtime dependency (and similar for `jetty-alpn`). The tests obviously are working, so that's a good start.

I'm currently attempting a reverse dependency lookup on it with IntelliJ, but I don't expect it to find anything. So maybe we just proceed unless @abidingotter knows the history.

cbeach47:
Sounds good @edalex-ian .  I plan to merge this in on my Tuesday unless I hear something.

For `jetty-alpn`, I'm not seeing that in the code.  Where is it referenced?

edalex-ian:
Ahhh, my bad, I was only looking at the dependency list in IntelliJ. `jetty-alpn` must be a transient. I wish IntelliJ would tell you better where a dep came from.

Issue #1280, opened 2019-09-27
Favourited searches added pre 2019.1 upgrades no longer work


mrblippy:
**Describe the bug**
Any favorite searches added before upgrading 2019.1 not result in a 
There is no tree for:/access/searching.do error

Caused by a change to FavouriteSearchServiceImpl.java in https://github.com/openequella/openEQUELLA/commit/e65af9bcf5d3c86c02b9396a9e833bc84333ec53 
Seems strange that this wouldn't have been found in an automated test though

The bit that's tripping it up is the fact that old favorite searches got stored in the database with /access/ at the start of the url, which no longer works. If we made a migration that removed it from existing entries in the favourite_search table we wcould probably get away with not having to do any other code changes

**To Reproduce**

1. Add a search to favorites in a pre 2019.1 oEQ 
2. Upgrade to 2019.1
3. Access the favorited search that you added in step 1

**Stacktrace**
`com.tle.web.sections.SectionsRuntimeException: There is no tree for:/access/searching.do
     at com.tle.web.sections.registry.AbstractSectionsController.createInfo(AbstractSectionsController.java:111)
     at com.tle.web.sections.registry.AbstractSectionsController.createForward(AbstractSectionsController.java:185)
     at com.tle.web.sections.generic.DefaultSectionInfo.createForwardForUri(DefaultSectionInfo.java:533)
     at com.tle.core.favourites.service.FavouriteSearchServiceImpl.executeSearch(FavouriteSearchServiceImpl.java:71)
     at com.tle.web.favourites.searches.SearchFavouritesList.runSearch(SearchFavouritesList.java:111)
     at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
     at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
     at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
     at java.lang.reflect.Method.invoke(Method.java:498)
     at com.tle.web.sections.events.js.MethodInvocationEventGenerator$1.fireDirect(MethodInvocationEventGenerator.java:94)
     at com.tle.web.sections.events.AbstractDirectEvent.fire(AbstractDirectEvent.java:86)
     at com.tle.web.sections.generic.DefaultSectionInfo.processEvent(DefaultSectionInfo.java:420)
[...]

PR #1495, opened 2020-02-20
Fix "no tree for xxx" error caused by favourite search links

PenghaiZhang:
#1280 
##### Checklist

##### Description of change
After discussing with @mrblippy , we think the best fix is to check and rewrite the URL of favourite searches because it doesn't require data migration. Maybe there are different approaches, but some significant changes would be required. 

In terms of why the redirection Servlet not working, it is probably because clicking the favourite search link triggers a tree check before it sends the request.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/core/favourites/bean/FavouriteSearch.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/core/favourites/bean/FavouriteSearch.java
@@ -18,7 +18,6 @@
 
 package com.tle.core.favourites.bean;
 
-import com.dytech.edge.web.WebConstants;
 import com.tle.beans.Institution;
 import java.util.Date;
 import javax.persistence.Column;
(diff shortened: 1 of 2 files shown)


Source/Plugins/Core/com.equella.core/src/com/tle/core/favourites/bean/FavouriteSearch.java
```diff
@@ -86,6 +87,13 @@ public void setName(String name) {
   }
 
   public String getUrl() {
+    // The value of 'url' starts with '/access' if the fav search was added in old oEQ versions,
```
abidingotter: The beans should really just be basic getters and setters.  I would move this logic into FavouriteSearchServiceImpl.executeSearch (the only place getUrl is called)

abidingotter: Also note that WebConstants.CLOUDSEARCH_PAGE and WebConstants.HIERARCHY_PAGE need to be checked too, since you can favourite those as well.

edalex-ian:
Looks like you may need to rerun your travis build @PenghaiZhang .

PenghaiZhang:
> Looks like you may need to rerun your travis build @PenghaiZhang .

Ah yes. Thanks for reminding!!!

Issue #1071, opened 2019-06-28
New ui: Errors in logs when performing actions after session has expired


mrblippy:
**Describe the bug**
Errors are getting logged in the resource-centre logs whenever a user tries to use legacy content in the new ui (like trying to perform a search). It's likely to confuse astute openequella administrators that like to go through their logs looking for issues.

An example of one of the errors:
![Capture](https://user-images.githubusercontent.com/4625498/60310391-0a644b80-9996-11e9-8a08-13f571fe8e61.PNG)

**To Reproduce**

1. Enable the new ui
2. Perform a search
3. Open up another tab pointing to the same institution
4. Logout of equella
5. Go back to the first tab, and do another search (it should fail because that user is logged out)
6. You should see a corresponding error in the logs

**Platform:**

- OpenEquella Version: 2019.1

nelson-edalex:
1. Login
2. Do a search
3. Open tab 2
4. Logout in tab 2
5. Go back to tab 1 and enter keyword (or add a criteria in facade)
Here, tab 1 did NOT redirect to logon.do but showing 0 result found.

PR #1502, opened 2020-02-23
Bugfix/errors in logs after session expiry

PenghaiZhang:
#1071 

##### Checklist

##### Description of change

The error we saw from the Resource-centre logs actually refers to an issue about LegacyContent not properly handling ajax requests. At the moment we can log this error in a better way and fix it in the future.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/LegacyContentApi.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/LegacyContentApi.scala
@@ -65,6 +65,7 @@ import javax.servlet.http.{HttpServletRequest, HttpServletResponse}
 import javax.ws.rs._
 import javax.ws.rs.core.Response.ResponseBuilder
 import javax.ws.rs.core.{CacheControl, Context, Response, UriInfo}
+import org.slf4j.LoggerFactory
 
 import scala.collection.JavaConverters._
 import scala.collection.mutable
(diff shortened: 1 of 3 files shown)


SammyIsConfused:
Looks good to me. I note that this fix doesn't display an error on the user side, but it at least stops the resource centre logs from filling up.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/LegacyContentApi.scala
```diff
@@ -656,6 +658,9 @@ class LegacyContentApi {
       case tr: TemplateResult    => tr.getNamedResult(context, "body")
       case sr: SectionRenderable => sr
       case pr: PreRenderable     => new PreRenderOnly(pr)
+      //Due to many unknowns of what could cause renderedBody being null, return a 500 error at the moment.
+      case _ =>
+        LOGGER.debug("Unknown error at renderedBody - ajaxResponse"); return Response.serverError();
```
edalex-ian: Cool, I see you're simply making the error trigger explicit rather than waiting for RestEasy to stumble over a `null`. As a result also, there is no change to the user experience.

One thing though, rather than a blanket 500, this is probably a good case for 501 - Not Implemented. Because essentially, you're saying if it's not one of the above matches (or types we know how to handle) then fail - i.e. because the code hasn't been implemented to support it (and indeed, probably never will be as we'll get the New UI done first).

So if you can tweak to a  501, that'd be fine.

edalex-ian:
Cool, let's await a clean build, and if it happens before the release branch for 2020.1 is created we'll discuss whether to merge.

Issue #1337, opened 2019-11-04
Add Searching configuration settings to oEQ


cathfitz:
Searching configuration settings are required for admins to be able to configure the new search UI (and still have the settings used for the old UI as well). These include:

* Search page display settings, such as what filter to show, default results order, showing gallery views etc.
* Content indexing config
* Boosting levels

PR #1509, opened 2020-02-26
Back-end for Search page settings

PenghaiZhang:
#1337 

##### Checklist

- [x] tests are included

##### Description of change

Provider REST endpoints to support Seach page settings.

Screenshot from 2020-02-26 12-03-59

Screenshot from 2020-02-26 12-04-56


Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/settings/SettingsResource.scala
```diff
@@ -69,4 +105,26 @@ class SettingsResource {
...
   )
+
+  @GET
+  @Path("searchpage")
+  def loadSearchPageSettings: SearchPageSettings = {
+    searchPrivProvider.checkAuthorised()
+    SearchPageSettings(loadSettings(new SearchSettings), loadSettings(new CloudSettings))
+  }
+
+  @POST
```
PenghaiZhang: Maybe PUT is more suitable ?

SammyIsConfused: I agree with this, PUT fits better. The difference is that PUT is idempotent; meaning if you were to try the same request multiple times the output would be the same as if you did it just once. That applies to changing settings, which is update rather than create operation.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/settings/SettingsResource.scala
```diff
@@ -45,12 +49,44 @@ object SettingTypeLinks {
...
 }
+
+case class GalleryViewSettings(disableImage: Boolean,
+                               disableVideo: Boolean,
+                               disableFileCount: Boolean)
+
+case class SearchPageSettings(showNoneLive: Boolean,
+                              disableGalleryViews: GalleryViewSettings,
+                              disableCloudSearching: Boolean,
+                              defaultResultOrder: String,
```
SammyIsConfused: could `defaultResultOrder` become `defaultSortOrder`?
Also, for `showNoneLive` we need the prefix version of the term (non) rather than none as a value. So it would end up being `showNonLive`

SammyIsConfused:
Tested locally, can confirm that the endpoint works. Looks good to me!

PenghaiZhang:
Hi @edalex-ian , can we have a different PR for SearchFilter? What Sam needs at the moment is SearchSetting and CloudSetting. So we can have these two merged in,  then she can keep moving.

PenghaiZhang:
I will add proper doco soon.

edalex-ian:
> Hi @edalex-ian , can we have a different PR for SearchFilter? What Sam needs at the moment is SearchSetting and CloudSetting. So if we can have these two merged in, then she can keep moving.

Sure can. Each endpoint indeed could (and maybe should - but cloud is so simple) be done one by one.

edalex-ian:
@PenghaiZhang - You mentioned validation... So what is the result at the moment if I do try and set an invalid value for say boosting? What does oEQ respond with?

Source/Plugins/Core/com.equella.base/src/com/tle/common/settings/standard/SearchSettings.java
```diff
@@ -73,6 +74,7 @@
   @Property(key = "search.defaultsort")
   private String defaultSearchSort;
 
+  @JsonIgnore
```
edalex-ian: When reading this class by itself it's going to seem a bit out of place to have a stray `@JsonIgnore`. So how about we add a comment:

```suggestion
  @JsonIgnore // The REST endpoints for this was split into two - one for general search settings, and one for filters - so we don't want this included in the general search settings.
```

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/settings/SettingsResource.scala
```diff
@@ -69,4 +87,46 @@ class SettingsResource {
   def setUISettings(in: UISettings): Unit = RunWithDB.executeWithPostCommit(
     SettingsDB.ensureEditSystem(UISettings.setUISettings(in))
   )
+
+  @GET
+  @Path("search")
+  @ApiOperation(
+    value = "Load Search settings",
+    notes =
+      "This endpoint is used to retrieve general search settings excluding search filter. The corresponding Model class is SearchSettings."
```
edalex-ian: No need to specify the Model classes here, as Swagger will document them for you based on method typings.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/settings/SettingsResource.scala
```diff
@@ -69,4 +87,46 @@ class SettingsResource {
...
+  )
+  def loadSearchSettings: SearchSettings = {
```
edalex-ian: As we add all these additional settings going forward we don't all want them in this one class. I suggest for each grouping we split them into their own class. So for below we'd probably add (in the same package):

* SearchSettingsResource
* CloudSearchSettingsResource

And later

* SearchFiltersResource

PenghaiZhang:
image

PR #1513, opened 2020-02-27
Add new Search Settings page

SammyIsConfused:
##### Checklist

- [x] the [contributor license agreement][] is signed
- [x] commit message follows [commit guidelines][]
- [x] tests are included
- [x] screenshots are included showing significant UI changes

##### Description of change
#1337 

image
image

Add a new "Search Page" subcategory of the "Searching" category of settings, built in material UI with TypeScript. Uses Penghai's new Rest endpoints.

Updated tests to work with the new page. Removed the tests pertaining to search filter settings (as these settings are being moved), and also added test coverage for the disableGalleryViews functionality which until now had none.

[contributor license agreement]: https://www.apereo.org/node/676
[commit guidelines]: https://chris.beams.io/posts/git-commit/

PR #1529, opened 2020-03-07
test: add unit tests for typescript utils

ChristianMurphy:
##### Checklist

- [x] tests are included

##### Description of change

Adds unit tests for typescript utils.
also adds support for writing tests in typescript.

```
-------------------|---------|----------|---------|---------|-------------------
File               | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
-------------------|---------|----------|---------|---------|-------------------
All files          |   95.18 |     87.1 |   86.67 |      96 |                   
 mainui            |   91.67 |      100 |      50 |     100 |                   
  ErrorPage.tsx    |   91.67 |      100 |      50 |     100 |                   
 util              |   95.77 |    86.21 |   92.31 |   95.52 |                   
  actionutil.ts    |      80 |      100 |      75 |      80 | 30,31             
  commonstrings.ts |     100 |      100 |     100 |     100 |                   
  dictionary.ts    |     100 |       50 |     100 |     100 | 15                
  encodequery.ts   |     100 |       90 |     100 |     100 | 17                
  langstrings.ts   |   96.55 |    81.82 |     100 |   96.43 | 57                
  validation.ts    |     100 |      100 |     100 |     100 |                   
-------------------|---------|----------|---------|---------|-------------------

Test Suites: 8 passed, 8 total
Tests:       29 passed, 29 total
Snapshots:   0 total
Time:        6.441s

```


Source/Plugins/Core/com.equella.core/js/tsrc/util/actionutil.test.ts
```diff
@@ -0,0 +1,38 @@
...
+      const action = actionFactory.started("payload");
+      expect(action).toMatchObject({
+        type: "example_STARTED",
+        payload: "payload"
+      });
+    });
+  });
+
+  describe("wrapAsyncWorker", () => {
+    it("should do something?", async () => {
```
ChristianMurphy: open to better descriptions, it's not entirely clear from the code why this util exists.

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/ErrorPage.test.tsx
```diff
@@ -0,0 +1,49 @@
+import * as React from "react";
+import { createMount } from "@material-ui/core/test-utils";
+import ErrorPage from "./ErrorPage";
+
+jest.mock("@material-ui/styles", () => ({
+  makeStyles: () => () => ({ errorPage: "mock-classname" })
+}));
+
+describe("<ErrorPage />", () => {
+  let mount: ReturnType<typeof createMount>;
```
ChristianMurphy: Note, when material ui styles on the page, use `createMount` to automatically setup the context providers.
https://material-ui.com/guides/testing/

ChristianMurphy: /cc @PenghaiZhang  re https://github.com/openequella/openEQUELLA/pull/1231#issuecomment-582915257
This is an example of an inline interface that could not be reused.

In the material ui side the typing is:

```ts
export default function createMount(
  options?: Partial<MountOptions>,
): typeof mount & {
  attachTo: HTMLElement;
  cleanUp(): void;
};
```

because that interface was done inline, rather than through creating a new exported `type` or `interface` the only way to get proper typing in oEQ is to do type extraction `ReturnType<typeof createMount>`

edalex-ian:
Heya @SammyIsConfused 

How's this align with the tests you're adding over in the Search Config work?

Source/Plugins/Core/com.equella.core/js/tsrc/util/langstrings.ts
```diff
@@ -7,8 +7,17 @@ export interface Sizes {
...
   more: string;
 }
+
+/**
+ * Get appropriate language string based off size of value
+ *
+ * @param size size or count of value
+ * @param strings language strings to choose from
+ *
+ * TODO: replace with https://github.com/formatjs/react-intl
```
edalex-ian: Nice! We should probably raise a Issue to do that - especially as we move to the New UI in earnest.

ChristianMurphy: Also be aware react-intl isn't the only option.
there is https://github.com/i18next/react-i18next which is also a good option.

Either way I'd recommend using ICU format
* http://userguide.icu-project.org/formatparse/messages
* https://formatjs.io/guides/message-syntax
* https://react.i18next.com/misc/using-with-icu-format

Because it comes baked in with awareness of different languages' pluralization rules (through CLDR)
http://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html is a nice reminder of the challenges it solves, Arabic and Czech are two languages which are prime examples of pluralization rules very different from English

ChristianMurphy:
Friendly ping @SammyIsConfused @edalex-ian

edalex-ian:
Heya @ChristianMurphy 

@SammyIsConfused is away for the week, but I'm keen to get this in and keep this moving. However just one question before we proceed.

I note this has used the practice of having the `.test.js` file alongside the source file. However the existing/initial jest stuff had used the `__tests__` and `__mocks__` folders. Could you comment as to why you've gone with a different approach?

ChristianMurphy:
> Could you comment as to why you've gone with a different approach?

Sure, a few reasons.
1. for projects which don't have coverage requirements the can fail the build, having the tests next to the source makes it easier for developers remember to update the test.
2. reduction in depth of imports, a simple `./filename` works
3. `versioncheck` and other legacy files already break with Jest testing convention, which has files being tested within the same folder where Jest lives, `versioncheck` lives in `resources` which is a sibling folder of `js`, using `__test__` may be only option for legacy files
4. personal preference

I'd be okay migrating the code to `__test__` if you have a strong preference.

ChristianMurphy:
It's also worth noting https://github.com/openequella/openEQUELLA/pull/1513
takes yet another approach, different from either mentioned above, using a `tssrc/components/test` folder.

https://github.com/openequella/openEQUELLA/blob/331551ecc8d9af79e3301bc671c6e3a817dde6f0/Source/Plugins/Core/com.equella.core/js/tsrc/components/tests/SearchSettingFormControl.test.tsx

edalex-ian:
Thanks for that @ChristianMurphy 

In that case, I think I'd prefer to stick with the `__tests__` directory pattern, and believe this will also become helpful when we start adding mocks and snapshots. Further, this also aligns with the [Jest example projects](https://github.com/facebook/jest/tree/master/examples).

But in typical fashion, this indeed is not simple. For starters we already have JS based code in three places:

* `js/src` - purescript mainly (with JS bridging and FFI stuff)
* `js/tsrc` - Typescript / React - ideally where all the new stuff goes
* `../resources/web/js` - legacy scripts for AJAXy type bits

And then the structure within them is also using different patterns. e.g. in `js/tsrc` we have a root `components` folder, but also then things like `mainui`, `api` and not to mention `utils` and then maybe page based folders....

hmmm, but even with all that (and maybe due to all that) I think it'd be good if we went with this for the new stuff:

* `js/tsrc/__tests__/`
* `js/tsrc/__tests__/snapshots`
* `js/tsrc/__mocks__/`

And then within each we can mirror folder structures as needed. So for example, the `tsrc/util/validation.ts` test would be at `tsrc/__tests__/util/validation.test.js`. This idea almost matches patterns familar from Maven based build tools - but alas we're already a bit hamstrung due to the existing three JS source folders.

Are you cool with that?

And as for #1513 I'll chat with @SammyIsConfused to see if we can do the same over there.

ChristianMurphy:
Can do, pushed c650fd3 which moves the files to `__test__` and also adds `describe` blocks to `versioncheck` test.

Source/Plugins/Core/com.equella.core/js/__tests__/versioncheck.test.ts
```diff
@@ -1,12 +1,8 @@
...
-const version_check_js_file = fs.readFileSync(
-  "../resources/web/js/versioncheck.js",
-  "utf8"
-);
-eval(version_check_js_file);
+import * as mockData from "../__mocks__/versioncheck_mock_data";
+import rewire = require("rewire");
+
+const versioncheck = rewire("../../resources/web/js/versioncheck.js");
+const createCheckResult = versioncheck.__get__("createCheckResult");
```
ChristianMurphy: rewire allows access to non-exported variables, without needing to run `eval`.
https://github.com/jhnns/rewire

PenghaiZhang:
Hi @ChristianMurphy , just remind you that in order to get Codebuild green, you need to merge develop to this branch.

ChristianMurphy:
synced with latest develop

ChristianMurphy:
:thinking: a few test failures, looking at the reasons:
1. a stale html reference? https://travis-ci.com/github/openequella/openEQUELLA/jobs/297598766#L1520-L1521
2. an attachment test failed https://travis-ci.com/github/openequella/openEQUELLA/jobs/297598769#L1789-L1790
3. another stale html reference? https://travis-ci.com/github/openequella/openEQUELLA/jobs/297598770#L1367
4. a broken jar https://travis-ci.com/github/openequella/openEQUELLA/jobs/297598773#L860
5. timeout https://travis-ci.com/github/openequella/openEQUELLA/jobs/297598778#L1199
6. timeout https://travis-ci.com/github/openequella/openEQUELLA/jobs/297598780#L1338

Most of these look like flakiness on CI.
The two the concern me are the attachment test failing and the broken jar.
Neither appear to be related to the changes in this PR though.

edalex-ian:
Hrmph. They're all just flaky I fear. We just need to rerun each job - and before the workspace expires. Are you happy to usher that through?

(This is especially seeing we've seen codebulid was happy.)

ChristianMurphy:
Sure, restarting test jobs now

ChristianMurphy:
> before the workspace expires

Out of curiosity,
Which workspace expires? Travis or the CodeBuild?
How long does it take the workspace to expire?

ChristianMurphy:
Refreshed tests, down to 1 failing test: the attachment test https://travis-ci.com/github/openequella/openEQUELLA/jobs/297598770#L1890

PenghaiZhang:
> Refreshed tests, down to 1 failing test: the attachment test https://travis-ci.com/github/openequella/openEQUELLA/jobs/297598770#L1890

I retried and a green Travis build is here now :)

ChristianMurphy:
Thanks @PenghaiZhang

edalex-ian:
> > before the workspace expires
> 
> Out of curiosity,
> Which workspace expires? Travis or the CodeBuild?
> How long does it take the workspace to expire?

The Travis Workspace. This is still a beta feature, but allows us to share assets from one job to another (such as the installer) without relying on third party storage (e.g. S3). As a result, doco is still a bit light on, but can be seen here: https://docs.travis-ci.com/user/using-workspaces/

Experience has shown us though that they only persist for so long (guessing around 24hours) before they're removed. At which point jobs will just fail to find expected files that would normally be pulled from the workspace.

ChristianMurphy:
thanks @edalex-ian! :bowing_man:

PR #1538, opened 2020-03-15
ci: auto retry failed test ng scripts

ChristianMurphy:
##### Checklist

- [x] tests are included

##### Description of change

these scripts can be flaky (https://github.com/openequella/openEQUELLA/pull/1529#issuecomment-598537057), travis provides a mechanism for auto retry for flaky scripts.
https://docs.travis-ci.com/user/common-build-problems/#travis_retry

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -135,7 +135,7 @@ jobs:
       after_script: ci/scalacheck-save-results.sh Tests-Serial
     # Admin
     - stage: functional test
-      script: ci/oldtests-run.sh
+      script: travis_retry ci/oldtests-run.sh
       workspaces:
         use:
           - oeq-installer
@@ -144,7 +144,7 @@ jobs:
       env: OLD_TEST_SUITE=admin
       after_script: ci/oldtests-save-results.sh
     - stage: functional test
-      script: ci/oldtests-run.sh
+      script: travis_retry ci/oldtests-run.sh
       workspaces:
         use:
           - oeq-installer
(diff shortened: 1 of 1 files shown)


ChristianMurphy:
The trade off here is that actual test failures would take longer to report as travis tries to rule out the failure being a false positive.

ChristianMurphy:
Interestingly enough even with this, I still needed to do some restarts :thinking: 
I wonder if this is at the wrong level, or if Selenium includes it's own retry mechanism that could be leveraged.

PenghaiZhang:
> Interestingly enough even with this, I still needed to do some restarts 
> I wonder if this is at the wrong level, or if Selenium includes it's own retry mechanism that could be leveraged.

I tried this locally not too long ago, with `TESTNG`'s retry. Basically, I just implemented `IRetryAnalyzer` and specified how many times I want to retry. Then, used the RetryAnalyzer in a few flaky tests.

edalex-ian:
> Interestingly enough even with this, I still needed to do some restarts thinking
> I wonder if this is at the wrong level, or if Selenium includes it's own retry mechanism that could be leveraged.

Heya @ChristianMurphy 

Yes I was interested when I saw this come through, but I was worried it might be a bit too heavy handed. As it happens (and as @PenghaiZhang ) captures, we have been looking into options around this and approaches that others take. That said though, we also hadn't come across this feature of travis, so we were interested.

First we were interested to see if SBT had any kind of support, but alas no. (As an aside, [Gradle does](https://blog.gradle.org/gradle-flaky-test-retry-plugin).) So then we looked into the test frameworks themselves, and as you can see it turns out that TestNG does offer us some options.

It seems generally the best practice is to ensure you only explicitly retry specific tests. This is to ensure various things but mainly that developers don't become lazy and that no actual quality issues are masked over. It also seems key that when such is done, you also take the effort to capture which tests are retried in a run and ideally monitor this - say with long term metrics etc.

So yes, maybe for now we should hold off on this approach and wait until we get something like the TestNG offering inplace.

PR #1234, opened 2019-09-09
test: add eslint config create react app to checks

ChristianMurphy:
This eslint config is provided by the react team with a set of recommendations to make react apps run smoothly.

https://github.com/facebook/create-react-app/tree/master/packages/eslint-config-react-app

These recommendations are non-stylistic, focusing on things like the rule of hooks.

https://reactjs.org/docs/hooks-rules.html
https://github.com/facebook/react/tree/master/packages/eslint-plugin-react-hooks

These rules do not overlap with, or interfere with, the existing prettier code style.

PR #1541, opened 2020-03-16
Feature/convert purs to tsx for setting menu

PenghaiZhang:
#1337 

##### Checklist

##### Description of change

Rewrite the Setting Menu and UI Setting Editor in Typescript.


Source/Plugins/Core/com.equella.core/js/tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "outDir": "target/ts",
     "module": "CommonJS",
     "target": "es5",
-    "lib": ["es6", "dom"],
+    "lib": ["es6", "dom", "es2018.promise"],
```
ChristianMurphy: Since TS is going to compile this down to ES5 would there be any issues setting this to support all features up to ES2020?

```suggestion
    "lib": ["es2020", "dom"],
```

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/index.tsx
```diff
@@ -245,10 +243,7 @@ export default function() {
     ReactDOM.render(
       <BrowserRouter basename={basePath} forceRefresh>
         <ThemeProvider theme={oeqTheme}>
-          <bridge.SettingsPage
-            refreshUser={() => {}}
-            updateTemplate={_ => {}}
-          />
+          <SettingsPage refreshUser={() => {}} updateTemplate={_ => {}} />
```
ChristianMurphy: for consistency

```suggestion
          <SettingsPage refreshUser={() => {}} updateTemplate={() => {}} />
```

Source/Plugins/Core/com.equella.core/js/tsrc/settings/SettingsPage.tsx
```diff
@@ -0,0 +1,203 @@
...
+    const groups: SettingGroup[] = [];
+
+    SETTING_CATEGORIES.forEach(category => {
+      const settingsOfCategory = settings
+        .filter(setting => setting.group === category.id)
+        .sort((s1, s2) => {
+          return s1.name > s2.name ? 1 : -1;
+        });
+      groups.push({ category: category, settings: settingsOfCategory });
+    });
```
ChristianMurphy: could `forEach` be replaced with a `map`?

```suggestion
    return SETTING_CATEGORIES.map(category => {
      const settingsOfCategory = settings
        .filter(setting => setting.group === category.id)
        .sort((s1, s2) => {
          return s1.name > s2.name ? 1 : -1;
        });
      return { category: category, settings: settingsOfCategory };
    });
```

Source/Plugins/Core/com.equella.core/js/tsconfig.json
edalex-ian: Heya @ChristianMurphy 

Just to be sure to be sure, I wonder if we should do that in a separate stand-alone PR? Just in case it triggers any other changes in the transpilation. Thoughts?

edalex-ian: Oh whoops. I just realised that @PenghaiZhang has brought in `es2018.promise`.

I guess in that case, maybe here is as good as any to bring in the whole thing and see what happens. It's a big enough change anyway.

ChristianMurphy: > Just to be sure to be sure, I wonder if we should do that in a separate stand-alone PR? Just in case it triggers any other changes in the transpilation. Thoughts?

I'd be okay either way.

> I guess in that case, maybe here is as good as any to bring in the whole thing and see what happens. It's a big enough change anyway.

My thoughts precisely

Source/Plugins/Core/com.equella.core/js/tsrc/settings/SettingsPage.tsx
```diff
@@ -0,0 +1,203 @@
...
+
+interface SettingCategory {
+  id: string;
+  details: {
+    name: string;
+    desc: string;
+  };
+}
+
+interface SettingGroup {
```
PenghaiZhang: Can we just use 'type' here because 'SettingCategory' and 'SettingGroup' probably do not need inheritance and other abilities of 'interface' ?

Source/Plugins/Core/com.equella.core/js/tsrc/settings/SettingsPage.tsx
```diff
@@ -0,0 +1,203 @@
...
+  const [loading, setLoading] = React.useState<boolean>(true);
+
+  // Update the title of App Bar
+  React.useEffect(() => {
+    updateTemplate(templateDefaults(languageStrings["com.equella.core"].title));
+  }, []);
+
+  // Fetch settings from the server
+  React.useEffect(() => {
+    let cancel = false;
```
PenghaiZhang: Help wanted !
Had a short discussion with Tom regarding the usage of this boolean flag. The intention is to avoid the warning: 
`Can only update a mounted or mounting component. This usually means you called setState, replaceState, or forceUpdate on an unmounted component. This is a no-op.`

It seems a lot developers have accepted this approach, but I wonder if we have any better approaches.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/SettingsPage.tsx
```diff
@@ -0,0 +1,203 @@
...
+    fetchSettings()
+      .then(result => {
+        if (!cancel) {
+          setSettings(result.data);
+        }
+      })
+      .catch(error => {
+        handleError(error);
```
PenghaiZhang: I remember we had short discussions about the generic error handling in oEQ. This one looks generic but there is only one usage.
https://github.com/openequella/openEQUELLA/blob/903ca6b8914d84542cafb8fb4a788c8c785cfbea/Source/Plugins/Core/com.equella.core/js/tsrc/components/index.ts#L10

Source/Plugins/Core/com.equella.core/js/tsrc/settings/SettingsPage.tsx
ChristianMurphy: It could be an option.
Generally I've seen `type` either used for managing primitive interfaces which can't be captured with `interface` or type projections ([often using util types](https://www.typescriptlang.org/docs/handbook/utility-types.html)) which again aren't easily captured as interfaces.

What benefit do you have in mind for preferring `type` over `interface` in all cases?

Source/Plugins/Core/com.equella.core/js/tsrc/settings/SettingsPage.tsx
ChristianMurphy: An approach I've liked is wrapping network calls into hooks.
This allows things like cancellation on unmount, suspense support or loading states, etc.

There are some nice general purpose premade hooks like https://use-http.com and https://github.com/zeit/swr (or in this particular case https://github.com/mauricedb/use-abortable-fetch)
Or a hook could be written wrapping a network utility like https://github.com/sindresorhus/ky or https://github.com/axios/axios automatically doing cleanup on unmount.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/SettingsPage.tsx
PenghaiZhang: Thanks @ChristianMurphy.  No, I am not preferring `type`. Yesterday, I  was confused about the mixed use of `type` and `interface` in this file: Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContent.tsx.
I did not work out why using interface for `ChangeRoute` and type for `LegacyContent`

ChristianMurphy: Yeah, I suspect there is a fair amount of that, if we can get ESLint into the project (similar to https://github.com/openequella/openEQUELLA/pull/1234)
We could add https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/consistent-type-definitions.md to normalize this across the project.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/UISettingEditor.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+  const [newUIEnabled, setNewUIEnabled] = React.useState<boolean>(true);
+  const [newSearchEnabled, setNewSearchEnabled] = React.useState<boolean>(
+    false
+  );
+
+  React.useEffect(() => {
+    fetchUISetting()
+      .then(result => {
+        const setting: UISetting = result.data;
```
edalex-tom: The data can be immediately destructed, saves needed to `settings.newUI`

```
const {enabled, newSearch} = result.data.newUI;
```

Source/Plugins/Core/com.equella.core/js/tsrc/settings/UISettingEditor.tsx
```diff
@@ -0,0 +1,121 @@
...
+    saveUISetting(newUIEnabled, enabled)
+      .then(_ => refreshUser())
+      .catch(error => {
+        handleError(error);
+      });
+  };
+
+  return (
+    <ExpansionPanelDetails>
+      <div>
```
edalex-tom: Not sure this `div` is needed

Source/Plugins/Core/com.equella.core/js/tsrc/settings/UISettingEditor.tsx
```diff
@@ -0,0 +1,121 @@
...
+    },
+    enableNewSearchColumn: {
+      flexBasis: "33.3%"
+    }
+  };
+});
+
+interface UISettingEditorProps {
+  refreshUser: () => void;
+  handleError: (error: AxiosError) => void;
```
edalex-tom: There's no guarantee the error will be an AxiosError as the catch could return a runtime error

Source/Plugins/Core/com.equella.core/js/tsconfig.json
edalex-tom: Do we have handling for legacy browsers?  E.g. message stating browser is not supported

Source/Plugins/Core/com.equella.core/js/tsrc/settings/UISettingEditor.tsx
```diff
@@ -0,0 +1,121 @@
...
+      .then(_ => refreshUser())
+      .catch(error => {
+        handleError(error);
+      });
+  };
+
+  return (
+    <ExpansionPanelDetails>
+      <div>
+        <div className={classes.enableNewUIColumn}>
```
edalex-tom: I think this can be achieved with a MUI Grid which will reduce the need for custom styles and cater for responsive design if desired at some point

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/routes.tsx
```diff
@@ -46,7 +47,7 @@ export const routes = {
   Settings: {
     path: "(/access/settings.do|/page/settings)",
     to: "/page/settings",
-    render: (p: OEQRouteComponentProps<any>) => <bridge.SettingsPage {...p} />
+    component: SettingsPage
```
edalex-tom: I take it properties are injected into a component at a higher level, as `routes` has no contract for properties (e.g. `component: "not valid"` is fine and won't cause errors here)

Source/Plugins/Core/com.equella.core/js/tsrc/settings/SettingsPage.tsx
ChristianMurphy: #1234 has been updated, no errors, only a few warnings, if you want to review that.
Once it's merged can start exploring adding additional checks like https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/consistent-type-definitions.md

Source/Plugins/Core/com.equella.core/js/tsrc/settings/SettingsPage.tsx
```diff
@@ -0,0 +1,203 @@
...
+  templateDefaults,
+  templateError,
+  TemplateUpdateProps
+} from "../mainui/Template";
+import { fetchSettings } from "./SettingsPageModule";
+import { GeneralSetting } from "./SettingsPageEntry";
+import { languageStrings } from "../util/langstrings";
+import ExpansionPanel from "@material-ui/core/ExpansionPanel";
+import ExpansionPanelSummary from "@material-ui/core/ExpansionPanelSummary";
+import ExpansionPanelDetails from "@material-ui/core/ExpansionPanelDetails";
```
edalex-tom: Most of the MUI components (excluding icons) can be imported all at once, tree shaking should still work

```
import { Theme, ExpansionPanelDetails, List, ListItem, ListItemText, ExpansionPanel, ExpansionPanelSummary, Typography } from "@material-ui/core";
```

Source/Plugins/Core/com.equella.core/js/tsrc/settings/SettingsPage.tsx
```diff
@@ -0,0 +1,203 @@
...
+    <div id="settingsPage">
+      <AdminDownloadDialog
+        open={adminDialogOpen}
+        onClose={() => setAdminDialogOpen(false)}
+      />
+      {// Display a circular Progress Bar or the Setting menu, depending on the state of 'loading'
+      loading ? (
+        <div className={classes.progress}>
+          <CircularProgress variant="indeterminate" />
+        </div>
```
edalex-tom: This could be a dedicated oEQ component, allowing it to  be defined once and potentially changed later without needing to update all the pages.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/UISettingEditor.tsx
```diff
@@ -0,0 +1,121 @@
+import * as React from "react";
+import { UISetting } from "./SettingsPageEntry";
+import { fetchUISetting, saveUISetting } from "./SettingsPageModule";
+import { ExpansionPanelDetails, Theme } from "@material-ui/core";
```
edalex-tom: Same as before, the MUI core components can be imported in a single line

Source/Plugins/Core/com.equella.core/js/tsrc/settings/SettingsPage.tsx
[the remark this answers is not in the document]
PenghaiZhang: OK I will add an inline comment first. Later, we can discuss this with @edalex-ian

Source/Plugins/Core/com.equella.core/js/tsrc/settings/SettingsPage.tsx
```diff
@@ -0,0 +1,203 @@
...
+  // Create a link for each setting
+  const settingLink = (setting: GeneralSetting): ReactElement => {
+    let link = <div />;
+    if (setting.links.route) {
+      link = <Link to={setting.links.route}>{setting.name}</Link>;
+    } else if (setting.links.href) {
+      link = <a href={setting.links.href}>{setting.name}</a>;
+    } else if (setting.id === "adminconsole") {
+      link = (
+        <a href="javascript:void(0)" onClick={() => setAdminDialogOpen(true)}>
```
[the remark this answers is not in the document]
PenghaiZhang: Hi Tom, this is not a MUI Link. I think it's a `react-router` Link.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/SettingsPage.tsx
```diff
@@ -0,0 +1,203 @@
...
+          {setting.name}
+        </a>
+      );
+    }
+
+    return link;
+  };
+
+  return (
+    <div id="settingsPage">
```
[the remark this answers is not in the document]
PenghaiZhang: I think so because quite a few Selenium tests need it.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/SettingsPage.tsx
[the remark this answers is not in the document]
edalex-tom: Sorry my question wasn't clear, it was more that the `a` hyperlinks may be better done as MUI `Link` as they will be tied into the theme rather than just be raw HTML links.  It will need an alias for the MUI link as it will collide with the router link

Source/Plugins/Core/com.equella.core/js/tsrc/settings/SettingsPage.tsx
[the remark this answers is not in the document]
edalex-tom: Different minds for this but I'd avoid having extra elements that just drive tests, it's adding something to the DOM just a for a test whereas I'd expect extra attributes on real elements (things that are meaningful outside of functional tests) to support tests

Source/Plugins/Core/com.equella.core/js/tsrc/settings/UISettingEditor.tsx
PenghaiZhang: Without this div, the UI Setting Editor will look different. Therefore, some extra styles are required to ensure the UI consistency. So at this stage I am more inclined to keep this div.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/UISettingEditor.tsx
PenghaiZhang: Good idea.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/UISettingEditor.tsx
PenghaiZhang: By using MUI Grid we can remove this div.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/SettingsPage.tsx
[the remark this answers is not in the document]
edalex-ian: Yup, definitely agree with that. But I'd prefer not to see this also trigger a whole lot of updates to tests. So for now, let's keep this as is, but not do this in the future. ;)

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/routes.tsx
PenghaiZhang: Yes, I think they are injected here : https://github.com/openequella/openEQUELLA/blob/220a60faa08bd0b34e3fd1dced038c339e7987d4/Source/Plugins/Core/com.equella.core/js/tsrc/mainui/index.tsx#L135

Source/Plugins/Core/com.equella.core/js/tsconfig.json
ChristianMurphy: this leaves the target for the output the same so browsers should be unaffected.

The closest I've seen oEQ having to an official browser matrix is https://github.com/openequella/openequella.github.io/blob/7001aad5156fd20a910acabb6e8314b239a30bb3/guides/FileManagerGuide.md#not-supported-without-reconfiguration-if-at-all

Source/Plugins/Core/com.equella.core/js/tsrc/settings/SettingsPage.tsx
PenghaiZhang: I have briefly discussed this point with Ian. It might be better that we create an oEQ component for `CircularProgress`  in another PR, and we can replace current usages of `CircularProgress` with the new component.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/SettingsPage.tsx
PenghaiZhang: Thanks @ChristianMurphy . Do you know if any of hooks mentioned above is used in oEQ ?

Source/Plugins/Core/com.equella.core/js/tsrc/settings/SettingsPage.tsx
```diff
@@ -0,0 +1,203 @@
...
+      })
+      .finally(() => setLoading(false));
+
+    return () => {
+      cancel = true;
+    };
+  }, []);
+
+  const handleError = (error: AxiosError) => {
+    updateTemplate(templateError(generateFromError(error)));
```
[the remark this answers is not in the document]
PenghaiZhang: As discussed, once we have successfully setup Redux to handle errors in a centralised way, we can remove these `handlerError`s. So keep it for a while.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/SettingsPage.tsx
ChristianMurphy: At this moment, no pre-built hooks or hooks packages are included in oEQ.
The utility function Axios is.

https://github.com/openequella/openEQUELLA/blob/220a60faa08bd0b34e3fd1dced038c339e7987d4/Source/Plugins/Core/com.equella.core/js/package.json#L31-L69

edalex-tom: Hi @ChristianMurphy , would the Axios example be the cancel token?   We have some internal code implements it but I'm curious to know if there are preferable alternatives.

ChristianMurphy: Yes, cancellation within Axios would be the cancel token https://github.com/axios/axios#cancellation

That could be wrapped as a react hook to automatically cancel on unmount.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/SettingsPageModule.ts
```diff
@@ -1,16 +1,22 @@
...
 
-export const fetchUISetting = (): Promise<UISetting> => {
-  return Axios.get<UISetting>(GET_UI_SETTINGS_URL).then(res => res.data);
+export const fetchUISetting = (token: CancelToken): Promise<UISetting> => {
+  return Axios.get<UISetting>(GET_UI_SETTINGS_URL, {
+    cancelToken: token
+  }).then(res => res.data);
 };
 
 export const saveUISetting = (
```
PenghaiZhang: I did not add cancel token here because cancelling this request will result in different user experience, and it is out of the scope of this PR.

edalex-ian:
Nice. I see Codebuild is green, hopefully Travis agrees and then you can merge on in. :+1:

Source/Plugins/Core/com.equella.core/js/tsrc/settings/SettingsPage.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+  const [adminDialogOpen, setAdminDialogOpen] = React.useState<boolean>(false);
+  const [loading, setLoading] = React.useState<boolean>(true);
+  const [settingGroups, setSettingGroups] = React.useState<SettingGroup[]>([]);
+
+  React.useEffect(() => {
+    updateTemplate(templateDefaults(languageStrings["com.equella.core"].title));
+  }, []);
+
+  React.useEffect(() => {
```
ChristianMurphy: Could `useEffect` and `useState` be destructured either on the import from react or at the beginning of the file?

PenghaiZhang: Sure can.

PR #1557, opened 2020-03-21
style: combine eslint and prettier

ChristianMurphy:
##### Checklist

- [x] the [contributor license agreement][] is signed
- [x] commit message follows [commit guidelines][]

##### Description of change

Builds on #1556 
Will allow issues like the one mentioned in https://github.com/openequella/openEQUELLA/pull/1541#discussion_r393370026 to be handled automatically by ESLint and Prettier in the future

[contributor license agreement]: https://www.apereo.org/node/676
[commit guidelines]: https://chris.beams.io/posts/git-commit/

PR #1558, opened 2020-03-21
style: normalize use of let, const, and interface

ChristianMurphy:
##### Checklist

##### Description of change

builds on #1557 
Sets a preference for using `let` and `const` over `var` in ts files.
Sets a preference for using `interface` over `type` for defining object structures.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "postinstall": "cross-env-shell \"./coursier bootstrap com.geirsson:scalafmt-cli_2.12:1.5.1 -f -o node_modules/.bin/scalafmt && ./coursier bootstrap com.google.googlejavaformat:google-java-format:1.7 -f -o node_modules/.bin/google-java-format\"",
     "format:scala": "find . -name \"*.scala\" -o -name \"*.sbt\" | xargs scalafmt",
     "format:java": "find . -name \"*.java\" | xargs google-java-format -r",
-    "format:ts": "eslint \"Source/Plugins/Core/com.equella.core/js/**/*.{js,ts,tsx}\"",
+    "format:ts": "eslint --fix \"Source/Plugins/Core/com.equella.core/js/**/*.{js,ts,tsx}\"",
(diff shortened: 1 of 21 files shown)


edalex-ian:
Ahhhh, okay. So this looks a lot bigger due to overlap with #1557 . I might wait until you merge that in to come back and review this one. ;)

ChristianMurphy:
@edalex-ian synced with develop, the diff should be easier to read now.

package.json
```diff
@@ -54,6 +54,27 @@
     ],
     "env": {
       "jquery": true
-    }
+    },
+    "overrides": [
+      {
+        "files": [
+          "Source/Plugins/Core/com.equella.core/js/tsrc/**/*.{ts,tsx}"
```
ChristianMurphy: for now these need to be scoped specifically to the typescript folder since they depend on the tsconfig in that folder.
Once everything migrates to ts, this may be able to be simplified.

ChristianMurphy:
friendly ping @edalex-ian, @PenghaiZhang, @SammyIsConfused this is ready for review

PenghaiZhang:
I am happy with these changes :)

ChristianMurphy:
> I'm also wondering how some of those non-null assertion operators could be removed if everything still builds.

In TypeScript non-null assertions loosen type checking, the linter found several places where that was not needed.
Ideally we could remove all uses of this assertion.

It is used in... interesting... ways in parts of the code.
For example, null being assigned as a value that is never null.
https://github.com/openequella/openEQUELLA/blob/91ff95aff6aa246c8031abf6f75d1ecfb2ff854a/Source/Plugins/Core/com.equella.core/js/tsrc/cloudcontrol/index.ts#L83-L84

I originally wanted to also apply https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unnecessary-condition.md in this PR, but backed off because there is a smattering of null checks on never null values in the code, caused by the `!` assertion being used.

PR #1564, opened 2020-03-23
Feature/rest api for enabling general filer

PenghaiZhang:
#1337 

##### Checklist

- [x] tests are included

##### Description of change

* Created a new class called `GeneralFilter` to represent both Owner filter and Date-modified filer, as an inner class of `SearchSetting`. 
* Created a GET and PUT REST APIs to support enabling and disabling the two filters.
* Tests for the new REST APIs.
* Properly rename some Scala methods related to the MIME type filters.

PUT: the response is the updated GeneralFilter
Screenshot from 2020-03-23 16-39-33

GET: the response is a GeneralFilter
Screenshot from 2020-03-23 16-39-52

Here is how these two settings are stored in database
image

### DIFF
--- a/Source/Plugins/Core/com.equella.base/src/com/tle/common/settings/standard/SearchSettings.java
+++ b/Source/Plugins/Core/com.equella.base/src/com/tle/common/settings/standard/SearchSettings.java
@@ -74,11 +74,7 @@ public class SearchSettings implements ConfigurationProperties {
   @Property(key = "search.defaultsort")
   private String defaultSearchSort;
 
-  @Property(key = "search.ownerfilter")
-  private boolean ownerFilter = true;
-
-  @Property(key = "search.datemodifiedfilter")
-  private boolean dateModifiedFilter = true;
+  @JsonIgnore private GeneralFilter generalFilter = new GeneralFilter();
 
   // The REST endpoints for this was split into two - one for general search settings,
   // and one for filters - so we don't want this included in the general search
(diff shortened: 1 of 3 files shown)


edalex-ian:
An initial pondering I have is the endpoint naming. :thinking: 

These are ultimately settings which are controlling what is seen on the Search UI - as opposed to the other filters which are a collection of filters that can be used to do a search. So maybe we need to somehow name it so it highlights that.

oEQ already has the practice of a `/settings/ui`, so maybe we should stick with that and go with: `/settings/search/ui`. Then you'd go with a endpoint description along the lines of: "This endpoint provides configuration of the oEQ Search User Interface - such as what controls are active."

Source/Plugins/Core/com.equella.base/src/com/tle/common/settings/standard/SearchSettings.java
```diff
@@ -226,4 +228,34 @@ public boolean isFileCountDisabled() {
   public void setFileCountDisabled(boolean fileCountDisabled) {
     this.fileCountDisabled = fileCountDisabled;
   }
+
+  public GeneralFilter getGeneralFilter() {
+    return generalFilter;
+  }
+
+  public static class GeneralFilter implements ConfigurationProperties {
```
edalex-ian: As per my ponderings on the naming of the endpoint, I don't think this should be a nested class here. As `SearchSettings` are about the execution of searches. Where as what's being added now is configuration items around the view layer.

As a result, we probably don't want the above getter (`getGeneralFilter`) and this class should probably be stand-alone and named something like `SearchUISettings`.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/settings/SearchFilterResource.scala
```diff
@@ -153,6 +154,42 @@ class SearchFilterResource {
...
+    : Response = {
+    searchPrivProvider.checkAuthorised()
+    val searchSettings = loadSettings(new SearchSettings)
+    val generalFilter  = searchSettings.getGeneralFilter
+    generalFilter.setOwnerFilter(enableOwnerFilter)
+    generalFilter.setDateModifiedFilter(enableDateModifiedFilter)
+    updateSettings(searchSettings)
+    Response.ok().entity(generalFilter).build()
+  }
+
```
edalex-ian: So I'm feeling this should really be on it's own in something like `SearchUISettingsResource`.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/settings/SearchFilterResource.scala
```diff
@@ -153,6 +154,42 @@ class SearchFilterResource {
...
+  @PUT
+  @Path("search/general-filter")
+  @ApiOperation(
+    value = "Update status of general filters",
+    notes = "This endpoint is used to update the status of Owner filter and Date-modified filer.",
+    response = classOf[GeneralFilter]
+  )
+  def updateGeneralFilter(
+      @ApiParam(required = true) @QueryParam("enableOwnerFilter") enableOwnerFilter: Boolean,
+      @ApiParam(required = true) @QueryParam("enableDateModifiedFilter") enableDateModifiedFilter: Boolean)
```
edalex-ian: Most of our other endpoints just expect this kind of thing as a JSON body rather than using params. I think we should just stick with one approach.

edalex-ian:
Actually, after reading the GitHub ticket etc. I think we've got a mix of things going on here. Will discuss with you soon @PenghaiZhang

edalex-ian:
Okay, as per our discussion, let's simply add two new flags to `SearchSettings` (and thereby accessed via the `/settings/search` endpoint. It seems the other flags are all named in the negative, so I guess they'd be

* `searchingDisableOwnerFilter`
* `searchingDisableDateModifiedFilter`

After that, you wont need a new endpoint or anything. However, we will need a fix to the newly merged Search Page Settings (#1513) to support the new model.

PenghaiZhang:
Open a new PR #1573 to support this feature. So close this one.

PR #1590, opened 2020-03-30
Add search filter category

PenghaiZhang:
#1337 

##### Checklist

##### Description of change

Create a new sub-category called Search filters under the new ‘Search’ category.
The Setting filter page will be developed in a different PR.
Also, I wonder if we need to change the description of this category (provided by Cath) because not only the MIME type filter but also the Owner filter and Date-modified filter will be included in the page.

Screenshot from 2020-03-30 11-00-52

### DIFF
--- a/Source/Plugins/Core/com.equella.core/resources/lang/i18n-resource-centre.properties
+++ b/Source/Plugins/Core/com.equella.core/resources/lang/i18n-resource-centre.properties
@@ -3651,7 +3651,7 @@ sessiontitle.tle_scrapbookpicker=Embed a resource attachment
 setting.searchpage.title=Search page
 setting.searchpage.desc=Configure options for the search page
 setting.searchfilter.title=Search filters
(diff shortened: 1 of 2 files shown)


Source/Plugins/Core/com.equella.core/resources/lang/i18n-resource-centre.properties
```diff
@@ -3650,6 +3650,8 @@ sessiontitle.tle_reslinker=Link to a resource
 sessiontitle.tle_scrapbookpicker=Embed a resource attachment
 setting.searchpage.title=Search page
 setting.searchpage.desc=Configure options for the search page
+setting.searchfilter.title=Search filters
+setting.searchfilter.desc=configuration of MIME type searches
```
edalex-ian:
```suggestion
setting.searchfilter.desc=Configure filter visibility and add filters based on resource MIME types
```

PR #1617, opened 2020-04-15
Change styles to allow SettingsListControl to support wider controls

SammyIsConfused:
Such as Sliders, for the Search boosting settings.

##### Checklist

##### Description of change
image
The amount of space dedicated to the ListItemSecondaryAction was prohibitively small - works great for things like checkboxes, but squashes wide components such as sliders. This PR changes this so there is much more room to work with.

Note: The slider item in the list isn't a change, simply a visual example of this change working with a wide component.

Also - add stories for the SettingsListControl.


Source/Plugins/Core/com.equella.core/js/tsrc/components/SettingsListControl.tsx
```diff
@@ -16,7 +16,12 @@ interface SettingsListControlProps {
 
 const useStyles = makeStyles({
   listItemText: {
-    maxWidth: "65%"
+    maxWidth: "40%"
+  },
+  secondaryAction: {
+    width: "400px",
```
edalex-ian, line 22: I'm wondering why you're using pixels here when above we're using percentage. Further, I'm concerned as to how this will play responsively - was that checked at all?

PR #1629, opened 2020-04-22
style: update prettier to version 2

ChristianMurphy:
##### Checklist

##### Description of change

https://prettier.io/blog/2020/03/21/2.0.0.html


edalex-ian:
Nice, I think I like it - I definitely like the trailing commas and parens around single argument arrow functions. Only issue I have is I don't personally like the space after `function`. But I could live with it. Or should we maybe change that?

ChristianMurphy:
I'm okay either way.
Here are the configuration settings for v2 if theres anything you want to tinker with: https://prettier.io/docs/en/options.html

edalex-ian:
Ah well, looks like there's not an option for that. Let's proceed. :+1:

Issue #1656, opened 2020-04-24
A simple login endpoint


edalex-ian:
**Is your feature request related to a problem? Please describe.**

Currently, the main way to use the REST API is via OAuth tokens and the like. However there is no direct means to do so with simple use credentials. When building a stand-alone user focused app (such as, an SPA like being done with the New UI) you really need a means for users to be able to login and then use the REST API to implement the functionality.

**Describe the solution you'd like**

To keep with the currently security architecture, I'd like to be able to use a simple HTTP endpoint to provide a username and password and setup a user session on the server. The result would be the normal returned JSESSIONID cookie which can then be used by all subsequent REST calls.

It'd also be handy if there was then a matching logout endpoint to be able to end that session.

**Describe alternatives you've considered**

Currently the New UI manages to achieve a kind of REST login by wrapping calls to the existing login page via the `/content/submit` endpoint. However this is somewhat convoluted when looking at a standard SPA kind of setup - and relies on the semantics of the existing UI.

edalex-ian:
@mrblippy tested this and said it worked. Closing.

PR #1667, opened 2020-04-29
Basic auth endpoint for REST

edalex-ian:
##### Checklist

- [x] tests are included
- [ ] documentation is changed or added - kinda of (apidocs)

##### Description of change

Provides a simple REST auth endpoint for user (and cookie) based authentication (rather than token / OAuth), utilising existing authentication and authorization mechanisms as referenced from the current form based authentication process.

Testing was a little interesting, as this did not really fit within the current oEQ REST testing paradigm. So I also pushed this out into it's own package (including root package reflecting current situation).

#1656 

### DIFF
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/auth/Auth.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/auth/Auth.scala
@@ -81,8 +81,8 @@ class Auth {
       "This is to logout sessions which were setup with the /api/auth/login endpoint, and will do so based on the JSESSIONID cookie."
   )
   def logout(@Context req: HttpServletRequest): Response = {
+    LegacyGuice.userSessionService.reenableSessionUse()
     val us = LegacyGuice.userService
-    us.logoutURI(CurrentUser.getUserState, null)
     us.logoutToGuest(us.getWebAuthenticationDetails(req), false)
     Response.ok().build()
   }
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/auth/Auth.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/auth/Auth.scala
@@ -1,0 +1,78 @@
+package com.tle.web.api.auth
+
+import com.tle.common.i18n.CurrentLocale
+import com.tle.common.usermanagement.user.{CurrentUser, WebAuthenticationDetails}
+import com.tle.exceptions.{
+  AccountExpiredException,
+  AuthenticationException,
+  BadCredentialsException
+}
+import com.tle.legacy.LegacyGuice
+import com.tle.web.resources.{PluginResourceHelper, ResourcesService}
+import io.swagger.annotations.{Api, ApiOperation}
+import javax.servlet.http.HttpServletRequest
+import javax.ws.rs.core.{Context, Response}
+import javax.ws.rs.{POST, PUT, Path, QueryParam}
+
+@Api("Authentication")
+@Path("auth")
+class Auth {
+  val RESOURCE_HELPER: PluginResourceHelper =
+    ResourcesService.getResourceHelper(classOf[Auth])
+
+  /**
+    * Provide simple username / password login as per a legacy oEQ form based authentication but for
+    * use with REST APIs - possible the start of an authenticated Single Page App. This basically
+    * mimics the existing form based login logic.
+    *
+    * @see com.tle.web.login.LogonSection#authenticate(SectionInfo)
+    */
+  @POST
+  @Path("login")
+  @ApiOperation(
+    value = "Login as a normal user.",
+    notes =
+      "Provides a means to establish a simple cookie based (JSESSIONID) session, for easy use of the REST API for user based operations.",
+    response = classOf[String]
+  )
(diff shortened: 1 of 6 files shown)


edalex-ian:
Failure in GHA is again a flaky test which is completely unrelated: com.tle.webtests.test.searching.hierarchy.HierarchyTopicTest

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/auth/Auth.scala
```diff
@@ -0,0 +1,96 @@
...
+      "This is to logout sessions which were setup with the /api/auth/login endpoint, and will do so based on the JSESSIONID cookie."
+  )
+  def logout(@Context req: HttpServletRequest): Response = {
+    LegacyGuice.userSessionService.reenableSessionUse()
+    val us = LegacyGuice.userService
+    us.logoutToGuest(us.getWebAuthenticationDetails(req), false)
+    Response.ok().build()
+  }
+
+  def loginFailedResponse(wad: WebAuthenticationDetails,
```
PenghaiZhang, line 90: We have a Scala object called `ApiErrorResponse` which generates responses for 404 and 400. What about we move this method to there ?

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/auth/Auth.scala
```diff
@@ -0,0 +1,96 @@
...
+      us.login(username, password, wad, true)
+      Response.ok().build()
+    } catch {
+      case _: BadCredentialsException => lfr("logon.invalid")
+      case e: AccountExpiredException => loginFailedResponse(wad, username, e.getMessage)
+      case _: AuthenticationException => lfr("logon.problems")
+    }
+  }
+
+  @PUT
```
PenghaiZhang, line 76: Can I ask Why use PUT for logout?

autotest/OldTests/testng-travis-webservices.yaml
```diff
@@ -28,6 +28,7 @@ tests:
       - com.tle.webtests.test.webservices.rest.TasksApiTest
       - com.tle.webtests.test.webservices.rest.TaxonomyApiTest
       - com.tle.webtests.test.webservices.rest.UserGroupManagementApiTest
+      - io.github.openequella.rest.AuthApiTest
```
PenghaiZhang, line 31: Will we use ` io.github.openequella.rest.XXX` as package names for REST Api tests created in the future ?

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/auth/Auth.scala
PenghaiZhang: But I realised this method is specific to `loginFailed` so it may be fine to stay here.

edalex-ian: Yeah I was wondering about that, but was then wondering if that's just on the component/search_config branch. :thinking: 

But yes, I wouldn't expect 401s anywhere else. It is unique to the auth process.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/auth/Auth.scala
edalex-ian: Yes, you may, as the choice of verbs was interesting. I went with POST for the login for two reasons:

1. So that it could be used if someone just wanted a simple form login kind of idea - and hence why I went with query params rather than JSON body; and
2. As it does essentially 'create' a new session on the server.

So with that in mind, using PUT at the other end seemed to make sense as it's updating a session. The other option could be DELETE as it's kind of removing a session, but really a session seems to remain in the oEQ Auth paradigm - as you can see we change the session to be as 'guest'.

But with that all now in mind, I'd be interested to hear your thoughts? Does the above play out, or should we change it?

autotest/OldTests/testng-travis-webservices.yaml
edalex-ian: We might, we'll have to see how things play out.

The current REST tests are all token based and are thereby possibly more focused towards system integration as opposed to end user use via a simple login. However the future extensions we'll be focusing on will be to support the new UI so it may make more sense to do them utilising this new login endpoint.

If things do play out that way, then we'll probably create a simpler version of AbstractRestTest and place them here. Plus, then we might be able to avoid also needing to interact with a browser and so then the tests can run much after.

An although it may not influence if we stay with this package name, it would also possibly make sense to write the new tests in Scala - and maybe with ScalaTest - seeing we plan to write all the endpoints with Scala. :thinking: 

So I'm afraid no black and white answer there. But definitely some things to keep in mind when we write the next one. ;)

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/auth/Auth.scala
PenghaiZhang: There have been a lot similar discussions, and one idea got my attention is If `session` is the resource in this scenario, does the path (`/Authentication`)  still match the resource?

PR #1690, opened 2020-05-04
Upgrade SBT to latest 1.1.x version

edalex-ian:
##### Checklist

- [ ] tests are included - NA
- [ ] documentation is changed or added - NA

##### Description of change

Upgrading to latest patch level of SBT. Regardless of what version of SBT you have installed locally, this file governs what is used for the build (unless explicitly set to override at runtime). As a result, we should have it up to date.

(Note: This however is not the latest version of SBT which is 1.3.10. It would be good to upgrade to that, but the build currently fails with anything greater than 1.1.x. I'll raise a ticket soon with further information.)

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version = 1.1.1
+sbt.version = 1.1.6


edalex-ian:
Well, disregarding Travis, that seems to have worked. Should really be fine to merge then.

PR #1701, opened 2020-05-05
Fix FavouriteItemsTest that fails frequently on GHA

PenghaiZhang:
##### Checklist

##### Description of change

`testAutoLoggedIn`, which is part of `FavouriteItemsTest`, has been failing quite frequently on GHA. Since the error is  `NoSuchElementException` and the test is really simple, the first attempt to fix this issue is adding a explicit wait.


PenghaiZhang:
I will re-run several times and see if this test fails again.

edalex-ian:
Well it's all green currently - other than our ol' mate Travis - so I reckon let's merge and see how it goes. It should definitely be an improvement.

PR #1706, opened 2020-05-06
Faceted search setting page

PenghaiZhang:
#1337

##### Checklist

- [ ] tests are included

##### Description of change

Add a new page that displays a list of Faceted search classifications and a dialog that is used to add/update a classification. Like Search filter page, this page does not save any change until click the Save button. But based on Tom's previous feedback, the approach for detecting changes is different (and easier and more readable for me), a few boolean flags are used to help that. 

The functionality of updating and deleting may be implemented in another PR, according to the Jira ticket. Also, once all functionality is implemented,  tests and stories will be added as well.


PenghaiZhang:
Screenshot from 2020-05-06 17-42-58

PenghaiZhang:
Screenshot from 2020-05-06 17-43-15

edalex-ian:
Should we convert this to a draft PR for now then?

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetedSearchSettingsPage.tsx
```diff
@@ -0,0 +1,218 @@
...
+      );
+    });
+  };
+
+  /**
+   * Save classifications in the queue to the Server.
+   * Show the message dialog if any error message is received otherwise show snackbar.
+   */
+  const save = () => {
+    const errorMessages: string[] = [];
```
PenghaiZhang, line 121: We don't need this array for now. When chained `then` returns one or more 207 responses, use this array to store all error messages, and set them to state in the last `then`.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/ClassificationDialog.tsx
```diff
@@ -0,0 +1,160 @@
...
+  const [maxResults, setMaxResults] = useState<number | undefined>();
+
+  const isNameInvalid = validateClassificationFields(name);
+  const isSchemaNodeInvalid = validateClassificationFields(schemaNode);
+
+  /**
+   * Initialise textfields' values, depending on 'onClose'.
+   */
+  useEffect(() => {
+    setName(classification ? classification.name : "");
```
edalex-tom, line 89: Alternative below, it's not much of a difference and the only real nicety is for setMaxResults and the not needing a ternary.
```
setName(classication?.name ?? '')
setSchemaNode(classification?.schemaNode ?? "");
setMaxResults(classification?.maxResults);
```

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/ClassificationDialog.tsx
```diff
@@ -0,0 +1,160 @@
...
+    setMaxResults(classification ? classification.maxResults : undefined);
+  }, [onClose]);
+
+  const onAddOrEdit = () => {
+    addOrEdit({
+      name: name,
+      schemaNode: schemaNode,
+      maxResults: maxResults,
+      orderIndex: highestOrderIndex + 1,
+    });
```
edalex-tom, line 100: Can this be simplified to 

```
addOrEdit({
      name,
      schemaNode,
      maxResults,
      orderIndex: highestOrderIndex + 1,
    });
```

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetedSearchSettingsModule.ts
```diff
@@ -0,0 +1,122 @@
...
+}
+
+const FACETED_SEARCH_API_URL = "api/settings/facetedsearch/classification";
+
+export const getClassificationsFromServer = (): Promise<
+  FacetedSearchClassification[]
+> => Axios.get(FACETED_SEARCH_API_URL).then((res) => res.data);
+
+export const batchUpdateOrAdd = (
+  classifitioncas: FacetedSearchClassification[]
```
edalex-tom, line 66: Should this be `classifications`?

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetedSearchSettingsModule.ts
```diff
@@ -0,0 +1,122 @@
...
+export const getHighestOrderIndex = (
+  classifications: ModifiedClassification[]
+) => {
+  const classification = classifications.reduce((prev, current) => {
+    return prev.orderIndex > current.orderIndex ? prev : current;
+  }, defaultClassifion);
+  return classification.orderIndex;
+};
+
+const defaultClassifion: FacetedSearchClassification = {
```
edalex-tom, line 118: This should be made a function or read only, otherwise it could be mutated by accident

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetedSearchSettingsModule.ts
```diff
@@ -0,0 +1,122 @@
...
+  });
+
+/**
+ * Convert an instace of ModifiedClassification to an instace of FacetedSearchClassification.
+ */
+export const removeFlags = (classification: ModifiedClassification) => {
+  delete classification.deleted;
+  delete classification.changed;
+  delete classification.dirty;
+  return classification;
```
edalex-tom, line 103: This approach tends to scale poorly (e.g. if a new field is added to ModifiedClassification then this will still appear to work), as discussed this won't provide any type safety (it's also missing the return type).

I'm not sure there's any better approaches though, I've had a brief look and there doesn't seem to be any nice way of converting interfaces and it can't be easily achieved with reflection on an interface

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetedSearchSettingsModule.ts
```diff
@@ -0,0 +1,122 @@
...
+  classification: FacetedSearchClassification,
+  changed: boolean,
+  deleted: boolean,
+  dirty: boolean
+): ModifiedClassification =>
+  Object.assign(classification, {
+    changed: changed,
+    deleted: deleted,
+    dirty: dirty,
+  });
```
edalex-tom, line 94: I'm not sure this provides any type safety, it may be better as a simple spread operator as that would 

1. not mutate the original object, the `classification` instance is being modified here
2. typescript should pickup on any key mismatches

It may be simpler to just spread the extra values rather than having a `addFlags` method at all, needing to pass multiple booleans in a method may cause confusion as it's hard to track what the correct order when calling the method.  JS doc for the arguments may help as they'd give code hints for development

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetedSearchSettingsModule.ts
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * Given a list of classifications, find the one that has the highest order index.
+ */
+export const getHighestOrderIndex = (
+  classifications: ModifiedClassification[]
+) => {
+  const classification = classifications.reduce((prev, current) => {
+    return prev.orderIndex > current.orderIndex ? prev : current;
+  }, defaultClassifion);
+  return classification.orderIndex;
```
edalex-tom, line 115: Could be simplified with a `.map` and `Math.max`. the reduce is better if there's a need to get the object itself rather than just the `orderIndex`

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetedSearchSettingsPage.tsx
```diff
@@ -0,0 +1,218 @@
...
+
+/**
+ * A page for setting Faceted search classifications.
+ */
+const FacetedSearchSettingsPage = ({ updateTemplate }: TemplateUpdateProps) => {
+  const facetedsearchsettingStrings =
+    languageStrings.settings.searching.facetedsearchsetting;
+  const classes = useStyles();
+
+  // Show the snackbar when all the status of a 207 response is 2xx.
```
edalex-tom, line 79: Not sure the comments should be here, it's dictating how the state is used when ideally this is just defining state.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetedSearchSettingsPage.tsx
```diff
@@ -0,0 +1,218 @@
...
+  const [resultMessages, setResultMessagesMessages] = useState<string[]>([]);
+
+  const [showEditingDialog, setShowEditingDialog] = useState<boolean>(false);
+  const [classifications, setClassifications] = useState<
+    ModifiedClassification[]
+  >([]);
+
+  const addOrEditQueue = classifications
+    .filter((classification) => classification.changed)
+    .map((classification) => removeFlags(classification));
```
edalex-tom, line 92: not sure `addOrEditQueue` is describing what this is actually doing

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetedSearchSettingsPage.tsx
```diff
@@ -0,0 +1,218 @@
...
+   * Save classifications in the queue to the Server.
+   * Show the message dialog if any error message is received otherwise show snackbar.
+   */
+  const save = () => {
+    const errorMessages: string[] = [];
+    batchUpdateOrAdd(addOrEditQueue)
+      .then((messages) => {
+        errorMessages.push(...messages);
+        if (errorMessages.length > 0) {
+          setResultMessagesMessages(errorMessages);
```
edalex-tom, line 126: Having trouble identifying why `errorMessages` is defined, it looks like the `messages` could be used directly

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetedSearchSettingsPage.tsx
```diff
@@ -0,0 +1,218 @@
...
+            })}
+        </List>
+        <CardActions className={classes.cardAction}>
+          <IconButton
+            onClick={() => setShowEditingDialog(true)}
+            aria-label={facetedsearchsettingStrings.add}
+            color={"primary"}
+          >
+            <AddCircleIcon fontSize={"large"} />
+          </IconButton>
```
edalex-tom, line 195: May be going into component overdrive (i.e. writing too many generic components) but this could be a common `AddButton` which would standardise it and reduce the need to import IconButton + Icon

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/ClassificationDialog.tsx
PenghaiZhang: I like this. I need to update my Intellij as the version I am using does not recognise this operator.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetedSearchSettingsPage.tsx
PenghaiZhang: I will talk to Ian in sprint review and see if we want to do another commonality refactoring.

Source/Plugins/Core/com.equella.core/js/tsrc/components/MessageDialog.tsx
```diff
@@ -30,25 +30,24 @@ import { commonString } from "../util/commonstrings";
 interface MessageDialogProps {
   open: boolean;
   title: string;
-  subtitle: string;
```
edalex-ian, line 33: Why was this removed?

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetDialog.tsx
```diff
@@ -0,0 +1,163 @@
...
+} from "@material-ui/core";
+import { commonString } from "../../../util/commonstrings";
+import { languageStrings } from "../../../util/langstrings";
+import { useEffect, useState } from "react";
+import {
+  FacetWithFlags,
+  validateFacetFields,
+} from "./FacetedSearchSettingsModule";
+
+interface facetDialogProps {
```
edalex-ian, line 35:
```suggestion
interface FacetDialogProps {
```

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetDialog.tsx
```diff
@@ -0,0 +1,163 @@
...
+    setSchemaNode(facet?.schemaNode ?? "");
+    setMaxResults(facet?.maxResults);
+  }, [onClose]);
+
+  const onAddOrEdit = () => {
+    addOrEdit({
+      name,
+      schemaNode,
+      maxResults,
+      orderIndex: highestOrderIndex + 1,
```
edalex-ian, line 99: Feels like this logic should be in the calling component. Actually, indeed if this is an 'edit' then this shouldn't change at all. So yes, this should be logic in the `addOrEdit` method and `highestOrderIndex` should be removed from the props interface.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetedSearchSettingsPage.tsx
```diff
@@ -0,0 +1,209 @@
...
+                      <EditIcon />
+                    </IconButton>
+                    |
+                    <IconButton color="secondary">
+                      <DeleteIcon />
+                    </IconButton>
+                  </ListItemSecondaryAction>
+                </ListItem>
+              );
+            })}
```
edalex-ian, line 177: For readability, could we please extract this out into a `const` and then simply place the variable in here? I'd like to start minimising the use of logic and functions within the return/render methods.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetedSearchSettingsPage.tsx
```diff
@@ -0,0 +1,209 @@
...
+      <FacetDialog
+        addOrEdit={addOrEdit}
+        open={showEditingDialog}
+        onClose={() => setShowEditingDialog(false)}
+        handleError={handleError}
+        facet={undefined}
+        highestOrderIndex={getHighestOrderIndex(facets)}
+      />
+
+      <MessageDialog
```
edalex-ian, line 199: Can you please remind me why this is here (rather than just using the message mechanism in the template)? Is it due to the batch CRUDing?

Source/Plugins/Core/com.equella.core/js/tsrc/components/MessageDialog.tsx
PenghaiZhang: I found the other two dialogs(`AdminDownloadDialog` and `ConfirmDialog`) do not have a `subtitle`, and none of the MUI Dialog examples use subtitle. So I removed it.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetDialog.tsx
```diff
@@ -0,0 +1,163 @@
...
+  const isNameInvalid = validateFacetFields(name);
+  const isSchemaNodeInvalid = validateFacetFields(schemaNode);
+
+  /**
+   * Initialise textfields' values, depending on 'onClose'.
+   */
+  useEffect(() => {
+    setName(facet?.name ?? "");
+    setSchemaNode(facet?.schemaNode ?? "");
+    setMaxResults(facet?.maxResults);
```
SammyIsConfused, line 91: What happens here if `facet` is undefined? Is that a possible situation?

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetDialog.tsx
```diff
@@ -0,0 +1,163 @@
...
+          margin="dense"
+          label={facetFieldStrings.schemanode}
+          value={schemaNode}
+          required
+          fullWidth
+          onChange={(event) => setSchemaNode(event.target.value)}
+          error={!!schemaNode && isSchemaNodeInvalid}
+        />
+        <TextField
+          type={"number"}
```
SammyIsConfused, line 136: Unnecessary braces

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetDialog.tsx
```diff
@@ -0,0 +1,163 @@
...
+          margin="dense"
+          label={facetFieldStrings.categorynumber}
+          value={maxResults}
+          fullWidth
+          onChange={(event) => {
+            const value = event.target.value;
```
SammyIsConfused, line 142: Rather than setting a `const`, couldn't you just refer to `event.target.value` directly in the `setMaxResults` function call?

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetedSearchSettingsModule.ts
```diff
@@ -0,0 +1,100 @@
...
+    FACETED_SEARCH_API_URL,
+    facets.map((facet) => removeFlags(facet))
+  ).then((res) => groupErrorMessages(res.data));
+
+/**
+ * Validate if trimmed name or schema node is empty.
+ * Return true if they are invalid.
+ */
+export const validateFacetFields = (field: string): boolean => {
+  return !field?.trim();
```
SammyIsConfused, line 78: Why the `?`? This function is typed such that field must be a string.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetedSearchSettingsPage.tsx
```diff
@@ -0,0 +1,209 @@
...
+    >
+      <Card className={classes.spacedCards}>
+        <List
+          subheader={
+            <ListSubheader disableGutters>
+              {facetedsearchsettingStrings.name}
+            </ListSubheader>
+          }
+        >
+          {facets
```
SammyIsConfused, line 160: Should be its own function.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetDialog.tsx
PenghaiZhang: my code was looking like this : 
`    let newFacet: FacetWithFlags;
    if(!facet){
      newFacet = {
        name,
        schemaNode,
        maxResults,
        orderIndex: highestOrderIndex + 1,
        updated: true,
        deleted: false,
        dirty: true,
      }
    }
    else{
      newFacet = {
        ...facet,
        name,
        schemaNode,
        maxResults,
        updated: true
      }
    }
    addOrEdit(newFacet);`

Based on whether `facet` is undefined or not we know this action is adding or updating. And then construct an object based on this action. 

But before I opened a PR I simplified this function as this PR only focuses on `adding`.

Source/Plugins/Core/com.equella.core/js/tsrc/util/langstrings.ts
```diff
@@ -420,9 +421,6 @@ export const languageStrings = {
         save: "save Search filter configurations",
         filternamelabel: "Name",
         mimetypelistlabel: "MIME types *",
-        messagedialogtitle: "Search filter setting results",
```
SammyIsConfused, line 423: Why are `messagedialogtitle` and `messagedialogsubtitle` removed? I can see that the `messagedialogsubtitle` is moved to `errors`, but the other one just seems to drop off.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetDialog.tsx
PenghaiZhang: But no worries I will move highestOrderIndex to the calling component.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetDialog.tsx
PenghaiZhang: Yes `facet` can be undefined. When this happens, empty strings will be set to `name` and `schemaNode`.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetDialog.tsx
edalex-ian: Well actually that highlights the main issue here. This dialog only needs to know about three things - the three fields it's displaying. So really really, that's all it should be putting in this object, everything else should be left to higher up where it knows the details.

As a result, these lines should really just be:

```
    addOrEdit({
      name,
      schemaNode,
      maxResults });
```

All the flags etc are only the concern of the calling component.

Source/Plugins/Core/com.equella.core/js/tsrc/components/MessageDialog.tsx
edalex-ian: Gotcha. Good plan. :+1:

Source/Plugins/Core/com.equella.core/js/tsrc/util/langstrings.ts
PenghaiZhang: Yes, as I mentioned above I found most `Dialog`s do not use subtitle so I removed it.

SammyIsConfused: When you say "most", does that mean all of them don't use the subtitle prop? If so then yes, feel free to remove.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetDialog.tsx
SammyIsConfused: And maxResults is allowed to be undefined. Gotcha. :+1: Thanks for clarifying.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetDialog.tsx
PenghaiZhang: I am afraid above may not work because the type of Facet requires orderIndex unless this field can be undefined. Alternatively I can create another interface which only includes these three fields.

Source/Plugins/Core/com.equella.core/js/tsrc/util/langstrings.ts
PenghaiZhang: I'd say yes. In oEQ existing dialogs and MUI Dialog examples I don't see any `subtitle`.  In MUI examples, they just put `title` in  `DialogTitle` and no `subtitle` is used.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetDialog.tsx
PenghaiZhang: I just recalled that in back-end this field is indeed nullable.

PenghaiZhang: I just recalled that in back-end the default value of `orderIndex` is 0, so it may be OK to allow it to be undefined in front-end.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetedSearchSettingsPage.tsx
```diff
@@ -141,6 +153,43 @@ const FacetedSearchSettingsPage = ({ updateTemplate }: TemplateUpdateProps) => {
...
   };
 
+  /**
+   * Only renders a ListItem for each non-deleted facet.
+   */
+  const facetListItems: ReactElement[] = facets
+    .filter((facet) => !facet.deleted)
+    .map((facet, index) => {
+      return (
+        <ListItem divider={true} key={index}>
```
SammyIsConfused, line 163: You can just use `divider` without `={true}`

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/facetedsearch/FacetedSearchSettingsPage.tsx
```diff
@@ -0,0 +1,231 @@
...
+  };
+
+  /**
+   * Only renders a ListItem for each non-deleted facet.
+   */
+  const facetListItems: ReactElement[] = facets
+    .filter((facet) => !facet.deleted)
+    .map((facet, index) => {
+      return (
+        <ListItem divider key={index}>
```
edalex-ian, line 163: Just a reminder, performance wise you normally don't want to simply use an item's array index as the key. Typically in this case - if I didn't have unique id from a DB for example - I'd go with something like `facet.name + index`.

This'll be more important once the editing/deleting/re-ordering is added.

Issue #1678, opened 2020-05-01
Contribution wizard with visibility scripted controls and multiple pages erasing metadata


cathfitz:
**Describe the bug**
A common configuration for clients with moderation is to have different moderators see different metadata controls.  Because of the way visibility scripting on pages works, if you only have one page control with all of your different controls on them, each with visibility scripting such that different controls only appear for specific roles or at different moderation stages, you will lose metadata.  This is because when you hide a control on a page, it erases that control’s metadata.  This is normal behaviour.

The work around for this is to construct different pages for different roles/moderation steps.  So Page 1/Draft contribution might have a bunch of controls.  Then Page 2 may have a subset, or superset of controls for the 1st moderator to perform ... then Page 3 and so on.

So when moderators look at their pages, previous pages don’t render … so visibility scripting can’t erase metadata that was entered in previous pages … until now.

An example is a client that has a collection where the initial contribution/draft if you're a manager, sets a specific user as an owner of sorts.  There is a separate page (page 2) for instructors that asks a subset of questions, including "do you know who the owner is" and if so, allows instructors to set the owner user.  In the case where a manager contributes, it then goes to the owner, bypassing the second page (the instructor).  When the owner gets it in their moderation, they have questions that trigger visibility scripts.  Just as soon as they trigger their visibility script, it wipes out the registered owner user.  They save their item and try to continue the moderation and get and AJAX error.  The logs show that the user is not allowed to moderate the item.  That's because the user for which that moderation step is configured is now null ... the metadata has been erased.  Note that the second page - the instructor's initial contribution never rendered.

**Platform:**

- OpenEquella Version: [e.g. 2019.2]
- OS: [e.g. Windows 10, MacOS 10.13]
- Browser [e.g. Chrome, Firefox]

**Additional context**

PR #1708, opened 2020-05-07
Contribution wizard lost metadata fix

SammyIsConfused:
To ensure metadata attributed to hidden pages is not wiped by the
mandatory field check.

##### Checklist

##### Description of change
#1678 
The lost metadata was caused by the mandatory fields check - it would initialize all pages of a wizard. Since wizards can have visibility scripting, this caused issues with hidden pages initializing when they shouldn't, which wipes their metadata. 
Surrounding this initialisation with a check against isViewable() solves the issue whilst retaining mandatory field checking. 

### DIFF
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/section/RootWizardSection.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/section/RootWizardSection.java
@@ -54,6 +54,7 @@ import com.tle.web.template.Breadcrumbs;
 import com.tle.web.template.Decorations;
 import com.tle.web.template.Decorations.MenuMode;
 import com.tle.web.viewurl.ItemSectionInfo;
+import com.tle.web.wizard.WebWizardPage;
 import com.tle.web.wizard.WizardExceptionHandler;
 import com.tle.web.wizard.WizardService;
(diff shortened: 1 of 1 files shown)


Source/Plugins/Core/com.equella.core/src/com/tle/web/wizard/section/RootWizardSection.java
```diff
@@ -305,6 +305,11 @@ void validateMandatoryFields(SectionInfo info, WizardState state) {
     PagesSection ps = info.lookupSection(PagesSection.class);
     wizardService
         .getWizardPages(state)
-        .forEach(p -> wizardService.ensureInitialisedPage(info, p, ps.getReloadFunction(), true));
+        .forEach(
+            p -> {
+              if (p.isViewable()) {
+                wizardService.ensureInitialisedPage(info, p, ps.getReloadFunction(), true);
+              }
+            });
```
edalex-ian, line 313: So as to stay within the functional paradigm a bit, how's about:

```suggestion
        .stream().filter(WebWizardPage::isViewable)
        .forEach(p -> wizardService.ensureInitialisedPage(info, p, ps.getReloadFunction(), true));
```

I think this also has a slight readability benefit.

PR #1710, opened 2020-05-07
REST endpoints for batch operations on Faceted search classifications

PenghaiZhang:
#1337 

##### Checklist

- [x] tests are included

##### Description of change

Add a PUT endpoint which create or update multiple classifications, and a DELETE endpoint which deletes multiple classifications.

Screenshot from 2020-05-07 10-24-49
Screenshot from 2020-05-07 10-47-30

### DIFF
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/ApiBatchOperationResponse.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/ApiBatchOperationResponse.scala
@@ -19,3 +19,9 @@
 package com.tle.web.api
 
 case class ApiBatchOperationResponse(id: String, status: Int, message: String)
+
+object ApiBatchOperationResponse {
+  def apply(id: Long, status: Int, message: String): ApiBatchOperationResponse = {
+    ApiBatchOperationResponse(id.toString, status, message)
+  }
+}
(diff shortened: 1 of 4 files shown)


PenghaiZhang:
We need to update package-lock in order to get GHA happy. We can probably do that directly on component branch.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/ApiBatchOperationResponse.scala
```diff
@@ -19,3 +19,9 @@
 package com.tle.web.api
 
 case class ApiBatchOperationResponse(id: String, status: Int, message: String)
+
+object ApiBatchOperationResponse {
+  def apply(id: Long, status: Int, message: String): ApiBatchOperationResponse = {
+    ApiBatchOperationResponse(id.toString, status, message)
+  }
+}
```
edalex-ian, line 27: With this being added, do we need to update the previous batch methods to also use this?

PenghaiZhang: Because Search filters use UUID as their IDs, the previous batch methods do not call this `apply` when constructing an instance of `ApiBatchOperationResponse`. They use the default one provided by case class.

edalex-ian: Excellent. Good to go.

Strange that package-lock.json needs updating. :thinking: But if so....

PR #1722, opened 2020-05-11
Setup a standalone JS module for OEQ REST API interaction

edalex-ian:
##### Checklist

- [x] tests are included

##### Description of change

This is the start of a JS module for interacting with OEQ via REST. Currently it is being built within the core repo for simplicity of integration. However, once it matures a bit, it will be split into its own repo and published (maybe just with GitHub packages).

This PR is targetting the `component/search_configuration` branch, as the next step will be to add support to it for the interrogation of oEQ Schema endpoints to be used in the faceted search configuration. The endpoints currently supported were kind of the bare minimum to prove the concept.

Although tests are there - and they do work locally - more effort is required to make it so that they can be part of the pipeline. Hopefully this can be worked on in subsequent commits. Indeed, more effort is due on the build in general. It was bootstrapped with TSDX, but bit by bit key areas are being substituted out and aligned with the rest of the OEQ JS build setup.

The plan going forward (other than the above mentioned Schema endpoint support) will be to refactor existing endpoint calls into this module and establish a good level of decoupling between the View and the REST calls. And further, all new REST calls would be added into this module from the start.

As patterns are emerging they're being refactored (such as the `GET` and `PUT`) and I expect this to continue to evolve as needed. Also note, that for now I've avoided cancel tokens, but if the need arises we can add them in (in an abstract form to maintain decoupling). I'm trying to balance DRY and YAGNI principles as well as minimal complexity, but there's still a few things that couple be improved.

Lastly, you'll note there is explicit handling of cookies. This is mainly for the tests, and in theory turns into a noop in browsers. I'm not intending this to be used by CLI node apps, but it may just end up being a side effect/benefit of running tests in Jest via Node. I did almost swap to Mocha tests, but for now things seem happy enough.


edalex-ian:
Oh, I should also say I'm keen to see how this works in the CI environment due to the interesting linking of builds between Core/js and the module via `prepare` script. :thinking:

oeq-ts-rest-api/src/LegacyContent.ts
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Retrieve details of the current user (based on JSESSIONID) including details for the UI such
+ * as menu structure and task and notification counts.
+ *
+ * @param apiBasePath Base URI to the oEQ institution and API
+ */
+export const getCurrentUserDetails = (
+  apiBasePath: string
+): Promise<CurrentUserDetails> =>
+  GET<CurrentUserDetails>(apiBasePath + '/content/currentuser');
```
ChristianMurphy, line 40: Really like that the types carry through here. :+1: 

Something which may be worth exploring, a lightway check that the data matches the expectation can be useful as an early warning system if the client get's out of sync with the API itself.
For example:
https://github.com/gcanti/io-ts
https://github.com/sindresorhus/ow
https://github.com/ianstormtaylor/superstruct

oeq-ts-rest-api/jest.config.js
```diff
@@ -0,0 +1,189 @@
...
+
+  // Indicates whether each individual test should be reported during the run
+  // verbose: null,
+
+  // An array of regexp patterns that are matched against all source file paths before re-running tests in watch mode
+  // watchPathIgnorePatterns: [],
+
+  // Whether to use watchman for file crawling
+  // watchman: true,
+};
```
ChristianMurphy, line 189: This was a bit tough to read, the comment at the top linking to the documentation helps.
100+ lines of comments for options not used make it difficult (for me) to read.

```suggestion
module.exports = {
  clearMocks: true,
  coverageDirectory: "coverage",
  preset: "ts-jest",
  testEnvironment: "node",
};
```

Source/Plugins/Core/com.equella.core/js/package.json
```diff
@@ -44,6 +45,7 @@
     "lodash": "4.17.15",
     "luxon": "1.16.0",
     "material-ui-pickers": "2.2.4",
+    "oeq": "file:../../../../../oeq-ts-rest-api",
```
ChristianMurphy, line 48: This should be fine for now, if this gets moved to the `npm` or `github` registry down the road, it might be good to match the package name.

oeq-ts-rest-api/.gitignore
```diff
@@ -0,0 +1,5 @@
+*.log
+.DS_Store
+node_modules
+dist
+.vscode
```
ChristianMurphy, line 5: Is this file included as an intermediate step to prep for moving this to another repo?

oeq-ts-rest-api/test/TestConfig.ts
```diff
@@ -0,0 +1,3 @@
+export const API_PATH = 'http://localhost:8080/rest/api';
+export const USERNAME = 'TLE_ADMINISTRATOR';
+export const PASSWORD = 'abc';
```
ChristianMurphy, line 3: Nice to have, a newline at end of file
```suggestion
export const PASSWORD = 'abc';

```

oeq-ts-rest-api/src/index.ts
```diff
@@ -0,0 +1,4 @@
+export * as Auth from './Auth';
+export * as LegacyContent from './LegacyContent';
+export * as Errors from './Errors';
+export * as Settings from './Settings';
```
ChristianMurphy, line 4: Nice to have, a newline at the end of the file

```suggestion
export * as Settings from './Settings';

```

oeq-ts-rest-api/package.json
```diff
@@ -0,0 +1,49 @@
...
+      "pre-commit": "npm run lint"
+    }
+  },
+  "prettier": {
+    "printWidth": 80,
+    "semi": true,
+    "singleQuote": true,
+    "trailingComma": "es5"
+  },
+  "name": "oeq-ts-rest-api",
```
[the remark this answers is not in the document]
edalex-ian: I like that idea. Will do. Thanks. :+1:

oeq-ts-rest-api/src/LegacyContent.ts
edalex-ian: I like the sounds of that. Will have a closer look and see if I can maybe slot something into the `GET`.

oeq-ts-rest-api/.gitignore
edalex-ian: That's correct. There's a couple of top level files that I've left there in prep for when it goes stand-alone (including things like the Husky set-up).

oeq-ts-rest-api/src/LegacyContent.ts
ChristianMurphy: A snippet I've used on past projects

```ts
import { Mixed, TypeOf } from 'io-ts';
import { either, pipeable } from 'fp-ts';

/**
 * Takes unknown data and returns a strongly typed data.
 * If data does not match expected API interface, it throws a TypeError.
 *
 * @throws TypeError
 */
export const validateDataShape = <T extends Mixed>(data: unknown, validateShape: T): TypeOf<T> =>
  pipeable.pipe(
    validateShape.decode(data),
    either.fold(
      /**
       * Where there is an error, throw exception and log a message.
       * Example message: 'value 3 is unexpected type, at root > property > 2 > isActive,
       * actual type "string" expected type "boolean"'
       */
      errors => {
        throw new TypeError(
          errors
            .map(({ message, context, value }) => {
              const overallErrorDescription = message || `value ${value} is unexpected type, at`;
              return `${overallErrorDescription} ${context
                .map(({ key }) => key || 'root')
                .join(' > ')}, actual type "${typeof context[context.length - 1]
                .actual}" expected type "${context[context.length - 1].type.name}"`;
            })
            .join('; '),
        );
      },
      /**
       * Pass valid value through
       */
      value => value,
    ),
  );

export const fetchWrapped = async <T extends Mixed>({
  url,
  shape,
  token,
  requestOptions,
}: UseFetchProps<T>) => {
  const response = await fetch(url, generateRequestOptions(requestOptions, token));

  let data;

  try {
    data = await response.json();
  } catch (e) {
    // the server might respond with an empty body
    // in some cases, which should be ignored
  }

  if (data?.status >= 400) {
    throw new Error(data.statusText);
  }

  return validateData(data, shape);
};
```

`fetch` is the stand in for axios,
and where shape can be any io-ts definition, for example:

```ts
import * as t from 'io-ts';

export const DefinedType = t.strict({
  type: t.string,
  isActive: t.boolean,
});
```

oeq-ts-rest-api/.gitignore
ChristianMurphy: Thanks! That makes sense

oeq-ts-rest-api/package.json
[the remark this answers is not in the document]
edalex-ian: Done.

Source/Plugins/Core/com.equella.core/js/package.json
edalex-ian: Good plan. Done. ;)

Source/Plugins/Core/com.equella.core/js/tsrc/settings/SettingsPage.tsx
```diff
@@ -81,22 +80,22 @@ const SettingsPage = ({ refreshUser, updateTemplate }: SettingsPageProps) => {
...
-          return; // Request was cancelled
-        }
         handleError(error);
         setLoading(false);
       });
 
     return () => {
-      cancelToken.cancel();
+      cleanupTriggered = true;
+      // cancelToken.cancel();
```
SammyIsConfused, line 98: Does this still need to be here? Or is it just a placeholder in case we need CancelTokens in the future?

oeq-ts-rest-api/README.md
```diff
@@ -0,0 +1,72 @@
+# openEQUELLA Typescript REST API SDK
+
+This module provides a code generated SDK for access to the openEQUELLA REST API.
+
+## Local Development
+
+This project was bootstrapped with [TSDX](https://github.com/jaredpalmer/tsdx) but the subsequently
```
SammyIsConfused, line 7:
```suggestion
This project was bootstrapped with [TSDX](https://github.com/jaredpalmer/tsdx) but then subsequently
```

oeq-ts-rest-api/README.md
```diff
@@ -0,0 +1,72 @@
...
+tweaked (especially for Jest and Eslint).
+
+Below is a list of commands you may find useful.
+
+### `npm start` or `yarn start`
+
+Runs the project in development/watch mode. Your project will be rebuilt upon changes. TSDX has a
+special logger for you convenience. Error messages are pretty printed and formatted for
```
SammyIsConfused, line 15:
```suggestion
special logger for your convenience. Error messages are pretty printed and formatted for
```

oeq-ts-rest-api/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+### Codegen just for the models
+
+This would be a great win, but the swagger/openapi codgen tools don't really provide a means to
+do this. But even if they did, again due to the way the oEQ API has been built over time (very much
+a code first approach) the resultant models would not necessarily be ideal - especially with a strong
+prevalence of the use of maps. (Maps are not really directly supported via swagger/openapi so you
+end up with some work arounds.)
+
+Ideally the we could just write our own codegen from the models. However, this has a couple of issues
```
SammyIsConfused, line 63:
```suggestion
Ideally then, we could just write our own codegen from the models. However, this has a couple of issues
```

oeq-ts-rest-api/README.md
```diff
@@ -0,0 +1,72 @@
...
+on it's own:
```
SammyIsConfused, line 64:
```suggestion
on its own:
```

oeq-ts-rest-api/README.md
```diff
@@ -0,0 +1,72 @@
+# openEQUELLA Typescript REST API SDK
```
SammyIsConfused, line 1: Not a suggestion, just a comment this README is really great. Gives lots of insight into not just how this works, but also how and why you got there.

oeq-ts-rest-api/test/auth.test.ts
```diff
@@ -0,0 +1,33 @@
...
+test('An attempt to login with bad credentials fails', () => {
+  expect.assertions(1);
+  return OEQ.Auth.login(TC.API_PATH, 'fakeusername', 'fakepassword').catch(
+    (error: OEQ.Errors.ApiError) => {
+      expect(error.status).toBe(401);
+    }
+  );
+});
+
+test("That having login, we're able to properly log out.", () =>
```
SammyIsConfused, line 23:
```suggestion
test("That having logged in, we're able to properly log out.", () =>
```

oeq-ts-rest-api/test/settings.test.ts
```diff
@@ -0,0 +1,40 @@
+import * as OEQ from '../src';
+import * as TC from './TestConfig';
+import { GeneralSetting, UISettings } from '../src/Settings';
+
+beforeAll(() => OEQ.Auth.login(TC.API_PATH, TC.USERNAME, TC.PASSWORD));
+
+afterAll(() => OEQ.Auth.logout(TC.API_PATH, true));
+
+test('Able to retrieve general settings', () =>
```
SammyIsConfused, line 9:
```suggestion
test('That we're able to retrieve general settings', () =>
```
Just for consistency with above.

oeq-ts-rest-api/jest.config.js
```diff
@@ -0,0 +1,10 @@
+// For a detailed explanation regarding each configuration property, visit:
+// https://jestjs.io/docs/en/configuration.html
+
+// eslint-disable-next-line no-undef
+module.exports = {
+  clearMocks: true,
+  coverageDirectory: "coverage",
+  preset: "ts-jest",
+  testEnvironment: "node",
```
SammyIsConfused, line 9: I know we were using jsdom before now - is there any reason for the change?

oeq-ts-rest-api/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "@openequella/rest-api-client",
+  "version": "2020.2.0-Alpha",
```
SammyIsConfused, line 3: Just curious - in future releases, part of the process of progressing the version will be to bump this number up to match the same on the backend? Also, is this documented anywhere?

oeq-ts-rest-api/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "src"
+  ],
+  "engines": {
+    "node": ">=10"
+  },
+  "scripts": {
+    "start": "tsdx watch",
+    "build": "tsdx build",
+    "test": "jest",
+    "lint": "eslint \"src/**/*.{ts,tsx}\" \"test/**/*.{ts,tsx}\"",
```
SammyIsConfused, line 18: is there some sort of `eslint --fix` command we could add?

oeq-ts-rest-api/src/LegacyContent.ts
edalex-ian: Well that ended up being an adventure, but I now have `typescript-is` plugged in - and indeed it picked up one error in my model. But to get there took migrating from TSDX to direct rollup and things definitely got interesting.

Although I quite like the look of the above I wasn't a fan of how with `io-ts` I couldn't just use a simple `interface`. It seems instead you define a `Type` and then derive your interfaces from that (and possibly use aliases) - based on https://lorefnon.tech/2018/03/25/typescript-and-validations-at-runtime-boundaries/ and https://github.com/gcanti/io-ts/blob/master/Type.md

That said, doing so I can see you get far more informative error messages - and when I had to debug my issue it was rather manual. But hopefully it's the exception rather than the norm, and just a bit of a guard.

So I've pushed for one more review, and based on current approvals I'll then move to merging tomorrow.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/SettingsPage.tsx
edalex-ian: Yup, that's just been missed. Will address.

oeq-ts-rest-api/jest.config.js
edalex-ian: Yup! Trying to get axios and cookies working under Jest was a real big pain. This fixes half the issues, and seeing this is a library for the back-end we have no need for a surrogate DOM to get in the way.

oeq-ts-rest-api/package.json
edalex-ian: That's the current tentative idea. This will come more into play when we split out into it's own repo. Consider this kind of a place holder for now.

oeq-ts-rest-api/package.json
edalex-ian: I've added it, but there is actually a lot of work still to do in this area - e.g. bring in the eslint prettier plugin, and also match configurations to core oEQ. That's a future.

edalex-ian:
Thank you everyone for your review and updates - most appreciated.

All feedback resolved, and have added runtime type checking via [typescript-is](https://github.com/woutervh-/typescript-is) which also resulted in the removal of the last bits of TSDX and full setup of Rollup.

Re-requesting review (of those updates).

edalex-ian:
Gah! It's broken the existing Jest tests now. This is some kind of JS build hell. I'll have to look at it tomorrow.

oeq-ts-rest-api/rollup.config.js
```diff
@@ -0,0 +1,20 @@
...
+import ttypescript from 'ttypescript';
+import typescript from '@rollup/plugin-typescript';
+
+export default {
+  input: 'src/index.ts',
+  output: {
+    dir: 'dist',
+    format: 'es',
+    sourcemap: true,
+  },
```
ChristianMurphy, line 13: ESM build is great! :+1: 
A lot of build tools look for a `main` entry in *package.json* which usually CJS format.
It might be worth having both formats?

https://babeljs.io/blog/2018/06/26/on-consuming-and-publishing-es2015+-packages talks a bit more about the formats.

oeq-ts-rest-api/src/LegacyContent.ts
```diff
@@ -0,0 +1,43 @@
...
+ * as menu structure and task and notification counts.
+ *
+ * @param apiBasePath Base URI to the oEQ institution and API
+ */
+export const getCurrentUserDetails = (
+  apiBasePath: string
+): Promise<CurrentUserDetails> =>
+  GET<CurrentUserDetails>(apiBasePath + '/content/currentuser', (data) =>
+    is<CurrentUserDetails>(data)
+  );
```
ChristianMurphy, line 43: newline at end of file
```suggestion
  );

```

oeq-ts-rest-api/src/LegacyContent.ts
```diff
@@ -0,0 +1,43 @@
...
+ * Retrieve details of the current user (based on JSESSIONID) including details for the UI such
+ * as menu structure and task and notification counts.
+ *
+ * @param apiBasePath Base URI to the oEQ institution and API
+ */
+export const getCurrentUserDetails = (
+  apiBasePath: string
+): Promise<CurrentUserDetails> =>
+  GET<CurrentUserDetails>(apiBasePath + '/content/currentuser', (data) =>
+    is<CurrentUserDetails>(data)
```
ChristianMurphy, line 42: Pretty slick that it can directly use the typing, nice! :+1:

oeq-ts-rest-api/src/index.ts
```diff
@@ -0,0 +1,7 @@
+// Needed to stop error of: "regeneratorRuntime is not defined"
+import "regenerator-runtime/runtime.js";
```
ChristianMurphy, line 2: :thinking: do we want to embed the runtime at the library level?
This would get included in the bundle right?
If there is ever another library added, which also needs a regenerator runtime, would that mean each bundle has it's own copy of the runtime?
Would that be the expected outcome?

oeq-ts-rest-api/rollup.config.js
edalex-ian: Ah!! I wonder if that is my current issue with Jest. I'll add commonJS output in and try.

oeq-ts-rest-api/src/index.ts
edalex-ian: I'm completely out of my depth with this bit. I'm definitely open to suggestions and help.

With vanilla TSDX builds this was not needed. But when I moved to doing Rollup myself suddenly oEQ's UI would throw the above mentioned error about regeneratorRuntime when attempting to use this module.

I spent hours trying to adding in babel to Rollup with the runtime plugins etc but nothing worked. Then I just tried hacking in the above import and oEQ started working again. And to be honest, I'm not even sure what dependency is making this available to the module's build.

If you can offer an insight to what TSDX might have been doing that meant this was never an issue I'd love to replicate it and not have this stray import.

oeq-ts-rest-api/.eslintrc.js
```diff
@@ -0,0 +1,23 @@
...
+    es6: true,
+  },
+  extends: [
+    'eslint:recommended',
+    'plugin:@typescript-eslint/eslint-recommended',
+    'plugin:@typescript-eslint/recommended',
+  ],
+  globals: {
+    Atomics: 'readonly',
+    SharedArrayBuffer: 'readonly',
```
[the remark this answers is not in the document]
edalex-ian: These lines just came from `eslint --init`. Should I remove?

(Ultimately I'm going to match the eslint/prettier setup with that done in core oEQ. Although may keep the config external to package.json.)

oeq-ts-rest-api/src/index.ts
ChristianMurphy: I'm not :100: sure,
I think the magic sauce is
https://github.com/jaredpalmer/tsdx/blob/17ffcd215f78a4e9d6936644cbeab332f6439088/src/createRollupConfig.ts#L62-L66

Which tells rollup not to include libraries which are dependencies in the bundle, but to instead to use `import` or `required` to resolve them.

oeq-ts-rest-api/rollup.config.js
edalex-ian: Aha! That was it. I've reworked the build to produce both, and in turn have now made Jest much happier. :+1:

oeq-ts-rest-api/src/LegacyContent.ts
edalex-ian: hmmm, I really need to get the linting and formatting fixed up on this module soon. Thanks.

oeq-ts-rest-api/src/LegacyContent.ts
edalex-ian: Yeah it's pretty tidy. :grin:

oeq-ts-rest-api/src/index.ts
edalex-ian: Okay, I'll have a little more of a play. As now this is causing issues on Codebuild. (Strange, seems there's a JS environment difference between CodeBuild and GHA.

edalex-ian:
Oh hello. The Jest tests are now working again on GHA, however failing on CodeBuild:

image

Looks like I'm back to trying to nut out that regenerator-runtime malarky.

oeq-ts-rest-api/src/index.ts
edalex-ian: Seems it was a side effect of not building a CommonJS format module and having main point to it. With that issue resolved, I could remove the line with no effect.

oeq-ts-rest-api/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "semi": true,
+    "singleQuote": true,
+    "trailingComma": "es5"
+  },
+  "dependencies": {
+    "axios": "0.19.2",
+    "axios-cookiejar-support": "1.0.0",
+    "tough-cookie": "4.0.0",
+    "typescript-is": "0.16.0"
+  },
```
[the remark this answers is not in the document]
edalex-ian: Okay, I'll have a read. Might have to save for a future todo for now though.

edalex-ian:
Okay, pending a green build (from GHA and Codebuild), this should now be good to go.

Hopefully I can then merge, and start on adding support for the Schema endpoint.

edalex-ian:
Alright, failed Travis but people are happy and the tests it failed on Travis are flaky and also are run on GHA and CodeBuild.

Merging.

oeq-ts-rest-api/.eslintrc.js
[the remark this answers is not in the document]
ChristianMurphy: I think we can keep them for now, I am a bit surprised `eslint` included them, but if they're default they shouldn't hurt.

PR #1729, opened 2020-05-13
build: removes esModuleInterop

ChristianMurphy:
##### Checklist

##### Description of change

esModuleInterop has cascading effects, using it here means any libraries that depend on it must also use esModuleInterop.
In this case it was a small enough change that it can be removed, and downstream adopter(s) can choose to use interop or not.

~externalizing the node_modules means that downstream bundlers will need to resolve the `import` or `require` statements for packages which are dependencies, like Axios.
This allows downstream builds to automatically deduplicate dependencies which may be used by multiple packages.~
_edit_: it looks like because the commonjs resolver isn't included, rollup can't find any of the node modules anyway and it's behavior is to keep the import and soft error, meaning external isn't needed.

### DIFF
--- a/oeq-ts-rest-api/package.json
+++ b/oeq-ts-rest-api/package.json
@@ -18,7 +18,7 @@
     "test": "jest",
     "lint": "eslint \"src/**/*.{ts,tsx}\" \"test/**/*.{ts,tsx}\"",
     "format": "eslint --fix \"src/**/*.{ts,tsx}\" \"test/**/*.{ts,tsx}\"",
-    "prepare": "rollup --config"
+    "prepare": "npm run build"
   },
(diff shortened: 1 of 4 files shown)


ChristianMurphy:
> I did originally have the CommonJS resolver in there (and pointing on node_modules), but in the end it seemed it made no difference for the current usage (i.e. it worked with or without it) so i dropped it for simplicity. So I guess if needs be maybe later we may end up adding it back in

Maybe,
I look at the commonjs resolver as well.
The only differences I saw was less warning in the logs, and it was able to tell ahead of time that it didn't need an esinterop polyfill.
I came to the same conclusion that right now it may not be worth the investment.

PR #1734, opened 2020-05-14
Add support to the REST Client for retrieval of schemas

edalex-ian:
##### Checklist

- [x] tests are included
- [ ] documentation is changed or added - NA

##### Description of change

This adds support to the REST Client to be able to simply:

* List available schemas (`listSchemas`); and
* Get a specific schema (`getSchema(uuid)`).

It also includes a slight improvement to the error handling to include the standard oEQ error response within `ApiError`.

But this was painful! First, what one sees in the Swagger/apidocs for the expected models returned is mostly wrong. And the same goes if you niavely look at the method signature in the code for an endpoint.

oEQ does a whole lot of customisation for its responses, so even though the method signature might say it will return a certain type, that can not be relied on - and in the case of listing schemas is mostly wrong (it says `SchemaBean` but you with really get either a stripped down `BaseEntityBean` or the more full fledged `EquellaSchemaBean`). Further, there are even some properties which are returned in the payload (e.g. `links`) which is purely synthetic and added dynamically via an internal `Map` - via `AbstractExtendableBean`.

All that means a few things:

1. Not attempting to force the route of codegen based on an OpenAPI or Swagger spec was the right course - indeed even attempting to codegen off the method signatures would've been folly; and
2. The models you see in this PR have been excruciatingly hand rolled, and sometimes are loser in typing then we'd ideally like; and
3. Although enforcing runtime types with `typescript-is` is hard, it does seem to be paying off and very much warranted with this module - and will pay dividends when we run the tests as part of the CI.

One last note: You'll see the NPM `test` script now calls Jest with `--no-cache`. This was due to finding inconsistent results, where sometimes IDEs (VS Code / IntelliJ) would show the tests as passing or failing, but the terminal would say the opposite. I guess that the codegen being done by `typescript-is` was being cached, and hence triggered the confusing results (the failures were always around the typechecking). Although this option is said to double the execution time, for now I'm willing to take the hit for sake of accuracy - it's hard enough divining the types as it is. (It still takes < 5sec to run the tests.)

### DIFF
--- a/oeq-ts-rest-api/src/Common.ts
+++ b/oeq-ts-rest-api/src/Common.ts
@@ -1,0 +1,71 @@
+import * as Security from './Security';
+import { is } from 'typescript-is';
+
+export type i18nString = string;
+
+export type I18nStrings = Record<string, string>;
+
+export interface User {
+  id: string;
+  username?: string;
+  firstName?: string;
+  lastName?: string;
+  emailAddress?: string;
+}
+
+export interface EntityLock {
+  uuid: string;
+  owner: User;
+  links: Record<string, string>;
+}
+
+export interface BaseEntityExport {
+  exportVersion: string;
+  lock: EntityLock;
+}
+
+export interface BaseEntityReadOnly {
+  granted: string[];
+}
+
+export interface BaseEntity {
+  uuid: string;
+  // Attempted to use Date here, but it broke checks with typescript-is. But Date could be used
+  // post processing
+  modifiedDate?: string;
+  createdDate?: string;
+  owner?: User;
+  name: i18nString;
+  nameStrings: I18nStrings;
+  description?: i18nString;
+  descriptionStrings?: I18nStrings;
+  security?: Security.BaseEntitySecurity;
+  exportDetails?: BaseEntityExport;
+  readonly?: BaseEntityReadOnly;
+  links: Record<string, string>;
+}
+
+export interface PagedResult<T> {
+  start: number;
+  length: number;
+  available: number;
+  results: T[];
+  resumptionToken?: string;
+}
+
+/**
+ * Helper function for a standard validator for BaseEntity instances via typescript-is.
+ *
+ * @param instance An instance to validate.
+ */
+export const isBaseEntity = (instance: unknown): boolean =>
+  is<BaseEntity>(instance);
+
+/**
+ * Helper function for a standard validator for BaseEntity  instances wrapped in a PagedResult
+ * via typescript-is.
+ *
+ * @param instance An instance to validate.
+ */
+export const isPagedBaseEntity = (instance: unknown): boolean =>
+  is<PagedResult<BaseEntity>>(instance);
--- a/oeq-ts-rest-api/src/Common.ts
+++ b/oeq-ts-rest-api/src/Common.ts
@@ -54,14 +54,6 @@ export interface PagedResult<T> {
 }
 
 /**
- * Helper function for a standard validator for BaseEntity instances via typescript-is.
- *
- * @param instance An instance to validate.
- */
-export const isBaseEntity = (instance: unknown): boolean =>
-  is<BaseEntity>(instance);
-
-/**
  * Helper function for a standard validator for BaseEntity  instances wrapped in a PagedResult
  * via typescript-is.
  *
--- a/oeq-ts-rest-api/src/Schema.ts
+++ b/oeq-ts-rest-api/src/Schema.ts
@@ -1,0 +1,112 @@
+import * as Common from './Common';
+import { GET } from './AxiosInstance';
+import { is } from 'typescript-is';
+
+export interface Citation {
+  name: string;
+  transformation: string;
+}
+
+export interface SchemaNode {
+  children: Record<string, SchemaNode>;
+  indexed: boolean;
+  field: boolean;
+  nested: boolean;
+  type: string;
+}
+
+export interface Schema extends Common.BaseEntity {
+  namePath: string;
+  descriptionPath: string;
+  /**
+   * Typically a tree of objects representing an XML schema - so first entry is normally "xml".
+   */
+  definition: Record<string, unknown>;
+}
+
+export interface EquellaSchema extends Schema {
+  citations: Citation[];
+  exportTransformsMap: Record<string, string>;
+  importTransformsMap: Record<string, string>;
+  ownerUuid: string;
+  serializedDefinition: string;
+}
+
+/**
+ * Query params for listing of schemas. All are optional!
+ */
+export interface ListSchemaParams {
+  /**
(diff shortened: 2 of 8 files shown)


oeq-ts-rest-api/src/Common.ts
```diff
@@ -0,0 +1,63 @@
+import * as Security from './Security';
+import { is } from 'typescript-is';
+
+export type i18nString = string;
+
+export type I18nStrings = Record<string, string>;
```
edalex-ian, line 6: These names (mostly) align with those in oEQ, and I figured they added reasonable documentation insight which seems to be the main justification for using `type` these days.

oeq-ts-rest-api/src/Schema.ts
```diff
@@ -0,0 +1,110 @@
...
+ * List all available schemas which the currently authenticated user has access to. Results can
+ * be customised based on params, and if the `full` param is specified then the return value is
+ * actually EquellaSchema with all details.
+ *
+ * @param apiBasePath Base URI to the oEQ institution and API
+ * @param params Query parameters to customize (and/or page) result
+ */
+export const listSchemas = (
+  apiBasePath: string,
+  params: ListSchemaParams | null
```
PenghaiZhang, line 82: Is there any particular reasons for using `null` ? Can we do `params?: ListSchemaParams` ? If yes, then at line 93 can just put `params` there.

oeq-ts-rest-api/test/schema.test.ts
```diff
@@ -0,0 +1,62 @@
...
+      (result: EquellaSchema) => {
+        expect(result.uuid).toBe(targetUuid);
+        // Better make sure we got a schema
+        expect(result.definition).toBeTruthy();
+      }
+    );
+  });
+
+  it('Should result in a 404 when attempting to retrieve an unknown UUID', () => {
+    expect.assertions(2);
```
SammyIsConfused, line 53: I don't think I've seen these `expect.assertions(int)` calls before. Is that just to say that this block should have 2 `expect` calls, not including the `expect.assertions` call?

oeq-ts-rest-api/src/Schema.ts
```diff
@@ -0,0 +1,110 @@
+import * as Common from './Common';
```
PenghaiZhang, line 1: Just a personal preference. I might do `import {BaseEntity, PagedResult, ...} from ./Common'`.

oeq-ts-rest-api/test/schema.test.ts
```diff
@@ -0,0 +1,62 @@
...
+
+beforeAll(() => OEQ.Auth.login(TC.API_PATH, TC.USERNAME, TC.PASSWORD));
+
+afterAll(() => OEQ.Auth.logout(TC.API_PATH, true));
+
+describe('Retrieving schemas', () => {
+  it('should be possible get schemas with no params', () => {
+    expect.assertions(2);
+
+    return OEQ.Schema.listSchemas(TC.API_PATH, null).then(
```
PenghaiZhang, line 13: Do we need return here?

oeq-ts-rest-api/test/schema.test.ts
edalex-ian: Yup, that's correct.

They're important when testing async things, especially if a different path is executed and a promise simply finishes down another track - which would mean a divergence from what the test is expecting to occur. If that occurs, and you don't have `expect.assertions` you'll still get a nice green tick / pass.

oeq-ts-rest-api/test/schema.test.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  it('should be possible to get schemas customised with params', () => {
+    expect.assertions(3);
+    const howMany = 2;
+
+    return OEQ.Schema.listSchemas(TC.API_PATH, {
+      length: howMany,
+      full: true,
+    }).then((pagedResult: OEQ.Common.PagedResult<OEQ.Common.BaseEntity>) => {
+      const result = pagedResult as OEQ.Common.PagedResult<EquellaSchema>;
```
PenghaiZhang, line 29: The validator here is `isPagedEquellaSchema` which ensures that type assertion is safe, or otherwise throws a mismatch error. Is my understanding right?

oeq-ts-rest-api/src/Schema.ts
edalex-ian: I'd actually like to see us start using more of this style so that we can improve readability through namespacing. It's for this reason in this module I've gone to lengths to ensure that everything is also exported in it's own name space etc (i.e. export * as Common from './Common').

oeq-ts-rest-api/src/Schema.ts
edalex-ian: I'll check, I did have issues at one point, but this could just be left over.

oeq-ts-rest-api/test/schema.test.ts
edalex-ian: Yup, because we want the promise to be explicitly returned so that Jest handles it as an async test.

oeq-ts-rest-api/test/schema.test.ts
edalex-ian: Well, the `isPagedEquellaSchema` is known here to be used during the GET call due to the the param `full: true`. As a result we know that the result should be `PagedResult<EquellaSchema>` and so we can use coersion on it to then be able to access full types.

I did want to set the return type on `listSchemas` to something like `PagedResult<BaseEntity | EquellaSchema>` but alas that resulted in some fun type errors - and I think even tripped up `typescript-is`. It was late by then, and Jest had caused enough grief so I stopped at this compromise.

oeq-ts-rest-api/src/Schema.ts
edalex-ian: Seems to be happy, no idea what I saw yesterday. :blush: 

Change pushed.

oeq-ts-rest-api/src/Schema.ts
ChristianMurphy: I'm okay either way.

I'll just throw out there that I personally tend toward the destructured approach to accessing data as well.
Most of the time when working with functions and classes I need to access a specific interface or function, and that having to lookup/remember both the namespace and the function name to be more a hindrance than a help.

edalex-ian:
This code really only comes into play at the moment during the build - as unfortunately the tests aren't yet setup to run as part of CI. So with everyone happy, and the GHA build and check successful, I'll merge.

oeq-ts-rest-api/src/Schema.ts
edalex-ian: Yup, definitely cool on the data side, just not so much on the cluttered global namespace side.

PR #1741, opened 2020-05-16
Add schema retrieval to SchemaModule.ts

edalex-ian:
##### Checklist

- [x] tests are included

##### Description of change

This adds a method to retrieve a specific schema (based on UUID) and generate a basic tree structure. It also provides a method to build the node path for a specified node.

The generated structure currently doesn't include the extra fields for leaf nodes (_type, _indexed, _field) as we wont yet be using them. But we can easily come back and add them when/if we need them.

I've included tests with the REST calls mocked out - including for the previous function rushed through to retrieve a list of schemas. They're basic, but at least give us some coverage for now.

I believe this completes the work needed for SchemaModule, (including related schema work in the REST client). So everything should be in place for @SammyIsConfused to finish off the schema browser.

#1337 

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/schema/SchemaModule.test.ts
+++ b/Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/schema/SchemaModule.test.ts
@@ -44,3 +44,21 @@ describe("SchemaModule", () => {
       expect(copyrightNode.children?.length).toEqual(14);
     }));
 });
+
+describe("pathForNode", () => {
+  const schemaDefinition = {
+    child1: { child2: { child3: { _type: "text" } } },
+  };
+  const testSchema = SchemaModule.buildSchemaTree(schemaDefinition, "xml");
+  const testNode = testSchema.children![0].children![0].children![0];
+
+  it("should correctly generate an oEQ path by default", () =>
+    expect(SchemaModule.pathForNode(testNode)).toEqual(
+      "/child1/child2/child3"
+    ));
+
+  it("should correctly include the XML prefix if desired", () =>
+    expect(SchemaModule.pathForNode(testNode, false)).toEqual(
+      "/xml/child1/child2/child3"
+    ));
+});
--- a/Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/schema/SchemaModule.test.ts
+++ b/Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/schema/SchemaModule.test.ts
@@ -1,0 +1,46 @@
+/*
+ * Licensed to The Apereo Foundation under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * The Apereo Foundation licenses this file to you under the Apache License,
+ * Version 2.0, (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


edalex-ian:
Unrelated flakey test failure on newui (true):

```
[error] Failed tests:
[error] 	com.tle.webtests.test.cal.CALActivationsRolloverTest
```

I think it's safe to ignore.

Source/Plugins/Core/com.equella.core/js/tsrc/schema/SchemaModule.ts
```diff
@@ -20,6 +20,22 @@ import { Config } from "../config";
...
+ * If a node has no/undefined 'parent' then it is the root. If a node has no/undefined
+ * children then it is a leaf node.
+ *
+ * Technically leaf nodes typically have a type (e.g. text) however that information
+ * (along with some other bits) are not retained in this representation. However it can be
+ * retrieved via the REST API.
+ */
+export interface SchemaNode {
+  name: string;
+  parent?: SchemaNode;
```
SammyIsConfused, line 35: Oh nice, they're doubly connected. I thought it would just be name, potentially an ID, and the children array. Neat.

Source/Plugins/Core/com.equella.core/js/tsrc/schema/SchemaModule.ts
```diff
@@ -39,3 +55,71 @@ export const schemaListSummary = (): Promise<Map<string, string>> =>
...
+    }
+  );
+
+/**
+ * Helper function to generate node path string from a node.
+ *
+ * @param node The node to generate a path for.
+ * @param stripXml Whether to include the leading /xml - although returned in schemas, not typically used in paths.
+ */
+export const pathForNode = (node: SchemaNode, stripXml = true): string => {
```
SammyIsConfused, line 111: When I was trying to do something along these lines, I assumed that parent wouldn't be part of this which made it much more confusing to traverse recursively. I needed to pass through the root XML as well as the node we are getting the path for. This makes it so much easier.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/schema/SchemaModule.test.ts
```diff
@@ -0,0 +1,64 @@
...
+import { getSchemaUuidResp } from "../../../__mocks__/getSchemaUuidResp";
+import * as OEQ from "@openequella/rest-api-client";
+
+jest.mock("@openequella/rest-api-client");
+(OEQ.Schema.listSchemas as jest.Mock<
+  Promise<OEQ.Common.PagedResult<OEQ.Common.BaseEntity>>
+>).mockResolvedValue(getSchemasResp);
+(OEQ.Schema.getSchema as jest.Mock<
+  Promise<OEQ.Schema.EquellaSchema>
+>).mockResolvedValue(getSchemaUuidResp);
```
PenghaiZhang, line 29: I am happy with how to mock functions here. 
I am thinking an alternative way for mocking these two functions could be 

jest.mock("@openequella/rest-api-client", () => (
  {
     listSchemas: jest.fn().mockImplementationOnce(() : 
            Promise<OEQ.Common.PagedResult<OEQ.Common.BaseEntity>> => {
         return Promise.resolve(mockFacets)
    })
  }
));

Never mind, just a thought.

edalex-ian: There was a lot of trial and error in this, but that's not one of the patterns I tried, might try it next time to see how it compares. I note it's returning an object, so I wonder how it merges things. :thinking: 

Because in your example the object just has `listSchemas`, but due to the import and exports you need to go a bit deeper. Actually, so this looks like it has potential to suffer other issues I encountered, but still... Might have a play one day, thanks. :+1:

Issue #1744, opened 2020-05-18
Canvas integration issue with new UI


cathfitz:
**Describe the bug**
There is a problem with the SelectOrAdd part of the new U/I.

Overview:  There are 3 points of integration.  The Course Navigation, Resource Selection, Editor Button.
Course navigation:  This is the Select Resources from openEQUELLA link in a canvas course left hand navigation.  It uses the structured integration endpoint.  This works in both old and new u/i.

The other two use the SelectOrAdd integration endpoint.  These work in the old U/I, but not in the new one.  In the new U/I, if you select a resource, you get a blank screen when trying to return the resource back to Canvas.

- OpenEquella Version: [e.g. 2020.1]
- OS: [e.g. Windows 10, MacOS 10.13]
- Browser [e.g. Chrome, Firefox]

**Additional context**

edalex-ian:
Possibly related to #1725.

PR #1749, opened 2020-05-20
Make Canvas Integration work with new UI

edalex-ian:
### Checklist

##### Description of change

Solves #1744. The problem was that the LegacyContentApi did not set up the form correctly in the returnResult method, and the frontend wasn't expecting it to contain the form snippet that caused content to be submitted to Canvas.


edalex-ian:
We're really just looking for a green codebuild - as it's running in New UI mode.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/LegacyContentApi.scala
```diff
@@ -448,14 +448,15 @@ class LegacyContentApi {
         info.preventGET()
         info.getRootRenderContext.setRootResultListener(new LegacyResponseListener(info))
         LegacyContentController.execute(info)
-        redirectResponse(info)
+        val r = redirectResponse(info)
```
edalex-ian, line 451: Obviously when we do this properly we don't need `r`.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/LegacyContentApi.scala
```diff
@@ -491,12 +492,19 @@ class LegacyContentApi {
...
+              SectionUtils.renderToString(context, form)
+            } else ""
+
           val upperbody =
             SectionUtils.renderToString(context, tr.getNamedResult(context, "upperbody"))
           val scrops = renderScreenOptions(context)
           val crumbs = renderCrumbs(context, decs).map(SectionUtils.renderToString(context, _))
           Iterable(
             Some("body"                                          -> body),
+            Some("form"                                          -> formString),
```
edalex-ian, line 507: I acknowledge I'm the guilty party for this, but when we polish it we should probably do similar to the below with Option / filter / map so that we only include it if there is a form.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/LegacyContentApi.scala
```diff
@@ -491,12 +492,19 @@ class LegacyContentApi {
           val body = SectionUtils.renderToString(
             context,
             wrapBody(context, tr.getNamedResult(context, "body")))
+          val form = context.getForm
+          val formString: String =
+            if (form.getAction != null) {
+              SectionUtils.renderToString(context, form)
+            } else ""
+
```
edalex-ian, line 500: Hopefully we can use a more idiomatic scala construct for this to. Maybe a for / yield, or an Option which will make it more straight forward below. (Again, I acknowledge I'm critiquing my own code.)

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/index.tsx
```diff
@@ -218,11 +218,20 @@ function IndexPage() {
...
-          );
+          const render = () => {
+            if (!content || content.noForm) {
+              return template;
+            } else {
+              const { form } = content.html;
+              return (
+                <>
+                  <LegacyForm state={content.state}>{template}</LegacyForm>
+                  {form && <div dangerouslySetInnerHTML={{ __html: form }} />}
```
edalex-ian, line 229: I wonder if it matters that this is after `LegacyForm` where as in the legacy UI this 'form' seemed to get rendered before the 'body'. :thinking:

PenghaiZhang: Do you mean the div that has an ID called `body`?  If yes, the `body` is part of this form in Old UI. 
image

edalex-ian: Oh, I see. I thought it was:

- `form`
- `div id=body`
- script

I just reformatted the code, and now see it's actually:

- `form`
  - `div id=body`
- `script`

Cool, so as long as we maintain that structure - and with the input fields etc at the top within the `form`, then that should be okay. Either-way, the test suite is passing, so....

SammyIsConfused:
@edalex-ian All feedback has been addressed.

edalex-ian:
Heya @SammyIsConfused , 

Before we all now do the proper review we'll want it point to the hotfix branch or the like.

edalex-ian:
Looks okay to me, but seeing I raised it I can't explicitly approve.

I do wonder if we can improve the Option/Some stuff - seems rather verbose - but that's okay. :)

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/index.tsx
```diff
@@ -218,11 +217,20 @@ function IndexPage() {
...
-          );
+          const render = () => {
+            if (!content || content.noForm) {
+              return template;
+            } else {
+              const { form } = content.html;
+              return (
+                <>
+                  <LegacyForm state={content.state}>{template}</LegacyForm>
+                  {form && HtmlParser(form)}
```
ChristianMurphy, line 228: Is cross site scripting a concern here?
This library does not sanitize inputs.
https://github.com/remarkablemark/html-react-parser#is-this-library-xss-safe

edalex-ian: I don't believe so, as this is just a form which is normally server side rendered and we're attempting to bring it across to be rendered as part of the New UI. But it's also deep in the depths of Sections and Legacy Content wrapping so I can't say I'm 100% across where else (and how) these forms are used.

In this case - that we're attempting to fix - it's a form with is pragmatically built up server side with some hidden fields (with LTI values) which are then submitted via JS so as to complete the Canvas submission. It feels like this is the primary way this 'form' element is used, and hence why to date nothing else has broken with it being absent from the legacy content wrapping. I think.... :crossed_fingers:

edalex-ian:
Oh @SammyIsConfused , don't forget we also need to 'forward' port this onto `develop`. (The process has definitely been skewed for this rush.

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/index.tsx
PenghaiZhang: BTW, the library we use is `react-html-parser` but not `html-react-parser`.  This library hasn't been updated for three years.

edalex-ian: heh, that naming's not confusing at all. :grimacing:

PR #1721, opened 2020-05-11
Fix the mandatory fields bug in moderation

SammyIsConfused:
##### Checklist

- [x] the [contributor license agreement][] is signed
- [x] commit message follows [commit guidelines][]
- [x] tests are included
- [ ] screenshots are included showing significant UI changes
- [ ] documentation is changed or added

##### Description of change

#1678  
There were a couple other areas in which the mandatory fields check happened - during SaveAndContinue, and during EditInWizard (moderation edits). These were missed during #1708, as the steps to reproduce didn't involve moderation.

A test has been added (VisibilityScriptingBugTest) to verifiy this has been fixed. 
This also includes editing the vanilla institution, which is where the large modified file count comes from. To review just relevant files, add a file filter just allowing .java and .yml files.

[contributor license agreement]: https://www.apereo.org/node/676
[commit guidelines]: https://chris.beams.io/posts/git-commit/

PR #1759, opened 2020-05-25
Simplify VisibilityScriptingBugTest and add test retries to TestNG tests

SammyIsConfused:
##### Checklist

- [x] tests are included

##### Description of change

- Removed unnecessary files from vanilla institution (added in during #1721). Isolated to 1ee5de7
- Made test collection simpler, removing the need for the taxonomy  and schema import. Isolated to eed28ea
- Simplified flakey VisibilityScriptingBugTest (removed all the test controls but one, for the minimum test case)
- Added new annotation @MaxRetryCount(int), which if added to a test will allow that test to automatically retry. This involved including a RetryAnalyzer as well as a TestAnnotationTransformer
- Added logging on retry (logs the test name, {current retry number}/{maximum retry number}, and a stacktrace of the failure that caused the retry)
image


autotest/OldTests/src/test/java/retry/MaxRetryCount.java
```diff
@@ -0,0 +1,9 @@
+package retry;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
+public @interface MaxRetryCount {
+  int value() default 0;
```
edalex-ian, line 8: How about a default of 3?

Seems odd to have 0 for a Max.

Or should we have a DefaultRetry annotation that stacks on this and has the value as zero?

autotest/OldTests/src/test/java/com/tle/webtests/test/contribute/bugs/VisibilityScriptingBugTest.java
```diff
@@ -27,23 +26,11 @@
...
-  private static final String RADIO2 = "Yes";
-  private static final String SHUFFLE1 = "Shuffle 1";
-  private static final String SHUFFLE2 = "Shuffle 2";
-
-  private static final String XML_PATH = "item/itembody/";
-  private static final String ROLE_ID = "ROLE_SYSTEM_ADMINISTRATOR";
-  private static final String USER_ID = "4026b550-9a1f-4521-9b56-9ec737c02526";
+  private static final String XML_PATH = "item/";
 
+  @MaxRetryCount(5)
```
edalex-ian, line 33: 5 seems rather high. 3?

autotest/OldTests/src/test/java/retry/FailureRetryAnalyzer.java
```diff
@@ -0,0 +1,38 @@
...
+  int currentRetry = 0;
+  Logger LOGGER = Logger.getLogger(FailureRetryAnalyzer.class.getName());
+
+  @Override
+  public boolean retry(ITestResult result) {
+    MaxRetryCount failureRetryCount =
+        result.getMethod().getConstructorOrMethod().getMethod().getAnnotation(MaxRetryCount.class);
+    int maxRetryCount = (failureRetryCount == null) ? 0 : failureRetryCount.value();
+    if (++currentRetry > maxRetryCount) {
+      currentRetry = 0;
```
edalex-ian, line 18: Is this required? If so, does that mean this is instantiated as a singleton? If so, does that mean we have concurrency issues with this?

autotest/OldTests/src/test/java/retry/FailureRetryAnalyzer.java
```diff
@@ -0,0 +1,38 @@
...
+    } else {
+      logRetryInfo(result, maxRetryCount);
+      return true;
+    }
+  }
+
+  private void logRetryInfo(ITestResult result, int maxRetryCount) {
+    LOGGER.warning("Test with retry analyser failed.\n\n");
+    // print failure stack trace
+    LOGGER.log(Level.WARNING, result.getThrowable().getMessage(), result.getThrowable());
```
edalex-ian, line 29:
```suggestion
    LOGGER.warning(result.getThrowable().getMessage(), result.getThrowable());
```

or can you

```suggestion
    LOGGER.warning(result.getThrowable());
```

Actually, do we even need to log this? Seems excessively verbose.

autotest/OldTests/src/test/java/retry/FailureRetryAnalyzer.java
```diff
@@ -0,0 +1,38 @@
...
+      currentRetry = 0;
+      return false;
+    } else {
+      logRetryInfo(result, maxRetryCount);
+      return true;
+    }
+  }
+
+  private void logRetryInfo(ITestResult result, int maxRetryCount) {
+    LOGGER.warning("Test with retry analyser failed.\n\n");
```
edalex-ian, line 27: This message is kind of redundant isn't it?

autotest/OldTests/src/test/java/retry/FailureRetryAnalyzer.java
```diff
@@ -0,0 +1,38 @@
...
+  @Override
+  public boolean retry(ITestResult result) {
+    MaxRetryCount failureRetryCount =
+        result.getMethod().getConstructorOrMethod().getMethod().getAnnotation(MaxRetryCount.class);
+    int maxRetryCount = (failureRetryCount == null) ? 0 : failureRetryCount.value();
+    if (++currentRetry > maxRetryCount) {
+      currentRetry = 0;
+      return false;
+    } else {
+      logRetryInfo(result, maxRetryCount);
```
edalex-ian, line 21: Does this get called before the first test run, of before the first retry?

(As we don't want any log message for a single successful run.)

autotest/OldTests/src/test/java/retry/TestAnnotationTransformer.java
```diff
@@ -0,0 +1,18 @@
...
+import org.testng.annotations.ITestAnnotation;
+
+public class TestAnnotationTransformer implements IAnnotationTransformer {
+
+  @Override
+  public void transform(
+      ITestAnnotation annotation, Class testClass, Constructor testConstructor, Method testMethod) {
+    if (testMethod == null) return;
+    MaxRetryCount maxRetryCount = testMethod.getAnnotation(MaxRetryCount.class);
+    if (maxRetryCount == null) return;
```
edalex-ian, line 15:
```suggestion
    if (maxRetryCount == null && maxRetryCount.value() < 1) return;
```

autotest/OldTests/src/test/java/retry/FailureRetryAnalyzer.java
```diff
@@ -0,0 +1,38 @@
+package retry;
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
```
edalex-ian, line 4: Do we have the log4j framework overhere? (Or Slf4j?)

edalex-ian:
Bugger, CodeBuild hit a different flaky test - com.tle.webtests.test.drm.DRMPrivilegeTest.

autotest/OldTests/src/test/java/retry/FailureRetryAnalyzer.java
SammyIsConfused: this gets called after a run that has @MaxRetryCount annotation fails, before it retries.

autotest/OldTests/src/test/java/retry/FailureRetryAnalyzer.java
SammyIsConfused: This was just for logging purposes - without this it would go straight to the stack trace, and that could potentially look too similar to a standard failed test.

autotest/OldTests/src/test/java/retry/FailureRetryAnalyzer.java
edalex-ian: Cool. :+1:

autotest/OldTests/src/test/java/retry/FailureRetryAnalyzer.java
edalex-ian: Gotcha. Well, let's remove this and the stack trace.

autotest/OldTests/src/test/java/retry/FailureRetryAnalyzer.java
SammyIsConfused: Your suggestion will not work. 
LOGGER.warning() helper function doesn't have a stack trace method like LOGGER.error does. In order to get a stack trace, and a warning label, you must LOGGER.log(level, message, throwable)

SammyIsConfused: The purpose here was so that when a TestNG method fails and retries, the initial failure will still log a stack trace.

autotest/OldTests/src/test/java/retry/FailureRetryAnalyzer.java
SammyIsConfused: I don't think it's required. Each individual test that requires one instantiates a FailureRetryAnalyzer of its own.

SammyIsConfused:
> Bugger, CodeBuild hit a different flaky test - com.tle.webtests.test.drm.DRMPrivilegeTest.

Oh! Actually it didn't flake - I may have gotten too delete happy and deleted an item we actually needed. My bad, I'll fix that up.

autotest/OldTests/src/test/java/retry/FailureRetryAnalyzer.java
edalex-ian: > Your suggestion will not work.

Yup, as it appears we're not using log4j, just vanilla JRE logging. We should be using log4j

> The purpose here was so that when a TestNG method fails and retries, the initial failure will still log a stack trace.

But what does that offer when we already have the report that test t from class c is retrying? We don't want our logs for testing even more verbose unless it serves a purpose.

edalex-ian:
> > Bugger, CodeBuild hit a different flaky test - com.tle.webtests.test.drm.DRMPrivilegeTest.
> 
> Oh! Actually it didn't flake - I may have gotten too delete happy and deleted an item we actually needed. My bad, I'll fix that up.

Ah yes, I forgot that this is not just about the retry. (I'll go and remove that from my flaky test list - was wondering why it was new.)

Thanks.

autotest/OldTests/src/test/java/retry/FailureRetryAnalyzer.java
SammyIsConfused: The idea I had for that is that if some test is flakey, we can look into why it was breaking without letting it fail the whole run. But if you don't think that's worthwhile I'm happy to remove it.

edalex-ian: Okay, so as discussed just now, this is to capture initial details of a test that eventually passes but we may want to go back at a later date an look into the flakiness.

Seeing this would really be the exception to the norm, let's log the stacktrace at 'debug' once the logger is changed over to log4j.

autotest/OldTests/src/test/java/retry/FailureRetryAnalyzer.java
edalex-ian: Phew, in that case might as well drop it.

autotest/OldTests/src/test/java/com/tle/webtests/test/contribute/bugs/VisibilityScriptingBugTest.java
SammyIsConfused: I'll rename this to RetryTest with a default of three. 
@RetryTest would just give the default.
@RetryTest(5) would override that.

autotest/OldTests/src/test/java/retry/MaxRetryCount.java
SammyIsConfused: I'll rename this to RetryTest with a default of three. 
@RetryTest would just give the default.
@RetryTest(5) would override that.

autotest/OldTests/src/test/java/retry/FailureRetryAnalyzer.java
SammyIsConfused: Changed over to slf4j.

SammyIsConfused:
All feedback should be addressed now.

SammyIsConfused:
The Travis build failure seems strange - all the tests are unrelated and are part of new additions in develop that are not included in this branch. After doing a local merge and running those tests they work fine for me.

Issue #1306, opened 2019-10-16
Create new Search UI using Material Design methodologies


cathfitz:
Create the new Search UI by applying Material Design methodologies to all Search components

PR #1765, opened 2020-05-27
Added basic datemodified component, and storybook test

mrblippy:
##### Checklist

- [x] tests are included

##### Description of change

A pretty basic component to display dates inside the oEQ new ui.  I'm currently not sure if the date parsing is as exhaustive as it could be. So feedback in that area in particular would be great

It includes a storybook test, is that enough testing-wise?
storybook


Source/Plugins/Core/com.equella.core/js/tsrc/components/DateModifiedDisplay.tsx
```diff
@@ -0,0 +1,52 @@
...
+  displayRelative: boolean;
+  date: Date;
+}
+export default function DateModifiedDisplay({
+  displayRelative,
+  date,
+}: DateModifiedDisplayProps) {
+  const classes = useStyles();
+  let primaryDate = moment(date).format("lll");
+  let hoverDate = moment(date).fromNow();
```
edalex-ian, line 38: I'm struggling to see if we need to bring in >50KB library just to do this - when you could write it in ~100 lines. Maybe it's okay, but.... primary date we could just go with `date.toLocaleString()`, it's only timeago support we'd need. I wonder what others think?

Source/Plugins/Core/com.equella.core/js/tsrc/components/DateModifiedDisplay.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+const useStyles = makeStyles({
+  dateModified: {
+    display: "inline-block",
+  },
+});
+export interface DateModifiedDisplayProps {
+  displayRelative: boolean;
+  date: Date;
+}
```
edalex-ian, line 31: JSDoc please.

Source/Plugins/Core/com.equella.core/js/tsrc/components/DateModifiedDisplay.tsx
edalex-ian: Okay, digging further I see we already have a dependency of `timeago-react` which under the hood apparently uses the [`timeago.js`](https://github.com/hustcc/timeago.js) library - which as a result should be in our dependencies (and is only ~3KB). So how about we use that and simple `date.toLocaleString()`?

Source/Plugins/Core/com.equella.core/js/tsrc/components/DateModifiedDisplay.tsx
```diff
@@ -0,0 +1,52 @@
...
+export default function DateModifiedDisplay({
```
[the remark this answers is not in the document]
mrblippy: oh yeah, that would probably be better

Source/Plugins/Core/com.equella.core/js/tsrc/components/DateModifiedDisplay.tsx
mrblippy: I didn't pay any attention to how big it was. But yeah, it's not really necessary, especially since we already have timeago-react

Source/Plugins/Core/com.equella.core/js/tsrc/components/DateModifiedDisplay.tsx
mrblippy: Oh yes. you raise a valid point

Source/Plugins/Core/com.equella.core/js/package.json
```diff
@@ -43,7 +44,7 @@
     "lodash": "4.17.15",
     "luxon": "1.16.0",
     "material-ui-pickers": "2.2.4",
-    "@openequella/rest-api-client": "file:../../../../../oeq-ts-rest-api",
+    "moment": "^2.26.0",
```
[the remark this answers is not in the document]
SammyIsConfused: Isn't luxon currently used for this purpose?

Source/Plugins/Core/com.equella.core/js/tsrc/components/DateModifiedDisplay.tsx
edalex-ian: Ah there you go, and @SammyIsConfused points out we also have Luxon - which can do both. Cheat sheet at https://moment.github.io/luxon/docs/manual/moment.html

Source/Plugins/Core/com.equella.core/js/tsrc/components/DateModifiedDisplay.tsx
```diff
@@ -0,0 +1,52 @@
...
+  displayRelative,
+  date,
+}: DateModifiedDisplayProps) {
+  const classes = useStyles();
+  let primaryDate = moment(date).format("lll");
+  let hoverDate = moment(date).fromNow();
+
+  if (displayRelative) {
+    //swap 'em around
+    [primaryDate, hoverDate] = [hoverDate, primaryDate];
```
PenghaiZhang, line 42: I was going to suggest you put this check inside a useEffect that takes an empty array as its dependency, to avoid running this check in each rendering. But then I realised this component does not even have any stateful value. So it's all good here.

PR #1784, opened 2020-06-04
Add support to the REST Client API to call the ‘normal' Search REST API

PenghaiZhang:
#1306 

##### Checklist

- [x] tests are included

##### Description of change

In `oeq-ts-rest-api`, add support to interact with the new search endpoint `search2`, including:

1. adding TS interfaces to help ensure the returned data structure is correct; and
2. adding functions to call the endpoint and validate data; and
3. adding a JEST test.


oeq-ts-rest-api/src/AxiosInstance.ts
```diff
@@ -19,7 +20,7 @@ export const GET = <T>(
   queryParams?: object
 ): Promise<T> =>
   axios
-    .get<T>(path, {params: queryParams})
+    .get<T>(path, {params: queryParams, paramsSerializer: params => stringify(params)}, )
```
PenghaiZhang, line 23: The reason to add this is because the default query string format for array is not accepted on oEQ Server. 
For example, `/search2/?status[]=xxx&status[]=yyy`  is not OK but `/search2/?status=xxx&status=yyy` is fine.
Anybody have better ideas?

oeq-ts-rest-api/src/Search.ts
```diff
@@ -0,0 +1,70 @@
+import * as Common from './Common';
+import { is } from 'typescript-is';
+import * as SearchResult from './SearchResult';
+import {GET} from './AxiosInstance';
+
+export interface SearchParams {
```
edalex-ian, line 6: Please add JSDoc at the interface level too.

oeq-ts-rest-api/src/Search.ts
```diff
@@ -0,0 +1,70 @@
...
+   */
+  owner?: string;
+  /**
+   * single dynamic collection uuid.
+   */
+  dynaCollection?: string;
+}
+
+export const isPagedSearchResultItem = (instance: unknown): boolean =>
+  is<Common.PagedResult<SearchResult.Item>>(instance);
```
edalex-ian, line 62: No need for this to be in its own stand-alone function. You'll see the others have just defined it where used. I see you've also exported it, but AFAICT it's not needed anywhere else.

oeq-ts-rest-api/test/search.test.ts
```diff
@@ -0,0 +1,56 @@
...
+import * as SearchResult from '../src/SearchResult';
+import * as Common from '../src/Common';
+import * as Auth from '../src/Auth';
+import * as TC from './TestConfig';
+import * as Errors from "../src/Errors";
+
+beforeAll(() => Auth.login(TC.API_PATH, TC.USERNAME, TC.PASSWORD));
+afterAll(() => Auth.logout(TC.API_PATH, true));
+
+describe("Search for items", () => {
```
PenghaiZhang, line 11: I feel this Jest test is more focused on validating the returned data type and handling errors rather than checking if the search result is correct or not. 
We have another test `Search2ApiTest` which check the search results.

oeq-ts-rest-api/src/SearchResult.ts
```diff
@@ -0,0 +1,50 @@
+import * as Common from './Common';
```
edalex-ian, line 1: File level comments:

* All the interfaces here should be in either `Common` and over in `Search`, there's no need for this file; and
* All of the interfaces in this file are missing JSDoc.

oeq-ts-rest-api/test/search.test.ts
```diff
@@ -0,0 +1,56 @@
...
+  const SEARCH_API_PATH = `${TC.API_PATH}/search2`;
+
+  it("should be able to search without params", () => {
+    expect.assertions(4);
```
edalex-ian, line 15: We believe (thanks to the recent updates from @ChristianMurphy ), that we don't need this if you define the test as async and use await for the API call - see the other tests.

oeq-ts-rest-api/src/Search.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export const search = (searchPath: string, params?: SearchParams): Promise<Common.PagedResult<SearchResult.Item>> => {
```
edalex-ian, line 64: Missing JSDoc.

Keep in mind the pattern is that `searchPath` is just meant to be `apiBasePath`. It then uses that to build the full path so that the caller does not need to worry about the routes etc. (See others as examples.)

oeq-ts-rest-api/test/search.test.ts
```diff
@@ -0,0 +1,56 @@
...
+import * as Common from '../src/Common';
+import * as Auth from '../src/Auth';
+import * as TC from './TestConfig';
+import * as Errors from "../src/Errors";
+
+beforeAll(() => Auth.login(TC.API_PATH, TC.USERNAME, TC.PASSWORD));
+afterAll(() => Auth.logout(TC.API_PATH, true));
+
+describe("Search for items", () => {
+  const SEARCH_API_PATH = `${TC.API_PATH}/search2`;
```
edalex-ian, line 12: As mentioned in `Search.ts`, this path should be internalised to `Search.ts`.

oeq-ts-rest-api/test/search.test.ts
```diff
@@ -0,0 +1,56 @@
...
+        expect(pagedResult).toHaveLength(pagedResult.results.length);
+        expect(uuid).toBeTruthy();
+        // Status returned is in lowercase so have to convert to uppercase.
+        expect(status.toUpperCase()).toBe(SearchResult.ItemStatus.LIVE);
+        expect(collectionId).toBe(collection);
+      }
+    )
+  });
+
+  it("should get 404 response if search for an non-existing collection", () => {
```
edalex-ian, line 47:
```suggestion
  it("should get 404 response if search for a non-existing collection", () => {
```

oeq-ts-rest-api/src/AxiosInstance.ts
```diff
@@ -2,6 +2,7 @@ import Axios, { AxiosResponse, AxiosError } from 'axios';
 import axiosCookieJarSupport from 'axios-cookiejar-support';
 import * as tough from 'tough-cookie';
 import { repackageError } from './Errors';
+import {stringify} from "querystring";
```
edalex-ian, line 5: I wonder if instead we should use https://www.npmjs.com/package/query-string

Only because it's been updated lately (2 months vs 7 years) and seems a bit more widely used.

It is also interesting that you could pull this in without adding anything to package.json. :thinking: 

(And as a real minor sidenote, interest that this is in double quotes where the others are in single quotes - I wonder if you ran Prettier over this...)

oeq-ts-rest-api/src/AxiosInstance.ts
edalex-ian: I'm fine with that. But see my comment, as we may want to use a slightly more active module.

oeq-ts-rest-api/test/search.test.ts
edalex-ian: Correct. The Jest tests here are to ensure the API comms work. The test of the actual search logic is a core oEQ testing concern. :+1:

oeq-ts-rest-api/src/AxiosInstance.ts
PenghaiZhang: No I did not. I just hit `alt`+`enter` to import this library.

oeq-ts-rest-api/src/Search.ts
```diff
@@ -0,0 +1,210 @@
...
+  DRAFT = "DRAFT",
+  LIVE = "LIVE",
+  REJECTED = "REJECTED",
+  MODERATING = "MODERATING",
+  ARCHIVED = "ARCHIVED",
+  SUSPENDED = "SUSPENDED",
+  DELETED = "DELETED",
+  REVIEW = "REVIEW",
+  PERSONAL = "PERSONAL",
+}
```
ChristianMurphy, line 73: Are these statuses fixed/static? Or are they customizable at an institution level?

edalex-ian: These are fixed/static and match an enum server side.

oeq-ts-rest-api/src/Search.ts
```diff
@@ -58,13 +60,151 @@ export interface SearchParams {
...
+/**
+ * Type of search result attachment.
+ */
+export interface SearchResultAttachment {
+  /**
+   * Attachment type.
+   */
+  attachmentType: string;
+  /**
+   * The unique ID of an attachment.
```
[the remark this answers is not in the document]
PenghaiZhang: Yes.

oeq-ts-rest-api/src/Search.ts
```diff
@@ -58,13 +60,151 @@ export interface SearchParams {
...
+   */
+  integrationOpen: boolean;
+}
+
+/**
+ * Type of search result attachment.
+ */
+export interface SearchResultAttachment {
+  /**
+   * Attachment type.
```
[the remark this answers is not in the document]
PenghaiZhang: Yes. We have quite a few `bean`s which extend `AttachmentBean` and these beans define their types.

oeq-ts-rest-api/src/Search.ts
```diff
@@ -58,13 +60,151 @@ export interface SearchParams {
...
+   */
+  status: string;
+  /**
+   * The date when item is created.
+   */
+  createdDate: string;
+  /**
+   * The last date when item is modified.
+   */
+  modifiedDate: string;
```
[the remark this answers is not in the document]
PenghaiZhang: I am a bit confused about this. Let's have a quick talk tomorrow.

The issue is using `Date` as the field's type will break type checks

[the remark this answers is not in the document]
edalex-ian: Yup, looks like we need support in the `GET` to take another function (say, `transformer`) which can modify the `response.data` as required, and then maybe a standard transformer called `convertDateFields` looking something like:

```typescript
const convertDateFields = (fields: string[], data: unknown) => T
```

That could sit in a new `Utils.ts`

oeq-ts-rest-api/src/AxiosInstance.ts
```diff
@@ -17,17 +17,19 @@ const catchHandler = (error: AxiosError | Error): never => {
...
   validator?: (data: unknown) => boolean,
-  queryParams?: object
+  queryParams?: object,
+  transformer?: (data: T) => unknown,
 ): Promise<T> =>
   axios
     .get<T>(path, {params: queryParams, paramsSerializer: params => stringify(params)}, )
     .then((response: AxiosResponse<T>) => {
-      if (validator && !validator(response.data)) {
+      const data: any = transformer? transformer(response.data) : response.data;
```
PenghaiZhang, line 26: I am concerned about if using `any` will break type safety.

oeq-ts-rest-api/src/Utils.ts
```diff
@@ -0,0 +1,29 @@
+import * as Lodash from "lodash";
+
+/**
+ * Return a copied object of which types of specified fields are converted to Date.
+ * @param fields Names of fields that need type conversion.
+ * @param data The object to be processed. Nested objects will be processed, too.
+ * @param cloneData True if deep clone is required.
+ */
+export const convertDateFields = <T>(fields: string[], data: T, cloneData: boolean): unknown => {
```
PenghaiZhang, line 9: The returning type is `unknown` because the object is modified so it might be no long of its original type. This is my understanding.

edalex-ian: I think that's right, but I'd also argue the input too is also unknown - or more so, does not have to be known consider we don't use any of the type information here - we've mostly resorted to metaprogramming.

edalex-ian, line 9: I'd suggest the target (`data`) should be the first param.

~~Why do we need an option around deep cloning?~~

Ahhh, now I see what's going on. You've kind of got two functions in one here. I suggest you split it out to something like:

```typescript
export const convertDateFields = (input: unknown, fields: string[]): unknown => {
  final inputClone = Lodash.cloneDeep(input);
  // Note: Not sure if we need to validate the date - so the converter returning `undefined` might be fine...
  convertFields(inputClone, fields, true, (value: string) => new Date( value));
  return inputClone;
}

/**
 * Performs inplace conversion of specified fields with supplied converter.
*/
const convertFields = (input: unknown, targetFields: string[], recursive: boolean, converter: (value: T) => R?): void =>
  Object.entries( input ).forEach((field, value}) => {
    if(typeof value === "object" && recursive) {
      convertFields(value, targetFields, recursive, converter);
    }
    else {
      targetFields.each( targetField => {
        if(field === targetField) {
          input[field] = converter(value);
        }
      });
    }
  }
```
Disclaimer: I've just hacked this code together here - probably completely broken with syntax errors.

I think there's a couple of benefits here: First makes for a cleaner interface on the converDateFields function, and second splits out a potentially reusable function - while also splitting out some recursive functionality (which is always wise to avoid accidents).

oeq-ts-rest-api/src/Utils.ts
```diff
@@ -0,0 +1,29 @@
+import * as Lodash from "lodash";
+
+/**
+ * Return a copied object of which types of specified fields are converted to Date.
+ * @param fields Names of fields that need type conversion.
+ * @param data The object to be processed. Nested objects will be processed, too.
+ * @param cloneData True if deep clone is required.
```
edalex-ian, line 7:
```suggestion
 * Return a clone of the provided object with specified fields converted to type Date. A deep clone
 * will be undertaken, and so nested fields with matching names will also be converted.
 *
 * @param fields List of the names of fields to convert.
 * @param data The object to be processed.
 * @param cloneData True if deep clone is required.
```

oeq-ts-rest-api/src/Utils.ts
```diff
@@ -1,29 +1,40 @@
...
     }
-    else{
-      fields.forEach(field => {
-        // Convert when a date string can be parsed to Date.
-        if(field === key && !isNaN(Date.parse(value))){
-          (clonedData as any)[field] = new Date(value)
+    else {
+      targetFields.forEach( targetField => {
+        if(field === targetField) {
+          (input as any)[field] = isNaN(Date.parse(value))? undefined: converter(value);
```
edalex-ian, line 21: Hold on, you've still got some of the specific date conversion logic here. This function should now be general purpose, and the logic goes down in the `convertDateFields()` function.

This line would now just be:

```suggestion
          (input as any)[field] = converter(value);
```

oeq-ts-rest-api/src/Utils.ts
```diff
@@ -1,29 +1,40 @@
...
         }
-      })
+      });
```
edalex-ian, line 23: I fear this block has now become a bit unreadable and unwieldly. But we can tidy it up with `filter` I feel. e.g.

```typescript
targetFields
  .filter(targetField => targetField === field)
  .forEach(field => (input as any)[field] = converter(value));
```

oeq-ts-rest-api/src/AxiosInstance.ts
ChristianMurphy: If I'm interpreting the previous comments for this line correctly right, this still needs to be updated

```suggestion
      const data: unknown = transformer? transformer(response.data) : response.data;
```

oeq-ts-rest-api/src/Utils.ts
```diff
@@ -0,0 +1,40 @@
+import * as Lodash from "lodash";
```
ChristianMurphy, line 1: Could we avoid importing all of lodash for one function?

Instead using

```suggestion
import { cloneDeep } from "lodash";
```

with https://github.com/lodash/babel-plugin-lodash

or
```suggestion
import * as cloneDeep from "lodash/cloneDeep";
```

oeq-ts-rest-api/src/Search.ts
```diff
@@ -0,0 +1,203 @@
+import * as Common from './Common';
+import { is } from 'typescript-is';
+import {GET} from './AxiosInstance';
```
ChristianMurphy, line 3: destructuring is inconsistent

```suggestion
import { is } from 'typescript-is';
import { GET } from './AxiosInstance';
```

PenghaiZhang:
Hi @ChristianMurphy , if we do ` data: unknown` at line 34, we probably also need to do `return data as T` at line 40. Do you think this is OK ?

oeq-ts-rest-api/src/AxiosInstance.ts
```diff
@@ -17,17 +17,19 @@ const catchHandler = (error: AxiosError | Error): never => {
...
+      if (validator && !validator(data)) {
         // If a validator is provided, but it fails to validate the provided data...
         throw new Error('Data format mismatch with data received from server.');
       }
 
-      return response.data;
+      return data;
```
ChristianMurphy, line 40: @PenghaiZhang cast could be avoided by using the type predicate to narrow `unknown` automatically before returning, with:
```suggestion
      if (validator) {
        if (validator(data)) {
          return data;
        }
        throw new Error('Data format mismatch with data received from server.');
      }

      return data;
```

oeq-ts-rest-api/src/AxiosInstance.ts
```diff
@@ -13,20 +14,30 @@ const catchHandler = (error: AxiosError | Error): never => {
...
+ * Executes a HTTP GET for a given path.
+ *
+ * @param path The URL path for the target GET
+ * @param validator A function to perform runtime type checking against the result - typically with typescript-is
+ * @param queryParams The query parameters to send with the GET request
+ * @param transformer A function which returns a copy of the raw data from the GET with any required values transformed - this should NOT mutate the input data (transforms should start on a copy/clone of the input)
+ */
 export const GET = <T>(
   path: string,
   validator?: (data: unknown) => boolean,
```
ChristianMurphy, line 27: to ensure `validator` is a predicate use
```suggestion
  validator?: (data: unknown) => data is T,
```

oeq-ts-rest-api/src/AxiosInstance.ts
ChristianMurphy: this may also need a change to the type for `validator` see https://github.com/openequella/openEQUELLA/pull/1784#pullrequestreview-431975228

PenghaiZhang: This function is expected to return a `Promise<T>`. As `data` is `unknown`, `return data` at line 42 may not be OK.

oeq-ts-rest-api/src/AxiosInstance.ts
ChristianMurphy: see: https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-type-predicates for more information on predicates

edalex-ian: Hold on though. This is not just meant to be a simple validator or type predicate. This is so that we can parse in a function that under the hood will do runtime validation via typescript-is.

oeq-ts-rest-api/src/AxiosInstance.ts
ChristianMurphy: That is close to what we want, if the type isn't validated it is `unknown`.

:thinking: the return type may need to be made conditional on the presence of `validator`

oeq-ts-rest-api/src/Search.ts
edalex-ian: Yeah, really need to get this setup with linting and formatting. Just with the plan to move to own repo I want to wait until then rather than messing things up here. So, I guess it's manual for now - which I do by manually triggering Prettier in my IDE as I work.

oeq-ts-rest-api/src/AxiosInstance.ts
edalex-ian: Ahhh, okay you're talking about the `typeof` predicate. My bad... Pondering... :thinking:

oeq-ts-rest-api/src/AxiosInstance.ts
edalex-ian: Okay, we had more of a play with this but it's getting a bit confusing. Perhaps what we've got is good enough for now, and we can try and improve it later. Or if you have time @ChristianMurphy  maybe once this branch is merged in you can have a hack at it then.

oeq-ts-rest-api/src/AxiosInstance.ts
```diff
@@ -2,6 +2,7 @@ import Axios, { AxiosResponse, AxiosError } from 'axios';
 import axiosCookieJarSupport from 'axios-cookiejar-support';
 import * as tough from 'tough-cookie';
 import { repackageError } from './Errors';
+import {stringify} from 'query-string';
```
ChristianMurphy, line 5:
```suggestion
import { stringify } from 'query-string';
```

oeq-ts-rest-api/src/AxiosInstance.ts
ChristianMurphy: @edalex-ian @PenghaiZhang https://github.com/PenghaiZhang/openEQUELLA/pull/2

oeq-ts-rest-api/src/AxiosInstance.ts
ChristianMurphy: See https://github.com/PenghaiZhang/openEQUELLA/pull/2 for how this would work

oeq-ts-rest-api/src/AxiosInstance.ts
ChristianMurphy: also resolved in https://github.com/PenghaiZhang/openEQUELLA/pull/2

oeq-ts-rest-api/src/AxiosInstance.ts
ChristianMurphy: see https://github.com/PenghaiZhang/openEQUELLA/pull/2 for another approach

oeq-ts-rest-api/src/AxiosInstance.ts
edalex-ian: Okay, we've had a look but are concerned about how that supports all the different possible permutations. e.g. What If I want a path, no validator, but params; or, path, no validator, maybe params, and a transformer. It's just getting a bit crazy, and we need to merge this one in so as not to block further work.

So let's proceed to merge for now, and then maybe once in your can repoint your PR and show us how all the permutations can be support cleanly... Sorry we can't accommodate just now, but we definitely tried.

ChristianMurphy: > but are concerned about how that supports all the different possible permutations. e.g. What If I want a path, no validator, but params; or, path, no validator, maybe params, and a transformer

I'm not sure I follow this, optional params are positional.
in the current implementation
```
(
  path: string,
  validator?: (data: unknown) => boolean,
  queryParams?: object,
  transformer?: (data: T) => unknown,
)
```
The parameters must be added in order (`path`, `path and validator`, `path and validator and query params`, or `path and validator and query params and transformer`).
It isn't possible to skip `validator` and go straight to `transformer` with this syntax.

```ts
export function GET<T>(path: string): Promise<unknown>;
export function GET<T>(
  path: string,
  validator: (data: unknown) => data is T,
  queryParams?: object,
  transformer?: (data: unknown) => T
): Promise<T>;
```

covers all possible combinations.

PR #1795, opened 2020-06-05
WIP Fetch BIRT plugins from Github Packages

abidingotter:
Just a WIP.  Ignore hard coded path in package.json (although it does need to get fixed)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -124,6 +124,10 @@ jobs:
 
   build_import_export_tool:
     runs-on: ubuntu-latest
+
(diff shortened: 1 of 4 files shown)


edalex-ian:
Yup, cross-env has been updated on `develop` just today (or yesterday) by renovate-bot.

PenghaiZhang:
So what are we going to do with this PR ? Aaron is gone and this PR is still WIP.

edalex-ian:
I need to review correspondence. From memory this did not work out - too complex due to OSGi etc. But will check first.

PR #1845, opened 2020-06-17
feat(typings): strengthen type checks for axios instance

ChristianMurphy:
##### Checklist

- [x] tests are included

##### Description of change

This includes a number of improvements:
1. It ensures that the return type of data is only narrowed when a validator is passed
2. it runs prettier on the rest client
3. it requires validators to be a type predicate, to ensure they match the assertion 

### DIFF
--- a/oeq-ts-rest-api/src/AxiosInstance.ts
+++ b/oeq-ts-rest-api/src/AxiosInstance.ts
@@ -24,17 +24,21 @@ const catchHandler = (error: AxiosError | Error): never => {
  */
 export const GET = <T>(
   path: string,
-  validator?: (data: unknown) => boolean,
-  queryParams?: object,
-  transformer?: (data: unknown) => T,
+  validator: (data: unknown) => data is T,
+  queryParams?: Parameters<typeof stringify>[0],
+  transformer?: (data: unknown) => T
 ): Promise<T> =>
   axios
-    .get(path, {params: queryParams, paramsSerializer: params => stringify(params)}, )
-    .then((response: AxiosResponse) => {
-      const data: any = transformer? transformer(response.data) : response.data;
-      if (validator && !validator(data)) {
-        // If a validator is provided, but it fails to validate the provided data...
-        throw new Error('Data format mismatch with data received from server.');
+    .get(path, {
+      params: queryParams,
+      paramsSerializer: (params) => stringify(params),
+    })
+    .then(({ data: rawData }: AxiosResponse<unknown>) => {
+      const data = transformer ? transformer(rawData) : rawData;
+      if (!validator(data)) {
+        throw new TypeError(
+          `Data format mismatch with data received from server, on request to: "${path}"`
+        );
       }
 
       return data;
--- a/oeq-ts-rest-api/src/Search.ts
+++ b/oeq-ts-rest-api/src/Search.ts
@@ -193,11 +193,19 @@ const SEARCH2_API_PATH = '/search2';
  * @param apiBasePath Base URI to the oEQ institution and API.
  * @param params Query parameters as search criteria.
  */
-export const search = (apiBasePath: string, params?: SearchParams): Promise<Common.PagedResult<SearchResultItem>> => {
+export const search = (
+  apiBasePath: string,
+  params?: SearchParams
+): Promise<Common.PagedResult<SearchResultItem>> => {
   return GET<Common.PagedResult<SearchResultItem>>(
     apiBasePath + SEARCH2_API_PATH,
-    (data) => is<Common.PagedResult<SearchResultItem>>(data),
+    (data): data is Common.PagedResult<SearchResultItem> =>
+      is<Common.PagedResult<SearchResultItem>>(data),
     params,
-    (data) => Utils.convertDateFields<Common.PagedResult<SearchResultItem>>(data, ["createdDate", "modifiedDate"] )
+    (data) =>
+      Utils.convertDateFields<Common.PagedResult<SearchResultItem>>(data, [
+        'createdDate',
+        'modifiedDate',
+      ])
   );
 };
--- a/oeq-ts-rest-api/src/Auth.ts
+++ b/oeq-ts-rest-api/src/Auth.ts
@@ -27,7 +27,7 @@ export const login = (
     })
     .then((response: AxiosResponse) => {
       const cookies = response.headers['set-cookie'] as Array<string>;
-      const sessionIdCookie = cookies?.find(c => c.startsWith('JSESSIONID'));
+      const sessionIdCookie = cookies?.find((c) => c.startsWith('JSESSIONID'));
       const sessionId = sessionIdCookie?.substring(
         sessionIdCookie.indexOf('=') + 1,
         sessionIdCookie.indexOf(';')
--- a/oeq-ts-rest-api/src/Common.ts
+++ b/oeq-ts-rest-api/src/Common.ts
@@ -46,15 +46,15 @@ export interface BaseEntity {
 }
 
 export enum ItemStatus {
-  DRAFT = "DRAFT",
-  LIVE = "LIVE",
-  REJECTED = "REJECTED",
-  MODERATING = "MODERATING",
-  ARCHIVED = "ARCHIVED",
-  SUSPENDED = "SUSPENDED",
-  DELETED = "DELETED",
-  REVIEW = "REVIEW",
-  PERSONAL = "PERSONAL",
+  DRAFT = 'DRAFT',
+  LIVE = 'LIVE',
+  REJECTED = 'REJECTED',
+  MODERATING = 'MODERATING',
+  ARCHIVED = 'ARCHIVED',
+  SUSPENDED = 'SUSPENDED',
+  DELETED = 'DELETED',
+  REVIEW = 'REVIEW',
+  PERSONAL = 'PERSONAL',
 }
 
 export interface PagedResult<T> {
@@ -71,5 +71,6 @@ export interface PagedResult<T> {
  *
  * @param instance An instance to validate.
  */
-export const isPagedBaseEntity = (instance: unknown): boolean =>
-  is<PagedResult<BaseEntity>>(instance);
+export const isPagedBaseEntity = (
+  instance: unknown
+): instance is PagedResult<BaseEntity> => is<PagedResult<BaseEntity>>(instance);
--- a/oeq-ts-rest-api/src/LegacyContent.ts
+++ b/oeq-ts-rest-api/src/LegacyContent.ts
@@ -38,6 +38,7 @@ export interface CurrentUserDetails {
 export const getCurrentUserDetails = (
   apiBasePath: string
 ): Promise<CurrentUserDetails> =>
-  GET<CurrentUserDetails>(apiBasePath + '/content/currentuser', (data) =>
-    is<CurrentUserDetails>(data)
+  GET<CurrentUserDetails>(
+    apiBasePath + '/content/currentuser',
+    (data): data is CurrentUserDetails => is<CurrentUserDetails>(data)
   );
--- a/oeq-ts-rest-api/src/Schema.ts
+++ b/oeq-ts-rest-api/src/Schema.ts
@@ -55,7 +55,7 @@ export interface ListSchemaParams {
  *
  * @param instance An instance to validate.
  */
-export const isEquellaSchema = (instance: unknown): boolean =>
+export const isEquellaSchema = (instance: unknown): instance is EquellaSchema =>
   is<EquellaSchema>(instance);
 
 /**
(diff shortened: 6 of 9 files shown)


oeq-ts-rest-api/src/AxiosInstance.ts
```diff
@@ -13,27 +13,45 @@ const catchHandler = (error: AxiosError | Error): never => {
...
+ * @param path The URL path for the target GET
+ * @param validator A function to perform runtime type checking against the result - typically with typescript-is
+ * @param queryParams The query parameters to send with the GET request
+ * @param transformer A function which returns a copy of the raw data from the GET with any required values transformed - this should NOT mutate the input data (transforms should start on a copy/clone of the input)
+ */
+export function GET<T>(path: string): Promise<unknown>;
+export function GET<T>(
   path: string,
-  validator?: (data: unknown) => boolean,
+  validator: (data: unknown) => data is T,
```
PenghaiZhang, line 27: Thanks @ChristianMurphy . So if I just want to search with a transformer, my intuition is I should do 
`GET('path..', ()=>transform())` . However, the type of the second parameter is `(data: unknown) => data is T` which does match the transformer passed in. And likewise, if I don't need a queryparam, I would do`GET('path..', ()=>validate(), ()=>transform())` , but the third parameter is expected to be object.

I found some people just simply use `any` as parameters types, or use union types. Is that what we also need to do ?  

Typescript's function overloading seems quite different with other languages that I can read. So please correct me if I am very wrong.

oeq-ts-rest-api/src/AxiosInstance.ts
```diff
@@ -13,27 +13,45 @@ const catchHandler = (error: AxiosError | Error): never => {
...
-export const GET = <T>(
+/**
+ * Executes a HTTP GET for a given path.
+ *
+ * @param path The URL path for the target GET
+ * @param validator A function to perform runtime type checking against the result - typically with typescript-is
+ * @param queryParams The query parameters to send with the GET request
+ * @param transformer A function which returns a copy of the raw data from the GET with any required values transformed - this should NOT mutate the input data (transforms should start on a copy/clone of the input)
+ */
+export function GET<T>(path: string): Promise<unknown>;
```
PenghaiZhang, line 24: Do we need a generic type if we just return `unknown` ?

edalex-ian:
Thanks for this @ChristianMurphy , definitely happy to explore this.

Currently my main concern is ensuring we're not locking ourselves out, and not ending up with obtuse unreadable code. There's a lot of extra ceremony happening here, so overall I want to make sure there's benefit vs complexity.

So back to my original question:

> Okay, we've had a look but are concerned about how that supports all the different possible permutations. e.g. What If I want a path, no validator, but params; or, path, no validator, maybe params, and a transformer. 

To which you replied (https://github.com/openequella/openEQUELLA/pull/1784#discussion_r441551101) 

> The parameters must be added in order (path, path and validator, path and validator and query params, or path and validator and query params and transformer).
It isn't possible to skip validator and go straight to transformer with this syntax.

The original implementation had the validator as optional. Therefore I could do something like:

    GET( '/some/path`, null, { param1: 'value1' } ).......

That is, provide no validator. However that now doesn't seem to be supported. We played around with adding another overload which supported that - ensuring the return type was then a Promise<unknown> - but then also looked at how we got the permutation in there for what if we didn't want query params either, but did want a transformer, and the code just seemed like it was going to grow exponentially.

We're familiar with method overloading from Java (and personally also C++), however this mechanism while trying to enforce type checking is not looking great. But perhaps we're still missing something.

Could you please show us how those previously optional parameters and thereby resulting permutations can still be supported?

ChristianMurphy:
> Currently my main concern is ensuring we're not locking ourselves out, and not ending up with obtuse unreadable code. There's a lot of extra ceremony happening here, so overall I want to make sure there's benefit vs complexity

I agree, the current solution is seeking to keep the same implementation, but remove the case where validator is not implemented, but a generic is still passed, which would open the door to runtime violations.
An arguably simpler implementation would be to make `validator` a required parameter, and add a utility function folks can use when they want to skip validation.
```ts
// utility
export const skipValidation = (data: unknown): data is unknown => true;

// example usage with skipValidation
GET("https://example.com", skipValidation);
```
I'd be happy to update the PR with this if you'd be okay with the approach.

---

> The original implementation had the validator as optional. Therefore I could do something like:

a tweak to:

```ts
export function GET<T>(
  path: string,
  validator?: undefined,
  queryParams?: object
): Promise<unknown>;
export function GET<T>(
  path: string,
  validator: (data: unknown) => data is T,
  queryParams?: object
): Promise<T>;
export function GET<T>(
  path: string,
  validator?: (data: unknown) => data is T,
  queryParams?: object
): Promise<T> | Promise<unknown> 
```

allows

```ts
GET('/some/path');
GET('/some/path', undefined, { param1: 'value1' });
```

to be supported.
I'd personally be more comfortable with making `validator` required, but would be okay with this as well.

---

> but did want a transformer

`transformer` was removed in a previous PR.
Could we avoid re-adding it to the `GET` function? If there needs to be an ETL layer in the client that's fine, but it probably shouldn't live in `GET`.

oeq-ts-rest-api/src/AxiosInstance.ts
ChristianMurphy: It isn't needed, and may be updated in one of several ways based off the outcome of https://github.com/openequella/openEQUELLA/pull/1845#issuecomment-645748264

oeq-ts-rest-api/src/AxiosInstance.ts
ChristianMurphy: > `GET('path..', ()=>transform())`
> `GET('path..', ()=>validate(), ()=>transform())`

We _could_ support these with lots of overloads.
I'm not sure this is the way we _should_ approach it though.
See https://github.com/openequella/openEQUELLA/pull/1845#issuecomment-645748264 for alternatives.

> `any`

`any` tells the compiler to skip type checking, while it is an option, it kinda defeats the purpose of using a typed language :sweat_smile:

edalex-ian:
So I don't mind the idea of a `skipValidation` function. The second option (i.e. with `undefined`) is basically what @PenghaiZhang and I played with yesterday, but then the `transformer` scenarios made it seem a bit OTT and we had to press on.

You mention the `transformer` being removed. Nope, that other PR was where it was added, and you'll see that over on the `component/new-search-page branch`. And I've just realised this PR is not targetted at. Could you please target there, that's where the active development on this stuff is happening.

To touch on why the `transformer`. We wanted this module to manage (first and foremost) deserialisation of dates stored in strings to Date. It didn't make sense for every consumer of the module to then have to go and do their own date management - especially when that's more so a protocol detail. Hence, we introduced the transformer on the premise there's every chance we might want to do more in the future to possibly massage bits of the oEQ API.

But again, in many places this will be optional - so hence, it's optional.

It has currently been added in the `GET` as it's the only place we need it. But as we see how things shape up as we progress, we may find a need for it elsewhere.

ChristianMurphy:
> `component/new-search-page branch`

I feel a bit out of the loop here, why is all the active development going to a feature branch instead of `develop`? :sweat_smile: 

> wanted this module to manage (first and foremost) deserialisation of dates stored in strings to Date. It didn't make sense for every consumer of the module to then have to go and do their own date management - especially when that's more so a protocol detail. Hence, we introduced the transformer on the premise there's every chance we might want to do more in the future to possibly massage bits of the oEQ API.

Again I understand why transforms are useful.
Why does it need to happen inside the network call function? Each instance of `GET` the client has is wrapped in a getter function, would it make sense for transforms to live there?

oeq-ts-rest-api/src/Search.ts
```diff
@@ -193,11 +193,19 @@ const SEARCH2_API_PATH = '/search2';
...
-    params,
-    (data) => Utils.convertDateFields<Common.PagedResult<SearchResultItem>>(data, ["createdDate", "modifiedDate"] )
+    (data): data is Common.PagedResult<SearchResultItem> =>
+      is<Common.PagedResult<SearchResultItem>>(data),
+    params
+  ).then((data) =>
+    Utils.convertDateFields<Common.PagedResult<SearchResultItem>>(data, [
+      'createdDate',
+      'modifiedDate',
+    ])
```
ChristianMurphy, line 209: this shifts the transform to after the `GET`

edalex-ian:
> I feel a bit out of the loop here, why is all the active development going to a feature branch instead of develop?

It's a bit more than a `feature` branch, it's a `component branch`. So we have:

```
--*-------------------------------------------------- master
  \--*---------------------------------------------*- develop
       \--*----------------*--*----------------*--/   component/big-feature
          \ --feature/x --/   \-- feature/y --/
```
But just note I said:

> active development on **this stuff**

So I'm talking about our work to add the New Search UI. Overall high level activity and general community activity is still focused towards `develop` for the next release.

But for any big features we're doing - such as completely removing and adding the New Search UI, and later when we try for the whole New UI - we're doing it off to the side to ensure we don't impact `develop` if anything changes in timelines/deliverables. This ensure `develop` only containing mostly complete components/features - just not necessarily polished or having undertaken full integration testing.

A key use case, is that if we're going down the path of adding a significant feature, then only get part way through and suddenly our direction changes, then at least `develop` can still be used to do the next feature release with out having to also back-out any partially implemented features.

We've done this across a few significant features lately - cloud provider support and new search UI configuration come to mind.

> Why does it need to happen inside the network call function?

Because it's in there that we're also attempting to do the type validation. So we currently have a flow of get -> transform -> validate so that the output from GET is the requested type. There we no need to repeat this pattern in each utilisation of GET when we can encapsulate it as part of the get process - just pass in the functions to build the pipeline.

oeq-ts-rest-api/src/Search.ts
edalex-ian: Except the validator will fail as `SearchResultItem` contains the `Date` fields which come back from the get requests as simple strings.

ChristianMurphy:
> So I'm talking about our work to add the New Search UI. Overall high level activity and general community activity is still focused towards develop for the next release.

Understood, thanks for clarifying :bow: 

> > Why does it need to happen inside the network call function?
>
> Because it's in there that we're also attempting to do the type validation. So we currently have a flow of get -> transform -> validate so that the output from GET is the requested type. There we no need to repeat this pattern in each utilisation of GET when we can encapsulate it as part of the get process - just pass in the functions to build the pipeline.

I walked back the `transformer` change from this PR, since it isn't needed for the typing changes.
There may be a follow up PR to discuss `transform` further, it feels like we may be able to simplify by working with a known data type instead of an unknown, and calling a function there isn't all that different from calling a function here, I could be wrong :man_shrugging: :slightly_smiling_face:

oeq-ts-rest-api/src/Search.ts
ChristianMurphy: walked this change back, see https://github.com/openequella/openEQUELLA/pull/1845#issuecomment-646022623

oeq-ts-rest-api/src/AxiosInstance.ts
```diff
@@ -24,17 +24,21 @@ const catchHandler = (error: AxiosError | Error): never => {
...
+    .get(path, {
+      params: queryParams,
+      paramsSerializer: (params) => stringify(params),
+    })
+    .then(({ data: rawData }: AxiosResponse<unknown>) => {
+      const data = transformer ? transformer(rawData) : rawData;
+      if (!validator(data)) {
+        throw new TypeError(
+          `Data format mismatch with data received from server, on request to: "${path}"`
+        );
```
ChristianMurphy, line 41: refined the error a bit, it is specifically a `TypeError`https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError
And including the url/path in the message for easier tracing.

oeq-ts-rest-api/src/AxiosInstance.ts
```diff
@@ -24,17 +24,21 @@ const catchHandler = (error: AxiosError | Error): never => {
  */
 export const GET = <T>(
   path: string,
-  validator?: (data: unknown) => boolean,
-  queryParams?: object,
-  transformer?: (data: unknown) => T,
+  validator: (data: unknown) => data is T,
+  queryParams?: Parameters<typeof stringify>[0],
```
ChristianMurphy, line 28: Since this type is passed directly to `query-string`, updated this to extract the exact typing the library uses.

PenghaiZhang: THanks Christian. `Parameters<F>` looks very handy to help extract types of parameters.

oeq-ts-rest-api/src/AxiosInstance.ts
PenghaiZhang: I like this. :+1:

ChristianMurphy:
@openequella/oeq-committers friendly ping, this is ready for review.

PR #1849, opened 2020-06-18
chore: reconfigure renovate bot - more throughput and automerge

edalex-ian:
##### Checklist

##### Description of change

This makes the following changes to the renovate config:

* Increase PR limit from 10 to 20
* Allow a maximum of 2 PRs per hour (used to be 1)
* Enable automerging - devDependencies only
* Remove no longer supported config option

I think this would be worthwhile, as we have an increasing number of 'stuck' PRs so it'd be good to have some more headroom to progress the backlog. Further, with better CI it seems a good time to start trialling automerging.

For now automerging is just limited to devDependencies, on the premise they should be rather safe to bump along but require frequent updates and we should be able to see the value.

I have validated the config with:

    npx -p renovate renovate-config-validator

(And indeed, it was with that I discovered with had a no longer supported option - so I removed that.)

But as a result, this is not so much waiting on a green build, as just acceptance from the community before merging.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -5,6 +5,10 @@
   "prHourlyLimit": 2,
   "packageRules": [
     {
+      "depTypeList": ["devDependencies"],
+      "automerge": true
+    },
+    {
       "packagePatterns": ["^com.fasterxml.jackson"],
       "groupName": "Jackson (XML) packages"
     },
--- a/renovate.json
+++ b/renovate.json
@@ -1,8 +1,8 @@
 {
   "extends": ["config:base", ":masterIssue"],
-  "prConcurrentLimit": 10,
+  "prConcurrentLimit": 20,
   "rebaseStalePrs": false,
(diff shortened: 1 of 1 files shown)


edalex-ian:
> disabling automerge on major version updates as these may need some additional manual review.

That sounds like it could be good. I'd hope that a major version update would break the CI if it was a significant change though, but good to be sure.

I saw your config is in package.json and I couldn't find the matching config in renovate.json. Well, more so what I saw in the [doco for major](https://docs.renovatebot.com/configuration-options/#major) seemed different to how you were using it. But, I don't find the renovate config doco the clearest, so maybe I've missed something.

I'll merge in (seeing you've approved), and then we can follow up from there.

(It'll be interesting to see it going, and could be interesting to look at a merge branch in the future to reduce noise - but was concerned it said it would merge to `master`...)

PR #1860, opened 2020-06-18
refactor: extract transforms out of GET, transform on data of known type

ChristianMurphy:
##### Checklist

- [ ] tests are included

##### Description of change

Follow up to https://github.com/openequella/openEQUELLA/pull/1845#issuecomment-646022623
The change is in 907203d 

### DIFF
--- a/oeq-ts-rest-api/src/Search.ts
+++ b/oeq-ts-rest-api/src/Search.ts
@@ -193,11 +193,19 @@ const SEARCH2_API_PATH = '/search2';
  * @param apiBasePath Base URI to the oEQ institution and API.
  * @param params Query parameters as search criteria.
  */
-export const search = (apiBasePath: string, params?: SearchParams): Promise<Common.PagedResult<SearchResultItem>> => {
+export const search = (
+  apiBasePath: string,
+  params?: SearchParams
+): Promise<Common.PagedResult<SearchResultItem>> => {
   return GET<Common.PagedResult<SearchResultItem>>(
     apiBasePath + SEARCH2_API_PATH,
-    (data) => is<Common.PagedResult<SearchResultItem>>(data),
+    (data): data is Common.PagedResult<SearchResultItem> =>
+      is<Common.PagedResult<SearchResultItem>>(data),
     params,
-    (data) => Utils.convertDateFields<Common.PagedResult<SearchResultItem>>(data, ["createdDate", "modifiedDate"] )
+    (data) =>
+      Utils.convertDateFields<Common.PagedResult<SearchResultItem>>(data, [
+        'createdDate',
+        'modifiedDate',
+      ])
   );
 };
--- a/oeq-ts-rest-api/src/Search.ts
+++ b/oeq-ts-rest-api/src/Search.ts
@@ -1,7 +1,7 @@
 import * as Common from './Common';
 import { is } from 'typescript-is';
 import { GET } from './AxiosInstance';
-import * as Utils from './Utils';
+import { toValidDate } from 'Utils';
 
 /**
  * Type of query parameters that can be used in a search.
@@ -130,7 +130,7 @@ export interface Attachment {
 /**
  * Type of search result item.
  */
-export interface SearchResultItem {
+interface BaseSearchResultItem {
   /**
    * Item's unique ID.
    */
@@ -147,14 +147,6 @@ export interface SearchResultItem {
    * Item's status
    */
   status: string;
-  /**
-   * The date when item is created.
-   */
-  createdDate: Date;
-  /**
-   * The last date when item is modified.
-   */
-  modifiedDate: Date;
   /**
    * The ID of item's collection.
    */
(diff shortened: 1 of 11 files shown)


oeq-ts-rest-api/src/Search.ts
```diff
@@ -197,15 +211,17 @@ export const search = (
...
+      is<Common.PagedResult<SearchResultItemFromApi>>(data),
+    params
+  ).then(({ results, ...rest }) => ({
+    ...rest,
+    results: results.map(({ createdDate, modifiedDate, ...rest }) => ({
+      ...rest,
+      createdDate: toValidDate(createdDate),
+      modifiedDate: toValidDate(modifiedDate),
+    })),
+  }));
```
ChristianMurphy, line 226: currently transform is designed with vanilla JS.
It could be updated to use https://github.com/gcanti/monocle-ts#readme

edalex-ian:
The original approach was attempting to keep things DRY. Hence the two smaller generic functions in Utils and also the functional composition-esque approach with GET.

I'm concerned with what you're offering here resulting in a lot more repetition long term. We will have a lot of calls in this module in the future which will need date/time transformations, we will also most likely have other data transforms too due to the nature of the existing APIs.

So with the bigger picture long term goal in place, I'd prefer to stick with the approach which was already there.

ChristianMurphy:
> The original approach was attempting to keep things DRY. Hence the two smaller generic functions in Utils and also the functional composition-esque approach with GET.

But it isn't DRY, DRY implies it is reused, which it isn't.
Currently it falls into another agile principle [YAGNI](https://www.martinfowler.com/bliki/Yagni.html)

> I'm concerned with what you're offering here resulting in a lot more repetition long term. We will have a lot of calls in this module in the future which will need date/time transformations, we will also most likely have other data transforms too due to the nature of the existing APIs.

Again, if the concern here is repetition, https://github.com/gcanti/monocle-ts (which is based inspired by the Scala library of the same name) addresses that concern.

My concern is writing transforms on unknown data types.
Which long term is maintenance concern.
The actual input from the backend isn't known on the frontend with this approach.
If (when) there is a data discrepancy, the only way to know if the issue is coming from the transform or from the data itself, would be to instrument the code and try to reverse engineer the intent.

When validation then transform are done sequentially, an error can be validation or transform, there is no ambiguity.

edalex-ian:
Heya @ChristianMurphy ,

Yes, with what we have currently you are strictly correct. It is not re-used and if you want to strictly follow the principals it falls under YAGNI. However, we need to be pragmatic with the principals, and in this case I'm prioritising DRY over YAGNI with the knowledge of the future work I and my team need to undertake.

In a similar vein I'm also happy to be pragmatic with type checking. There is great value with type checking, but there are also times when you can use your experience and judgement (in such a context as Typescript) to take a different approach. That's what's been done here while we watch how things shape up.

Later, the idea of bringing in monicle would be nice - as indeed would a whole lot from ts-fp. And so I appreciate that suggestion and will keep that in mind for once we get a few more of the endpoints in here. 

But even before we get to that point, I **really** want to see this module get to the point where it is split to it's own repo and CI pipeline. Then the intention is primarily to use that to also validate the integration between this and the back-end, and then as a result that provides part of the framework for the validation between the two.

I realise this is not necessarily ideal or pure. But I do believe it is pragmatic. And with the oEQ codebase in it's current state, we need a lot of pragmatism as we transition to where we're aiming for. (Keep in mind, for a long while yet the resulting build of this module - even when split - will be directly delivered from the server which matches the same data version.)

However, as always I am happy to discuss further. Either in the CDM, or drop me an email and we'll tee up a call.

ChristianMurphy:
>  I really want to see this module get to the point where it is split to it's own repo and CI pipeline. 

I agree

> Then the intention is primarily to use that to also validate the integration between this and the back-end, and then as a result that provides part of the framework for the validation between the two.

also agreed

> I'm prioritising DRY over YAGNI with the knowledge of the future work I and my team need to undertake.

Since this hasn't been discussed at a CDM, I'll take your word for it.

> But I do believe it is pragmatic. And with the oEQ codebase in it's current state, we need a lot of pragmatism as we transition to where we're aiming for

I also agree that a pragmatic approach is needed.

> However, as always I am happy to discuss further. Either in the CDM, or drop me an email and we'll tee up a call.

Let's discuss at the next CDM.
We're 90% on the same page here, I suspect some missing context on both sides is leading to some the confusion.

edalex-ian:
Heya @ChristianMurphy 

I quickly re-reviewed this in prep for today's CDM, so I thought I might as well catch my thoughts here.

First, you'll want to update this branch with all the updates from `component/new-search-page` to address the conflicts.

But I'm wondering if we could do something like:

* Keep your new additional interface definitions for `SearchResultItem`; and
* Keep the original functions in Utils, and use them in place of the long form vanilla JS conversion you've done - this'd be in place of bringing in monocle for now. We could then possibly further tighten that up, by removing the `unknown` and adding a generic and/or an additional typescript-is validator.

What do you think?

ChristianMurphy:
> First, you'll want to update this branch with all the updates from component/new-search-page to address the conflicts

Can do

> Keep your new additional interface definitions for SearchResultItem

:heavy_check_mark: 

> Keep the original functions in Utils, and use them in place of the long form vanilla JS conversion you've done

I can look into this, IIRC the reason I started looking into the transform/monocle route is/was that the structure of the transform as is (targeting properties anywhere, at any depth) is complex to represent as a type.

> this'd be in place of bringing in monocle for now. We could then possibly further tighten that up, by removing the unknown and adding a generic and/or an additional typescript-is validator.

I'm confused, do we want to keep the original transform or replace it with monocle?

edalex-ian:
> I'm confused, do we want to keep the original transform or replace it with monocle?

Sorry. Keep the original transform, but with your new `interface`s we can tighten them up a bit by removing the unknown and adding a generic and/or an additional typescript-is validator.

ChristianMurphy:
thanks @edalex-ian, opened #2432 to address your comments

PR #1896, opened 2020-06-22
Integrate search page with search result component

PenghaiZhang:
#1306 

##### Checklist

- [ ] tests are included

##### Description of change

Integrate search page with search result component and search module.

When opening the search page, search settings are retrieved first.  Then, a default search is done so that a list of items are displayed. 

To match the existing functionality in old Search UI, a couple of things are considered.
1. The default sort order is handled on server side (view `getOrderType` in `DefaultSearch.java` to get more details)  so client does not need to implement it again.
2. The default item status is 'live' and 'review' and it's handled on client.

image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
@@ -35,11 +35,6 @@ import {
 import SearchIcon from "@material-ui/icons/Search";
 import { searchItems } from "./SearchModule";
 import * as OEQ from "@openequella/rest-api-client";
-import {
-  defaultSearchSettings,
-  getSearchSettingsFromServer,
-  SearchSettings,
-} from "../settings/Search/SearchSettingsModule";
 import SearchResult from "./components/SearchResult";
 import { generateFromError } from "../api/errors";
 
@@ -50,8 +45,7 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
   >([]);
 
   /**
-   * What is done in this hook include: updating the page title, retrieving search settings, and
-   * doing a default search.
+   * Update the page title and do a default search.
    */
   useEffect(() => {
     updateTemplate((tp) => ({
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
@@ -16,63 +16,115 @@
  * limitations under the License.
  */
 import * as React from "react";
-import { templateDefaults, TemplateUpdateProps } from "../mainui/Template";
-import { useEffect } from "react";
+import { useEffect, useState } from "react";
+import {
+  templateDefaults,
+  templateError,
+  TemplateUpdateProps,
+} from "../mainui/Template";
 import { languageStrings } from "../util/langstrings";
 import {
-  Card,
   IconButton,
   List,
   ListItem,
-  ListItemText,
   ListSubheader,
   TextField,
+  Grid,
+  Card,
 } from "@material-ui/core";
 import SearchIcon from "@material-ui/icons/Search";
+import { searchItems } from "./SearchModule";
+import * as OEQ from "@openequella/rest-api-client";
+import {
+  defaultSearchSettings,
+  getSearchSettingsFromServer,
+  SearchSettings,
+} from "../settings/Search/SearchSettingsModule";
+import SearchResult from "./components/SearchResult";
+import { generateFromError } from "../api/errors";
 
 const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
   const searchStrings = languageStrings.searchpage;
-  const placeholderData: string[] = [
-    "item1",
-    "item2",
-    "item3",
-    "item4",
-    "item5",
-  ];
+  const [searchResultItems, setSearchResultItems] = useState<
+    OEQ.Search.SearchResultItem[]
+  >([]);
+  const [searchSettings, setSearchSettings] = useState<SearchSettings>(
+    defaultSearchSettings
+  );
 
+  /**
+   * What is done in this hook include: updating the page title, retrieving search settings, and
+   * doing a default search.
+   */
   useEffect(() => {
     updateTemplate((tp) => ({
       ...templateDefaults(searchStrings.title)(tp),
     }));
+
+    getSearchSettingsFromServer()
+      .then((settings: SearchSettings) => {
+        setSearchSettings(settings);
+        search({
+          status: [OEQ.Common.ItemStatus.LIVE, OEQ.Common.ItemStatus.REVIEW],
+        });
+      })
+      .catch((error: Error) => handleError(error));
   }, []);
 
-  const searchResults = placeholderData.map((data: string) => {
-    return (
-      <ListItem>
-        <ListItemText primary={data} />
-      </ListItem>
+  const handleError = (error: Error) => {
+    updateTemplate(templateError(generateFromError(error)));
+  };
+
+  /**
+   * Search items with specified search criteria.
+   * @param params Search criteria
+   */
+  const search = (params?: OEQ.Search.SearchParams): void => {
+    searchItems(
+      params
+    ).then((items: OEQ.Common.PagedResult<OEQ.Search.SearchResultItem>) =>
+      setSearchResultItems(items.results)
     );
-  });
+  };
+
+  /**
+   * A SearchResult that represents one of the search result items.
+   */
+  const searchResults = searchResultItems.map(
+    (item: OEQ.Search.SearchResultItem) => (
+      <ListItem key={item.uuid}>
+        <SearchResult
+          resultData={item}
+          onClick={() => (window.location.href = item.links.view)}
+        />
+      </ListItem>
+    )
+  );
+
+  /**
+   * A list that consists of search result items.
+   */
   const searchResultList = (
-    <List
-      subheader={
-        <ListSubheader disableGutters>{searchStrings.subtitle}</ListSubheader>
-      }
-    >
+    <List subheader={<ListSubheader>{searchStrings.subtitle}</ListSubheader>}>
       {searchResults}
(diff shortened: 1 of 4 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -16,63 +16,113 @@
...
-    return (
-      <ListItem>
-        <ListItemText primary={data} />
+  const handleError = (error: Error) => {
+    updateTemplate(templateError(generateFromError(error)));
+  };
+
+  /**
+   * Search items with specified search criteria.
+   * @param params Search criteria
```
SammyIsConfused, line 78: Perhaps have an @see pointing towards SearchParams for this one?

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -16,63 +16,113 @@
...
+
+  /**
+   * A SearchResult that represents one of the search result items.
+   */
+  const searchResults = searchResultItems.map(
+    (item: OEQ.Search.SearchResultItem) => (
+      <ListItem key={item.uuid}>
+        <SearchResult
+          resultData={item}
+          onClick={() => (window.location.href = item.links.view)}
```
SammyIsConfused, line 96: Is there a way to do this redirect with Router rather than window.location?

oeq-ts-rest-api/src/Search.ts
```diff
@@ -124,7 +124,16 @@ export interface Attachment {
   /**
    * Links to the attachment.
    */
-  links: Record<string, string>;
+  links: {
+    /**
+     * The URL of viewing this attachment.
```
SammyIsConfused, line 129:
```suggestion
     * The URL for viewing this attachment.
```

oeq-ts-rest-api/src/Search.ts
```diff
@@ -124,7 +124,16 @@ export interface Attachment {
...
+     */
+    view: string;
+    /**
+     * The URL of viewing this attachment's thumbnail.
```
SammyIsConfused, line 133:
```suggestion
     * The URL for viewing this attachment's thumbnail.
```

oeq-ts-rest-api/src/Search.ts
```diff
@@ -182,7 +191,16 @@ export interface SearchResultItem {
   /**
    * Links to an item.
    */
-  links: Record<string, string>;
+  links: {
+    /**
+     * The URL of viewing this item.
```
SammyIsConfused, line 196:
```suggestion
     * The URL for viewing this item.
```

oeq-ts-rest-api/src/Search.ts
```diff
@@ -182,7 +191,16 @@ export interface SearchResultItem {
...
+     */
+    view: string;
+    /**
+     * The REST API path used to get details of this item.
```
SammyIsConfused, line 200:
```suggestion
     * The REST API path used to get this item's details.
```

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -16,63 +16,113 @@
...
   const searchStrings = languageStrings.searchpage;
-  const placeholderData: string[] = [
-    "item1",
-    "item2",
-    "item3",
-    "item4",
-    "item5",
-  ];
+  const [searchResultItems, setSearchResultItems] = useState<
+    OEQ.Search.SearchResultItem[]
```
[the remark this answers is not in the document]
SammyIsConfused: Or does that not apply to simply evaluating types?

PenghaiZhang:
Remove the retrieval of search settings due to this error : 'searchSettings' is declared but its value is never read.

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
PenghaiZhang: I guess we can't do that at the moment because we don't have a route for this specific URL. But maybe we can try  hook `useHistory` to do.

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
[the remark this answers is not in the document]
PenghaiZhang: I think we can reuse these interfaces. We use modules because we may need to do some checks before calling the function. So the module is the best place to do these checks.

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
PenghaiZhang: I tried   `* @param params Search crierial, also see {@see OEQ.Search.SearchParams }` but IDE can't resolve `Search.SearchParams` from namespace `OEQ` :thinking:

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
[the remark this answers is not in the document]
edalex-ian: Yup, for the types it is fine. It should act as the authoritative source for the raw oEQ types. Now and then the various Modules might choose to have their own representation to better fit the domain, and that's fine too.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResult.tsx
```diff
@@ -158,9 +158,9 @@ export default function SearchResult({
   );
 
   const customDisplayMetadata = displayFields.map(
-    (element: OEQ.Search.DisplayFields) => {
+    (element: OEQ.Search.DisplayFields, index: number) => {
       return (
-        <ListItem disableGutters dense>
+        <ListItem disableGutters dense key={element.name + index}>
```
mrblippy, line 163: Good pickup here 👍 , I thought i had added keys to all of the ListItem components, but must've missed this one

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
edalex-ian: Ahh well, in that case it's okay as you can still see the type on the function signature and find out more from there.

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
edalex-ian: Yeah, it's an interesting this one - as will be the attachments - as it's where we still need to straddle over into the Legacy UI. :thinking: 

Yes, maybe the `useHistory` hook could be the goer for now.

edalex-ian:
Aaaahhhh! This kind of highlights/reminds that we don't have any tests around the New Search UI.

We'll want to work on that soonish. Maybe swap out the current search test when NewUI is specified to use a new selenium test targetted at the New Search UI.

edalex-ian:
(Ah! I see we already have that in our backlog to add functional tests for new search UI.)

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -16,63 +16,101 @@
...
+      <Grid item xs={9}>
+        <Card>
+          <IconButton>
+            <SearchIcon fontSize={"large"} />
+          </IconButton>
+          <TextField />
+        </Card>
+      </Grid>
+
+      <Grid item xs={9}>
```
mrblippy, line 110: Should we make it so the SearchResults takes up the full width of the page when there is no refine search panel? Probably something to address once we implement the refine search panel component.

edalex-ian: Yeah, true, we could conditionally do that. For now I guess this is just a nice default while we lay things out. But I guess it's especially in the case of selection sessions we wont have a refine search panel?

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
PenghaiZhang: I just had a talk with Tom. A better way could be to tweak `SearchResult` to make it be a hyperlink.

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
mrblippy: Yeah, mainly for selection sessions. But also in the regular ui, when the horizontal screen size goes below a certain level. 
Either way, it makes sense to revisit this when we get to actually implementing the refine panel component

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -16,63 +16,98 @@
...
-        </IconButton>
-        <TextField />
-      </Card>
-      <Card>{searchResultList}</Card>
-    </>
+    <Grid container direction={"column"} spacing={2}>
+      <Grid item xs={9}>
+        <Card>
+          <IconButton>
+            <SearchIcon fontSize={"large"} />
```
ChristianMurphy, line 101:
```suggestion
            <SearchIcon fontSize="large" />
```

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -16,63 +16,98 @@
...
-    <>
-      <Card>
-        <IconButton>
-          <SearchIcon fontSize={"large"} />
-        </IconButton>
-        <TextField />
-      </Card>
-      <Card>{searchResultList}</Card>
-    </>
+    <Grid container direction={"column"} spacing={2}>
```
ChristianMurphy, line 97:
```suggestion
    <Grid container direction="column" spacing={2}>
```

PR #1937, opened 2020-06-25
chore(ci): Prepackage coverage report

edalex-ian:
##### Checklist

##### Description of change

The uploading of the coverage reports was taking a very long time and
sometimes failing. I believe this is because the artefact upload action
wasn't really designed for uploading directories with tens of thousands
of files.

Considering the upload of the primary artefacts only takes a few minutes
(and is a few hundred megabytes), I'm assuming this will save us
significant time and shorten our test runs by about 50%.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -255,8 +255,15 @@ jobs:
           ./sbt -jvm-opts autotest/.jvmopts "project autotest" \
             coverageReport
 
+      # To save significant time on file upload, we first tar up the directory
+      # as it contains > 10k files and the artefact upload was taking almost
+      # 1hr.
+      - name: Package Coverage Report
+        run: |
+          tar cvf coverage_report.tar autotest/target/coverage-report/
+
       - name: Save Coverage Report
         uses: actions/upload-artifact@v2.0.1
         with:
           name: CoverageReport-newui-${{ matrix.newui }}
-          path: autotest/target/coverage-report/
+          path: coverage_report.tar


.github/workflows/ci.yaml
```diff
@@ -255,8 +255,15 @@ jobs:
           ./sbt -jvm-opts autotest/.jvmopts "project autotest" \
             coverageReport
 
+      # To save significant time on file upload, we first tar up the directory
+      # as it contains > 10k files and the artefact upload was taking almost
+      # 1hr.
+      - name: Package Coverage Report
+        run: |
+          tar cvf coverage_report.tar autotest/target/coverage-report/
```
ChristianMurphy, line 263: :thinking: if upload is where the script is taking most of the hit, it may be worth adding some additional compression to reduce the size

```suggestion
          tar cvzf coverage_report.tar.gz autotest/target/coverage-report/
```

.github/workflows/ci.yaml
```diff
@@ -255,8 +255,15 @@ jobs:
...
+
       - name: Save Coverage Report
         uses: actions/upload-artifact@v2.0.1
         with:
           name: CoverageReport-newui-${{ matrix.newui }}
-          path: autotest/target/coverage-report/
+          path: coverage_report.tar
```
ChristianMurphy, line 269:
```suggestion
          path: coverage_report.tar.gz
```

.github/workflows/ci.yaml
edalex-ian: I felt it was more the IO of opening and closing all the files - especially seeing a lot are tiny. Plus, the artefact upload action compresses to zip as well.

edalex-ian:
Wowsers!!

Package coverage report: 1s
Save coverage report: 3s

That's in place of a [previous run](https://github.com/openequella/openEQUELLA/runs/805588130?check_suite_focus=true) which was a _fast_ normal example: 45m 22s

And so total build time:

* Before (roughly): 2hr 31m
* After: 1hr 55m

That actually seemed too good to be true. So I downloaded the coverage report to check it was still okay, and it was. ;)

Issue #1527, opened 2020-03-06
New UI: Some contribution wizard controls are half width


mrblippy:
**Describe the bug**
I must've messed up some of the contribution wizard styling while I was rewriting the css for legacy content in the new ui. 
As a result of this some contribution wizard controls are not full width. Controls affected are: 
- Shuffle list
- Auto-complete taxonomy
- Grouping control

**Expected behavior**
All wizard controls should be the full width, just as they were when the New Ui is disabled.

**Screenshots**
![old v new](https://user-images.githubusercontent.com/4625498/76045349-e2ab9580-5fb0-11ea-8cfd-be2a2fa50f9f.jpeg)

**Stacktrace**
If applicable, a stacktrace to help explain your problem.

**Platform:**

- OpenEquella Version: 2019.2, 2020.1
- OS: [e.g. Windows 10, MacOS 10.13]
- Browser [e.g. Chrome, Firefox]

edalex-ian:
Adding Edalex tag, as it looks like we've worked on this and it will be in 2020.2.

PR #1964, opened 2020-06-29
Improve CSS styles for Wizard Taxonomy control

PenghaiZhang:
#1527 

##### Checklist

- [ ] tests are included

##### Description of change

Change Wizard Taxonomy control's styles so that it displays properly with different resolution.


Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -1922,23 +1922,34 @@ div.skinnysearch {
...
 }
 
+@media only screen and (min-width: 1801px){
+  .autocompleteControl .autocomplete-container {
+    width: 92%;
+  }
+}
+.autocompleteControl .btn {
+  float: right;
+}
```
edalex-ian, line 1948: nit: be nice to have a blank line on either side of this block, but hey.

edalex-ian:
Hmmm, looks like your commit hooks are broken @PenghaiZhang . This failed the CSS format check, which is normally formatted as part of the pre-commit hooks.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -1922,23 +1922,36 @@ div.skinnysearch {
...
 
+.autocompleteControl .btn {
+  float: right;
+}
+
 #wizard-controls .autocompleteControl input[type="text"] {
   height: 34px;
-  width: 98%;
+  width: 100%;
+  padding-right: 10px;
```
ChristianMurphy, line 1954: This would have an overall width or `100% + 10px` could this cause an overflow issue?

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -1922,23 +1922,36 @@ div.skinnysearch {
...
+
+@media only screen and (max-width: 1500px) {
   .autocompleteControl .autocomplete-container {
-    float: left;
     width: 85%;
   }
 }
 
-@media only screen and (min-width: 1201px) {
+@media only screen and (min-width: 1501px) and (max-width: 1800px) {
```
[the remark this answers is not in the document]
edalex-ian: I think for this PR - seeing it's driven to just quickly get this hotfix done - let's leave as is for now. (As we really wanted to finish the hotfix last week.)

However, let's do this across the file, as I did want to start seeing greater usage of variables in here so as to better leverage Sass. And if you have a look, we do have extensive use of these media queries and it looks like we have four sizes in play:

* 768 (small)
* 1200 (medium)
* 1500 (large)
* 1800 (extra-large)

When we refactor that out, then we can do some more testing to make sure all the bits are still working. I think this is a fair fit with the EBP MkII we're doing for 2020.2.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
ChristianMurphy: ref: https://css-tricks.com/the-css-box-model/

PenghaiZhang: Thanks @ChristianMurphy . We came up with another solution so we no longer need this padding.

PR #1967, opened 2020-06-30
Add debounced query bar for searches

SammyIsConfused:
##### Checklist

- [x] the [contributor license agreement][] is signed
- [x] commit message follows [commit guidelines][]
- [x] tests are included
- [x] screenshots are included showing significant UI changes
- [x] documentation is changed or added

##### Description of change

- Added Searchbar component (no search button, automatic type-ahead. Includes clear button and half second onChange debounce)
- Add story for SearchBar
- Add Searchbar to the search page
- Add tests for the search page
image

[contributor license agreement]: https://www.apereo.org/node/676
[commit guidelines]: https://chris.beams.io/posts/git-commit/

PR #1970, opened 2020-06-30
Ensure the default search order is included in the first search

PenghaiZhang:
#1306 

##### Checklist

- [ ] tests are included

##### Description of change

The search triggered by opening the Search page should include the default search order configured in the Search settings page. Therefore, in the Search page, retrieve Search settings first, and then add `order` to search criteria.

For some unknown reasons (maybe just unknown to me), the order values have been inconsistent  for a long time. For example, in the context of Search settings, the value of order `Last modified date` is `datemodified`. However in the context of Search, the accepted value is 'modified'. Previously in Purescript, this was handled by a Purscript Map. More details can be seen from these two links [old search setting page](https://github.com/openequella/openEQUELLA/blob/master/Source/Plugins/Core/com.equella.core/src/com/tle/web/search/settings/SearchSettingsSection.java) (line 249-254), [purs version search page](https://github.com/openequella/openEQUELLA/blob/master/Source/Plugins/Core/com.equella.core/js/src/Search/OrderControl.purs).

My approach for handling this is to slightly change the method 'getOrderType' because this requires least changes, but I am happy to discuss other approaches.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
@@ -38,6 +38,10 @@ import { searchItems } from "./SearchModule";
 import * as OEQ from "@openequella/rest-api-client";
 import SearchResult from "./components/SearchResult";
 import { generateFromError } from "../api/errors";
+import {
+  getSearchSettingsFromServer,
+  SearchSettings,
+} from "../settings/Search/SearchSettingsModule";
 
 const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
   const searchStrings = languageStrings.searchpage;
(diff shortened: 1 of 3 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -53,9 +57,12 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
       ...templateDefaults(searchStrings.title)(tp),
     }));
 
-    search({
-      status: [OEQ.Common.ItemStatus.LIVE, OEQ.Common.ItemStatus.REVIEW],
-    });
+    getSearchSettingsFromServer().then((settings: SearchSettings) =>
+      search({
+        status: [OEQ.Common.ItemStatus.LIVE, OEQ.Common.ItemStatus.REVIEW],
+        order: settings.defaultSearchSort,
```
edalex-ian, line 63: Hmm, this is a bit interesting...

As really, we'll want to set a sort order control to the default, and then on searches simply read in it's current value. But we don't have that control yet. However, consideration will also need to be given to this value in the work being done over in #1967 .

Mind you, the same also applies for the current hardcoded `status`.

PenghaiZhang: The idea in my mind is that we will save Search settings in state and then render other controls and side panels depending on the settings.

Also, pass `(query)=>search({...params, query}) ` to `SearchBar`.

edalex-ian: Yup, sounds about right as it will then trigger a reload of the content on changing any control. It'll be interesting to see if we end up finding it's too dynamic. :)

PR #1985, opened 2020-07-01
Add pagination to Search page and tweak SearchResult

PenghaiZhang:
#1306 

##### Checklist

- [ ] tests are included

##### Description of change

Add the feature of pagination to Search page. When change the page size or move to next/previous pages, a search is triggered.

image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
@@ -44,17 +44,8 @@ import {
   SearchSettings,
   SortOrder,
 } from "../settings/Search/SearchSettingsModule";
-import { makeStyles } from "@material-ui/core/styles";
-
-const useStyles = makeStyles({
-  cardAction: {
-    display: "flex",
-    justifyContent: "center",
-  },
-});
 
 const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
-  const classes = useStyles();
   const searchStrings = languageStrings.searchpage;
 
   const [rowsPerPage, setRowsPerPage] = useState<number>(10);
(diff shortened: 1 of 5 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -16,55 +16,88 @@
...
-    );
+    getSearchSettingsFromServer().then((settings: SearchSettings) => {
+      setSortOrder(settings.defaultSearchSort);
+    });
   }, []);
 
+  /**
+   * Trigger a search when state values change, but skip the initial values.
+   */
+  const isInitialSearch = useRef(true);
```
PenghaiZhang, line 92: The idea is that we don't need to do a search for initial state values.
So we can use this hook to create an object that persists until the component is unmounted.  And then use this object to help skip calling below useEffect for initial state values.
It seems this approach is recommended by react, but I am happy to discuss better approaches.

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -16,55 +16,88 @@
...
+  useEffect(() => {
+    if (isInitialSearch.current) {
+      isInitialSearch.current = false;
+    } else {
+      search(standardParams);
+    }
+  }, [rowsPerPage, currentPage, sortOrder]);
```
PenghaiZhang, line 99: I had a pondering that if we should move `const [query, setQuery] = React.useState<string>("");` done in #1967 to here,  pass `setQuery` to `SearchBar`, and then add `query` as one dependency for this hook.

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -76,19 +109,17 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
...
       .catch((error: Error) => handleError(error));
   };
 
   /**
    * A SearchResult that represents one of the search result items.
    */
-  const searchResults = searchResultItems.map(
+  const searchResults = pagedSearchResult.results.map(
     (item: OEQ.Search.SearchResultItem) => (
-      <ListItem key={item.uuid} divider>
```
PenghaiZhang, line 89: This ListItem is redundant as `SearchResult` already returns a ListItem.

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
edalex-ian: And so are you saying that initially, we wont see any search results until at least the above effect to getSearchSettingsFromServer completes and calls setSortOrder?

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
edalex-ian: I like your thinking. :+1: 

I do wonder whether we may need to consider having a single bit of state though which contains an object with all our search options. I feel this will make things clearer and to ensure we don't miss any. Hey, maybe it's even the params that ultimately get passed to `search`. But I see there is some misalignment there, and you may need an intermediate object. However that object could then be passed to `search` and perhaps it has to the log to transform it to `OEQ.Search.SearchParams`. :thinking: 

What do you think? Could we set that up here now? And then when #1967 gets done it just plugs on in with `query`?

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -16,55 +16,88 @@
...
+  const [sortOrder, setSortOrder] = useState<SortOrder>(SortOrder.RANK);
+  const [pagedSearchResult, setPagedSearchResult] = useState<
+    OEQ.Common.PagedResult<OEQ.Search.SearchResultItem>
+  >(defaultPagedSearchResult);
 
   /**
-   * Update the page title and do a default search.
+   * Construct a standard search criteria.
+   */
+  const standardParams: OEQ.Search.SearchParams = {
```
edalex-ian, line 70: Are there any non-standard params? Maybe I'm missing something, but it seems like this could just be:

```suggestion
  const searchParams: OEQ.Search.SearchParams = {
```

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
ChristianMurphy: Is the goal of this similar to https://ahooks.js.org/hooks/life-cycle/use-update-effect ?

PenghaiZhang: > And so are you saying that initially, we wont see any search results until at least the above effect to getSearchSettingsFromServer completes and calls setSortOrder?

Yes, but I wonder if we want a initial search no matter `getSearchSettingsFromServer` succeeds or fails.

PenghaiZhang: > Is the goal of this similar to https://ahooks.js.org/hooks/life-cycle/use-update-effect ?

Yes!!

PR #1990, opened 2020-07-02
Refactor Theme Editor page to use the new standard SettingsPageTemplate

edalex-ian:
##### Checklist

- [ ] tests are included

##### Description of change

Well, this was one of those small little refactors that suddenly became a bit more... Not too much more, but.. you'll see.

The intention was to make `ThemePage` use the new `SettingsPageTemplate` and `SettingsList` so as to be consistent with our new settings work. That was achieved, however getting there got a bit tricker due to some bumps with the the `ColorPickerComponent`, so that ended up also getting refactored as well. It went from a class based component to a functional one, and rather than a pop-over is now a dialog (which is also more consistent), as well as some of it's oddities were removed and tidied up.

But also did some tweaking to `SettingsList` for a stronger heading, while also providing the heading as a stand-alone component. This was triggered by noting some of the other new settings pages had copy and pasted the `Typography` block to match what was in `SettingsList`.

In Summary:

* `ThemePage` refactored to use `SettingsPageTemplate` and `SettingsList`
* `ColorPickerComponent` refactored to functional component and dialog based
* Changed heading style of `SettingsList`
* Added new component `SettingsListHeading` (primarily for easy re-use)
* Removed display of current logo from `ThemePage` as it was redundant - current logo is always shown top left of the page
* Added back navigation to `SettingsPage` (for consistency, easy of use)

I also wanted to at least add `SettingsListHeading` to Storybook. However Storybook is currently broken on `component/new-search-page`. (I'll attempt to look at this on the side.)

Updated `ThemePage`

image

Updated `ColorPickerComponent`

image

Example of modified headings on other settings page

image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/theme/ThemePage.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/theme/ThemePage.tsx
@@ -18,36 +18,34 @@
 import * as React from "react";
 import {
   Button,
-  CardContent,
-  CardActions,
   Card,
-  Divider,
-  FormControl,
+  CardActions,
+  CardContent,
+  createStyles,
+  Grid,
   Typography,
   WithStyles,
   withStyles,
-  createStyles,
-  Grid,
-  Snackbar,
-  IconButton,
 } from "@material-ui/core";
-import CloseIcon from "@material-ui/icons/Close";
 import ColorPickerComponent from "./ColorPickerComponent";
 import axios, { AxiosError } from "axios";
 import { API_BASE_URL } from "../config";
 import { languageStrings } from "../util/langstrings";
 import { commonString } from "../util/commonstrings";
 import {
+  ErrorResponse,
   generateFromError,
   generateNewErrorID,
-  ErrorResponse,
 } from "../api/errors";
 import {
   templateDefaults,
-  TemplateUpdate,
   templateError,
+  TemplateUpdate,
 } from "../mainui/Template";
 import { IThemeSettings } from ".";
+import SettingPageTemplate from "../components/SettingPageTemplate";
+import SettingsListControl from "../components/SettingsListControl";
+import SettingsList from "../components/SettingsList";
 
 declare const themeSettings: IThemeSettings;
 declare const logoURL: string;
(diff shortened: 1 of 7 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/theme/ThemePage.tsx
```diff
@@ -122,6 +130,7 @@ class ThemePage extends React.Component<
         menuIcon: "#000000",
         primaryText: "#000000",
         secondaryText: "#444444",
+        changesUnsaved: true,
```
PenghaiZhang, line 133: If all color settings are already default and I press the `reset` button again, will the Save button become enabled and allow me to submit?

Source/Plugins/Core/com.equella.core/js/tsrc/theme/ThemePage.tsx
```diff
@@ -184,58 +168,52 @@ class ThemePage extends React.Component<
...
+
+  submitLogo = () =>
+    this.state.logoToUpload &&
+    axios.put(`${API_BASE_URL}/theme/logo/`, this.state.logoToUpload);
+
+  async saveChanges() {
+    try {
+      await this.submitLogo();
+      await this.submitTheme();
+      this.setState({ changesUnsaved: false, showSuccess: true });
```
PenghaiZhang, line 198: I like this `try...catch` block because it makes this function easy to read. Only when the two promises get resolved,  the state is updated.

edalex-ian: Yeah, the async function syntax is definitely a big win for readability.

Source/Plugins/Core/com.equella.core/js/tsrc/theme/ThemePage.tsx
edalex-ian: Yup, it will. It's very simple.

Arguably it could have some more smarts in there. However there's no real harm as a result, and for that kind of stuff I've tried to keep the logic pretty well the same.

Indeed, the logo reset was designed to reload when pushing the button, and I've left that as is for now. But in the future, we may want to come back and refactor that side of things.

edalex-ian:
Thanks @PenghaiZhang 

Let me know if you're happy with the response, and if so I'll merge on - and then build on that for the refactoring of the login notice pages.

PR #1998, opened 2020-07-03
Add sorting options for Search page

PenghaiZhang:
#1306 

##### Checklist

- [ ] tests are included

##### Description of change

Added a sorting control on Search page to allow sorting search results. The initially selected sort order is consistent with the default sort order configured on the Search setting page.

image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchModule.ts
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchModule.ts
@@ -18,34 +18,7 @@
 import * as OEQ from "@openequella/rest-api-client";
 import { API_BASE_URL } from "../config";
 import { SortOrder } from "../settings/Search/SearchSettingsModule";
-
-/**
- * A function that takes search options and converts search options to search params,
- * and then does a search and returns a list of Items.
- * @param searchOptions  Search options selected on Search page.
- */
-export const searchItems = (
-  searchOptions: SearchOptions
-): Promise<OEQ.Common.PagedResult<OEQ.Search.SearchResultItem>> => {
-  const { rowsPerPage, currentPage, sortOrder } = searchOptions;
-  const searchParams: OEQ.Search.SearchParams = {
-    start: currentPage * rowsPerPage,
-    length: rowsPerPage,
-    status: [
-      "LIVE" as OEQ.Common.ItemStatus,
-      "REVIEW" as OEQ.Common.ItemStatus,
-    ],
-    order: sortOrder,
-  };
-  return OEQ.Search.search(API_BASE_URL, searchParams);
-};
-
-export const defaultPagedSearchResult: OEQ.Common.PagedResult<OEQ.Search.SearchResultItem> = {
-  start: 0,
-  length: 10,
-  available: 10,
-  results: [],
-};
+import { languageStrings } from "../util/langstrings";
 
 /**
  * Type of all search options on Search page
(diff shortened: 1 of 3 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -135,6 +135,26 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
 
       <Grid item xs={9}>
         <Card>
+          <CardHeader
+            title={searchStrings.subtitle}
+            action={
+              <Select
```
PenghaiZhang, line 141: Do we want a separate component for this control?

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -135,6 +135,26 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
 
       <Grid item xs={9}>
         <Card>
+          <CardHeader
+            title={searchStrings.subtitle}
```
PenghaiZhang, line 139: I noted that the style of CardHeader's title seems to be consistent with `SettingsListHeading`.

edalex-ian: Nice work. Not sure how I missed that. May need to go back and refactor my refactors once they're done. ;)

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
edalex-ian: I think at least pulling it out to a component within this function would be worthwhile for now. That way we have this key component at least stand-alone and here we can have `<CardHeader ... action={<SearchOrderSelect />}/>}` which is more readable.

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchModule.ts
```diff
@@ -40,33 +72,22 @@ export const searchItems = (
...
-   * Selected search result sorting order.
-   */
-  sortOrder: SortOrder | undefined;
-}
-
-export const defaultSearchOptions: SearchOptions = {
-  rowsPerPage: 10,
-  currentPage: 0,
-  sortOrder: undefined,
+export const getSortingOptions = (): Record<string, SortOrder> => {
```
edalex-ian, line 78: This really should be a return type of `Map<SortOrder, string>`, and maybe then a name to better reflect it's purpose like: `getSortingOptionStrings()`. Then, it'd also possibly worthwhile to memoize it - as it is completely static for each run (no need to build the data structure again).

SammyIsConfused, line 78: Is it worth having a test for this function?

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchModule.ts
```diff
@@ -40,33 +72,22 @@ export const searchItems = (
...
+  const {
+    relevance,
+    lastModified,
+    dateCreated,
+    title,
+    userRating,
+  } = languageStrings.settings.searching.searchPageSettings;
+  return {
+    [relevance]: SortOrder.RANK,
```
SammyIsConfused, line 87: Not sure I quite understand this return - so depending on language strings the shape of the return data will change? If someone changed these values in a language pack would the data remain the same or map to the new strings? And would that break anything if so?

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchModule.ts
PenghaiZhang: I am curious that if the cost of using `useMem`o is more expensive than rebuilding that simple map. I feel it's more worthwhile to memorize this component.

edalex-ian: Not sure I'm seeing the cost you mention, so I'd be interested to discuss that. But really, that is a bit of a premature optimisations, and so requires no action right now. ;)

edalex-ian:
For the test, let's first wait for #1967 to get done so that we don't clash excessively. We can do that in a latter PR, let's not have that hold up this one.

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchModule.ts
PenghaiZhang: I have refactored this to a simple `map` so we probably do not need a test.

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchModule.ts
PenghaiZhang: I think it would be fine as long as the values provided by eunm `SortOrder` keep consistent. But anyway, this function has been removed.

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -120,6 +115,16 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
...
   const searchResultList = <List>{searchResults}</List>;
 
+  /**
+   * Provide a memorized callback for updating sort order in order to avoid re-rendering
+   * component SearchOrderSelect when other search control values are changed.
+   */
+  const handleSortOrderChanged = useCallback(
+    (order: SortOrder) =>
+      setSearchOptions({ ...searchOptions, sortOrder: order }),
+    []
```
edalex-ian, line 125: This gave me pause, until I saw it's to maintain referential consistency when passed into your memoised SearchOrderSelect. :+1:

PenghaiZhang:
> Looks good @PenghaiZhang
> 
> I was going to ask why you haven't added a story for the new component.... Then I remembered storybook is broken. :'(

Yeah, when I run Storybook I see an error about the logo URL being missing.

PR #2029, opened 2020-07-08
Add CircularProgress spinner to LegacyPage

mrblippy:
##### Checklist

##### Description of change

legacyspinner

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/mainui/LegacyPage.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/mainui/LegacyPage.tsx
@@ -31,6 +31,7 @@ import {
   LegacyContentProps,
 } from "../legacycontent/LegacyContent";
 import { LegacyContentRenderer } from "../legacycontent/LegacyContentRenderer";
+import { CircularProgress, makeStyles } from "@material-ui/core";
 
 interface LegacyPageProps extends TemplateUpdateProps {
   location: Location;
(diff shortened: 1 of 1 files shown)


mrblippy:
> <Grid container direction="column" justify="center" alignItems="center">
>    <Grid item><CircularProgress/></Grid>
> </Grid>

Yeah, that's very true

PenghaiZhang:
When I open the logon page or when I logout I can see this spinner as well. Is that also what we want ?

PenghaiZhang:
Another thing I noted is when I navigate from New UI to Legacy pages the title bar, for example, from Search to Contribute, the title on AppBar show `Search` until Contribute page is open. Would it be better to show `Contribute` while the spinner is displayed?

PenghaiZhang:
> Another thing I noted is when I navigate from New UI to Legacy pages the title bar, for example, from Search to Contribute, the title on AppBar is `Search` until Contribute page is open. Would it be better to show `Contribute` while the spinner is displayed?

OK after a quick look I think this would be hard to achieve because `title` is part of  `content`. So we don't know what `title` would be until we get `content` .

mrblippy:
> > Another thing I noted is when I navigate from New UI to Legacy pages the title bar, for example, from Search to Contribute, the title on AppBar is `Search` until Contribute page is open. Would it be better to show `Contribute` while the spinner is displayed?
> 
> OK after a quick look I think this would be hard to achieve because `title` is part of `content`. So we don't know what `title` would be until we get `content` .

Yeah, that might be out of the scope of this ticket. It feels like a 'nice to have' feature, as opposed to essential

mrblippy:
> When I open the logon page or when I logout I can see this spinner as well. Is that also what we want ?

Hmm, I haven't visited the login page since implementing this spinner. I will take a look to see how it looks

mrblippy:
> > When I open the logon page or when I logout I can see this spinner as well. Is that also what we want ?
> 
> Hmm, I haven't visited the login page since implementing this spinner. I will take a look to see how it looks

Yeah, i think this is fine

PR #2030, opened 2020-07-08
Fix missing icons in RichTextEditor

edalex-ian:
##### Checklist

##### Description of change

It seems that from TinyMCE (which we upgraded to recently) you needed to
explicitly include the icons. This can be done simply with an import
such as:

    import 'tinymce/icons/default';

But while at it I noticed all the TinyMCE imports had been done CommonJS
style rather then ES2015, so I updated them all for consistency.

**Before:**

image

**With this PR:**

image


ChristianMurphy:
Aside it may be worth enabling https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-require-imports.md

PR #2040, opened 2020-07-13
Add "Raw search" mode to new search UI

SammyIsConfused:
##### Checklist

- [x] tests are included

##### Description of change
- Adds a toggle to the search page for "raw search" mode. 
- When raw search is off, search queries are de-bounced type-ahead, sent to the server side with an appended '*' wildcard
- When raw search is on, search queries are sent as-is, and only when you hit enter
- Helper-text to remind the user to hit Enter appears when in raw search mode
- Tool-tip appears when hovering on the toggle to explain the purpose of raw search mode
- Jest tests to assert that type-ahead and wildcard append happen in simple search mode
- Jest test to assert that type-ahead and wildcard append do not happen in raw search mode
image
image
image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
@@ -184,16 +184,19 @@ describe("<SearchPage/>", () => {
   it("should not debounce and send query as-is when in raw search mode", async () => {
     const input = component.find(SEARCHBAR_ID);
     const rawModeSwitch = component.find(RAW_SEARCH_TOGGLE_ID);
-    await awaitAct(() => {
-      //turn raw search mode on, add a search query and hit enter
-      rawModeSwitch.simulate("change", { target: { checked: true } });
-      jest.advanceTimersByTime(1000);
-      querySearch("raw search test");
+
+    //turn raw search mode on, add a search query and hit enter
+    await awaitAct(() =>
+      rawModeSwitch.simulate("change", { target: { checked: true } })
+    );
+    await awaitAct(() =>
+      input.simulate("change", { target: { value: "raw search test" } })
+    );
+    await awaitAct(() =>
       input.simulate("keyDown", {
         keyCode: ENTER_KEYCODE,
-      });
-      jest.advanceTimersByTime(1000);
-    });
+      })
+    );
 
     //assert that the query was passed in as-is
     expect(SearchModule.searchItems).toHaveBeenLastCalledWith({
--- a/Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
@@ -56,7 +56,6 @@ describe("<SearchPage/>", () => {
   let component: ReactWrapper<any, Readonly<{}>, React.Component<{}, {}, any>>;
 
   beforeEach(async () => {
-    jest.useFakeTimers("modern");
     component = mount(
       <BrowserRouter>
         <SearchPage updateTemplate={jest.fn()} />{" "}
@@ -88,6 +87,7 @@ describe("<SearchPage/>", () => {
    * @param searchTerm The specified search term.
    */
   const querySearch = async (searchTerm: string) => {
+    jest.useFakeTimers("modern");
     const input = component.find(SEARCHBAR_ID);
     await awaitAct(() => {
       input.simulate("change", { target: { value: searchTerm } });
--- a/Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
@@ -27,7 +27,7 @@ import { defaultSearchSettings } from "../../../tsrc/settings/Search/SearchSetti
 import { BrowserRouter } from "react-router-dom";
 import { CircularProgress } from "@material-ui/core";
 
-jest.useFakeTimers();
+jest.useFakeTimers("modern");
 const mockSearch = jest.spyOn(SearchModule, "searchItems");
 const mockSearchSettings = jest.spyOn(
   SearchSettingsModule,
@@ -107,5 +107,52 @@ describe("<SearchPage/>", () => {
       await searchPromise;
     });
     expect(component.find(CircularProgress)).toHaveLength(0);
+  });
+  it("should debounce append an * when not in raw search mode", async () => {
+    await act(async () => {
+      const input = component.find("input.MuiInputBase-input");
+      await input.simulate("change", { target: { value: "new query" } });
+      await jest.advanceTimersByTime(1000);
+      expect(searchPromise).toHaveBeenLastCalledWith({
+        currentPage: 0,
+        rowsPerPage: 10,
+        sortOrder: "RANK",
+        query: "new query*",
+      });
+    });
+  });
+  it("should not debounce when in raw search mode", async () => {
+    await act(async () => {
+      const input = component.find("input.MuiInputBase-input");
+      const rawModeSwitch = component.find("input[type='checkbox']");
+
+      //turn raw search mode on, add a search query and hit enter
+      await rawModeSwitch.simulate("change", { target: { checked: true } });
+      await input.simulate("change", { target: { value: "raw search test" } });
+      await input.simulate("keyDown", {
+        keyCode: 13,
+      });
+      await jest.advanceTimersByTime(1000);
+
+      //assert that the simple search wildcard was not appended
+      expect(searchPromise).not.toHaveBeenCalledWith({
+        currentPage: 0,
+        rowsPerPage: 10,
+        sortOrder: "RANK",
+        query: "raw search test*",
+      });
+      //assert that the query was passed in as-is
+      expect(searchPromise).toHaveBeenLastCalledWith({
+        currentPage: 0,
+        rowsPerPage: 10,
+        sortOrder: "RANK",
+        query: "raw search test",
+      });
+      /*assert that there were only two calls - initial search and Enter key triggered.
+      if the debounce was still on, this would be three calls -
+      the initial search, the search triggered by the debounce,
+      and the search triggered by the Enter key.*/
+      expect(searchPromise).toHaveBeenCalledTimes(2);
+    });
   });
 });
--- a/Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
@@ -95,6 +95,32 @@ describe("<SearchPage/>", () => {
     });
   };
 
+  /**
+   * Do a raw query search with fake timer. Turns on raw search mode, enters a search and hits enter.
+   * Waits for the debounce after the enter key.
+   * @param searchTerm The specified search term.
+   */
+  const rawQuerySearch = async (searchTerm: string) => {
+    jest.useFakeTimers("modern");
+    const input = component.find(SEARCHBAR_ID);
+    const rawModeSwitch = component.find(RAW_SEARCH_TOGGLE_ID);
+    //turn raw search mode on
+    await awaitAct(() =>
+      rawModeSwitch.simulate("change", { target: { checked: true } })
+    );
+    //add the searchTerm
+    await awaitAct(() => {
+      input.simulate("change", { target: { value: searchTerm } });
+    });
+    //Hit Enter and wait for debounce
+    await awaitAct(() => {
+      input.simulate("keyDown", {
+        keyCode: ENTER_KEYCODE,
+      });
+      jest.advanceTimersByTime(1000);
+    });
+  };
+
   it("should retrieve search settings and do a search when the page is opened", () => {
     expect(
       SearchSettingsModule.getSearchSettingsFromServer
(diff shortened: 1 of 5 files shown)


Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
```diff
@@ -177,4 +177,43 @@ describe("<SearchPage/>", () => {
...
+    await querySearch("new query");
+    expect(searchPromise).toHaveBeenLastCalledWith({
+      ...defaultSearchOptions,
+      query: "new query*",
+    });
+  });
+
+  it("should not debounce and send query as-is when in raw search mode", async () => {
+    const input = component.find("input.MuiInputBase-input");
+    const rawModeSwitch = component.find("input[type='checkbox']");
```
edalex-ian, line 191: I wonder if we should start putting IDs on these controls we wish to use in tests. We're doing this a bit, but it seems fragile. Further, we also have a repetitions now with with the string on line 190.

Could you perhaps add an ID for the raw mode switch, and push the duplicate string into a constant?

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
```diff
@@ -177,4 +177,43 @@ describe("<SearchPage/>", () => {
...
+    await awaitAct(() => {
+      //turn raw search mode on, add a search query and hit enter
+      rawModeSwitch.simulate("change", { target: { checked: true } });
+      jest.advanceTimersByTime(1000);
+      querySearch("raw search test");
+      input.simulate("keyDown", {
+        keyCode: 13,
```
edalex-ian, line 198: Again, good to put magic numbers into constants.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
```diff
@@ -177,4 +177,43 @@ describe("<SearchPage/>", () => {
...
+    //assert that the simple search wildcard was not appended
+    expect(searchPromise).not.toHaveBeenLastCalledWith({
+      ...defaultSearchOptions,
+      query: "raw search test*",
+    });
+    //assert that the query was passed in as-is
+    expect(searchPromise).toHaveBeenLastCalledWith({
+      ...defaultSearchOptions,
+      query: "raw search test",
+    });
```
edalex-ian, line 212: Being that these are mutually exclusive, we only need one.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchBar.tsx
```diff
@@ -37,30 +46,44 @@ interface SearchBarProps {
...
+      case ESCAPE_KEY_CODE:
+        event.preventDefault();
+        handleQueryChange("");
+        break;
+      case ENTER_KEY_CODE:
+        event.preventDefault();
+        if (rawSearchMode) {
+          onChange(query);
+        } else {
+          onChange(query + "*");
```
edalex-ian, line 71: It'd be good to have this small bit of repeated logic centralised - seeing it's rather key to expected functionality. If there was a function like:

```typescript
const callOnChange = (query: string) =>
  onChange(query + (rawSearchMode ? "" : "*"));
```

... then you could simply call that function in this `case` block, so:

```typescript
case ENTER_KEY_CODE:
   event.preventDefault();
  callOnChange(query);
```

... and obviously you'd use that same function up in your debounce call.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
```diff
@@ -177,4 +177,43 @@ describe("<SearchPage/>", () => {
...
     expect(component.find(CircularProgress)).toHaveLength(0);
   });
+
+  it("should debounce and append an * when not in raw search mode", async () => {
+    await querySearch("new query");
+    expect(searchPromise).toHaveBeenLastCalledWith({
+      ...defaultSearchOptions,
+      query: "new query*",
+    });
+  });
```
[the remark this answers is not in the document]
SammyIsConfused: Yup, so it does. Removed.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
SammyIsConfused: Done. I still need to refer to the tagname too as adding an ID to a MUI control makes some of it's wrappers share the ID.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
SammyIsConfused: Fixed

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
SammyIsConfused: Removed the .not call.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchBar.tsx
SammyIsConfused: Fixed

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
```diff
@@ -177,4 +177,43 @@ describe("<SearchPage/>", () => {
...
+
+  it("should not debounce and send query as-is when in raw search mode", async () => {
+    const input = component.find("input.MuiInputBase-input");
+    const rawModeSwitch = component.find("input[type='checkbox']");
+    await awaitAct(() => {
+      //turn raw search mode on, add a search query and hit enter
+      rawModeSwitch.simulate("change", { target: { checked: true } });
+      jest.advanceTimersByTime(1000);
```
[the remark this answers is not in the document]
SammyIsConfused: It was because `advanceTimersByTime` had been called before `useFakeTimers("modern")`. To fix this, I've simply put that line in the beforeEach and removed it from elsewhere in the test.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchBar.tsx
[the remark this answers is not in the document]
SammyIsConfused: In simple search mode, this is what happens.
Type a query, 500ms delay, onChange called
Press enter, onChange called
No call after this.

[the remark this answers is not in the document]
PenghaiZhang: `Type a query, 500ms delay, onChange called`  seems not working.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
```diff
@@ -56,6 +56,7 @@ describe("<SearchPage/>", () => {
   let component: ReactWrapper<any, Readonly<{}>, React.Component<{}, {}, any>>;
 
   beforeEach(async () => {
+    jest.useFakeTimers("modern");
```
PenghaiZhang, line 59: We don't really need this for each test. Only tests in relation to debounce search need it,

SammyIsConfused: Surely it makes sense to put it in just one place though? 
There will be more tests in here later on and they will need this too. And just having this option set doesn't negatively affect anything. It just means its set up for async testing if it comes up.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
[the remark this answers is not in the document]
PenghaiZhang: Sorry I still don't understand why we need fake timer for this one because debounce does not get involved in this use case. Once we turn on raw search, no search is triggered until `enter` is pressed.  

Please correct me if I am wrong.
What this test is supposed to do, in my mind, are: 
1. turn on the switch; and
2. change the query value; and
3. press `enter`; and
4. expect.

The third point should be wrap inside a `act` because it updates Search page state.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchBar.tsx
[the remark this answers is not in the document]
SammyIsConfused: It is? I just checked it.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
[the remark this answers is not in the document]
SammyIsConfused: Without the waits the switch and change to the query value seem to be too close together - we still get the debounce call. We need the wait to ensure that state has been update and percolated through after turning on the switch.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchBar.tsx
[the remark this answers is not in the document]
PenghaiZhang: No, I tried this when I thought `simple search ` = `raw search`.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchBar.tsx
```diff
@@ -37,30 +46,40 @@ interface SearchBarProps {
...
 
+  const handleKeyDown = (event: React.KeyboardEvent<HTMLDivElement>) => {
+    switch (event.keyCode) {
+      case ESCAPE_KEY_CODE:
+        event.preventDefault();
+        handleQueryChange("");
+        break;
+      case ENTER_KEY_CODE:
+        event.preventDefault();
+        callOnChange(query);
```
edalex-ian, line 67: Thanks for the tidy-up. Makes it nice and clear. :+1: 

But I do now wonder if this should simply be calling `delayedQuery()` so as to avoid someone typing and then pressing enter resulting in two queries to the server. As each search results in an entry into the Audit Log in the DB, pressing enter could just spam the audit log.

Plus if this is then done, things can be further changed, as the body of `callOnChange` could just be placed in `delayedQuery`.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchBar.tsx
```diff
@@ -37,30 +46,40 @@ interface SearchBarProps {
...
+  const callOnChange = (query: string) =>
+    onChange(query + (rawSearchMode ? "" : "*"));
   /**
    * uses lodash to debounce the search query by half a second
    */
-  const delayedQuery = useCallback(
-    debounce((query: string) => onChange(query), 500),
-    [onChange]
-  );
+  const delayedQuery = useCallback(debounce(callOnChange, 500), [onChange]);
```
edalex-ian, line 57: As an aside, this name should probably be `debouncedQuery`. There is no delay.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchBar.tsx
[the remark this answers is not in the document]
SammyIsConfused: Ahh I getcha. Sorry for the confusion. 
Raw search on = raw search.
Raw search off = simple search.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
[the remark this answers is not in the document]
PenghaiZhang: I have played with Ian this test and we think we can refactor this test a bit.
We just need to wait for these three actions without the use of fake timer.
    await awaitAct( () => rawModeSwitch.simulate("change", { target: { checked: true } }) );
    await awaitAct( () => input.simulate("change", { target: { value: "raw search test" } }));
    await awaitAct( () => input.simulate("keyDown", {
      keyCode: ENTER_KEYCODE,
    }));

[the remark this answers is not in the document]
SammyIsConfused: Oooh nice work that looks really good in comparison. I just tested it and it works great. Pushed.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
PenghaiZhang: I am not sure if we would have more features requiring debounce. If no, I suggest we put `jest.useFakeTimers("modern")` back to `querySearch` so the fake timer is available for tests that check the debounce query search.

SammyIsConfused: Yep, I agree. My previous comment was based upon the assumption we needed them for non-debounce - like the test that was fixed in 99640bc.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchBar.tsx
SammyIsConfused: Fixed

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchBar.tsx
SammyIsConfused: I like that idea, and it works too. I've fixed that up and updated the test.

edalex-ian:
Sorry @SammyIsConfused . Could you please make one more cosmetic change.

Could we please have the new control in the same grid row as the search bar. It feels that'd be more visually balanced, remove wasted white-space, and in some ways a more common approach.

PR #2048, opened 2020-07-14
Add thumbnail styles to the placeholder thumbnail

SammyIsConfused:
##### Checklist

##### Description of change
image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResult.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResult.tsx
@@ -128,7 +128,7 @@ export default function SearchResult({
     } else {
       result = (
         <Subject
-          className={`${classes.placeholderThumbnail} ${classes.thumbnail}`}
+          className={`MuiPaper-elevation1 MuiPaper-rounded ${classes.placeholderThumbnail} ${classes.thumbnail}`}
           fontSize="large"
         />
       );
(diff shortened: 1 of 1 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResult.tsx
```diff
@@ -60,6 +60,9 @@ const useStyles = makeStyles((theme: Theme) => {
     },
     placeholderThumbnail: {
       color: theme.palette.text.secondary,
+      marginRight: theme.spacing(2),
+      width: "88px",
+      height: "auto",
```
edalex-ian, line 65: Rather than duplicating, why wouldn't we just put both `thumbnail` and `placeholderThumbnail` classes on the placeholder?

SammyIsConfused: Fixed! That makes much more sense.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResult.tsx
```diff
@@ -128,7 +128,7 @@ export default function SearchResult({
     } else {
       result = (
         <Subject
-          className={`${classes.placeholderThumbnail} ${classes.thumbnail}`}
+          className={`MuiPaper-elevation1 MuiPaper-rounded ${classes.placeholderThumbnail} ${classes.thumbnail}`}
```
PenghaiZhang, line 131: Just a query: Is it common to use MUI specific styles in `className`?

SammyIsConfused: This was the only way we were able to get it to inherit those styles without actually using Paper. And since Paper doesn't look right in this scenario, its a special case.

PenghaiZhang: I see. Thanks for explanation!

PR #2055, opened 2020-07-15
Trim the search query and only append wildcard to non-empty queries

SammyIsConfused:
##### Checklist

##### Description of change
WIldcard for simple search was getting incorrectly appended to empty search queries. This fix trims the query before sending it and only appends the wildcard for non-empty searches.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchBar.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchBar.tsx
@@ -53,8 +53,10 @@ export default function SearchBar({ onChange }: SearchBarProps) {
   const [rawSearchMode, setRawSearchMode] = useState<boolean>(false);
   const [queryString, setQuery] = React.useState<string>("");
   const strings = languageStrings.searchpage;
-  const callOnChange = (query: string) =>
-    onChange(query + (rawSearchMode ? "" : "*"));
+  const callOnChange = (query: string) => {
+    const trimmedQuery = query.trim();
+    onChange(trimmedQuery + (rawSearchMode || !trimmedQuery ? "" : "*"));
+  };
   /**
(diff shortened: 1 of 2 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchBar.tsx
```diff
@@ -53,8 +53,10 @@ export default function SearchBar({ onChange }: SearchBarProps) {
   const [rawSearchMode, setRawSearchMode] = useState<boolean>(false);
   const [queryString, setQuery] = React.useState<string>("");
   const strings = languageStrings.searchpage;
-  const callOnChange = (query: string) =>
-    onChange(query + (rawSearchMode ? "" : "*"));
+  const callOnChange = (query: string) => {
+    const trimmedQuery = query.trim();
+    onChange(trimmedQuery + (rawSearchMode || !trimmedQuery ? "" : "*"));
```
edalex-ian, line 58: The readability here is suffering. And in trying to untangle it, it led me to a more basic question. If there is no query after trimming, then should we call onChange at all? If not, that may indeed enable simplifying this logic quite readily.

SammyIsConfused: Yes, I think it should. 
For example, you have a search query "search 1" and you press the X button to clear the query. OnChange gets called, trimmedQuery would be "". It still needs calling so that you get results for no query rather than "search1".

edalex-ian: Cool, in that case could you please break it out a bit and improve the readability?

SammyIsConfused: Sure thing. That should be done now.

PR #2074, opened 2020-07-19
Remember search options when going 'back' to the search page

edalex-ian:
##### Checklist

- [x] tests are included

##### Description of change

The main feature which was being implemented here was to remember what options a user has selected on a search page when they use the back button to return to a search. But also snuck in a quick couple of others (reduced prominence of raw mode toggle; and remove need to press enter when in raw search mode) following team review.

This became a bit more involved, as it was realised `SearchBar` was handling the state around raw mode itself - rather than pushing state up. As a result, `SearchBar` was refactored to use two way binding to support pushing this state up allowing it to be stored in history. However `SearchBar` does still maintain it's own internal state of the current query value to support debouncing of sending the new query value up.

In addition to state, it was also noted that `SearchBar` had taken on the business logic around what raw mode meant. This then was also pushed out back to the Module layer, which in turn then also meant the test should be moved - and others added.

(A also swapped the debounce used by the `SearchBar` to the one included in Material-UI to reduce dependency on lodash.)

Of note, is that with the two way binding being used for the new Collections control, it automatically had those options stored in history. Therefore, if this pattern is observed going forward, the additional settings will also readily be persisted for 'on back'.

A couple of screenshots to show the change in the raw toggle:

image

image

#1306 

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchModule.test.ts
+++ b/Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchModule.test.ts
@@ -20,7 +20,7 @@ import * as SearchModule from "../../../tsrc/search/SearchModule";
 import { getSearchResult } from "../../../__mocks__/getSearchResult";
 
 jest.mock("@openequella/rest-api-client");
-(OEQ.Search.search as jest.Mock<
+const mockedSearch = (OEQ.Search.search as jest.Mock<
   Promise<OEQ.Common.PagedResult<OEQ.Search.SearchResultItem>>
 >).mockResolvedValue(getSearchResult);
 
@@ -31,5 +31,41 @@ describe("SearchModule", () => {
     );
     expect(searchResult.available).toBe(12);
     expect(searchResult.results).toHaveLength(12);
+  });
+
+  const validateSearchQuery = (expectedQuery: string) => {
+    const calls = mockedSearch.mock.calls;
+    const params = calls[0][1]; // Second parameter of the call is the 'params'
+    expect(params.query).toEqual(expectedQuery);
+  };
+
+  it("should not append a wildcard for a search which is empty when trimmed", async () => {
+    mockedSearch.mockReset();
+    await SearchModule.searchItems({
+      ...SearchModule.defaultSearchOptions,
+      query: "   ",
+    });
+    validateSearchQuery("");
+  });
+
+  it("should append a wildcard for a search non-rawMode, non-empty query", async () => {
+    mockedSearch.mockReset();
+    const queryTerm = "non RAW";
+    await SearchModule.searchItems({
+      ...SearchModule.defaultSearchOptions,
+      query: queryTerm,
+    });
+    validateSearchQuery(`${queryTerm}*`);
+  });
+
+  it("should NOT append a wildcard for a rawMode search with a non-empty query", async () => {
+    mockedSearch.mockReset();
+    const queryTerm = "RAW mode!";
+    await SearchModule.searchItems({
+      ...SearchModule.defaultSearchOptions,
+      query: queryTerm,
+      rawMode: true,
+    });
+    validateSearchQuery(`${queryTerm}`);
   });
 });
--- a/Source/Plugins/Core/com.equella.core/js/__stories__/components/SearchBar.stories.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/__stories__/components/SearchBar.stories.tsx
@@ -24,6 +24,17 @@ export default {
   component: SearchBar,
 };
 
-export const BasicSearchPage = () => (
-  <SearchBar onChange={action("OnChange called")} />
-);
+const actions = {
+  onQueryChange: action("onQueryChange called"),
+  onRawModeChange: action("onRawModeChange called"),
+  doSearch: action("doSearch called"),
+};
+
+const values = {
+  query: "",
+  rawMode: false,
+};
+
+export const NonRawMode = () => <SearchBar {...actions} {...values} />;
+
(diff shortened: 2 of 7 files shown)


edalex-ian:
Ah drats! Looks like I have a Storybook story to fix up. Apologies, didn't realise it was there. Need to come up with a way to readily check that. :thinking:

edalex-ian:
Alright, I've updated the `SearchBar` story, now hopefully all is okay.

(But I did note unrelated errors in Storybook, so I'll fix them separately if this branch passes the checks.)

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchModule.test.ts
```diff
@@ -32,4 +32,40 @@ describe("SearchModule", () => {
...
+    expect(params.query).toEqual(expectedQuery);
+  };
+
+  it("should not append a wildcard for a search which is empty when trimmed", async () => {
+    mockedSearch.mockReset();
+    await SearchModule.searchItems({
+      ...SearchModule.defaultSearchOptions,
+      query: "   ",
+    });
+    validateSearchQuery("");
```
PenghaiZhang, line 48: May I ask why not use `toHaveBeenCalledWith({...defaultSearchOptions, query: ""})` ?

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchModule.test.ts
```diff
@@ -32,4 +32,40 @@ describe("SearchModule", () => {
     expect(searchResult.available).toBe(12);
     expect(searchResult.results).toHaveLength(12);
   });
+
+  const validateSearchQuery = (expectedQuery: string) => {
+    const calls = mockedSearch.mock.calls;
+    const params = calls[0][1]; // Second parameter of the call is the 'params'
```
PenghaiZhang, line 38: I am not familiar with `mock.calls`.  May I ask if 0 means this is the first call ? And what does 1 stand for ?

PenghaiZhang: OK now I know `calls[0][1]` means `The second arg of the first call`

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchModule.test.ts
edalex-ian: You may. :)

It's because under the hood in the search module, we then go and add additional values to make up the search params. And so I'm actually validating against the search params which includes things like statuses.

Indeed, I initially did the exact thing you mention there - and was confused where all these `LIVE`, `REVIEW` etc ones were coming from... And then it clicked. See `statuses` at https://github.com/openequella/openEQUELLA/blob/component/new-search-page/Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchModule.ts#L54

(As an aside, I do wonder if some point we'll want to support filtering on those. :thinking: )

edalex-ian:
Well that GHA failure is wierd. It's as though it didn't pick up the updated `SearchBar.stories.tsx`. I'll kick off the GHA again.

edalex-ian:
Okay, I'm at a loss for this. GHA is still failing check with:

```
Source/Plugins/Core/com.equella.core/js/__stories__/components/SearchBar.stories.tsx(28,14): error TS2322: Type '{ onChange: HandlerFunction; }' is not assignable to type 'IntrinsicAttributes & SearchBarProps'.
  Property 'onChange' does not exist on type 'IntrinsicAttributes & SearchBarProps'.
```

Even though that line no longer talks about that, as it was removed/fixed in 774c7ea. It's as though the GHA build is stuck on the old code for some reason. :thinking:

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchModule.test.ts
PenghaiZhang: Ah I see! Thanks for explaining!

edalex-ian:
Okay, confirmed. Rather than using 774c7ea it's using 2e5c6c6: 

image

I may need to just proceed with merge based on CodeBuild being happy. :disappointed:

PenghaiZhang:
> Okay, confirmed. Rather than using [774c7ea](https://github.com/openequella/openEQUELLA/commit/774c7ea8db4c46180a81aefb375568302d841f84) it's using [2e5c6c6](https://github.com/openequella/openEQUELLA/commit/2e5c6c623ad21ee90d6b33a20c424f580fff7305):
> 
> image
> 
> I may need to just proceed with merge based on CodeBuild being happy.

I agree.

PenghaiZhang:
> Okay, I'm at a loss for this. GHA is still failing check with:
> 
> ```
> Source/Plugins/Core/com.equella.core/js/__stories__/components/SearchBar.stories.tsx(28,14): error TS2322: Type '{ onChange: HandlerFunction; }' is not assignable to type 'IntrinsicAttributes & SearchBarProps'.
>   Property 'onChange' does not exist on type 'IntrinsicAttributes & SearchBarProps'.
> ```
> 
> Even though that line no longer talks about that, as it was removed/fixed in [774c7ea](https://github.com/openequella/openEQUELLA/commit/774c7ea8db4c46180a81aefb375568302d841f84). It's as though the GHA build is stuck on the old code for some reason.

This is strange. Even retry did not help.

edalex-ian:
Yup, it's wack. And I've just confirmed CodeBuild has indeed built the right commit:

image

I'll merge now.

PR #2076, opened 2020-07-20
Add a new npm build task to run checks in core npm build

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

Our npm check tasks are defined in the root folder `package.json`. As a result,  we are required to create a CI job on both GHA and Codebuild to get these tasking running . This PR attempts to integrate these tasks into the `core npm build` so each CI will run them without  an extra job.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/package.json
+++ b/Source/Plugins/Core/com.equella.core/js/package.json
@@ -18,7 +18,7 @@
     "dev": "pulp build && cross-env-shell parcel watch --no-autoinstall --public-url=. --out-dir=${npm_package_config_devjs} entrydev/*.html entrydev/scripts/*.js",
     "dev:build": "run-s bundleps && cross-env-shell parcel build --public-url=. --out-dir=${npm_package_config_devjs} entrybuild/*.html entrybuild/scripts/*.js",
     "build": "npm-run-all build:*",
-    "build:check": "cd ../../../../../ && npm run check",
+    "build:check": "cd ../../../../../ && npm run check:license && npm run check:ts && npm run check:ts-types-source",
     "build:bundles": "run-s bundleps && cross-env-shell parcel build --public-url=. --out-dir=${npm_package_config_dist} entrybuild/*.html entrybuild/scripts/*.js",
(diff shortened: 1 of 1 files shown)


PenghaiZhang:
I can confirm ts checks have been undertaken on Codebuild. Also, I think we no longer need to run `npm run check` on GHA.

PenghaiZhang:
> Looks alright,
> will this have any impact on CI? Causing any check duplication?
> 
> edit: it looks like it may
> 
> > I think we no longer need to run npm run check on GHA.
> 
> do we want to remove the duplication in CI as part of this PR?

I think we want.

Source/Plugins/Core/com.equella.core/js/package.json
```diff
@@ -18,6 +18,7 @@
     "dev": "pulp build && cross-env-shell parcel watch --no-autoinstall --public-url=. --out-dir=${npm_package_config_devjs} entrydev/*.html entrydev/scripts/*.js",
     "dev:build": "run-s bundleps && cross-env-shell parcel build --public-url=. --out-dir=${npm_package_config_devjs} entrybuild/*.html entrybuild/scripts/*.js",
     "build": "npm-run-all build:*",
+    "build:check": "cd ../../../../../ && npm install && npm run check:license && npm run check:ts && npm run check:ts-types-source",
```
edalex-ian, line 21: Hmmm, a few things:

1. We should probably wrap this in `cross-env-shell` to support the Windows peeps; and
2. We should really use `install ci` up top - so that we have consistent results, and no accidential package-lock.json churn; but
3. Seeing those check tasks are primarily targetting down here, should we simply move their logic etc down into here and change the top level package.json to simply call these ones? (In that fashion, this package.json becomes more self contained - which will help in the one day hope we split it out.)

PenghaiZhang:
After discussing, we determined to implement this feature through the easiest way - adding one more command on `buildspec.yml`. The reason is properly refactoring both root and core npm tasks require more efforts and at the moment we have higher priority work to do. We may discuss this in next CDM.

PR #2087, opened 2020-07-22
Refine control: Last modified date selector

PenghaiZhang:
#1306 

##### Checklist

- [x] tests are included

##### Description of change

Created another refine control: Last modified date selector. This control shows a Dropdown of Quick date options or two custom Date pickers, depending on the value of the Quick mode switch which is also included in this control. Also, this control is compatible with the feature of `back to Search page`. 

`Luxon` is used to deal with date formats and calculations,. 

In Search module, added some utils to help transform data between a date range and a Quick date option. However, I wonder if we should move these utils to a new module dedicated for `Last modified date`.

Lastly, added a new Storybook for this control.

One more side thing is during developing I found the dependency of `material-ui picker` we are using is very old. I tried to update it to the latest version, which resulted in some errors in other pages due to significant changes. So I did not proceed.

image

image


PenghaiZhang:
I also noted that in the screenshot for two Date pickers, the space between them is quite large. This is because each of them is wrapped in a MUI Grid item and the Grid takes full width of Storybook's Canvas area. So when it comes to the Search page refine panel,  the space will look fine.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/ModifiedDateSelector.tsx
```diff
@@ -0,0 +1,155 @@
...
+    </FormControl>
+  );
+
+  const customDateRange: ReactNode = (
+    <Grid container>
+      <Grid item xs={6}>
+        <InlineDatePicker
+          disableFuture
+          variant="outlined"
+          format="dd/MM/yyyy"
```
edalex-ian, line 111: A couple of things:

* These props are duplicates - maybe put them in an object and spread them; and
* Should `format` be using the locale format or if internal than ISO? (I'm not sure what this prop controls.)

PenghaiZhang:
Updates for Storybook screenshots.
image

image

PenghaiZhang:
Here are two updated storybook screenshots.  I removed some unnecessary knobs.

image

image

PR #2089, opened 2020-07-22
Make window scroll to top when search is performed

mrblippy:
##### Checklist

##### Description of change

search scroll

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
@@ -101,7 +101,7 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
       .then((items: OEQ.Common.PagedResult<OEQ.Search.SearchResultItem>) => {
         setPagedSearchResult(items);
         history.replace({ ...history.location, state: searchOptions });
-        //scroll back up to the top of the page
+        // scroll back up to the top of the page
         window.scrollTo(0, 0);
       })
       .catch(handleError)
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
@@ -101,6 +101,8 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
       .then((items: OEQ.Common.PagedResult<OEQ.Search.SearchResultItem>) => {
         setPagedSearchResult(items);
         history.replace({ ...history.location, state: searchOptions });
+        //scroll back up to the top of the page
+        window.scrollTo(0, 0);
       })
       .catch(handleError)
       .finally(() => setShowSpinner(false));


Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -101,6 +101,8 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
       .then((items: OEQ.Common.PagedResult<OEQ.Search.SearchResultItem>) => {
         setPagedSearchResult(items);
         history.replace({ ...history.location, state: searchOptions });
+        //scroll back up to the top of the page
+        window.scrollTo(0, 0);
```
PenghaiZhang, line 105: This change means changing `rows per page` also scrolls window to top. Is that what we want? I feel it's better not to scroll to top in this case.

edalex-ian: Well that's an interesting thought.

:thinking: 

But on thinking it through, I came to this scenario: I'm on page three of my results, and now I've changed the page size from 10 to 50. So previously I was looking at items 21-30, but now I'm looking at items 101-150.

So I wonder... Should we actually then go back to the first page and the top of the screen?

PenghaiZhang: I believe ` go back to the first page after changing page size ` is what we have now.

edalex-ian: Well in that case, I'd say it definitely seems to make sense to go back to the top, as everything you're looking at has changed.

But you didn't really say _why_ you felt it was better not to go to the top. Could you elaborate please?

PenghaiZhang: With this change, if I stay in page 1 and I change the page size from 10 to 50, I will need to scroll down again to view item 11.

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -101,6 +101,8 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
       .then((items: OEQ.Common.PagedResult<OEQ.Search.SearchResultItem>) => {
         setPagedSearchResult(items);
         history.replace({ ...history.location, state: searchOptions });
+        //scroll back up to the top of the page
```
ChristianMurphy, line 104:
```suggestion
        // scroll back up to the top of the page
```

cathfitz: I don't feel strongly either way, however I do agree with Penghai that it's annoying when you're browsing a list and you change the no. of results, you have to find where you were up to.

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
edalex-ian: Okay, so how about we conditionally do it in that scenario? e.g. If changing page size and NOT on page 1, then scrollTop?

cathfitz: I don't feel strongly either way, however I do agree with Penghai that it's annoying when you're browsing a list and you change the no. of results, you have to find where you were up to.

PenghaiZhang: I am happy to get this merged in now because doing that conditionally would bring in more complexity.

edalex-ian: Cool, let's press on then and see how it goes.

PR #2093, opened 2020-07-23
Remove dead code / redux

edalex-ian:
##### Checklist

- [ ] tests are included

##### Description of change

We've be holding onto a growing collection of code from an earlier approach of the data access layer for the New UI which also used Redux. We've gone down a different path, so removing this no longer used code. (And indeed, some of this code is now duplicated/overlapping with what's in the REST API module - so some more work to do there too.)

There are possibly still some other bits of dead code (maybe in `utils` and `components`) but they can be addressed later.

Also tweaked the `dev:build` tasks so that it does a `tsc --noEmit` to validate the Typescript types. Definitely helps with these kinds of activities to always run it, plus assists in general day to day dev work if you're doing a dev build each time.

Undertook local building and running of jest tests, as well as some manually smoke testing of the UI. Hopefully all is still fine.


PenghaiZhang:
> ##### Checklist
> * [x]  the [contributor license agreement](https://www.apereo.org/node/676) is signed
> * [x]  commit message follows [commit guidelines](https://chris.beams.io/posts/git-commit/)
> * [ ]  tests are included
> 
> ##### Description of change
> We've be holding onto a growing collection of code from an earlier approach of the data access layer for the New UI which also used Redux. We've gone down a different path, so removing this no longer used code. (And indeed, some of this code is now duplicated/overlapping with what's in the REST API module - so some more work to do there too.)
> 
> There are possibly still some other bits of dead code (maybe in `utils` and `components`) but they can be addressed later.
> 
> Also tweaked the `dev:build` tasks so that it does a `tsc --noEmit` to validate the Typescript types. Definitely helps with these kinds of activities to always run it, plus assists in general day to day dev work if you're doing a dev build each time.
> 
> Undertook local building and running of jest tests, as well as some manually smoke testing of the UI. Hopefully all is still fine.

Yes, some components can be deleted, such as `AppBarQuery` .

PR #2108, opened 2020-07-27
Feature/pagination actions

mrblippy:
##### Checklist

- [x] tests are included

##### Description of change

This PR adds in additional Paging options to the mui TablePagination component, allowing for navigating to the first and last page of search results

paginations


Source/Plugins/Core/com.equella.core/js/__mocks__/getSearchResult.js
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck
```
SammyIsConfused, line 1: Is this necessary? What caused this to be added?

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
```diff
@@ -150,7 +151,11 @@ describe("<SearchPage/>", () => {
 
   it("should support changing the number of items displayed per page", async () => {
     // Initial items per page is 10
-    expect(component.html()).toContain("1-10 of 12");
+    const pageCount = component
+      .find(".MuiTablePagination-toolbar")
```
SammyIsConfused, line 155: This .find chain seems very fragile. Is there a better way to find it with IDs?

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
```diff
@@ -161,20 +166,52 @@ describe("<SearchPage/>", () => {
...
+  });
+
+  it("should support navigating to first/last page of results", async () => {
+    mockSearch.mockImplementation(() =>
+      Promise.resolve(getSearchResultsCustom(30))
+    );
+    await querySearch("");
+    component.update();
+    const firstPageButton = component
+      .find(".MuiTablePagination-actions button")
```
SammyIsConfused, line 200: This string is used multiple times. Should be a constant.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
```diff
@@ -161,20 +166,52 @@ describe("<SearchPage/>", () => {
       ...defaultSearchOptions,
       rowsPerPage: 25,
     });
-    expect(component.html()).toContain("1-12 of 12");
+    expect(pageCount.text()).toContain("1-12 of 12");
   });
 
   it("should support navigating to previous/next page", async () => {
+    await querySearch("");
+    component.update();
```
SammyIsConfused, line 174: I'm not sure the component update is necessary here. The querysearch already has a wait.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
```diff
@@ -161,20 +166,52 @@ describe("<SearchPage/>", () => {
...
+      .at(2);
+    const pageCount = component
+      .find(".MuiTablePagination-toolbar")
+      .find("p")
       .at(1);
     await awaitAct(() => nextPageButton.simulate("click"));
-    expect(component.html()).toContain("11-12 of 12");
+    expect(pageCount.text()).toContain("11-12 of 12");
+    await querySearch("");
+    component.update();
```
SammyIsConfused, line 188: As above.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
```diff
@@ -161,20 +166,52 @@ describe("<SearchPage/>", () => {
...
-    expect(component.html()).toContain("1-10 of 12");
+    expect(pageCount.text()).toContain("1-10 of 12");
+  });
+
+  it("should support navigating to first/last page of results", async () => {
+    mockSearch.mockImplementation(() =>
+      Promise.resolve(getSearchResultsCustom(30))
+    );
+    await querySearch("");
+    component.update();
```
SammyIsConfused, line 198: As above.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
```diff
@@ -161,20 +166,52 @@ describe("<SearchPage/>", () => {
...
+      .find(".MuiTablePagination-actions button")
+      .at(3);
+    const pageCount = component
+      .find(".MuiTablePagination-toolbar")
+      .find("p")
+      .at(1);
+    await awaitAct(() => lastPageButton.simulate("click"));
+    expect(pageCount.text()).toContain("21-30 of 30");
+    await querySearch("");
+    component.update();
```
SammyIsConfused, line 212: As above.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchPagination.tsx
```diff
@@ -46,6 +59,68 @@ export const SearchPagination = ({
...
+            onClick={() => onPageChange(currentPage + 1)}
+            aria-label={paginationStrings.nextPageButton}
+            disabled={isLastPage}
+          >
+            <KeyboardArrowRight />
+          </IconButton>
+        </Grid>
+        <Grid item>
+          <IconButton
+            onClick={() => onPageChange(Math.max(0, numberOfPages() - 1))}
```
SammyIsConfused, line 116: Similar to the consts on 76 and 77, could these values passed into onPageChange be named? 
Eg
const lastPage = Math.max(0, numberOfPages() -1;
......
onClick = {() => onPageChange(lastPage))}

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchPagination.tsx
```diff
@@ -46,6 +59,68 @@ export const SearchPagination = ({
...
+      ActionsComponent={PaginationActions}
     />
   );
+
+  /**
+   * Provides pagination navigation controls.
+   * @return {ReactElement} IconButtons for navigating to First, Previous, Next, and Last Page of results
+   */
+  function PaginationActions() {
+    const theme = createMuiTheme();
```
ChristianMurphy, line 71: other parts of the code leverage `makeStyles`, would it make sense to use that here as well?

PenghaiZhang:
The Search page test is getting more and more test cases and mock data in. Almost all of tests are reliant on `simulate` which seems to be [deprecated](https://github.com/enzymejs/enzyme/issues/2173). So I am wondering we need to consider another approach...

Source/Plugins/Core/com.equella.core/js/__mocks__/getSearchResult.js
mrblippy: If that's not added Typescript gives type errors for the getSearchResultsCustom arrow function, (even though it's a .js file)

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
mrblippy: It doesn't appear to be straightforward way to give that element an ID. We are going to revisit this SearchPage test in the future. So this code isn't likely to stay the same for very long

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
mrblippy: That's what I thought too. But this is the only way I could get it to work

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
```diff
@@ -161,20 +166,52 @@ describe("<SearchPage/>", () => {
...
-    expect(component.html()).toContain("11-12 of 12");
+    expect(pageCount.text()).toContain("11-12 of 12");
+    await querySearch("");
+    component.update();
     await awaitAct(() => prevPageButton.simulate("click"));
-    expect(component.html()).toContain("1-10 of 12");
+    expect(pageCount.text()).toContain("1-10 of 12");
+  });
+
+  it("should support navigating to first/last page of results", async () => {
```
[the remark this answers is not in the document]
mrblippy: Yeah, this test isn't as exhaustive as it could be. In the interests of time. I'm leaving this as-is for now. The SerachPage test likely needs to be rewritten to support the paging. But we can revisit in thre future

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchPagination.tsx
mrblippy: I didn't include makeStyles in this case, because there was only one single inline style, but i needed to get the spacing unit for that inline style

mrblippy:
> The Search page test is getting more and more test cases and mock data in. Almost all of tests are reliant on `simulate` which seems to be [deprecated](https://github.com/enzymejs/enzyme/issues/2173). So I am wondering we need to consider another approach...

Yes. I think this test needs an overhaul in general. But in the interests of time we should revisit this in the future

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
SammyIsConfused: You should be about to do `".MuiTablePagination-toolbar > p"` in the first `.find()` rather than two finds. That returns all `<p/>` tags that are direct children of `.MuiTablePagination-toolbar` (unless it isn't a direct child in which case your solution is better).

Source/Plugins/Core/com.equella.core/js/__mocks__/getSearchResult.js
```diff
@@ -285,3 +287,30 @@ exports.getSearchResult = {
...
+    collectionId: "9a1ddb24-6bf5-db3d-d8fe-4fca20ecf69c",
+    commentCount: 0,
+    attachments: [],
+    thumbnail: "initial",
+    displayFields: [],
+    links: {
+      view:
+        "http://localhost:8080/rest/items/266bb0ff-a730-4658-aec0-c68bbefc227c/1/",
+      self:
+        "http://localhost:8080/rest/api/item/266bb0ff-a730-4658-aec0-c68bbefc227c/1/",
```
edalex-ian, line 313: I'm must've missed this before. But.... Should this URL use the UUID being used for this item? Or maybe these URLs aren't really relevant in this context. :thinking:

Source/Plugins/Core/com.equella.core/js/__mocks__/getSearchResult.js
```diff
@@ -285,3 +287,30 @@ exports.getSearchResult = {
     },
   ],
 };
+
+const { v4: uuidv4 } = require("uuid");
+const _ = require("lodash");
+exports.getSearchResultsCustom = (numberOfResults) => ({
+  start: 0,
+  length: 10,
```
edalex-ian, line 295:
```suggestion
  length: numberOfResults,
```

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
```diff
@@ -161,20 +166,45 @@ describe("<SearchPage/>", () => {
...
+    );
+
+    await querySearch("");
+    component.update();
+    const firstPageButton = component.find("#firstPageButton button");
+    const lastPageButton = component.find("#lastPageButton button");
+    const pageCount = component
+      .find(".MuiTablePagination-toolbar")
+      .find("p")
+      .at(1);
```
edalex-ian, line 201: This is repeated from above. How about a function such as:

```typescript
const paginatorControls = (component: ReactWrapper) => [
  component.find("#firstPageButton button"),
  component.find("#lastPageButton button"),
  component
    .find(".MuiTablePagination-toolbar")
    .find("p")
    .at(1)
];
```

Then you could simply use it with

```typescript
const [firstPageButton, lastPageButton, pageCount] = paginatorControls(component);
```

Arguably, that function could sit over in `SearchPagination.test.tsx` and also be used over there. And now if that paginator changes later, we have all this handling in one nice place.

Source/Plugins/Core/com.equella.core/js/__mocks__/getSearchResult.js
mrblippy: At the moment they aren't relevant, or used at all in this context

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
mrblippy: That is indeed a good idea

Source/Plugins/Core/com.equella.core/js/__mocks__/getSearchResult.js
mrblippy: Also, the hardcoded results used inside the pre-existing getSearchResult function has the same issue

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
mrblippy: And is easier to read than what I had

mrblippy: I tried adding/exporting that function in SearchPagination.test

But eslint is yelling at me:
  22:1  error  Do not export from a test file  jest/no-export

I guess that function could just be in a regular .tsx file?

edalex-ian: Ah drats!

How about we create a plain ol' TS file alongside (no need for TSX). `SearchPaginationTestHelper.ts`? I wonder if the eslint rule will be cool with that. :thinking: 

As it happens, I'm hoping to do the same for the OwnerSelector so that I can also share back to SearchPage.test.tsx.

mrblippy: Yeah, I will give that a go

mrblippy: Yep, that worked

Source/Plugins/Core/com.equella.core/js/__mocks__/getSearchResult.js
edalex-ian: It seems you've done it to `available` rather than here on `length`. `available` from server represents total available on server, however `length` === `results.length`. So hence here `length` === `numberOfResults`.

Source/Plugins/Core/com.equella.core/js/__mocks__/getSearchResult.js
edalex-ian: Cool, fine to leave for now then.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/components/SearchPaginationTestHelper.ts
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { ReactWrapper } from "enzyme";
+
+export const paginatorControls = (component: ReactWrapper) => [
```
edalex-ian, line 20: Needs JSDoc.

Also, seeing how many there are now, this would probably be helped to have an interface and return an object. It also make it more readable where you're chosing to skip the first few and you're doing things like `[, , , nextPageButton] = ...`.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
edalex-ian: Did it work when you did a full `npm run test`?

Because when I just did similar locally that failed with Jest trying to run the utility class as a test. :'(

PR #2113, opened 2020-07-28
oeq-ts-rest-api: Add support for /userquery/search

edalex-ian:
##### Checklist

- [x] tests are included

##### Description of change

Primarily the change here was to add support for the `/userquery/search` endpoint in the form of `UserQuery.ts`. But then when I noticed the formatting on my commits was not correct I added that support to the `oeq-ts-rest-api` module by applying the final tweaks to`oeq-ts-rest-api/package.json` and `oeq-ts-rest-api/.eslintrc.js`. On running that, suddenly I found many other files were also a bit out of wack.

So unfortunately this PR now contains a lot more noise than I'd like. Here is a run down to the help:

Feature (ccf789a):

* `UserQuery.ts`
* `UserQuery.test.ts`
* Auxillary: `Common.ts`, `index.ts`, `TestConfig.ts`

**Ran the tests:**

```
Test Suites: 7 passed, 7 total
Tests:       23 passed, 23 total
Snapshots:   0 total
Time:        4.905 s
Ran all test suites.
```

### DIFF
--- a/oeq-ts-rest-api/.eslintrc.js
+++ b/oeq-ts-rest-api/.eslintrc.js
@@ -1,6 +1,7 @@
 module.exports = {
   root: true,
   env: {
+    node: true,
     browser: true,
     es6: true,
   },
@@ -10,6 +11,7 @@ module.exports = {
     'plugin:@typescript-eslint/recommended',
     'plugin:jest/recommended',
     'plugin:jest/style',
+    'plugin:prettier/recommended',
   ],
   globals: {
     Atomics: 'readonly',
--- a/oeq-ts-rest-api/src/Common.ts
+++ b/oeq-ts-rest-api/src/Common.ts
@@ -5,6 +5,8 @@ export type i18nString = string;
 
 export type I18nStrings = Record<string, string>;
 
+export type UuidString = string;
+
 export interface User {
   id: string;
   username?: string;
--- a/oeq-ts-rest-api/src/UserQuery.ts
+++ b/oeq-ts-rest-api/src/UserQuery.ts
@@ -1,6 +1,6 @@
 import { GET } from './AxiosInstance';
 import { is } from 'typescript-is';
-import {UuidString} from "./Common";
+import { UuidString } from './Common';
 
 export interface UserDetails {
   id: UuidString;
@@ -59,4 +59,8 @@ export const search = (
   apiBasePath: string,
   params: SearchParams
 ): Promise<SearchResult> =>
-  GET<SearchResult>( apiBasePath + USERQUERY_ROOT_PATH + '/search', isSearchResult, params);
+  GET<SearchResult>(
+    apiBasePath + USERQUERY_ROOT_PATH + '/search',
+    isSearchResult,
+    params
+  );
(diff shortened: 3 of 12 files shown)


oeq-ts-rest-api/src/Common.ts
```diff
@@ -5,6 +5,8 @@ export type i18nString = string;
 
 export type I18nStrings = Record<string, string>;
 
+export type UuidString = string;
```
ChristianMurphy, line 8: If uuid the string is different from other strings, would it make sense to make it on opaque type?
https://codemix.com/opaque-types-in-javascript
https://github.com/sindresorhus/type-fest

edalex-ian: Thanks Christian. I'm only using the `type` here for documentation purposes.

oeq-ts-rest-api/.eslintrc.js
```diff
@@ -1,6 +1,7 @@
 module.exports = {
   root: true,
   env: {
+    node: true,
```
ChristianMurphy, line 4: what parts of the node env are being used?

edalex-ian: The Jest tests are running in node mode.

But more so.... This was added due to an eslint error from the precommit hook about module does not exist. SO seemed to point to adding this line which indeed then solved the error and all was fine. :thinking: 

Odd, as when in the directory and simply doing `npm run format` all was fine, but the root setup was not happy until I did this.

ChristianMurphy: Interesting :thinking: 
Thanks for the context!

oeq-ts-rest-api/src/Common.ts
PenghaiZhang, line 8: Since this type is specific to `uuid`, should we update types of `uuid` for other objects later ? such as `uuid` of `BaseEntity` and `SearchResultItem`

oeq-ts-rest-api/src/UserQuery.ts
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Search for users and related entities (i.e. groups and roles).
+ *
+ * @param apiBasePath Base URI to the oEQ institution and API
+ * @param params Query parameters to customize result
+ */
+export const search = (
+  apiBasePath: string,
+  params: SearchParams
```
PenghaiZhang, line 60: Is `params` optional ?  If we want to get all users, groups and roles we probably don't need to pass params

oeq-ts-rest-api/src/Common.ts
edalex-ian: Yeah, I was also thinking that'll be something good to do - just didn't want to blur things too much in this PR. Will need to go back and confirm too which are UUIDs. (Guess that's another good argument for adding this type.)

oeq-ts-rest-api/src/UserQuery.ts
edalex-ian: We could, and probably get away with that... But I feel that'd be accidentally supported. It'd work for now, but if we changed things then people's code could easily break. It feels like here it'd just be better if people were deliberate. :thinking:

edalex-ian:
Cool. I'm going to merge even with that failed PackageAndNavigation test - as these changes have no impact on the core code at this point.

PR #2116, opened 2020-07-29
Improve Date range selector

PenghaiZhang:
#1306

##### Checklist

- [ ] tests are included

##### Description of change

Here are some improvements made for the Date range selector. However, the component becomes a lot more complicated accordingly. So I am still thinking if these improvements are worth to have. 
 
1. Used `KeyboardDatePicker` to allow type and clear a date. But in order to properly use it, `state` is required, which increases complexity.
2. Changed to default picker labels to `from` and `to`.
3. Added a util function to help convert a date object to a string in ISO format. The reason is Javascript native function `toISOString` changes the time to UTC time,  which may result in incorrect search results.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/components/DateRangeSelector.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/components/DateRangeSelector.tsx
@@ -131,8 +131,20 @@ export const DateRangeSelector = ({
   useEffect(() => {
     const start = stateDateRange?.start;
     const end = stateDateRange?.end;
-    // Call onDateRangeChange when the range is open-ended or when start <= end
-    if (!start || !end || start <= end) {
+    const isStartValid = start && DateTime.fromJSDate(start).isValid;
+    const isEndValid = end && DateTime.fromJSDate(end).isValid;
+
+    // start is undefined and end is a valid date
+    const openStart = !start && isEndValid;
+    // End is undefined and start is a valid date
+    const openEnd = !end && isStartValid;
+    // Both are undefined
+    const openRange = !start && !end;
+    // Both are valid dates and start is equal or less than end
+    const closedRange = isStartValid && isEndValid && start! <= end!;
+
+    // Call onDateRangeChange for above four cases
+    if (openStart || openEnd || openRange || closedRange) {
       onDateRangeChange(stateDateRange);
     }
   }, [stateDateRange]);
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/components/DateRangeSelector.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/components/DateRangeSelector.tsx
@@ -28,7 +28,7 @@ import {
   MuiPickersUtilsProvider,
 } from "@material-ui/pickers";
 import SettingsToggleSwitch from "./SettingsToggleSwitch";
-import { ReactNode, useEffect, useState } from "react";
+import { ReactNode, useEffect, useRef, useState } from "react";
 import { languageStrings } from "../util/langstrings";
 import LuxonUtils from "@date-io/luxon";
 import { DateTime } from "luxon";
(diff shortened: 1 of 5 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/components/DateRangeSelector.tsx
```diff
@@ -118,6 +120,38 @@ export const DateRangeSelector = ({
...
 
+  /**
+   * MUI KeyboardDatePicker requires its value to be controlled by state,
+   * in order to properly parse and validate TextField inputs.
+   */
+  const [stateDateRange, setStateDateRange] = useState<DateRange | undefined>(
+    dateRange
+  );
+
+  const isInitialRender = useRef(true);
```
PenghaiZhang, line 131: The initial date range always meets one of the below four conditions and therefore `onDateRangeChange` is called. This results in an extra render which should not be performed. So here I want  this `effect` to skip the first render.

Source/Plugins/Core/com.equella.core/js/tsrc/components/DateRangeSelector.tsx
```diff
@@ -118,6 +120,38 @@ export const DateRangeSelector = ({
...
+  useEffect(() => {
+    if (isInitialRender.current) {
+      isInitialRender.current = false;
+      return;
+    }
+    const start = stateDateRange?.start;
+    const end = stateDateRange?.end;
+    const isStartValid = start && DateTime.fromJSDate(start).isValid;
+    const isEndValid = end && DateTime.fromJSDate(end).isValid;
```
edalex-ian, line 140: How would either of these fail the `isValid` check? If they're not `undefined` then they're a `Date` and shouldn't they be valid?

Source/Plugins/Core/com.equella.core/js/tsrc/components/DateRangeSelector.tsx
```diff
@@ -163,18 +197,18 @@ export const DateRangeSelector = ({
    *
    * @param dateRange A date range to be converted to a Quick date option label.
    */
-  const dateRangeToDateOptionConverter = (dateRange?: DateRange): string => {
+  const dateRangeToDateOptionConverter = (): string => {
```
edalex-ian, line 200: I would've been inclined to leave this as was (taking a parameter) so that you have a nice simple pure stand-alone function (easier to maintain and reason out). But it's done now, so up to you.

Source/Plugins/Core/com.equella.core/js/tsrc/components/DateRangeSelector.tsx
```diff
@@ -224,36 +258,47 @@ export const DateRangeSelector = ({
...
+            variant="dialog"
+            clearable
+            inputVariant="outlined"
+            autoOk
+            // Show date in locale string, or nothing if date is null.
+            labelFunc={(date, _) => {
+              return date?.toLocaleString() ?? "";
+            }}
+            // TextField inputs are parsed to this format.
+            format="dd/MM/yyyy"
```
edalex-ian, line 285: I think here we should either:

a. Use the Locale again; or
b. Go with the more internationally acceptable ISO format.

Source/Plugins/Core/com.equella.core/js/tsrc/components/DateRangeSelector.tsx
PenghaiZhang: The Date picker returns a value whose type is `DateTime` rather than `Date`.  An object of `DateTime` could be an invalid date. For example, users can put in incomplete date '01/01/202` in the TextField.


image

Source/Plugins/Core/com.equella.core/js/tsrc/components/DateRangeSelector.tsx
PenghaiZhang: I have not found any approach to change the format depending on locale. If we go with ISO, should we give some help texts ?

Source/Plugins/Core/com.equella.core/js/tsrc/components/DateRangeSelector.tsx
edalex-ian: Aaaahhhh gotcha! I thought it was post that. :+1:

Source/Plugins/Core/com.equella.core/js/tsrc/components/DateRangeSelector.tsx
edalex-ian: Okay, as discussed, let's just go ISO on both sides and leave this for now.

PenghaiZhang:
Merged in although a flaky test `MyResourcesTest` failed.

Issue #1735, opened 2020-05-15
Broken resumption tokens


edalex-ian:
**Describe the bug**

Returned resumption tokens for endpoints like `GET /schema` and `GET /collection` are incorrectly formatted.

**To Reproduce**

1. Go to apidocs.do
2. Use the 'Try it out' feature on the `GET /schema` endpoint
3. Set a length of 2
4. Click 'execute'
5. See in the resulting response body the resumption token will be `4:2`

**Expected behavior**

For the above example, the first number should only be `2` and so the token should be `2:2`.

**Platform:**

- OpenEquella Version: 2020.1 (but I imagine all the way back to when this was added)

**Additional context**

Seems to have come about due to the change of this line:

https://github.com/openequella/openEQUELLA/commit/4caadc3d1de1af0bf6a75565bf1bb5208ec387c0#diff-f9b2c7213842a2a6fb84553bfbb78199R56

There is no explanation as to why `len` is being doubled (`len * 2`), especially when further down there is still a `.take(len)` but then `nextOffset` continues to be used with the doubling.

It's also questionable what the value of the second number really is in the resumptionToken, and if it wouldn't be better to just change this query param to the standard `offset` and drop the resumptionToken from the response.

It's also curious why for a simply paging function there is recursion going on. Possibly `Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/entity/PagedResults.scala` should have a full review and maybe wrapped in a test or two.

PR #2139, opened 2020-08-04
Fix resumption token used in REST APIs

PenghaiZhang:
#1735 

##### Checklist

- [ ] tests are included

##### Description of change

Fixed the resumption token used in some REST APIs. 

The testing environment for below screenshots has 5 items totally.

First request with initial resumption token being `0:2`. The length specified in params is ignored.
Screenshot from 2020-08-04 14-31-02

Second request with the token returned from last request
Screenshot from 2020-08-04 14-31-15

Third request with the token returned from last request. No more token is returned.
Screenshot from 2020-08-04 14-31-24

Another request with an initial token which can retrieve all items so no token returned.
Screenshot from 2020-08-04 14-31-43

Another request with an invalid token. So use the length from query params instead.
Screenshot from 2020-08-04 14-32-35

### DIFF
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/entity/PagedResults.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/entity/PagedResults.scala
@@ -102,16 +102,15 @@ object PagedResults {
 
     val (_, results) = collectMore(_length, firstOffset, 0, Vector.empty)
     val pb           = new PagingBean[BEB]
-    val actualLen    = results.length
     val available = res.getEntityService
       .countAll(new EnumerateOptions(q, 0, -1, system, if (includeDisabled) null else false))
       .toInt
     pb.setStart(firstOffset)
-    pb.setLength(actualLen)
+    pb.setLength(results.length)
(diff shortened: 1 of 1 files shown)


Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/entity/PagedResults.scala
```diff
@@ -99,16 +100,18 @@ object PagedResults {
       b
     }
 
-    val (nextOffset, results) = collectMore(length, firstOffset, 0, Vector.empty)
-    val pb                    = new PagingBean[BEB]
-    val actualLen             = results.length
-    pb.setStart(start)
+    val (_, results) = collectMore(_length, firstOffset, 0, Vector.empty)
+    val pb           = new PagingBean[BEB]
+    val actualLen    = results.length
```
edalex-ian, line 105: I think putting `results.length` into its own variable impinges on readability later on. Let's not have `actualLen` and just use `results.length`.

PenghaiZhang:
Considering the last commit has just one minor change and builds for  previous commit were all green, so merge this PR without waiting for green builds.

PR #2141, opened 2020-08-05
Update Schema module and Collection module

PenghaiZhang:
#1306 

##### Checklist

- [ ] tests are included

##### Description of change

As the resumption token has been fixed, this PR attempts to update Schema module and Collection module to use resumption tokens to retrieve data.

Created a new function to list entities that extend `BaseEntity`, and then updated the two modules to use this function. Also updated mock data and renamed a field of `CommonParams`.

Screenshot from 2020-08-05 15-55-32

Screenshot from 2020-08-05 15-56-20

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/modules/OEQHelpers.ts
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/modules/OEQHelpers.ts
@@ -45,12 +45,11 @@ export const listEntities = async <T extends OEQ.Common.BaseEntity>(
   getData: (resumptionToken: string) => Promise<OEQ.Common.PagedResult<T>>
 ): Promise<T[]> => {
   const entities: T[] = [];
-  let token: string = DEFAULT_RESUMPTION_TOKEN;
+  let token: string | undefined = DEFAULT_RESUMPTION_TOKEN;
   while (token) {
-    await getData(token).then((pagedResult) => {
-      entities.push(...pagedResult.results);
-      token = pagedResult.resumptionToken ?? "";
-    });
+    const pagedResult: OEQ.Common.PagedResult<T> = await getData(token);
+    entities.push(...pagedResult.results);
+    token = pagedResult.resumptionToken;
   }
   return entities;
 };
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/modules/OEQHelpers.ts
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/modules/OEQHelpers.ts
@@ -46,7 +46,6 @@ export const listEntities = <T extends OEQ.Common.BaseEntity>(
 ): Promise<T[]> => {
   const entities: T[] = [];
   return getData(DEFAULT_RESUMPTION_TOKEN).then(async (pagedResult) => {
-    console.log(pagedResult);
     entities.push(...pagedResult.results);
     // If a resumption token is returned, recursively call listEntities to retrieve more entities.
     if (pagedResult.resumptionToken) {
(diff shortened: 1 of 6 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/modules/OEQHelpers.ts
```diff
@@ -24,11 +24,38 @@ import * as OEQ from "@openequella/rest-api-client";
...
+ *
+ * @param getData A function which takes a resumption token to retrieve specific entities
+ *        such as schemas and collections.
+ */
+export const listEntities = <T extends OEQ.Common.BaseEntity>(
+  getData: (resumptionToken: string) => Promise<OEQ.Common.PagedResult<T>>
+): Promise<T[]> => {
+  const entities: T[] = [];
+  return getData(DEFAULT_RESUMPTION_TOKEN).then(async (pagedResult) => {
+    console.log(pagedResult);
```
SammyIsConfused, line 49: Spare log.

PenghaiZhang: Ah thanks for reminding!

Source/Plugins/Core/com.equella.core/js/tsrc/modules/OEQHelpers.ts
```diff
@@ -24,11 +24,37 @@ import * as OEQ from "@openequella/rest-api-client";
...
+ * @param getData A function which takes a resumption token to retrieve specific entities
+ *        such as schemas and collections.
+ */
+export const listEntities = <T extends OEQ.Common.BaseEntity>(
+  getData: (resumptionToken: string) => Promise<OEQ.Common.PagedResult<T>>
+): Promise<T[]> => {
+  const entities: T[] = [];
+  return getData(DEFAULT_RESUMPTION_TOKEN).then(async (pagedResult) => {
+    entities.push(...pagedResult.results);
+    // If a resumption token is returned, recursively call listEntities to retrieve more entities.
```
edalex-ian, line 50: A couple of things here:

1. If you make a recursive function, please put it in the JSDoc so it's immediately obvious to any consumers. This is important, as recursive functions can be a source of all kinds of wonderful issues. That leads to;
2. Although I like you're thinking, using a recursive function to iteratively retrieve an unknown size of results from a server is _dangerous_! Here in Javascript land we don't have tail-recursion and so you easily run the risk of blowing the stack. this is a pretty simple case, so a while loop will be fine, or you want you could play with JS Iterators or Generators.

So, we need to rewrite this without the recursion.

PenghaiZhang:
> Thanks @PenghaiZhang
> 
> A few bits:
> 
> * I'm not sure if we need to local functions of `getCollections` and `getSchemas`, but I don't mind.
> * I'm surprised the tests are still fine.... Or is that because there's no resumption token in the mocked data...
> * Overall though, we need to replace that recursive call - if we have a thousand or more results our stack will be at risk.

Yeah the mocked date do not have resumption tokens, but we should be able to add a mocked implementation to add resumption tokens back.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/OEQHelpers.ts
```diff
@@ -24,11 +24,33 @@ import * as OEQ from "@openequella/rest-api-client";
...
+export const listEntities = async <T extends OEQ.Common.BaseEntity>(
+  getData: (resumptionToken: string) => Promise<OEQ.Common.PagedResult<T>>
+): Promise<T[]> => {
+  const entities: T[] = [];
+  let token: string = DEFAULT_RESUMPTION_TOKEN;
+  while (token) {
+    await getData(token).then((pagedResult) => {
+      entities.push(...pagedResult.results);
+      token = pagedResult.resumptionToken ?? "";
+    });
```
edalex-ian, line 53: If you're going to use await, wouldn't you just do:

```typescript
    const pagedResult = await getData(token);
    entities.push(...pagedResult.results);
    token = pagedResult.resumptionToken ?? "";
```

Or are you doing it for the error handling?

Also, shouldn't it be fine for:

```typescript
    token = pagedResult.resumptionToken ?? "";
```

Just to be:

```typescript
    token = pagedResult.resumptionToken;
```

Issue #1531, opened 2020-03-09
docker-compose database issue


cbeach47:
**Describe the bug**
`docker-compose up -d` succeeds, but the logs are throwing an error with the postgre database (below).  Is there something missing from the documentation?

**To Reproduce**
Using docker-compose v 1.25.4 :
Run `docker-compose up -d`
Run `docker-compose logs` and note the errors with the database

**Expected behavior**
docker-compose should start up cleanly without external dependencies.

**Stacktrace**
```
oeq_1        | Caused by: java.net.UnknownHostException: database
oeq_1        | 	at java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:184)
oeq_1        | 	at java.net.SocksSocketImpl.connect(SocksSocketImpl.java:392)
oeq_1        | 	at java.net.Socket.connect(Socket.java:589)
oeq_1        | 	at org.postgresql.core.PGStream.<init>(PGStream.java:75)
oeq_1        | 	at org.postgresql.core.v3.ConnectionFactoryImpl.tryConnect(ConnectionFactoryImpl.java:91)
oeq_1        | 	at org.postgresql.core.v3.ConnectionFactoryImpl.openConnectionImpl(ConnectionFactoryImpl.java:192)
oeq_1        | 	... 16 more
database_1   | Error: Database is uninitialized and superuser password is not specified.
database_1   |        You must specify POSTGRES_PASSWORD to a non-empty value for the
database_1   |        superuser. For example, "-e POSTGRES_PASSWORD=password" on "docker run".
database_1   | 
database_1   |        You may also use "POSTGRES_HOST_AUTH_METHOD=trust" to allow all
database_1   |        connections without a password. This is *not* recommended.
database_1   | 
database_1   |        See PostgreSQL documentation about "trust":
database_1   |        https://www.postgresql.org/docs/current/auth-trust.html

```

**Platform:**

- OpenEquella Version: [2020.1.0-RC4]
- OS: [Ubuntu 16]

edalex-ian:
Heya @PenghaiZhang 

Could you have a look at this please?

Thanks.

PenghaiZhang:
> Heya @PenghaiZhang
> 
> Could you have a look at this please?
> 
> Thanks.

Yeah I will have a look today.

PenghaiZhang:
[...]

PR #2147, opened 2020-08-06
Update docker compose configuration

PenghaiZhang:
#1531 

##### Checklist

- [ ] tests are included

##### Description of change

Since Postgres docker image requires the environment variable POSTGRES_PASSWORD, added this variable to `docker-compose.yml `. Also added a `.env` file which is used to declare default environment variables.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -49,6 +49,8 @@ services:
     restart: always
     networks:
       - oeq-cluster
+    environment:
+      - POSTGRES_PASSWORD="${POSTGRES_PASSWORD}"
 
   traefik:
     image: traefik:1.7.26
--- a/docker/.env
+++ b/docker/.env
@@ -1,0 +1,1 @@
+POSTGRES_PASSWORD=postgres


docker/docker-compose.yml
```diff
@@ -49,6 +49,8 @@ services:
     restart: always
     networks:
       - oeq-cluster
+    environment:
+      - POSTGRES_PASSWORD="${POSTGRES_PASSWORD}"
```
ChristianMurphy, line 53: :thinking: would it make sense to leverage secrets here to encode/protect the password? https://docs.docker.com/engine/swarm/secrets/#build-support-for-docker-secrets-into-your-images
Or will this always be used for demo purposes so it isn't considered a secret?

PenghaiZhang: I thought it would be used for development purpose, but I am happy to add Docker secrets.

edalex-ian: Yeah, probably okay. Really only for dev/demo. I would not advise this for a production setup.

Mind you, I do hope that in the future we can start looking at an official 'docker' offering with images and compose files.

Issue #2107, opened 2020-07-27
Updates to legacy.scss do not show up properly until theme changes happen


SammyIsConfused:
**Describe the bug**
The intended way for this to work was for the date modified of the legacy.scss compared to the applied one to determine which copy gets applied. This doesn’t work, instead the only way to get it to update is to make a change in the Theme Editor to force a change. 

**To Reproduce**
1.  Make a change to legacy.scss and rebuild your oEQ. 
2.  Note that the change has not applied.
3. Go to the Theme Editor and click “Reset To Default” for the theme.
4. Note that the change has applied.

**Expected behavior**
Any change to the legacy.scss file should be automatically applied upon rebuild. We sh ould not have to rely upon changing the theme editor to get that to work. 

**Platform:**
- OpenEquella Version: 2020.2

SammyIsConfused:
The fix for this has been merged, and tested by @cathfitz.  Closing.

PR #2158, opened 2020-08-10
EBP: Updates to legacy.scss do not show up properly until theme changes happen

SammyIsConfused:
##### Checklist

##### Description of change
#2107 
The issue was caused by the lastModified check on the legacy.scss file. 
Creating a new File() with getResource().getFile() worked fine in a dev environment - but in an installed oEQ
this file would be inside a jar and wouldn't work. The lastModified date would always return 0. 
This PR changes to use an URLConnection to the same file to grab the value and then close the stream, and I have tested this via installerZip. 

### DIFF
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/service/impl/ThemeSettingsServiceImpl.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/service/impl/ThemeSettingsServiceImpl.java
@@ -41,6 +41,7 @@ import java.awt.*;
 import java.awt.image.BufferedImage;
 import java.awt.image.RenderedImage;
 import java.io.*;
+import java.net.URLConnection;
 import java.util.Collections;
 import javax.imageio.ImageIO;
 import javax.inject.Inject;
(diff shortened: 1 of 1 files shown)


edalex-ian:
Looks like your java format failed. Wonder if your pre-commit hooks are unhappy. :thinking:

SammyIsConfused:
> Looks like your java format failed. Wonder if your pre-commit hooks are unhappy. thinking

Right you are - I had accidentally CTRL-C'd out of the `npm ci` call in the root. Should be fixed up now.

Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/service/impl/ThemeSettingsServiceImpl.java
```diff
@@ -165,16 +166,22 @@ private String themeToJSONString(NewUITheme theme) throws JsonProcessingExceptio
...
+    long lastMod = conn.getLastModified();
+    conn.getInputStream().close();
+
+    // compare against the modified date of the css file if it exists
+    boolean legacyCssExists = fileSystemService.fileExists(customisationFile, LEGACY_CSS_FILENAME);
+    if (legacyCssExists) {
+      long cssLastMod = fileSystemService.lastModified(customisationFile, LEGACY_CSS_FILENAME);
+      boolean baseSassUpdated = lastMod > cssLastMod;
+      if (baseSassUpdated) {
+        compileSass();
```
edalex-ian, line 183: The logic for this used to be:

* !`legacyCssExists` OR `baseSassUpdated`

Now it's:

* `legacyCssExists` AND `baseSassUpdated`

I believe that means on the first run of a fresh system, this will not get generated.

SammyIsConfused: Gah! Yup, I forgot the else for the initial legacyCssExists check. Sorry about that.

PR #2167, opened 2020-08-11
Make date range end inclusive in searches

PenghaiZhang:
#1306  

##### Checklist

- [ ] tests are included

##### Description of change

This PR attempts to make the end of a date range inclusive in searches without making any change on the old Search page.

Also during the work, I found that in the old Search page, selected dates are always convert to UTC dates, which means the search results may be wrong. For example, if I select 1st Aug, the real date passed to Lucene is `31 July 14:00:00` because AEST is 10 hours ahead of UTC time. More info can be found in screenshots.

In new Search page: 
image
image

In old Search page:
image
image

### DIFF
--- a/Source/Plugins/Core/com.equella.base/src/com/tle/common/searching/Search.java
+++ b/Source/Plugins/Core/com.equella.base/src/com/tle/common/searching/Search.java
@@ -86,4 +86,6 @@ public interface Search {
   String getPrivilegeToCollect();
 
   Collection<DateFilter> getDateFilters();
+
+  boolean isFromSearch2API();
 }
--- a/Source/Plugins/Core/com.equella.base/src/com/tle/common/searching/Search.java
+++ b/Source/Plugins/Core/com.equella.base/src/com/tle/common/searching/Search.java
@@ -88,7 +88,7 @@ public interface Search {
   Collection<DateFilter> getDateFilters();
 
   /**
-   * Indicate if the server time zone is needed in searches.
+   * Indicates if the server time zone should be included in searches.
    *
    * @return True if server time zone is used or otherwise false.
    */
--- a/Source/Plugins/Core/com.equella.base/src/com/tle/common/searching/Search.java
+++ b/Source/Plugins/Core/com.equella.base/src/com/tle/common/searching/Search.java
@@ -87,5 +87,10 @@ public interface Search {
 
   Collection<DateFilter> getDateFilters();
 
-  boolean isFromSearch2API();
+  /**
+   * Indicate if the server time zone is needed in searches.
+   *
+   * @return True if server time zone is used or otherwise false.
+   */
+  boolean useServerTimeZone();
 }
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
@@ -61,7 +61,7 @@ object SearchHelper {
     */
   def createSearch(params: SearchParam): DefaultSearch = {
     val search = new DefaultSearch
-    search.setFromSearch2API(true)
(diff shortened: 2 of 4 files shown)


Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
```diff
@@ -118,17 +120,22 @@ object SearchHelper {
...
     * Parse a string to a new instance of Date in the format of "yyyy-MM-dd".
-    * @param date The string to parse.
-    * @return An option which wraps an instace of Date.
+    * @param dateString The string to parse.
+    * @param isStart True if the date is the start of a date range.
+    * @return An option which wraps an instance of Date.
     */
-  def handleModifiedDate(date: String): Option[Date] = {
-    if (Check.isEmpty(date)) {
+  def handleModifiedDate(dateString: String, isStart: Boolean = false): Option[Date] = {
```
edalex-ian, line 127: I feel the `isStart` flag is obfuscating things unnecessarily here. Could this not just be `time: LocalTime`?

If it was, where it's used would become much clearer, as you'd simply have:

```scala
    // The After date is the start of the range and so should be at the first possible moment of the day - LocalTime.MIN
    val modifiedAfter  = handleModifiedDate(params.modifiedAfter, LocalTime.MIN)
    // The Before date is the end of the range and so should be in the last possible moment of the day - LocalTime.MAX
    val modifiedBefore = handleModifiedDate(params.modifiedBefore, LocalTime.MAX)
```

(As an aside, you'll also note I've re-ordered the declaration of these two values to be slightly more logical.)

This change then also simplifies the code down below.

Source/Plugins/Core/com.equella.base/src/com/tle/common/searching/Search.java
```diff
@@ -86,4 +86,6 @@ public SortField getSortField(boolean reverseTheDefault) {
   String getPrivilegeToCollect();
 
   Collection<DateFilter> getDateFilters();
+
+  boolean isFromSearch2API();
```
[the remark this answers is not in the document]
PenghaiZhang: I actually came up with this name when I was working on the multiple Dynamic Collections because I wanted to indicate some code is only written for the new Search page. However, as we won't do multiple Dynamic Collections, perhaps this feature becomes the only one that requires changes on legacy code. So I am happy to rename the variable to `useServerTimeZone`.

Source/Plugins/Core/com.equella.base/src/com/tle/common/searching/Search.java
```diff
@@ -87,5 +87,10 @@ public SortField getSortField(boolean reverseTheDefault) {
 
   Collection<DateFilter> getDateFilters();
 
-  boolean isFromSearch2API();
+  /**
+   * Indicate if the server time zone is needed in searches.
```
edalex-ian, line 91:
```suggestion
   * Indicates if the server time zone should be included in searches.
```

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
```diff
@@ -118,17 +121,20 @@ object SearchHelper {
 
   /**
     * Parse a string to a new instance of Date in the format of "yyyy-MM-dd".
-    * @param date The string to parse.
-    * @return An option which wraps an instace of Date.
+    * @param dateString The string to parse.
+    * @param time The time added to a date.
+    * @return An option which wraps an instance of Date.
```
edalex-ian, line 126: Be good to further detail the return value. That way a user can rely on their IDE's documentation feature to display useful guidance without needing to review the function.

```suggestion
    * @return An Option which wraps an instance of Date, combining the successfully parsed dateString and provided time (based on the system's default timezone).
```

PR #2171, opened 2020-08-12
Apply Search filter visibility settings to refine controls

PenghaiZhang:
#1306

##### Checklist

- [x] tests are included

##### Description of change

The visibility of the Owner and Date Modified filters are now controlled by Search settings.  Also, some tweaks were made for  tests that use `react-testing-library`.

Both are enabled.
Screenshot from 2020-08-12 14-22-16

Owner filter is disabled.
Screenshot from 2020-08-12 14-22-29

Date modified filter is disabled, too.
Screenshot from 2020-08-12 14-22-41

Screenshot from 2020-08-12 14-16-17


Source/Plugins/Core/com.equella.core/js/tsrc/search/components/RefineSearchPanel.tsx
```diff
@@ -41,6 +41,10 @@ export interface RefinePanelControl {
    * Refine control.
    */
   component: ReactNode;
+  /**
+   * True if the control is configured not to be showed.
+   */
+  disabled: boolean;
```
edalex-ian, line 47: Should we make this optional, so it can kind of default to `false`?

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -234,6 +239,7 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
           value={searchPageOptions.owner}
         />
       ),
+      disabled: searchSettings.searchingDisableOwnerFilter,
```
edalex-ian, line 242: hmmmm, because these are getting changed via an effect, if the default is `true` but then the server says false will there be a momentary rendering of them? (I think I've seen this kind of visual effect with how we show the settings - they initially render in one state, and then swap to what we get from the server.) Maybe none should be rendered until we've heard from the server to ensure smooth UX - rather than disappearing controls.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/RefineSearchPanel.tsx
```diff
@@ -57,20 +61,23 @@ export const RefineSearchPanel = ({ controls }: RefinePanelProps) => {
...
-                <Grid item>
-                  <RefinePanelControlHeading title={control.title} />
-                </Grid>
-                <Grid item>{control.component}</Grid>
-              </Grid>
-            </ListItem>
-          ))}
+          {controls.map(
+            (control) =>
+              !control.disabled && (
```
edalex-ian, line 66: To reduce nesting and increase readability, could we:

```typescript
{controls
.filter( c => !c.disabled )
.map( (control) => (
  <ListItem ...>
    ...
  </ListItem>
)}
```

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
edalex-ian: So it turns out they all currently have a default of false and as a result this should just be a case of a delay in showing the components. That should be AOK. :+1:

PR #2172, opened 2020-08-12
Refine Search - StatusSelector

edalex-ian:
##### Checklist

- [x] tests are included

##### Description of change

Another Refine Search control, this time for status.

Also needed to refactor the `OEQ.Common.ItemStatus` from an `enum` to essentially a string literal union. Triggered due to issues with Jest losing visibility of the enum post transpilation. To provide the full power I was after, rather than just a plain ol' string literal union, I brought in [runtypes](https://www.npmjs.com/package/runtypes).

Further auxilary items:

* A wording change for the configuration of whether to display this control - previously worded around non-live statuses;
* For IntelliJ project turned on sort imports by module - so that <kbd>CTRL</kbd>+<kbd>ALT</kbd>+<kbd>o</kbd> gives a nice tidy up (which was done in some files); and
* Sorted the sub groups of the Search Page language strings, as it was become rather unwieldy.

image 
image

#1306 

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
@@ -15,39 +15,45 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import { CircularProgress } from "@material-ui/core";
+import * as OEQ from "@openequella/rest-api-client";
+import "@testing-library/jest-dom/extend-expect";
+import {
+  fireEvent,
+  getByText,
+  render,
+  RenderResult,
+  screen,
+  waitFor,
+} from "@testing-library/react";
+import { mount, ReactWrapper } from "enzyme";
+import * as React from "react";
+import { act } from "react-dom/test-utils";
+import { BrowserRouter } from "react-router-dom";
+import { getCollectionMap } from "../../../__mocks__/getCollectionsResp";
 import {
   getEmptySearchResult,
   getSearchResult,
   getSearchResultsCustom,
 } from "../../../__mocks__/getSearchResult";
-import { getCollectionMap } from "../../../__mocks__/getCollectionsResp";
-import * as React from "react";
-import SearchPage, { SearchPageOptions } from "../../../tsrc/search/SearchPage";
-import { mount, ReactWrapper } from "enzyme";
-import { act } from "react-dom/test-utils";
-import * as SearchModule from "../../../tsrc/modules/SearchModule";
+import * as UserSearchMock from "../../../__mocks__/UserSearch.mock";
+import { DateRangeSelector } from "../../../tsrc/components/DateRangeSelector";
 import * as CollectionsModule from "../../../tsrc/modules/CollectionsModule";
+import {
+  liveStatuses,
+  nonLiveStatuses,
+} from "../../../tsrc/modules/SearchModule";
+import * as SearchModule from "../../../tsrc/modules/SearchModule";
 import * as SearchSettingsModule from "../../../tsrc/modules/SearchSettingsModule";
 import * as UserModule from "../../../tsrc/modules/UserModule";
-import { BrowserRouter } from "react-router-dom";
-import { CircularProgress } from "@material-ui/core";
 import { CollectionSelector } from "../../../tsrc/search/components/CollectionSelector";
+import SearchPage, { SearchPageOptions } from "../../../tsrc/search/SearchPage";
+import { languageStrings } from "../../../tsrc/util/langstrings";
 import { paginatorControls } from "../components/SearchPaginationTestHelper";
-import { DateRangeSelector } from "../../../tsrc/components/DateRangeSelector";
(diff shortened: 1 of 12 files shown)


Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
```diff
@@ -75,36 +81,112 @@ const defaultSearchPageOptions: SearchPageOptions = {
...
+};
+
+describe("Refine search by status", () => {
+  const {
+    live: liveButtonLabel,
+    all: allButtonLabel,
+  } = languageStrings.searchpage.statusSelector;
+
+  const expectSearchItemsCalledWithStatus = (status: OEQ.Common.ItemStatus[]) =>
+    expect(mockSearch).toHaveBeenCalledWith({
```
PenghaiZhang, line 139: We have mixed use of  `toHaveBeenCalledWith` and `toHaveBeenLastCalledWith` ? Is it something we want to unify ? 

Also I noted a small difference. This function checks `mockSearch` whereas others check `SearchModule.searchItems`, but I think the result is the same.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
```diff
@@ -75,36 +81,112 @@ const defaultSearchPageOptions: SearchPageOptions = {
...
+    });
+
+  const getStatusSelector = (container: Element): HTMLElement =>
+    getRefineSearchComponent(container, "StatusSelector");
+
+  const selectStatus = (container: Element, status: string) =>
+    fireEvent.click(getByText(getStatusSelector(container), status));
+
+  afterEach(() => {
+    // Needed to keep Enzyme tests below happy
```
PenghaiZhang, line 151: According to the doco of `render`, it might be good to call `cleanup` here.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -22,11 +22,31 @@ import { Collection } from "./CollectionsModule";
...
+ * Predicate for checking if a provided status is not one of `liveStatuses`.
+ * @param status a status to check for liveliness
+ */
+export const nonLiveStatus = (status: OEQ.Common.ItemStatus): boolean =>
+  !liveStatuses.find((liveStatus) => status === liveStatus);
+
+/**
+ * List of statuses which are considered non-live.
+ */
+export const nonLiveStatuses: OEQ.Common.ItemStatus[] = OEQ.Common.ItemStatuses.alternatives
```
PenghaiZhang, line 40: May I ask why we need to a const representing non-live Statuses rather than all Statuses ? If we have something like `allStatuses` then we can skip some array concatenation.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
edalex-ian: Heh, I don't know. Funny you mention it, as I queried it over in your PR:

https://github.com/openequella/openEQUELLA/pull/2171#discussion_r469028916

Be interested to hear your thoughts. :thinking:

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
edalex-ian: I guess it depends on context. Here I'm not concerned with whether it was first, last or otherwise. Just that it was called with that value at one point.

I decided to go with the `mockSearch` here as we do have that variable capturing the situation, and indeed it provides us an abstraction - should the search function used by SearchPage change later. I think arguably we should use these constants in place of the explicit function reference when we've mocked them out. Maybe as part of the refactor we standardise on that. Thoughts?

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
edalex-ian: I was wondering if anyone would pick this up, as I'm not sure either way. But here goes my thinking...

If you have have one subset on hand, it'd be useful to have the complement too. As in this case now I can readily say here are the 'live' status, and here are the 'non-live' statuses for any business logic we need to implement. And then if we need all I can be explicit and say here are all the live and non-live statuses. (But I've probably over thought it at this point.)

Then as an aside, we do now technically have all the status on hand thanks to `runtypes` - via `OEQ.Common.ItemStatuses.alternatives.map( i => i.value)`.

But so that is the thinking. What do you think? Should I drop the `nonLiveStatuses` and just use the above code snippet where we need all?

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
edalex-ian: Actually, I am concerned here that it was the last one, as due to all the effects I can't be sure. Fixed. :+1:

PR #2181, opened 2020-08-13
Support being able to set the theme for the paper / card elements

edalex-ian:
##### Checklist

##### Description of change

I noticed it was not possible to properly setup a dark theme as the cards were fixed at white. So this change supports being able to theme the colour of the cards, as well as some minor tweaks to the look of EBP cards.

(We may want to cherry-pick the EBP commit 45f43de back to `develop` with the other EBP updates.)

image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/theme/ThemePage.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/theme/ThemePage.tsx
@@ -15,7 +15,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import * as React from "react";
 import {
   Button,
   Card,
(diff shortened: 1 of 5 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/theme/ThemePage.tsx
```diff
@@ -27,26 +26,27 @@ import {
   WithStyles,
   withStyles,
 } from "@material-ui/core";
-import ColorPickerComponent from "./ColorPickerComponent";
 import axios, { AxiosError } from "axios";
-import { API_BASE_URL } from "../config";
-import { languageStrings } from "../util/langstrings";
-import { commonString } from "../util/commonstrings";
+import * as React from "react";
+import { IThemeSettings } from ".";
```
ChristianMurphy, line 31: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export could be useful here

```suggestion
import type { IThemeSettings } from ".";
```

edalex-ian: This syntax is indeed new to me.

From what I understand, all this will do is ensure that nothing remains for the interface after compilation time. But isn't that the case already?

ChristianMurphy: In some cases.
If a module is side-effect free, tree shakers and bundlers can automatically import only the types and exclude the code.
However, if there is a side-effect in the module, a tree-shaker can't determine if the side-effect is intended or not, so it will include the code just in case.
The `import type` annotation tells the tooling to ignore any side effects and only pull the types.

edalex-ian: Not sure I follow though. There is no code for an `interface`. Is there any chance you could point to an example or something?

ChristianMurphy: _side-effect.ts_

```ts
import axios from 'axios'

// side effect, this runs when the script is imported
axios.get('https://example.com')

interface Example {}
```

_example.ts_

```ts
// any import will run the side effect
// meaning example.com will be called, and all of axios gets pulled in
import 'side-effect'

// including if the import is just a typing
import { Example } from 'side-effect'

// unless you tell the compiler to skip the side effects and only get the types
import type { Example } from 'side-effect'
```

edalex-ian: Aaaaahhhh!!! Now I got it. Thank you. :+1:

PR #2183, opened 2020-08-14
Add the support of whether to search attachments to Search2 API

PenghaiZhang:
#1306 

##### Checklist

- [x] tests are included

##### Description of change

This PR includes back-end changes for the feature of whether to search attachments or not. As we don't want to use `User Preference` to check if the option of searching attachments is enabled, added a new search param which represents the option, and overloaded method  `search` to accept a boolean.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/core/freetext/service/FreeTextService.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/core/freetext/service/FreeTextService.java
@@ -43,6 +43,18 @@ public interface FreeTextService {
   <T extends FreetextResult> FreetextSearchResults<T> search(
       Search searchreq, int nStart, int nCount);
 
+  /**
+   * Another search method which allows users to specify if they want to search attachments.
+   *
+   * @param searchReq An search request
+   * @param start The first record of a search result
+   * @param count The number of records of a search result
+   * @param searchAttachments Whether to search attachments
+   * @return An instance of SearchResults
+   */
+  <T extends FreetextResult> FreetextSearchResults<T> search(
+      Search searchReq, int start, int count, boolean searchAttachments);
+
   SearchResults<ItemIdKey> searchIds(Search searchreq, int nStart, int nCount);
 
   LongSet searchIdsBitSet(Search searchreq);
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/core/freetext/service/FreeTextService.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/core/freetext/service/FreeTextService.java
@@ -35,22 +35,22 @@ public interface FreeTextService {
 
   /**
    * @param <T>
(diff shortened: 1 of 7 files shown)


Source/Plugins/Core/com.equella.core/src/com/tle/core/freetext/service/FreeTextService.java
```diff
@@ -43,6 +43,18 @@
   <T extends FreetextResult> FreetextSearchResults<T> search(
       Search searchreq, int nStart, int nCount);
 
+  /**
+   * Another search method which allows users to specify if they want to search attachments.
+   *
+   * @param searchReq An search request
```
edalex-ian, line 49:
```suggestion
   * @param searchReq A search request
```

Source/Plugins/Core/com.equella.core/src/com/tle/core/freetext/service/FreeTextService.java
```diff
@@ -43,6 +43,18 @@
...
+   * @param start The first record of a search result
+   * @param count The number of records of a search result
```
edalex-ian, line 51:
```suggestion
   * @param count The maximum number of results requested, or -1 for all.
```
Actually, should this param be called `max` or `length` or `maxCount`/`maxLength`? Ooooh, perhaps not, this is meant to be an overloaded method). In that case these should probably be `nStart` and `nCount` to match the method you're overloading. (Mind you, if you wanted too you could update the other method too so as to improve the names.)

Source/Plugins/Core/com.equella.core/src/com/tle/core/freetext/service/FreeTextService.java
```diff
@@ -43,6 +43,18 @@
...
+   * @param searchAttachments Whether to search attachments
+   * @return An instance of SearchResults
```
edalex-ian, line 53: The method signature already tells us this, so how about:

```suggestion
   * @return Search results matching the specified criteria, with a maximum length of count.
```

Source/Plugins/Core/com.equella.core/src/com/tle/core/freetext/service/impl/FreeTextServiceImpl.java
```diff
@@ -263,6 +263,13 @@ public void indexAll() {
     return new StdFreetextResults<T>(itemService, fTextResults, searchreq);
   }
 
+  @Override
+  public <T extends FreetextResult> FreetextSearchResults<T> search(
+      Search searchReq, int start, int count, boolean searchAttachments) {
+    SearchResults<T> fTextResults = indexer.search(searchReq, start, count, searchAttachments);
```
edalex-ian, line 269: Is the `f` in `fTextResults` for `free`? If so, let's just use the whole word. But then again, we do have the type here, so why not just `results`?

PR #2188, opened 2020-08-17
Search attachment Selector

PenghaiZhang:
#1306 

##### Checklist

- [x] tests are included

##### Description of change

So here is another Refine Search control - Search attachment Selector. By default, the option of 'yes' is selected. And because there are no Search settings controlling this component's visibility, it's always visible. 

The structure of this component is almost same as `StatusSelector`, and so are its tests and stories. So we can create an abstract component called `TwoOptionsSelector` , but I am not sure how much value we will gain at this stage so I did not implement this.

image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
@@ -154,6 +154,7 @@ const getRefineSearchComponent = (
 
 describe("Refine search by searching attachments", () => {
   let page: RenderResult;
+
   beforeEach(async () => {
     page = await renderSearchPage();
   });
@@ -166,10 +167,7 @@ describe("Refine search by searching attachments", () => {
     getRefineSearchComponent(container, "SearchAttachmentsSelector");
   const changeOption = (selector: HTMLElement, option: string) =>
     fireEvent.click(getByText(selector, option));
-  const {
-    yes: yesLabel,
-    no: noLabel,
-  } = languageStrings.searchpage.searchAttachmentsSelector;
+  const { yes: yesLabel, no: noLabel } = languageStrings.common.action;
 
   it("Should default to searching attachments", async () => {
     expect(mockSearch).toHaveBeenLastCalledWith(defaultSearchPageOptions);
--- a/Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
@@ -152,6 +152,45 @@ const getRefineSearchComponent = (
   return e as HTMLElement;
 };
 
+describe("Refine search by searching attachments", () => {
+  let page: RenderResult;
+  beforeEach(async () => {
+    page = await renderSearchPage();
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  const getSearchAttachmentsSelector = (container: Element): HTMLElement =>
+    getRefineSearchComponent(container, "SearchAttachmentsSelector");
+  const changeOption = (selector: HTMLElement, option: string) =>
+    fireEvent.click(getByText(selector, option));
+  const {
+    yes: yesLabel,
+    no: noLabel,
+  } = languageStrings.searchpage.searchAttachmentsSelector;
+
+  it("Should default to searching attachments", async () => {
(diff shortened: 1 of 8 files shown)


edalex-ian:
> The structure of this component is almost same as `StatusSelector`, and so are its tests and stories. So we can create an abstract component called `TwoOptionsSelector` , but I am not sure how much value we will gain at this stage so I did not implement this.

Yup I was ponder this too when you started on it. However it'd be a very thin layer on top of ButtonGroup. But maybe in the future we may want a component that is called something like `ButtonToggle` which could take a collection of button definitions so you could have two or more in the group. The definition would probably specify a label and what the output value would be when it's clicked/selected. And the component would have the logic to ensure only one can be selected at a time.

But for now, let's just wait and see if we end up with at least one more.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchAttachmentsSelector.tsx
```diff
@@ -0,0 +1,51 @@
...
+   * True if 'yes' is selected.
+   */
+  value?: boolean;
+  /**
+   * Fired when a different option is selected.
+   * @param value A boolean representing the selected option.
+   */
+  onChange: (value: boolean) => void;
+}
+export const SearchAttachmentsSelector = ({
```
edalex-ian, line 34: Nit: could we have a blank line above here please?

oeq-ts-rest-api/src/Search.ts
```diff
@@ -59,6 +59,10 @@ export interface SearchParams {
    * single dynamic collection uuid.
    */
   dynaCollection?: string;
+  /**
+   * A flag indicating whether to search attachments or not.
+   */
+  searchAttachments?: boolean;
```
edalex-ian, line 65: Good to see this was done in its own commit. May've also been good to be in it's own PR, but considering it's only a one liner I can see why not. Hopefully soon we can look at moving this module to stand-alone. (Probably after we merge back to `develop`.

But... This should probably have a test to confirm that the definition of this param is correct - and is harnessed into the future. Would it be much effort to do such a test? (Ideally it'd be a simple positive/negative test calling with true/false.)

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
```diff
@@ -152,6 +152,45 @@ const getRefineSearchComponent = (
   return e as HTMLElement;
 };
 
+describe("Refine search by searching attachments", () => {
+  let page: RenderResult;
+  beforeEach(async () => {
```
SammyIsConfused, line 157: Perhaps leave a newline before beforeEach

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchAttachmentsSelector.tsx
```diff
@@ -0,0 +1,51 @@
...
+ * limitations under the License.
+ */
+import { Button, ButtonGroup } from "@material-ui/core";
+import * as React from "react";
+import { languageStrings } from "../../util/langstrings";
+
+interface SearchAttachmentsSelectorProps {
+  /**
+   * A boolean value indicating whether 'yes' or 'no' is selected.
+   * True if 'yes' is selected.
```
SammyIsConfused, line 25: I think this line is potentially redundant.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchAttachmentsSelector.tsx
```diff
@@ -0,0 +1,51 @@
...
+}
+export const SearchAttachmentsSelector = ({
+  value,
+  onChange,
+}: SearchAttachmentsSelectorProps) => {
+  const variant = (option?: boolean) => (option ? "contained" : "outlined");
+  return (
+    <ButtonGroup color="secondary">
+      <Button variant={variant(value)} onClick={() => onChange(true)}>
+        {languageStrings.searchpage.searchAttachmentsSelector.yes}
```
SammyIsConfused, line 42: I could have sworn there was already language strings for yes and no somewhere :thinking: Not exactly sure where but it would be good to reduce duplication.

SammyIsConfused: As discussed we couldn't find one, but it makes more sense for it to be in common strings (yes and no will definitely be used by more than just the searchAttachmentsSelector)

Issue #1983, opened 2020-07-01
Improvements and tweaking of the default theme (EBP) for the new UI


cathfitz:
**Is your feature request related to a problem? Please describe.**
Improvements, fixes and tweaks to the new default theme used when switching on the new UI.

**Describe the solution you'd like**
A seamless UI that matches the new Search UI across the balance of oEQ

PR #2196, opened 2020-08-18
EBP: Add margin to the right of portlets

SammyIsConfused:
##### Checklist

##### Description of change
#1983  
Margins were off for side-by-side portlets- the leftmost portlet merged into the rightmost one. This simply adds a margin to the right of each portlet.

image


SammyIsConfused:
> Definitely seems to fix the issue, however... Now there appears to be double padding on the right-most portlet. Does this perhaps need a left and right margin of 8px each? But what is managing the padding between portlets vertically?

@edalex-ian The vertical padding is caused by the .box class - which is used everywher so isn't the best target for this. It has default vertical padding but not horizontal. 
I could do left and right. That looks like this. 
image

edalex-ian:
> I could do left and right. That looks like this.

Wow, that looks much nicer. Notice the consistent space around the whole dashboard now. :+1:

PR #2199, opened 2020-08-18
EBP: improve contribution page text styling

SammyIsConfused:
##### Checklist

##### Description of change
#1983 
Aligns contribution page text styles to MUI typography. 
image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
+++ b/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
@@ -133,7 +133,6 @@ $activeBackgroundColor: transparentize(#000, 0.7);
 
 @mixin typography-headline-5 {
   font-size: 1.5rem;
-  font-family: "Roboto", "Helvetica", "Arial", sans-serif;
   font-weight: 400;
   line-height: 1.334;
   letter-spacing: 0em;
@@ -141,7 +140,6 @@ $activeBackgroundColor: transparentize(#000, 0.7);
 
 @mixin typography-headline-6 {
   font-size: 1.25rem;
-  font-family: "Roboto", "Helvetica", "Arial", sans-serif;
   font-weight: 500;
   line-height: 1.6;
   letter-spacing: 0.0075em;
@@ -171,7 +169,6 @@ $activeBackgroundColor: transparentize(#000, 0.7);
 
 @mixin typography-body-1 {
   font-size: 1rem;
-  font-family: "Roboto", "Helvetica", "Arial", sans-serif;
   font-weight: 400;
   line-height: 1.5;
   letter-spacing: 0.00938em;
--- a/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
+++ b/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
@@ -1114,7 +1114,7 @@ Generic layout
 }
 
 .area h2 {
-  @include typography-headline-4;
+  @include typography-headline-5;
   color: #444444;
 }
 
@@ -3494,7 +3494,7 @@ Item summary styling
 }
 
 .itemsummary-layout #col2 h2 {
-  @include typography-headline-4;
+  @include typography-headline-5;
 }
 
 .itemsummary-layout #col2 ul {
--- a/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
+++ b/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
@@ -1274,14 +1274,6 @@ div[class*="fancy-bg"] {
   span[class="resumable"],
   span[class="removelink"] {
     @include typography-body-1;
-
-    a {
-      @include typography-body-1;
-    }
-
-    abbr {
-      @include typography-body-1;
-    }
   }
 
   h2 {
--- a/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
+++ b/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
@@ -2920,7 +2920,7 @@ a.add {
 }
 
 .resumeables {
-  padding-bottom: 16px;
+  padding-bottom: 20px;
 }
 
 .contribution-selection-page .resumeables ul li span.resumable {
--- a/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
+++ b/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
@@ -2040,7 +2040,7 @@ input[type="text"] {
 #selection-content-inner.wizard-layout
 .indent2
 .autocompleteControl
-  input[type="text"] {
+input[type="text"] {
   width: 350px;
 }
 
@@ -4825,37 +4825,37 @@ input[type="text"],
 #selection-content-inner.wizard-layout
 .multieditbox
 div.singletranslation
-  input[type="text"],
+input[type="text"],
 #selection-content-inner.wizard-layout
 .multieditbox
 div.singletranslation
-  textarea {
+textarea {
   width: 512px;
 }
 
 #selection-content-inner.wizard-layout
 .indent1
 .multieditbox
-  div.singletranslation
-  input[type="text"],
+div.singletranslation
+input[type="text"],
 #selection-content-inner.wizard-layout
 .indent1
 .multieditbox
-  div.singletranslation
-  textarea {
+div.singletranslation
+textarea {
   width: 472px;
 }
 
 #selection-content-inner.wizard-layout
 .indent2
 .multieditbox
-  div.singletranslation
-  input[type="text"],
+div.singletranslation
+input[type="text"],
 #selection-content-inner.wizard-layout
 .indent2
 .multieditbox
-  div.singletranslation
-  textarea {
+div.singletranslation
+textarea {
   width: 432px;
 }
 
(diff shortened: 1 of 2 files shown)


Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -1279,6 +1270,41 @@ div[class*="fancy-bg"] {
...
+  }
+
+  h2 {
+    @include typography-headline-5;
+  }
+
+  table {
+    thead > tr > th {
+      @include typography-headline-6;
+      color: $menuTextColor;
```
edalex-ian, line 1294: Shouldn't this be `$primaryTextColor`?

(The menu colours should really only be used in the New UI menu.)

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -1279,6 +1270,41 @@ div[class*="fancy-bg"] {
...
+    }
+
+    tbody > tr > td {
+      @include typography-body-1;
+      color: $menuTextColor;
```
edalex-ian, line 1299: Shouldn't this be `$primaryTextColor`?

(The menu colours should really only be used in the New UI menu.)

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -1279,6 +1270,41 @@ div[class*="fancy-bg"] {
...
+  span[class="resumable"],
+  span[class="removelink"] {
+    @include typography-body-1;
+
+    a {
+      @include typography-body-1;
+    }
+
+    abbr {
+      @include typography-body-1;
```
edalex-ian, line 1283: These are interesting, as seeing you've already got it on the parent span I'd expect it to flow through. Is this explicitly over ridding something?

Also, do you still need the `@include` at the parent level if you're mainly targetting these two?

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -4806,8 +4942,9 @@ Login Notice Formatting
 *****************************************************************************/
 #loginNotice p {
   font-family: "Arial", "georgia", "impact", "Tahoma", "Times", "Verdana",
-    monospace;
+  monospace;
```
edalex-ian, line 4945: The original indenting here was correct.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -1279,6 +1270,41 @@ div[class*="fancy-bg"] {
...
+      @include typography-body-1;
+    }
+
+    abbr {
+      @include typography-body-1;
+    }
+  }
+
+  h2 {
+    @include typography-headline-5;
```
edalex-ian, line 1288: Interesting. Our standard `h2` typography is `typography-headline-4`. With all the other changes, have we re-reviewed this one to see if it's okay as 'standard'?

edalex-ian (commented):
Thanks @SammyIsConfused 

Mostly there, just a few tweaks needed, and a number of queries.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -132,25 +132,19 @@ $activeBackgroundColor: transparentize(#000, 0.7);
 }
 
 @mixin typography-headline-5 {
-  -moz-osx-font-smoothing: grayscale;
-  -webkit-font-smoothing: antialiased;
   font-size: 1.5rem;
-  line-height: 2rem;
+  font-family: "Roboto", "Helvetica", "Arial", sans-serif;
```
mrblippy, line 136: Do these need to be explicitly defined in each mixin? I thought the font-family rule here https://github.com/openequella/openEQUELLA/blob/5ca1b1da8ef4e9a6f1fba84d693a3fc28e9f2aea/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss#L310 would be enough?

SammyIsConfused: I just copied exactly what Material UI had.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
SammyIsConfused: Correct me if I'm wrong, but menuTextColor is the name of the variable that controls secondary text colour. I know that in your reworked theme page it is. I just wanted to use the secondary text color.

SammyIsConfused: When I originally did the theme editor I named the variables really badly, the menu text was the only thing that controlled at the time.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
SammyIsConfused: The span has inner text, but it also has the link and the time tags in there. We have default CSS for abbr and a, it only seemed to work when I put it on both and the span itself. 
I could split it out of the hierarchy the perhaps make it easier to read?

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
edalex-ian: Nope. There's a `secondaryTextColor` - which I think has been there for quite some time too. :thinking:

SammyIsConfused: image
I really don't think that's true - are you sure it's not component you're thinking of?

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
edalex-ian: Readability is fine. Just wanted to ensure it was needed. All good. :+1:

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
edalex-ian: I guess I've been misled by the UI:

image

So the question to you then is: What's "Secondary Text Color" vs "Sidebar Text Color".

From the above screenshot and your above description are you saying you originally mapped `menuTextColor` to "Secondary Text Color", and `menuItemTextColor` to "Sidebar Text Color"?

SammyIsConfused: That's correct, yeah. I vaguely remember a discussion as to what to call them - since clients wouldn't know their way around MUI we named them based upon what changed when that setting changed.

edalex-ian: But it seems to have been done the other way around. The UI is show the clients the Material Design names, and then the data model has this odd name. At the very least, we're going to need to fix the data model.

I guess though to get this over the line we're stuck with what is.

SammyIsConfused: They were renamed at some point - but we couldn't rename the variables themselves because they were stored in the DB as a JSON blob - it would have broken existing themes.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
SammyIsConfused: Here is what it looks like with 5 instead. 
image

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
edalex-ian: Interesting. So, the `font-family` didn't work? (Was overridden?)

If we do need to repeat it, should probably be in a variable.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
SammyIsConfused: image
Seems like the standard on component for this is headline-5 - where are you getting 4 from?

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
SammyIsConfused: It's removable - doing so now. 
It was more I was copying the exact rulesets for each typography from MUI.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
edalex-ian: From this file you've edited. You see it here:

https://github.com/openequella/openEQUELLA/blob/5ca1b1da8ef4e9a6f1fba84d693a3fc28e9f2aea/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss#L1120

And here:

https://github.com/openequella/openEQUELLA/blob/5ca1b1da8ef4e9a6f1fba84d693a3fc28e9f2aea/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss#L3508

The first one being the most important, as that's any h2 in the card equivalents.

SammyIsConfused: Ahhh I getcha now. Interesting. 
Well the review session we had the other day we saw it as headline-5. And it seems like the one used in the new UI in component is 5. 
Well, I could either change those two to lines 5 as well, or change mine to 4. Either way we'll need some new screenshots. Which would you suggest?

edalex-ian (approved):
Merge on. :+1:

PR #2202, opened 2020-08-19
Removed not-null assertions, as our eslint config no longer allows them.

mrblippy:
Another change that has to be made was removing a string type assertion, as eslint was also erroring on that

##### Checklist

##### Description of change

#1306 
This is to fix up some eslint errors that made their way into this branch, after Develop changes were merged into it

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/components/DateRangeSelector.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/components/DateRangeSelector.tsx
@@ -150,7 +150,8 @@ export const DateRangeSelector = ({
     // Both are undefined.
     const openRange = !start && !end;
     // Both are valid dates and start is equal or less than end.
-    const closedRange = isStartValid && isEndValid && start! <= end!;
+    const closedRange =
+      start && end ? isStartValid && isEndValid && start <= end : false;
     // Call onDateRangeChange for above four cases.
     if (openStart || openEnd || openRange || closedRange) {
       onDateRangeChange(stateDateRange);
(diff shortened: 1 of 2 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/components/DateRangeSelector.tsx
```diff
@@ -248,8 +249,8 @@ export const DateRangeSelector = ({
         value={dateRangeToDateOptionConverter(stateDateRange)}
         id="date_range_selector"
         labelId="date_range_selector_label"
-        onChange={(event) =>
-          handleQuickDateOptionChange(event.target.value as string)
+        onChange={(event: any) =>
```
ChristianMurphy, line 252: event should automatically have a type here:

```suggestion
        onChange={(event) =>
```

mrblippy: Putting any as the type was to get around event.target.value now being an Unknown type. 
ESLint complains aboutthe string assertion 'This assertion is unnecessary since it does not change the type of the expression'

But i'm open to other ways to get around this. Since using `any` feels like a bit of a hack

ChristianMurphy: try switching `target` to `currentTarget`.
`target` bubbles events up, if there were another input inside the input it would catch that, so it's type is dynamic and unknown.
`currentTarget` refers to the element which the listener is attached to, so it can be statically typed.

mrblippy: That seems to work nicely, thank you for your suggestion  @ChristianMurphy :+1: :+1:

SammyIsConfused (approved):

edalex-ian (approved):

PR #2216, opened 2020-08-21
Feature/Facet Selector

PenghaiZhang:
#1306 

##### Checklist

- [x] tests are included

##### Description of change

Here is another refine control - Facet Selector.  A new type representing Search page specific Classifications was firstly added in `SearchPage.tsx`.  A function transforming standard Classifications to the Search page Classifications was also added in and will be consumed in the next integration. 

The component does not have its own state so at this stage only props changes can trigger a re-render. 
Jest tests and Storybooks for this component are also included.

image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/__mocks__/FacetSelector.mock.ts
+++ b/Source/Plugins/Core/com.equella.core/js/__mocks__/FacetSelector.mock.ts
@@ -15,7 +15,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import type { SearchPageClassification } from "../tsrc/search/SearchPage";
+import type { SearchPageClassification } from "../tsrc/search/components/FacetSelector";
 
 export const classifications: SearchPageClassification[] = [
   {
--- a/Source/Plugins/Core/com.equella.core/js/__mocks__/FacetSelector.mock.ts
+++ b/Source/Plugins/Core/com.equella.core/js/__mocks__/FacetSelector.mock.ts
@@ -15,7 +15,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { SearchPageClassification } from "../tsrc/search/SearchPage";
+import type { SearchPageClassification } from "../tsrc/search/SearchPage";
 
 export const classifications: SearchPageClassification[] = [
   {
--- a/Source/Plugins/Core/com.equella.core/js/__mocks__/FacetSelector.mock.ts
+++ b/Source/Plugins/Core/com.equella.core/js/__mocks__/FacetSelector.mock.ts
@@ -1,0 +1,79 @@
+/*
+ * Licensed to The Apereo Foundation under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * The Apereo Foundation licenses this file to you under the Apache License,
+ * Version 2.0, (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { SearchPageClassification } from "../tsrc/search/SearchPage";
+
+export const classifications: SearchPageClassification[] = [
+  {
+    name: "Language",
+    categories: [
+      {
+        term: "scala",
+        count: 312,
+      },
+      {
+        term: "java",
+        count: 1212,
+      },
+      {
+        term: "python",
+        count: 222,
+      },
+      {
+        term: "php",
+        count: 612,
+      },
+      {
+        term: "c#",
+        count: 888,
+      },
+      {
+        term: "purescript",
+        count: 1,
+      },
+    ],
+    maxDisplay: 7,
+    orderIndex: 1,
+    showMore: false,
+  },
+  {
+    name: "City",
+    categories: [
+      {
+        term: "Hobart",
+        count: 111,
+      },
+      {
+        term: "Sydney",
+        count: 222,
+      },
+      {
+        term: "Adelaide",
+        count: 333,
+      },
+    ],
+    maxDisplay: 2,
+    orderIndex: 0,
+    showMore: true,
+  },
+  {
+    name: "Color",
+    categories: [],
+    orderIndex: 2,
+    showMore: false,
+  },
+];
--- a/Source/Plugins/Core/com.equella.core/js/__mocks__/FacetSelector.mock.ts
+++ b/Source/Plugins/Core/com.equella.core/js/__mocks__/FacetSelector.mock.ts
@@ -42,7 +42,8 @@ export const classifications: SearchPageClassification[] = [
         count: 888,
       },
       {
-        term: "purescript",
+        term:
+          "purescript - the last version is xxxxxxxxx because maintenance is so hard!!!",
         count: 1,
       },
     ],
--- a/Source/Plugins/Core/com.equella.core/js/__mocks__/FacetSelector.mock.ts
+++ b/Source/Plugins/Core/com.equella.core/js/__mocks__/FacetSelector.mock.ts
@@ -19,6 +19,7 @@ import type { SearchPageClassification } from "../tsrc/search/components/FacetSe
 
 export const classifications: SearchPageClassification[] = [
   {
+    id: 766942,
     name: "Language",
     categories: [
       {
@@ -52,6 +53,7 @@ export const classifications: SearchPageClassification[] = [
     showMore: false,
   },
   {
+    id: 766943,
     name: "City",
     categories: [
       {
@@ -72,6 +74,7 @@ export const classifications: SearchPageClassification[] = [
     showMore: true,
   },
   {
+    id: 766944,
     name: "Color",
     categories: [],
     orderIndex: 2,
--- a/Source/Plugins/Core/com.equella.core/js/__stories__/search/FacetSelector.stories.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/__stories__/search/FacetSelector.stories.tsx
@@ -1,0 +1,36 @@
+/*
+ * Licensed to The Apereo Foundation under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * The Apereo Foundation licenses this file to you under the Apache License,
+ * Version 2.0, (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { action } from "@storybook/addon-actions";
+import * as React from "react";
+import { FacetSelector } from "../../tsrc/search/components/FacetSelector";
+import * as FacetSelectorMock from "../../__mocks__/FacetSelector.mock";
+
+export default {
+  title: "Search/FacetSelector",
+  component: FacetSelector,
+};
+const commonProps = {
+  classifications: FacetSelectorMock.classifications,
+  onSelectTermsChange: action("on select terms"),
+  onShowMore: action("on Show more"),
+};
+export const termsSelected = () => (
+  <FacetSelector selectedTerms={["scala", "QLD"]} {...commonProps} />
+);
+
+export const noTermsSelected = () => <FacetSelector {...commonProps} />;
--- a/Source/Plugins/Core/com.equella.core/js/__stories__/search/FacetSelector.stories.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/__stories__/search/FacetSelector.stories.tsx
@@ -32,39 +32,43 @@ export default {
   },
 } as Meta<FacetSelectorProps>;
 
+const FacetSelectorTemplate = (args: FacetSelectorProps) => (
+  <FacetSelector {...args} />
+);
 const selectedClassificationTerms = new Map<string, string[]>([
   ["Language", ["scala"]],
   ["City", ["Hobart"]],
 ]);
-export const termsSelected: Story<FacetSelectorProps> = (args) => (
-  <FacetSelector {...args} />
-);
-termsSelected.args = {
-  classifications: FacetSelectorMock.classifications,
-  selectedClassificationTerms: selectedClassificationTerms,
+const showMoreClassification = {
+  ...FacetSelectorMock.classifications[0],
+  showMore: true,
 };
 
-export const noTermsSelected: Story<FacetSelectorProps> = (args) => (
-  <FacetSelector {...args} />
+export const noTermsSelected: Story<FacetSelectorProps> = FacetSelectorTemplate.bind(
+  {}
 );
 noTermsSelected.args = {
   classifications: FacetSelectorMock.classifications,
 };
 
-export const showMore: Story<FacetSelectorProps> = (args) => (
-  <FacetSelector {...args} />
+export const termsSelected: Story<FacetSelectorProps> = FacetSelectorTemplate.bind(
+  {}
 );
-showMore.args = {
-  classifications: [
-    { ...FacetSelectorMock.classifications[0], showMore: true },
-  ],
+termsSelected.args = {
+  ...noTermsSelected.args,
+  selectedClassificationTerms: selectedClassificationTerms,
 };
 
-export const notShowMore: Story<FacetSelectorProps> = (args) => (
-  <FacetSelector {...args} />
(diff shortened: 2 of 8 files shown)


Source/Plugins/Core/com.equella.core/js/__mocks__/FacetSelector.mock.ts
```diff
@@ -0,0 +1,79 @@
...
+ * limitations under the License.
+ */
+import { SearchPageClassification } from "../tsrc/search/SearchPage";
+
+export const classifications: SearchPageClassification[] = [
+  {
+    name: "Language",
+    categories: [
+      {
+        term: "scala",
```
SammyIsConfused, line 25: It's probably worth having a really long name in one of these to show the wrapping.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/components/FacetSelector.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+  fireEvent,
+  RenderResult,
+  getAllByRole,
+} from "@testing-library/react";
+import { FacetSelector } from "../../../../tsrc/search/components/FacetSelector";
+import * as FacetSelectorMock from "../../../../__mocks__/FacetSelector.mock";
+import "@testing-library/jest-dom/extend-expect";
+
+describe("<FacetSelector />", () => {
+  // Mocked callbacks.
```
SammyIsConfused, line 32: nit: comments don't need full stops

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/FacetSelector.tsx
```diff
@@ -0,0 +1,189 @@
...
+import * as OEQ from "@openequella/rest-api-client";
+import { SearchPageClassification } from "../SearchPage";
+
+const useStyles = makeStyles({
+  classificationList: {
+    maxHeight: 500,
+    overflow: "auto",
+  },
+  facetListItem: {
+    padding: 0,
```
SammyIsConfused, line 38: For single styles, didn't we decide to just add a style prop inline?

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/FacetSelector.tsx
```diff
@@ -0,0 +1,189 @@
...
+  const showMoreButton = (classificationName: string) => (
+    <ListItem>
+      <Grid container justify="center">
+        <Grid item>
+          <Button
+            id={`${classificationName}_show_more`}
+            variant="text"
+            onClick={() => onShowMore(classificationName)}
+          >
+            Show more
```
SammyIsConfused, line 101: Language string missing here.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/FacetSelector.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+  /**
+   * Render a list for a Classification's categories. Some categories
+   * may have facets not displayed due to the configured maximum display number.
+   * @param classification A Classification..
+   */
+  const renderCategories = (classification: SearchPageClassification) => {
+    const { categories, maxDisplay, showMore } = classification;
+    const facets = categories.map((facet) => renderFacet(facet));
+    // The number of displayed facets Depends on whether to show more
```
SammyIsConfused, line 154:
```suggestion
    // The number of displayed facets depends on whether to show more
```

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/FacetSelector.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+  /**
+   * Sort and render Classifications that have categories.
+   * For each Classification, a scroll bar and a 'Show more' button may or may not
+   * be added, depending on whether a classification has more categories to show or not.
+   */
+  const renderClassifications = classifications
+    .filter((c) => c.categories.length > 0)
+    .sort((c1, c2) => c1.orderIndex - c2.orderIndex)
+    .map((c) => {
```
SammyIsConfused, line 167: I'd avoid these single character parameter names - it would improve readability to have a meaningful identifier.

PenghaiZhang: Thanks @SammyIsConfused . All addressed!

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/FacetSelector.tsx
```diff
@@ -0,0 +1,185 @@
...
+  const renderCategories = (classification: SearchPageClassification) => {
+    const { categories, maxDisplay, showMore } = classification;
+    const facets = categories.map((facet) => renderFacet(facet));
+    // The number of displayed facets depends on whether to show more
+    // and the maximum display number.
+    return facets.slice(0, showMore ? maxDisplay : undefined);
+  };
+
+  /**
+   * Sort and render Classifications that have categories.
```
edalex-ian, line 153:
```suggestion
   * Sort and build Classifications that have categories.
```

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/FacetSelector.tsx
```diff
@@ -0,0 +1,185 @@
...
+   * For each Classification, a scroll bar and a 'Show more' button may or may not
+   * be added, depending on whether a classification has more categories to show or not.
+   */
+  const renderClassifications = classifications
```
edalex-ian, line 157:
```suggestion
  const buildClassifications = classifications
```

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/FacetSelector.tsx
```diff
@@ -0,0 +1,185 @@
...
+    .filter((classification) => classification.categories.length > 0)
+    .sort(
+      (prevClassification, nextClassification) =>
+        prevClassification.orderIndex - nextClassification.orderIndex
+    )
+    .map((classification) => {
+      const { name, showMore } = classification;
```
edalex-ian, line 164: Why not combine these two lines?

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/FacetSelector.tsx
```diff
@@ -0,0 +1,185 @@
...
+      return (
+        <ListItem divider key={name} id={`classification_${name}`}>
```
edalex-ian, line 166: What happens if `name` has spaces?

edalex-ian (changes requested):
Thanks @PenghaiZhang 

A few things:

* Let's start using the new Args and Controls for storybook;
* Not I've remove all most all words of render, as at that point all you are doing is generating data structures in mem;
* A few other bits of tidy up and consideration.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/FacetSelector.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+interface FacetSelectorProps {
+  /**
+   * A list of Classifications.
+   */
+  classifications: SearchPageClassification[];
+  /**
+   * A list of selected facets.
+   */
+  selectedTerms?: string[];
```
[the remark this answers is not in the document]
PenghaiZhang: No enough :smile: .
I did not commit another type for this field which is a Map<string, string[]>. I will bring that in this PR.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/components/FacetSelector.test.tsx
```diff
@@ -0,0 +1,128 @@
...
+    );
+
+  // Return a 'li' that represents a Classification.
+  const getClassificationByName = (container: HTMLElement, name: string) => {
+    const id = `#classification_${name}`;
+    const classification = container.querySelector(id);
+    if (!classification) {
+      throw new Error(`Unable to find Classification ${name}`);
+    }
+    return classification as HTMLElement;
```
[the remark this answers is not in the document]
PenghaiZhang: This is not working on my machine. Does it work on yours ?

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/components/FacetSelector.test.tsx
```diff
@@ -0,0 +1,128 @@
...
+  };
+
+  // Return a Classification's 'SHOW MORE' button.
+  const getShowMoreButton = (
```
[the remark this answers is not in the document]
PenghaiZhang: Good idea. If the functions throws an error then it's name should be `getByxxx`.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/components/FacetSelector.test.tsx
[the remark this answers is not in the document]
edalex-ian: hmmm, I thought it did on one of my new functions in the SearchPage.test.tsx refactor.

This it's this one (snipped off comments for readability here):

```typescript
const queryRefineSearchComponent = (
  container: Element,
  componentSuffix: string
): HTMLElement | null => {
  const id = `#RefineSearchPanel-${componentSuffix}`;
  return container.querySelector(id);
};

const getRefineSearchComponent = (
  container: Element,
  componentSuffix: string
): HTMLElement => {
  const e = queryRefineSearchComponent(container, componentSuffix);
  if (!e) {
    throw new Error(`Failed to find ${componentSuffix}`);
  }

  return e;
};
```

`getRefineSearchComponent` used to do the coercion. (Probably still does on this branch if you want to take a look/have a play.)

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/FacetSelector.tsx
```diff
@@ -0,0 +1,187 @@
...
+      />
+    </ListItem>
+  );
+
+  /**
+   * Render a list for a Classification's categories. Some categories
+   * may have facets not displayed due to the configured maximum display number.
+   * @param classification A Classification..
+   */
+  const renderCategories = (classification: SearchPageClassification) => {
```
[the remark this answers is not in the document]
PenghaiZhang: `ListItem` is not a type so do you like `ReactElement[]` instead ?

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/FacetSelector.tsx
PenghaiZhang: Because function `listCategories`, which is called inside this `map`,  takes a classification as its parameter. But I can change its signature.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/FacetSelector.tsx
```diff
@@ -0,0 +1,187 @@
...
+  /**
+   * Render a 'SHOW MORE' button for each Classification.
+   * @param classificationName The name of a Classification.
+   */
+  const showMoreButton = (classificationName: string) => (
+    <ListItem>
+      <Grid container justify="center">
+        <Grid item>
+          <Button
+            id={`${classificationName}_show_more`}
```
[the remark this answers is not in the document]
PenghaiZhang: hmmm, it turns out we don't need IDs for these buttons. So I will remove them.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/components/FacetSelector.test.tsx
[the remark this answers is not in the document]
PenghaiZhang: I think if you assign `container.querySelector(id)` to a new variable you would see the error. But, if we do 
`container.querySelector<HtmlElement>(id)` or specify the type of the variable, then the error is gone.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/FacetSelector.tsx
PenghaiZhang: Can we replace spaces with `_` ? Actually if there are another approach that can help me find a ListItem(a `li` in DOM) then I don't need these IDs.

PenghaiZhang: Ah I can use `.closest("li")` !!!

edalex-ian: That'd be okay, and we can do similar in Selenium. At least we've got an ID at the top level of the selector, so we narrow in anyway and don't search the _whole_ DOM.

Source/Plugins/Core/com.equella.core/js/__stories__/search/FacetSelector.stories.tsx
```diff
@@ -15,22 +15,52 @@
...
+  <FacetSelector {...args} />
 );
+showMore.args = {
+  classifications: [
+    { ...FacetSelectorMock.classifications[0], showMore: true },
+  ],
+};
 
-export const noTermsSelected = () => <FacetSelector {...commonProps} />;
+export const notShowMore: Story<FacetSelectorProps> = (args) => (
```
PenghaiZhang, line 63: I am repeating this pattern to create stories:  creating a story and configuring its arguments. I probably can create a `template` to reduce repeating.

PenghaiZhang:
Two more Storybooks

1. Show more button is visible.
image

2. The button is invisible.
image

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/FacetSelector.tsx
```diff
@@ -0,0 +1,221 @@
...
+export interface FacetSelectorProps {
+  /**
+   * A list of Classifications.
+   */
+  classifications: SearchPageClassification[];
+  /**
+   * A map where the key is a Classification's name and value is
+   * a list of terms.
+   */
+  selectedClassificationTerms?: Map<string, string[]>;
```
edalex-ian, line 60: Hmmm, I'm reluctant to mention this - as this is so close. But... I wonder if in `SearchFacetsModule.Classification` we should be bringing in the UUID from the server. :thinking: (I'm not sure why I didn't in the first place.)

Then we'd use the UUID rather than the name for the key here, but also when doing the handlers etc. That way if someone does decide to have two Classifications/Facets with the same name, we'll be safe.

What do you think?

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/FacetSelector.tsx
```diff
@@ -0,0 +1,221 @@
...
+   * @param classificationName The name of a Classification
+   * @param facet A facet
+   */
+  const facetListItem = (
+    classificationName: string,
+    facet: OEQ.SearchFacets.Facet
+  ): ReactElement => {
+    const { term, count } = facet;
+    return (
+      <ListItem key={`${term} ${count}`} style={{ padding: 0 }}>
```
edalex-ian, line 150: Interesting. Did realise you could have spaces in the `key`, but I guess at the end of the day internally React is probably just using the string.

I assume this all works, so no need to change anything. Just observing.

edalex-ian (approved):
Looking good @PenghaiZhang . Just wonder if we can get away with using Classification name. :thinking:

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/FacetSelector.tsx
PenghaiZhang: :thinking:  I would like to avoid any potential errors. What about adding a `-` between term and count.

edalex-ian: Up to you. I don't know for 100% sure if terms are always devoid of spaces (and/or punctuation).

PenghaiZhang: Just had a conversation with Tom regarding this. As we can't guarantee `${term} ${count}` is unique, in this case maybe we have to use `index` . What do you think @edalex-ian ?

edalex-ian: Well, now that you have `id` from Classification in, how about `${id}:${term}`?

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchFacetsModule.ts
```diff
@@ -94,7 +98,10 @@ export const listClassifications = async (
 ): Promise<Classification[]> =>
   Promise.all(
     (await getFacetsFromServer()).map<Promise<Classification>>(
-      async (settings) => ({
+      async (settings, index) => ({
+        // We know IDs won't be undefined here, but due to its type being number | undefined,
+        // we have to do a nullish coalescing
+        id: settings.id ?? index,
```
edalex-ian, line 104: Hmmm, so it's only undefined if it's dirty. We should not have any 'dirty' entries here, so perhaps we should throw an error. If we leave like this it will probably be fine, but if we ever get corrupt data this could compound the situation.

Up to you.

edalex-ian (approved):
Good stuff @PenghaiZhang . Up to you with the `id`..

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchFacetsModule.ts
PenghaiZhang: I prefer leave it like this for now.

PR #2240, opened 2020-08-27
Refactor SearchPage.test.tsx and fix DateModifiedSelector

edalex-ian:
##### Checklist

- [x] tests are included

##### Description of change

First and foremost this was about refactoring `SearchPage.test.tsx` to use React Testing Library (RTL) rather than Enzyme. As part of doing this, there was a need to understand how to interact with MUI `<Select>` components. To build this understanding `SearchPagination.test.tsx` provided a good platform and so it was also refactored.

Fixing `SearchPage.test.tsx` highlighted that it was returning a false positive for selecting a quick date via the `DateModifiedSelector` component. This was broken recently due to the work done in #2202 . It feels this was a good example of the benefits of the move to RTL.

With the refactor I also brought in the user-events module, and then have attempted to use this in most places - there are more which later can be refactored to do the same.

There was one test I had to drop, and that was the checking for the display of the progress bar on loading results. This relied on the Enzyme way of `component.update()` to check before and after, but I think in RTL we more so need to do real-time and I couldn't quite unravel it. However we do have a Storybook story for it, and see it's non-critical I think this is fine.


Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
```diff
@@ -582,22 +561,11 @@ describe("<SearchPage/>", () => {
     });
   });
 
-  it("should display a spinner when search is in progress", async () => {
```
edalex-ian, line 585: This is the one test I did not replicate.

Source/Plugins/Core/com.equella.core/js/tsrc/components/DateRangeSelector.tsx
```diff
@@ -250,7 +250,7 @@ export const DateRangeSelector = ({
         id="date_range_selector"
         labelId="date_range_selector_label"
         onChange={(event) =>
-          handleQuickDateOptionChange(event.currentTarget.value as string)
+          handleQuickDateOptionChange(event.target.value as string)
```
edalex-ian, line 253: Fix for issue from #2202

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchOrderSelect.tsx
```diff
@@ -58,18 +58,28 @@ export const SearchOrderSelect = ({
...
-        </MenuItem>
-      ))}
-    </Select>
+    <>
+      <InputLabel id={labelId} style={{ display: "none" }}>
+        Search Order
+      </InputLabel>
+      <Select
+        id={baseId}
+        labelId={labelId}
```
edalex-ian, line 71: To improve accessibility, but also testability. Follows MUI recommendations - as far as using `InputLabel` etc. However I put the `display: none` so as to ensure it looked no different to before.

edalex-ian:
That's odd. The GHA `build_import_export_tool` job failed due to missing SBT dep - which is obviously nothing to do with this. I'm going to ignore that. (Especially seeing the `build_and_check` passed which actually does the TS side of things, including running the tests.)

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/components/SearchPagination.test.tsx
```diff
@@ -39,87 +41,103 @@ describe("<SearchPagination/>", () => {
     );
 
   it("Goes back to the first page of results when First Page Button is clicked", () => {
-    //currently on page 3 of 3
-    const lastPage = searchPagination(30, 1, 10);
-    const { firstPageButton } = paginatorControls(lastPage);
+    // currently on page 3 of 3
+    const { container } = searchPagination(30, 1, 10);
```
PenghaiZhang, line 45: The comment for this line seems wrong.  It's on page 2 of 3 because the second parameter is 1.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/components/SearchPagination.test.tsx
```diff
@@ -39,87 +41,103 @@ describe("<SearchPagination/>", () => {
...
+    // currently on page 1 of 2
+    const { container } = searchPagination(20, 0, 10);
+    const { getNextPageButton } = queryPaginatorControls(container);
 
-    nextPageButton.simulate("click");
+    userEvent.click(getNextPageButton());
     expect(mockOnPageChange).toHaveBeenCalledWith(1);
   });
 
   it("Goes to the last page of results when Last Page Button is clicked", () => {
```
PenghaiZhang, line 70: It seems like we can do a parameterised test for above 4 test cases ? But that probably does not bring in many benefits.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/components/SearchPaginationTestHelper.ts
```diff
@@ -15,29 +15,131 @@
...
 
-/**
- * Provides a helper function for east access to pagination controls
- * Takes a ReactWrapper, representing SearchPagination component
- * @return {ReactWrapper} Wrappers for navigating to First, Previous, Next, and Last Page, and PageCount of pagination controls
- */
-export const paginatorControls = (
-  component: ReactWrapper
-): PaginatorControls => {
+interface PaginatorQueries extends PaginatorBaseQueries {
```
PenghaiZhang, line 47: Is there any reason why you want to have another interface extending the base one rather than put below getXXX in the base interface ?  And They have the same comment. Should they have different comments to let people know the get ones throw exceptions and the query ones don't ?

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchOrderSelect.tsx
PenghaiZhang: I actually feel  without `display: none` the UI looks better for me. This is because when the default sorting has not been configured, it is undefined. As a result, this control shows nothing. But if a lable is showed there, it helps me know this control is about sorting
image

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
```diff
@@ -143,13 +141,48 @@ const queryRefineSearchComponent = (
...
+  if (rawMode) {
+    const rawModeSwitch = container.querySelector("#rawSearch");
+    if (!rawModeSwitch) {
+      throw new Error("Failed to find the raw mode switch!");
+    }
+    userEvent.click(rawModeSwitch);
+  }
+  const _queryBar = () => getQueryBar(container);
+  // Would be nice to replace this with a userEvent.type like:
+  //   await act(async () => await userEvent.type(_queryBar(), query, {delay: 1}));
```
PenghaiZhang, line 179: Have you tried calling `advanceTimersByTime(1000)` after this await ?

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
```diff
@@ -611,44 +579,31 @@ describe("<SearchPage/>", () => {
...
 
-  it("should support selecting a date range through Quick options", async () => {
-    component.update();
-    const quickOptions = component.find("#date_range_selector input");
-    await awaitAct(() =>
-      quickOptions.simulate("change", { target: { value: "Today" } })
+  it("filters by selected collection", async () => {
+    const targetCollection: Collection = getCollectionMap[0];
+    userEvent.click(
+      page.getByLabelText(languageStrings.searchpage.collectionSelector.title)
```
PenghaiZhang, line 598: I was very struggling to do this with Enzyme!  So `userEvent.click` works for MUI `<Autocomplete />`!!

PenghaiZhang (commented):
Thanks for this big refactoring @edalex-ian !  Looks good to me. Just some queries.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchOrderSelect.tsx
edalex-ian: Yup, I agree. Get's a bit wacky though when you do then select a value, and I was wondering if that empty case is actually just an anomaly from our dev environments.

Maybe we should raise an issue around this to fix next sprint. :thinking:

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/MuiTestHelpers.ts
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Use to 'select' an option from a MUI `<Select>`. This is done through a series of
+ * userEvent.click() calls.
+ *
+ * @param container The base container from which to start a search
+ * @param selector A CSS selector to pass to `HTMLElement.querySelector()` to find the `<Select>`
+ * @param optionText The text which represents the option you wish to select - user visible.
+ */
+export const selectOption = (
```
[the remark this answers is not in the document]
edalex-ian: From what I saw, it looks like this could possibly work with `<Autocomplete>`s. But for the one I was testing it didn't have an ID but did have a rather specific text label ('Collections') so I didn't end up using `selectOption`. But the code I did use was basically the same - just started with a `getByText` or `getByLabel` on the rendered component.

(The other issue I had trying to check with `<Autocomplete>`, is when attempting to inspect the DOM to check, the autocomplete would close the drop down list and so I couldn't see for sure.)

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/MuiTestHelpers.ts
```diff
@@ -0,0 +1,48 @@
...
+) => {
+  const muiSelect = container.querySelector(selector);
+  if (!muiSelect) {
+    throw Error("Unable to find MUI Select.");
+  }
+
+  // Click the <Select>
+  userEvent.click(muiSelect);
+  // .. then click the option in the list
+  userEvent.click(
```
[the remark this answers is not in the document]
edalex-ian: That's correct. I wasted a good number of hours until I discovered MUI selects register `onMouseDown` not `onClick`. `userEvent` allows you to not worry about this - just like a user wouldn't worry about it.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/components/SearchPagination.test.tsx
edalex-ian: Yeah, I agree. I left them as is, as I thought I might be missing something. But now that you mention it too. I'm going to go and rework them.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/components/SearchPagination.test.tsx
edalex-ian: You may be right. I have tried to avoid significant structural changes in that regard. I think that'd be a subsequent refactor - as then it's easier to ensure we're not losing bits and pieces.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/components/SearchPaginationTestHelper.ts
edalex-ian: Sure is a reason.

You'll note that first I setup so called 'baseQueries', and then I build the other functions off that. Having the two interfaces allows me to build up the first (without the need for optionals) and then build up the other.

For example, say I had:

```typescript
interface SomeInterface {
  a: () => HTMLElement | null;
  b: () => HTMLElement
}
```

How do I then build that object such that `b` is created based on `a`. AFAIK I can't do:

```typescript
return {
  a: () => {container.querySelector('body');
  b: () => {
    const e = a();
    if( !e ) {
      throw new Error("no e");
    }

    return e;
  }
}
```

???

As for the comment, you are correct. I could work these more.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
edalex-ian: It does that down in the waitFor down on line 184 doesn't it? Or do you mean something else?

Of note, it's the `expect` in the `waitFor` which fails. It only ever sees the last character of `query`. For example, if `query` was `abcd` it says the `expect` fails as the value is only `"d"` rather than `"abcd"`. :confused: 

(Keep in mind the waitFor loops, so it possibly even does it more.)

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
edalex-ian: Sure does! Nice an easy. (And this is almost the same pattern as the `selectOption` helper.)

edalex-ian:
Hmmm, flaky GHA newUI failure: `com.tle.webtests.test.myresources.MyResourcesTest -> testAddMyResource`.

I'm going to address some feedback, so I'll wont retry yet.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/MuiTestHelpers.ts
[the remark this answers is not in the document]
PenghaiZhang: This is the same issue I have !

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/components/SearchPaginationTestHelper.ts
PenghaiZhang: I get your idea! Cool.

edalex-ian:
Alright. Fixed up the comments. If you're happy, and if I can manage a green build, I'll get this merged in. :crossed_fingers:

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
PenghaiZhang: I mean in your attempt with userEvent.type

PenghaiZhang (approved):

PR #2245, opened 2020-08-28
Facet Selector integration

PenghaiZhang:
#1306 

##### Checklist

- [x] tests are included

##### Description of change

This PR attempts to integrate the Facet Selector with Search page. Changes include:
1. Facet Selector manages the 'Show more/less' buttons by its own state so the extra interface was no longer needed; and
2. Selected terms are used to generate a Where clause for Search and Facet Search; and
3. Move the component to a separate MUI Card; and
4. Only add non-empty Collections to facet search param; and
5. Modify  interface `Classification`; and
6. Strip out the starting `/xml` for selected Schema node in the Facet dialog.

From the commit history, you may find that some significant changes were made, modified, and reverted. But most importantly, in the end the logic is not so complicated - as long as a term is selected, a search and an update of Classification list will be performed. We also worked out quite a few issues such as the Lucene query relationship between each term and how to avoid the `lowercase query` issue.

image


Source/Plugins/Core/com.equella.core/js/__stories__/search/FacetSelector.stories.tsx
```diff
@@ -58,17 +54,3 @@ termsSelected.args = {
   ...noTermsSelected.args,
   selectedClassificationTerms: selectedClassificationTerms,
 };
-
-export const showMore: Story<FacetSelectorProps> = FacetSelectorTemplate.bind(
```
PenghaiZhang, line 62: Now this is controlled by the component's state so we don't need these two stories (the remaining stories can also show/hide the buttons. I also did some research about how to directly modify state with Storybook to and this seems not to be encouraged (and well supported).

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -116,40 +116,32 @@ export const defaultPagedSearchResult: OEQ.Common.PagedResult<OEQ.Search.SearchR
...
-
-  if (!classificationTerms || classificationTerms.length === 0) {
-    return textQuery;
+  if (!classificationTerms) {
+    return undefined;
   }
 
-  const or = " OR ";
+  // Append '/xml' back to the schema node and generate a search condition for each term.
+  const processNodeTerms = ({ node, terms }: NodeAndTerms) =>
```
PenghaiZhang, line 138: Do we want to move this function outside ?

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -115,6 +120,15 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
...
   }, [searchPageOptions]);
 
+  // When Search options get changed, also update the Classification list.
+  useEffect(() => {
+    if (!isInitialSearch.current) {
+      listClassifications(searchPageOptions).then((classifications) =>
+        setClassifications(classifications)
+      );
+    }
+  }, [searchPageOptions]);
```
PenghaiZhang, line 130: Separate the update of Classification list from the first `useEffect` to help understand the logic more easily.

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -143,13 +157,15 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
       ...searchPageOptions,
       query: query,
       currentPage: 0,
+      classificationTerms: undefined,
```
PenghaiZhang, line 160: You will see more `classificationTerms: undefined,` below, but the benefit of doing this is we don't need to put  `classificationTerms` into state, which does not break 'New search' and 'Back to Search page'. So I feel the repeating is fine.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/components/FacetSelector.test.tsx
```diff
@@ -41,16 +40,19 @@ describe("<FacetSelector />", () => {
...
   const HOBART = "Hobart";
-  const mockedSelectedTerms = new Map([[CITY_ID, [HOBART]]]);
-  // The text of 'SHOW MORE' button
+  const SCHEMA_NODE = "item/city";
+  const mockedSelectedTerms = new Map([
+    [CITY_ID, { node: SCHEMA_NODE, terms: [HOBART] }],
+  ]);
+  // The text of 'SHOW MORE' and 'SHOW LESS' buttons
   const SHOW_MORE = languageStrings.searchpage.facetSelector.showMoreButton;
+  const SHOW_LESS = languageStrings.searchpage.facetSelector.showLessButton;
```
mrblippy, line 49: These two language strings will likely be used throughout the app. They should probably go inside languageStrings.common.action

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/FacetSelector.tsx
```diff
@@ -39,78 +39,98 @@ const useStyles = makeStyles({
...
    * A list of Classifications.
    */
-  classifications: SearchPageClassification[];
+  classifications: Classification[];
   /**
    * A map where the key is a Classification's ID and value is
    * a list of terms.
    */
-  selectedClassificationTerms?: Map<number, string[]>;
+  selectedClassificationTerms?: Map<number, NodeAndTerms>;
```
edalex-ian, line 64: Can we reworking the naming here? It was confusing down below when reviewing `listCategories()` where you then start using the agreed terminology of 'categories' (rather than the server side, 'terms').

```typescript
/**
 * The currently selected categories, where the key of the map is the ID of
 * the `Classification` they below to.
 */
selectedCategories?: Map<number, NodeAndTerms>;
```

Hmmm, I wonder why you even need `NodeAndTerms` here. I mention this further down, I think you should only need the Classification ID and the string array of terms. :thinking:

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/FacetSelector.tsx
```diff
@@ -171,19 +195,27 @@ export const FacetSelector = ({
...
-  }: SearchPageClassification): ReactElement[] =>
-    categories
+  const listCategories = (
+    { id, categories, schemaNode, maxDisplay }: Classification,
+    showMore: boolean
+  ): ReactElement[] => {
+    const selectedTerms = selectedClassificationTerms?.get(id)?.terms ?? [];
+    const selectedCategories = categories.filter((c) =>
+      selectedTerms.includes(c.term)
+    );
```
edalex-ian, line 209: As mentioned above on `selectedClassificationTerms`, let's normalise on our terminology. 

But... Is this even all required? Doesn't `selectedClassificationTerms` already hold all this value... Maybe it's because you need it in a different shape for TS, in which case maybe all you need is a `.map` over it. :thinking:

Source/Plugins/Core/com.equella.core/js/__mocks__/FacetSelector.mock.ts
```diff
@@ -15,9 +15,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import type { SearchPageClassification } from "../tsrc/search/components/FacetSelector";
 
-export const classifications: SearchPageClassification[] = [
+import { Classification } from "../tsrc/modules/SearchFacetsModule";
```
SammyIsConfused, line 19:
```suggestion
import type { Classification } from "../tsrc/modules/SearchFacetsModule";
```

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchFacetsModule.ts
```diff
@@ -49,6 +49,10 @@ export interface Classification {
    * The configured order in which this classification should be displayed.
    */
   orderIndex: number;
+  /**
+   * The configured Schema node of this classification.
```
SammyIsConfused, line 53:
```suggestion
   * The configured schema node of this classification.
```

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchFacetsModule.ts
```diff
@@ -59,18 +63,36 @@ export interface Classification {
...
+    } = options;
+    let searchFacetsParams: OEQ.SearchFacets.SearchFacetsParams = {
+      nodes: [],
+      q: query ? formatQuery(query, !rawMode) : undefined,
+      modifiedAfter: getISODateString(lastModifiedDateRange?.start),
+      modifiedBefore: getISODateString(lastModifiedDateRange?.end),
+      owner: owner?.id,
+      showall: isEqual(
+        status?.sort(),
+        OEQ.Common.ItemStatuses.alternatives.map((i) => i.value).sort()
```
SammyIsConfused, line 84: I'd suggest you rename `i` to something more meaningful.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchFacetsModule.ts
```diff
@@ -59,18 +63,36 @@ export interface Classification {
...
+      ),
+      where: generateWhereQuery(classificationTerms),
+    };
+    if (collections && collections.length > 0) {
+      searchFacetsParams = {
+        ...searchFacetsParams,
+        collections: collections.map((c) => c.uuid),
```
SammyIsConfused, line 91: I'd suggest you rename `c` to something more meaningful.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchFacetsModule.ts
```diff
@@ -81,7 +103,9 @@ const convertSearchOptions: (
 export const listCategories = async (
   options: OEQ.SearchFacets.SearchFacetsParams
 ): Promise<OEQ.SearchFacets.Facet[]> =>
-  (await OEQ.SearchFacets.searchFacets(API_BASE_URL, options)).results;
+  (await OEQ.SearchFacets.searchFacets(API_BASE_URL, options)).results.filter(
+    (r) => r.term
```
SammyIsConfused, line 107: As above - single letter identifier.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -116,40 +116,32 @@ export const defaultPagedSearchResult: OEQ.Common.PagedResult<OEQ.Search.SearchR
...
+    terms.map((t) => `/xml${node}='${t}'`);
```
SammyIsConfused, line 139: As above.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/FacetSelector.tsx
```diff
@@ -39,78 +39,98 @@ const useStyles = makeStyles({
...
     }
     const copiedMap = new Map(selectedClassificationTerms ?? []);
-    copiedMap.set(classificationID, copiedTerms);
+    copiedMap.set(classificationID, { node: schemaNode, terms: copiedTerms });
     onSelectTermsChange(copiedMap);
   };
 
   /**
-   * Render a 'SHOW MORE' button for each Classification.
+   * Create a button to help show more/less categories for each Classification.
```
SammyIsConfused, line 116:
```suggestion
   * Create a button to show more/less categories for each Classification.
```

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/FacetSelector.tsx
```diff
@@ -171,19 +195,27 @@ export const FacetSelector = ({
...
-    showMore,
-    maxDisplay,
-  }: SearchPageClassification): ReactElement[] =>
-    categories
+  const listCategories = (
+    { id, categories, schemaNode, maxDisplay }: Classification,
+    showMore: boolean
+  ): ReactElement[] => {
+    const selectedTerms = selectedClassificationTerms?.get(id)?.terms ?? [];
+    const selectedCategories = categories.filter((c) =>
```
SammyIsConfused, line 207: As above.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/FacetSelector.tsx
```diff
@@ -171,19 +195,27 @@ export const FacetSelector = ({
...
+      selectedTerms.includes(c.term)
+    );
+    const unselectedCategories = categories.filter(
+      (c) => !selectedTerms.includes(c.term)
```
SammyIsConfused, line 211: As above.

Source/Plugins/Core/com.equella.core/js/tsrc/util/langstrings.ts
```diff
@@ -339,7 +339,9 @@ export const languageStrings = {
       day: "Day",
     },
     facetSelector: {
+      title: "Classifications",
       showMoreButton: "Show more",
+      showLessButton: "Show less",
```
SammyIsConfused, line 344: These should probably be common strings.

SammyIsConfused (commented):
Looking good! Just a few tweaks.

edalex-ian (changes requested):
Wowsers! This Ended up a big PR hey. :) I'm really excited to see this coming in.

A few things to do.

* Looks like SearchModule.test.ts needs a bit of a tidy
* I think types need a bit of a tidy-up/rethink. The Map<Number,NodeAndTerms> is getting confusing, and indeed there is probably overlap here with existing types (see other comments).
* We may also want to move away from adding the newer 'Terms' on the front-end code and try and go with `category` or `facet`. It's getting pretty confusing in places.
* We've started to end up with really tight coupling with bidirectional type references between the presentation layer and the business layer. We need to address this, going forward, but here there is a case where it's further been done so while fixing that DateRange should be fixed at the same time.
* `FacetSelector`: Seems to be doing a whole lot of stuff with `schemaNode`, but doesn't really need to, and as a result that's complicating some function signatures unnecessarily.

BTY.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchFacetsModule.ts
edalex-ian: Single character is okay if it's in close context - like here and below in single use anonymous functions. So these are all good. :+1:

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/components/FacetSelector.test.tsx
edalex-ian: I pondered this too. Then I was wondering... What if people do want to control the text differently specifically on these buttons (vs. other show more/less buttons). If we want to support that, we'd want to keep it specific like this.

What do peeps think?

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchFacetsModule.ts
SammyIsConfused: Sure thing! I'll resolve them.

Source/Plugins/Core/com.equella.core/js/__stories__/search/FacetSelector.stories.tsx
PenghaiZhang: Yes, they are interactive in storybook now.

Source/Plugins/Core/com.equella.core/js/__mocks__/FacetSelector.mock.ts
PenghaiZhang: Thanks Sam. I always forget this :disappointed_relieved:

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchFacetsModule.ts
```diff
@@ -59,18 +63,36 @@ export interface Classification {
...
+      modifiedAfter: getISODateString(lastModifiedDateRange?.start),
+      modifiedBefore: getISODateString(lastModifiedDateRange?.end),
+      owner: owner?.id,
+      showall: isEqual(
+        status?.sort(),
+        OEQ.Common.ItemStatuses.alternatives.map((i) => i.value).sort()
+      ),
+      where: generateWhereQuery(classificationTerms),
+    };
+    if (collections && collections.length > 0) {
```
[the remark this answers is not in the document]
PenghaiZhang: The reason for dong this is to make sure `collections` is not included in the search params when it's undefined or it is empty.  If we pass `undefined` or an empty collections to the facet search api, we will get nothing returned. (but we can  make some changes on the api side to make it happy with `undefined` or an empty collections)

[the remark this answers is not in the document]
PenghaiZhang: And Typescript is not happy with `if (collections?.length > 0)` because collections?.length can be undefined. So we can do `if (collections?.length ?? 0 > 0)` instead.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/FacetSelector.tsx
PenghaiZhang: I think we may need because generating each category's label text needs full information of a `facet` (term and count) , but `selectedClassificationTerms` only knows the `term`.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/components/FacetSelector.test.tsx
PenghaiZhang: I prefer they stay here :smile: , what about other peeps ?

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
```diff
@@ -185,6 +198,10 @@ const changeQuery = async (
   });
 };
 
+const selectTerm = (container: HTMLElement, term: string) => {
```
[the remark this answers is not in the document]
PenghaiZhang: I prefer change to name to `clickCategory` because we use MUI `FormControlLabel` which should ensure clicking the label will tick the box.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -17,6 +17,7 @@
  */
 import * as OEQ from "@openequella/rest-api-client";
 import { API_BASE_URL } from "../config";
+import { NodeAndTerms } from "../search/components/FacetSelector";
```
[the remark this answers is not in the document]
PenghaiZhang: I agree. I also feel `DateRange` should be sitting in a `utility module` because it's not specific to Search page.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/components/FacetSelector.test.tsx
edalex-ian: I see @mrblippy changed them over in his PR. So I'll leave it for you guys to reach agreement. ;)

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/CategorySelector.tsx
```diff
@@ -0,0 +1,266 @@
...
+  /**
+   * Build a list for a Classification's categories. Some categories may have facets
+   * not displayed due to the configured maximum display number.
+   *
+   * @param id The ID of a Classification
+   * @param categories A list of terms to build into a list
+   * @param maxDisplay Default maximum number of displayed facets
+   * @param expanded Whether to show more categories or not
+   */
+  const listCategories = (
```
edalex-ian, line 206: As discussed, I'm finding this one a little tricky to read. Seems mainly due to variable names, and variable shadowing, but I think also compounded by the imperative nature of it.

I've come up with this alternative more declarative implementation:

```typescript
  const listCategories = (
    { id, categories, maxDisplay }: Classification,
    expanded: boolean
  ): ReactElement[] => {
    const selectedTerms: string[] =
      selectedCategories?.find((c) => c.id === id)?.categories ?? [];
    const selectedFacets: OEQ.SearchFacets.Facet[] = selectedTerms.map((selectedTerm) => {
      const facet = categories.find((c) => selectedTerm === c.term);
      if (!facet) {
        throw new Error(
          "Data integrity issue, we're unable to match a selected term."
        );
      }

      return facet;
    });
    const orderedFacets = selectedFacets.concat(
      categories.filter((c) => !selectedFacets.includes(c))
    );

    return orderedFacets
      .slice(0, expanded ? undefined : maxDisplay)
      .map((facet) => categoryListItem(id, facet));
  };
```

Two things interesting here:

1. Although it should never happen, this showed the potential for a data integrity issue that would be glossed over; and
2. Against my own advice I've been drawn into having to use Categories, Terms and Facets in the one function. But this was to align with data types and existing variable names. So, I dunno. :shrug: 

(Also, because this is declarative it highlights the data linkages/flow. I could've chained all this together, but I erred on the side of readability. :wink:  )

Source/Plugins/Core/com.equella.core/js/__stories__/search/CategorySelector.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+export default {
+  title: "Search/FacetSelector",
+  component: CategorySelector,
+  argType: {
+    onSelectedCategoriesChange: { action: "on select categories" },
+  },
+} as Meta<GategorySelectorProps>;
+
+const FacetSelectorTemplate = (args: GategorySelectorProps) => (
```
edalex-ian, line 35:
```suggestion
const CategorySelectorTemplate = (args: GategorySelectorProps) => (
```

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/modules/SearchModule.test.ts
```diff
@@ -25,72 +26,81 @@ const mockedSearch = (OEQ.Search.search as jest.Mock<
...
 describe("SearchModule", () => {
-  it("should provide an list of items", async () => {
-    const searchResult = await SearchModule.searchItems(
-      SearchModule.defaultSearchOptions
-    );
-    expect(searchResult.available).toBe(12);
-    expect(searchResult.results).toHaveLength(12);
-  });
+  describe("searchItems", () => {
+    it("should provide an list of items", async () => {
```
edalex-ian, line 30:
```suggestion
    it("should provide a list of items", async () => {
```

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/modules/SearchModule.test.ts
```diff
@@ -25,72 +26,81 @@ const mockedSearch = (OEQ.Search.search as jest.Mock<
...
+        schemaNode: "/item/language",
+        categories: ["Java", "Scala"],
+      },
+      {
+        id: 766943,
+        schemaNode: "/item/city",
+        categories: ["Hobart"],
+      },
+    ];
+    it("should generate a where clause for one category", () => {
```
edalex-ian, line 88:
```suggestion

    it("should generate a where clause for one category", () => {
```

edalex-ian (changes requested):
Good stuff @PenghaiZhang I think it's pretty well there now.

You'll see my feedback is really just formatting, typos and naming. Other than that just the listCategories function.

It looks like you've not refactored DateRange at this time. If that's not the case (please confirm) I'll put it on the backlog for next sprint or so.

edalex-ian (approved):
Looks GTG @PenghaiZhang 

Thanks for all your hard work on this. :+1:

PR #2255, opened 2020-09-03
EBP: improve edit attachment dialog styles

SammyIsConfused:
##### Checklist

##### Description of change
#1983 
- Increase size of the dialog
- Add border shadow and curved edges
- Hide redundant string
- Fix padding issues
- Hide unnecessary string that appears twice( "Upload files from your computer"
- Refactor all 8px and 16px margin/padding values into variables

Peek 2020-09-03 13-44


SammyIsConfused:
Gah, there's a few broken tests. Probably other dialogs I've inadvertently borked with these styles. I'll fix those up.

SammyIsConfused:
I don't think the build failures are caused by this branch - the same error happened in old and new UI in a test I don't touch. :thinking:

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -4167,10 +4206,9 @@ div.universaliframe {
   @include buttonShadowBorder;
   overflow: hidden;
   height: 650px;
-  width: 1250px;
-  // workaround - offset leftwards to centre panel. Can't use justify-content on fancybox-inner,
-  // as this breaks the popup taxonomy selector
-  margin-left: -240px;
+  width: 1175px;
+  //workaround to avoid editing fancybox-styles
+  margin-left: -200px;
```
edalex-ian, line 4211: This has been repeated. Looks like a good candidate for a mixin.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -4167,10 +4206,9 @@ div.universaliframe {
   @include buttonShadowBorder;
   overflow: hidden;
   height: 650px;
-  width: 1250px;
-  // workaround - offset leftwards to centre panel. Can't use justify-content on fancybox-inner,
-  // as this breaks the popup taxonomy selector
-  margin-left: -240px;
+  width: 1175px;
```
edalex-ian, line 4209: This appears to be the same width as `.fileHandler`. I'm gathering that is deliberate, and so should share a var?

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -4167,10 +4206,9 @@ div.universaliframe {
   @include buttonShadowBorder;
   overflow: hidden;
   height: 650px;
```
edalex-ian, line 4208: This is an interesting one, seeing the width here is matching the `.fileHanlder` I'm gathering this is also essentially matching the height but with an offset of 18px. It may therefore be worthwhile having a var for that standard height, and then expressing this as that height minus 18px. Although 18px seems very similar to 16px, so that makes me wonder if the offset should be the standard `doubleMargin`?

edalex-ian (commented):
Looks good @SammyIsConfused . Looks as though couple more variables could be used, as well as a mixin.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
SammyIsConfused: Refactored to use iframeDialogHeight + doubleMargin.

edalex-ian (approved):
Great job on the rework with the vars and additional mixin. Very readable. :+1:

Issue #2210, opened 2020-08-20
Cannot upgrade 2019.2.4 to later version


nelson-edalex:
**Describe the bug**
When oEQ is upgraded to 2019.2.4, further upgrade to 2020.1.X fails with the following error:

![image](https://user-images.githubusercontent.com/28520211/90834683-67327a80-e37d-11ea-9a8e-66f788a0945c.png)

**To Reproduce**

1. Upgrade to 2019.2.4.
2. Apply 2020.1.3 hotfix.

**Expected behavior**
EQUELLA upgrade successfully.

**Platform:**

- OpenEquella Version: 2020.1.3

PR #2257, opened 2020-09-04
Ensure no more complaints about migration for last known owner

PenghaiZhang:
#2210 

##### Checklist

- [ ] tests are included

##### Description of change

Previous fix can still cause trouble.
For example, on unstable, `v20202` of this migration has been applied,  but `v20192` and `v20201` are never applied. So database does not have records for these two migrations. As a result, the migration task finds two new migrations  `v20192` and `v20201` and try to run them, but fails because we don't actually provide matched files (beans) for them.

The builds in previous PR were green because none of `v20192`, `v20201` and `v20202` exist in the database. I am still not 100% sure the logic behind. From my understanding, the migration task should have found three migrations to run and failed due to the same reason, but it actually did succeed...

But anyway, the solution is to make sure  `v20192` and `v20201` have an available bean in 2020.2.

Manually done testing includes:
1. from 2019.1 to the latest; 
2. from 2019.2.4 to the latest; 
3. from 2020.1.3 to the latest;
4. from 2020.2 which is a few commits behind the latest to the latest (most importantly, did not apply 2020.1.3 and 2019.2.4 before);

### DIFF
--- a/Source/Plugins/Core/com.equella.core/plugin-jpf.xml
+++ b/Source/Plugins/Core/com.equella.core/plugin-jpf.xml
@@ -6024,16 +6024,20 @@
     <parameter id="bean" value="bean:com.tle.core.institution.migration.v20202.CreateFacetedSearchClassificationTable" />
     <parameter id="date" value="2020-04-22" />
   </extension>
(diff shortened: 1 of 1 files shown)


edalex-ian (approved):
LGTM! :+1:

PR #2258, opened 2020-09-07
Add support to REST Module for new Search term highlights 

edalex-ian:
##### Checklist

- [x] tests are included

##### Description of change

This mainly adds support to `oeq-ts-rest-api` to support the new `highlight` array which comes back from the search2 endpoint. In doing this though I noticed the return type for `search` was the `Common.PagedResult<T>` which does not match the server - and even has the `resumptionToken` which search2 does not use.

So inline with the Server definition I also created a specialised interface (`SearchResult`), one that didn't have a `resumptionToken` and instead does have a `highlight[]`. I could have looked at changing `PagedResult` to have a base class without `resumptionToken` and build from there, however server side these are not related so it made no sense to build in that relationship this side of the fence.

Once that was also done I tweaked the front-end to use it. Interestingly, because the types structurally are very similar TS was happy and everything still worked. However if these diverge further in the future we want to make sure things line up. And to further support this I changed the mocked data to TS with types.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/__mocks__/SearchResult.mock.ts
+++ b/Source/Plugins/Core/com.equella.core/js/__mocks__/SearchResult.mock.ts
@@ -17,7 +17,7 @@
  * limitations under the License.
  */
 import * as OEQ from "@openequella/rest-api-client";
-import * as _ from "lodash";
+import { range } from "lodash";
 import { v4 as uuidv4 } from "uuid";
 
 export const getEmptySearchResult: OEQ.Search.SearchResult<OEQ.Search.SearchResultItem> = {
@@ -309,7 +309,7 @@ export const getSearchResultsCustom = (
   start: 0,
   length: 10,
   available: numberOfResults,
-  results: _.range(numberOfResults).map((i) => ({
+  results: range(numberOfResults).map((i) => ({
     uuid: uuidv4(),
     name: `item ${i}`,
     version: 1,
--- a/Source/Plugins/Core/com.equella.core/js/__mocks__/SearchResult.mock.ts
+++ b/Source/Plugins/Core/com.equella.core/js/__mocks__/SearchResult.mock.ts
@@ -16,15 +16,19 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import * as OEQ from "@openequella/rest-api-client";
+import * as _ from "lodash";
+import { v4 as uuidv4 } from "uuid";
 
-exports.getEmptySearchResult = {
+export const getEmptySearchResult: OEQ.Search.SearchResult<OEQ.Search.SearchResultItem> = {
   start: 0,
   length: 0,
   available: 0,
   results: [],
+  highlight: [],
 };
 
-exports.getSearchResult = {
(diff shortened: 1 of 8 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -102,11 +102,12 @@ export const defaultSearchOptions: SearchOptions = {
   searchAttachments: true,
 };
 
-export const defaultPagedSearchResult: OEQ.Common.PagedResult<OEQ.Search.SearchResultItem> = {
+export const defaultPagedSearchResult: OEQ.Search.SearchResult<OEQ.Search.SearchResultItem> = {
   start: 0,
   length: 10,
   available: 10,
   results: [],
+  highlight: [],
```
PenghaiZhang, line 110: I wonder if `highlight` can be optional or not because a search may not have a query.

oeq-ts-rest-api/src/Search.ts
```diff
@@ -240,6 +240,32 @@ export interface SearchResultItem {
...
+   */
+  available: number;
+  /**
+   * The individual items which match the search
+   */
+  results: T[];
+  /**
+   * List of words to use to highlight when displaying the results
+   */
+  highlight: string[];
```
PenghaiZhang, line 266: As mentioned above, can `hightlight` be optional ?

PenghaiZhang (approved):
Looks good to me. Just one query.

Source/Plugins/Core/com.equella.core/js/__mocks__/SearchResult.mock.ts
```diff
@@ -16,15 +16,19 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import * as OEQ from "@openequella/rest-api-client";
+import * as _ from "lodash";
```
ChristianMurphy, line 20: is all of lodash needed in this scope?
Would
```suggestion
import { range } from "lodash";
```

work? (would also need a change from `_.range` to `range` below, on line 312)

ChristianMurphy (approved):

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
edalex-ian: The server always returns it.... Which I like as then there doesn't need to be the additional handling for values that may or may not be there.

oeq-ts-rest-api/src/Search.ts
edalex-ian: The server always returns it, just leaves it  as an empty array if there are no highlights - say, from an empty query.

Source/Plugins/Core/com.equella.core/js/__mocks__/SearchResult.mock.ts
edalex-ian: I'd say so. I just replaced like for like. I'll give it a tweak. :wink:

edalex-ian:
I'll merge on green unless anyone has anything else they'd like addressed.

PR #2259, opened 2020-09-09
Feature/facet selector integration

PenghaiZhang:
#1306 

##### Checklist

- [ ] tests are included

##### Description of change

According to the latest discussions, we want to refactor the Category Selector for below changes: 
1. Each Classification is joined by `AND`; 
2. Each category of a Classification is joined by `OR`; 
3. Selected categories that do not apply to current search criteria are still displayed and their counts are 0 so that users can deselect them;
4. To generate a where clause for updating one Classification, only categories selected from other Classifications are used.
4. The Classification panel do not show if all Classifications do not have categories for selecting.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/search/components/CategorySelector.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/search/components/CategorySelector.tsx
@@ -213,26 +213,29 @@ export const CategorySelector = ({
     const selectedTerms: string[] =
       selectedCategories?.find((c) => c.id === id)?.categories ?? [];
 
-    // Previously selected categories that still apply to current search criteria.
-    const selectedAndExist = categories.filter((c) =>
-      selectedTerms.includes(c.term)
+    // Generate two arrays for previously selected categories. One for those still
+    // applicable and the other one for those not applicable.
+    const [selectedApplicable, selectedNotApplicable] = selectedTerms.reduce<
+      [OEQ.SearchFacets.Facet[], OEQ.SearchFacets.Facet[]]
+    >(
+      ([applicable, notApplicable], term) => {
+        const applicableCategory = categories.find((c) => c.term === term);
+        return applicableCategory
+          ? [applicable.concat(applicableCategory), notApplicable]
+          : [applicable, notApplicable.concat({ term: term, count: 0 })];
+      },
+      [[], []]
     );
-
-    // Previously selected categories that do not apply to current search criteria.
-    // Their counts should be 0.
-    const selectedButDisappear = selectedTerms
-      .filter((t) => categories.every((c) => c.term !== t))
-      .map((t) => ({ term: t, count: 0 }));
 
     // Categories that apply to current search criteria but have not been selected.
-    const noSelected = categories.filter(
+    const notSelected = categories.filter(
       (c) => !selectedTerms.includes(c.term)
     );
-    // The concatenating order is selectedAndExist -> selectedButDisappear -> noSelected.
+    // The concatenating order is selectedApplicable -> selectedNotApplicable -> notSelected.
     const orderedFacets: OEQ.SearchFacets.Facet[] = [
-      ...selectedAndExist,
-      ...selectedButDisappear,
-      ...noSelected,
+      ...selectedApplicable,
+      ...selectedNotApplicable,
+      ...notSelected,
     ];
 
     return orderedFacets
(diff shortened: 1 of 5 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/search/components/CategorySelector.tsx
```diff
@@ -212,21 +212,28 @@ export const CategorySelector = ({
...
-        if (!facet) {
-          throw new Error(
-            "Data integrity issue, we're unable to match a selected term."
-          );
-        }
 
-        return facet;
-      }
+    // Previously selected categories that still apply to current search criteria.
+    const selectedAndExist = categories.filter((c) =>
```
PenghaiZhang, line 217: It was struggling for me to give proper names in this function. so HELP wanted.

edalex-ian:
@PenghaiZhang 

When you say:

> 1. Each Classification is joined by AND;
> 2. Each category of a Classification is joined by OR;

Are you just referring to when they're used for the queries for the Facet Search, not the item search?

PenghaiZhang:
> @PenghaiZhang
> 
> When you say:
> 
> > 1. Each Classification is joined by AND;
> > 2. Each category of a Classification is joined by OR;
> 
> Are you just referring to when they're used for the queries for the Facet Search, not the item search?

They are used for both.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/CategorySelector.tsx
edalex-ian: You could go with:

* `selectedApplicable`
* `selectedNotApplicable`
* `notSelected`

And to further this idea, it'd be great to maintain the relationships here strictly back to `selectedTerms`. So how about changing the first two sets to be processed with something like:

```typescript
const [selectedApplicable, selectedNotApplicable] =
  selectedTerms.reduce<[OEQ.SearchFacets.Facet[], OEQ.SearchFacets.Facet[]]>(
    ([applicable, notApplicable], term) => {
      const applicableCategory = category.find( c => c.term === term );
      return applicableCategory
        ? [applicable.concat(applicableCategory), notApplicable]
        : [applicable, notApplicable.concat({term: term, count: 0});
    }, [[], []]);
```

This is a bit more verbose, but it ties it the data structure it is being generated from, and so I think it's worth that.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/CategorySelector.tsx
```diff
@@ -212,21 +212,28 @@ export const CategorySelector = ({
...
-      categories.filter((c) => !selectedFacets.includes(c))
+
+    // Previously selected categories that do not apply to current search criteria.
+    // Their counts should be 0.
+    const selectedButDisappear = selectedTerms
+      .filter((t) => categories.every((c) => c.term !== t))
+      .map((t) => ({ term: t, count: 0 }));
+
+    // Categories that apply to current search criteria but have not been selected.
+    const noSelected = categories.filter(
```
edalex-ian, line 228: I'm guess this is a typo and should be:

```suggestion
    const notSelected = categories.filter(
```

edalex-ian (changes requested):
Fair effort @PenghaiZhang ! :+1: 

A couple of things:

* Need to rework JSDoc description for `generateCategoryWhereQuery`. The doco there is currently the old way, and I guess this might be the place to capture the business rules around what's been done here.
* Feedback given to your "HELP wanted" query. Up to you if you want to drop the `reduce` in there. :wink:

edalex-ian:
> > @PenghaiZhang
> > When you say:
> > > 1. Each Classification is joined by AND;
> > > 2. Each category of a Classification is joined by OR;
> > 
> > 
> > Are you just referring to when they're used for the queries for the Facet Search, not the item search?
> 
> They are used for both.

Ah yes, of course. So in my review feedback where I said putting the business rules on `generateCategoryWhereQuery`, probably half go there, but the other half are on `listClassifications` in `SearchFacetsModule` I guess.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/CategorySelector.tsx
PenghaiZhang: This is a very nice good use of reduce! I like it!

PenghaiZhang:
Verbally approved by Ian.

PR #2284, opened 2020-09-15
ci(github-actions): use actions setup node with nvmrc

ChristianMurphy:
##### Checklist

- [x] the [contributor license agreement][] is signed
- [x] commit message follows [commit guidelines][]
- [x] tests are included
- [ ] screenshots are included showing significant UI changes
- [ ] documentation is changed or added

##### Description of change

leverages github actions setup-node to remove need for `shell: bash -l -eo pipefail {0}`

based off https://github.com/actions/setup-node/issues/32#issuecomment-539794249

[contributor license agreement]: https://www.apereo.org/node/676
[commit guidelines]: https://chris.beams.io/posts/git-commit/

PR #2297, opened 2020-09-17
Include REST module's tests in GHA

PenghaiZhang:
#1480 ### Checklist

- [ ] tests are included

##### Description of change

The purpose of this PR is to include tests of module `oeq-ts-rest-api` in GHA.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -292,13 +292,6 @@ jobs:
         with:
           java-version: 1.8
 
-      - name: Install Chromium
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y --no-install-recommends \
-            chromium-browser \
-            chromium-chromedriver \
-
       - name: Download installer
         uses: actions/download-artifact@v1
         with:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -262,7 +262,9 @@ jobs:
 
   rest-module-testing:
     needs: build_and_check
+
     runs-on: ubuntu-latest
+
     services:
       postgres:
         image: postgres:9.6
@@ -278,6 +280,10 @@ jobs:
         ports:
           # Maps tcp port 5432 on service container to the host
           - 5432:5432
+
+    env:
+      AUTOTEST_CONFIG: autotest/github-actions.conf
+
     steps:
       - uses: actions/checkout@v1
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -318,17 +318,14 @@ jobs:
           ./sbt -jvm-opts autotest/.jvmopts "project autotest" \
             installEquella startEquella configureInstall setupForTests
 
-      - name: Set up NVM
-        shell: bash -l {0}
-        run: |
-          curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.3/install.sh | bash
-          export NVM_DIR="$HOME/.nvm"
-          echo "source $NVM_DIR/nvm.sh" >> "$HOME/.bash_profile"
+      - name: Read .nvmrc
+        run: echo ::set-output name=NVMRC::$(cat .nvmrc)
+        id: nvm
 
-      - name: Install NodeJS
-        shell: bash -l {0}
-        run: |
-          nvm install
+      - name: Setup node
+        uses: actions/setup-node@v1
+        with:
+          node-version: "${{ steps.nvm.outputs.NVMRC }}"
 
       - name: Run tests
         working-directory: oeq-ts-rest-api
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -167,8 +167,6 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install -y --no-install-recommends \
-            chromium-browser \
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yaml
```diff
@@ -259,3 +259,78 @@ jobs:
...
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+
+      - name: Install Chromium
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends \
+            chromium-browser \
+            chromium-chromedriver \
```
ChristianMurphy, line 300: FWIW the github linux virtual environment comes with Chrome + ChromeDriver and Firefox + GeckoDriver pre-installed https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#ubuntu-20041-lts

.github/workflows/ci.yaml
```diff
@@ -259,3 +259,78 @@ jobs:
...
+        shell: bash -l {0}
+        run: |
+          curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.3/install.sh | bash
+          export NVM_DIR="$HOME/.nvm"
+          echo "source $NVM_DIR/nvm.sh" >> "$HOME/.bash_profile"
+
+      - name: Install NodeJS
+        shell: bash -l {0}
+        run: |
+          nvm install
```
ChristianMurphy, line 331: https://github.com/openequella/openEQUELLA/pull/2284 may work here as well

ChristianMurphy (commented):
Thanks @PenghaiZhang!
It's good to see more tests making their way into CI! :+1:

.github/workflows/ci.yaml
PenghaiZhang: Thanks!!  I should action `setup-node`!

.github/workflows/ci.yaml
PenghaiZhang: Nice! Thanks @ChristianMurphy !

ChristianMurphy: hmmm it looks like oEQ may be hard coded to look for chromium, even if chrome is present.

PenghaiZhang: Yes, in `github-actions.conf` we specify where to look for chromedriver. Do you know where is the pre-installed chromedriver ? I am about to do some search.

edalex-ian: It's in the autotest.conf. So if you change this, make sure you also change it for the functional test.

ChristianMurphy: > Do you know where is the pre-installed chromedriver ?

I'm not sure the exact location, but the docs mention

> Chrome Driver is available via CHROMEWEBDRIVER environment variable

so I'm guessing using that directly, or echo'ing it and the coding it in would work.

https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#ubuntu-20041-lts

PenghaiZhang: Yes I also found this. I will push a commit soon.

PenghaiZhang: https://github.com/actions/virtual-environments/issues/248

edalex-ian (commented):
LGTM... mostly. :grin: 

Ideally it'd be great to see if we could create a custom action which did the four steps of setting up oEQ - i.e. setting up the DB, downloading the artefacts and setting up oEQ. We could put this in its own repo, and then later when we split this out we could share it. But, maybe that's a bit much for here...

... hmmm, actually, then again, those four steps are now duplicated so we should probably do something.. Either do a custom action just in this repo, or maybe use a YAML reference.

What do you think?

PenghaiZhang:
> LGTM... mostly. 
> 
> Ideally it'd be great to see if we could create a custom action which did the four steps of setting up oEQ - i.e. setting up the DB, downloading the artefacts and setting up oEQ. We could put this in its own repo, and then later when we split this out we could share it. But, maybe that's a bit much for here...
> 
> ... hmmm, actually, then again, those four steps are now duplicated so we should probably do something.. Either do a custom action just in this repo, or maybe use a YAML reference.
> 
> What do you think?

I prefer to create a composite run steps action. Do you it's good to try ?

edalex-ian:
Yup, definitely! That's what I meant by this clumsy bit of wording:

> custom action just in this repo

PenghaiZhang:
> Yup, definitely! That's what I meant by this clumsy bit of wording:
> 
> > custom action just in this repo
> 
> Oh, other than it has to go in another repo... But ah well, that'll set us up for when we split out.

Yes! So would you like to create a new repo under `openequella`?

edalex-ian:
Hmmm, I'm not sure it'll work.. Based on [a comment from a GH staffer](https://github.community/t/reusing-sharing-inheriting-steps-between-jobs-declarations/16851/17) you may only be able to use these for `run` actions:

> You can use composite run steps to make your run steps reusable across actions 101 (but not use actions within actions)

Maybe we're facing a fundamental DRY issue with GHA. But there's lots of discussion out there about them adding YAML Anchors etc, so maybe in the future we'll have better options. :disappointed: 

Perhaps we leave this (as is) for now. :thinking:

edalex-ian:
So... Yeah, I reckon let's merge now. :+1:

PR #2298, opened 2020-09-17
Highlight search terms in the results

edalex-ian:
##### Checklist

- [x] tests are included

##### Description of change

This change highlights search terms in the title and description of search result items in the same way (algorithmically) as the Legacy Search UI. The algorithm around this was mainly in java class `com.tle.web.sections.render.TextUtils`. As a result, a key part of this PR was to convert that to TS (`utils/TextUtils.ts`).

Once that was in place, it was just a matter of weaving it in, integrating with the 'highlight' list returned as part of #2225. 

To assist testing, I also refactored the `SearchResult.stories.tsx` to Storybook 6 before adding an additional story.

#1306 

image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/__stories__/search/SearchResult.stories.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/__stories__/search/SearchResult.stories.tsx
@@ -15,131 +15,43 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import * as OEQ from "@openequella/rest-api-client";
+import type { Story } from "@storybook/react";
 import * as React from "react";
 import * as mockData from "../../__mocks__/searchresult_mock_data";
 import SearchResult from "../../tsrc/search/components/SearchResult";
-import { boolean, number, object, text } from "@storybook/addon-knobs";
 
 export default {
   title: "Search/SearchResult",
   component: SearchResult,
 };
-const defaultItemName = "Item Name";
-const defaultItemDescription = "Item Description";
 
-export const BasicSearchResultComponent = () => (
-  <SearchResult
-    name={text("name", mockData.basicSearchObj.name || defaultItemName)}
-    version={number("version", mockData.basicSearchObj.version)}
-    uuid={text("uuid", mockData.basicSearchObj.uuid)}
-    description={text(
-      "description",
-      mockData.basicSearchObj.description || defaultItemDescription
-    )}
-    displayFields={object("display fields", [
-      ...mockData.basicSearchObj.displayFields,
-    ])}
-    modifiedDate={object("modified date", mockData.basicSearchObj.modifiedDate)}
-    createdDate={object("created date", mockData.basicSearchObj.createdDate)}
-    status={text("item status", mockData.basicSearchObj.status)}
-    displayOptions={object(
-      "display options",
-      mockData.basicSearchObj.displayOptions
-    )}
-    attachments={object("attachments", [
-      ...mockData.basicSearchObj.attachments,
-    ])}
-    links={object("links", mockData.basicSearchObj.links)}
-    collectionId={text("collection ID", mockData.basicSearchObj.collectionId)}
-    commentCount={number("comment count", mockData.basicSearchObj.commentCount)}
-    thumbnail={text("thumbnail", mockData.basicSearchObj.thumbnail)}
-    keywordFoundInAttachment={boolean(
-      "keywordFoundInAttachment",
-      mockData.basicSearchObj.keywordFoundInAttachment
-    )}
-  />
+export const BasicSearchResult: Story<OEQ.Search.SearchResultItem> = (args) => (
+  <SearchResult {...args} />
 );
+BasicSearchResult.args = {
+  ...mockData.basicSearchObj,
+};
 
-export const AttachmentSearchResultComponent = () => (
-  <SearchResult
-    name={text("name", mockData.attachSearchObj.name || defaultItemName)}
-    version={number("version", mockData.attachSearchObj.version)}
-    uuid={text("uuid", mockData.attachSearchObj.uuid)}
-    description={text(
-      "description",
-      mockData.attachSearchObj.description || defaultItemDescription
-    )}
-    displayFields={object("display fields", [
-      ...mockData.attachSearchObj.displayFields,
-    ])}
-    modifiedDate={object(
-      "modified date",
-      mockData.attachSearchObj.modifiedDate
-    )}
-    createdDate={object("created date", mockData.attachSearchObj.createdDate)}
-    status={text("item status", mockData.attachSearchObj.status)}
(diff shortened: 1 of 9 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResult.tsx
```diff
@@ -79,34 +80,51 @@ const useStyles = makeStyles((theme: Theme) => {
...
+    keywordFoundInAttachment,
+  },
+  highlights,
+}: SearchResultProps) {
   const classes = useStyles();
 
   const searchResultStrings = languageStrings.searchpage.searchresult;
 
-  const [attachExapanded, setAttachExpanded] = React.useState(
+  const [attachExpanded, setAttachExpanded] = React.useState(
```
edalex-ian, line 120: Just fixing a typo.

mrblippy (commented):
Storybook 6 FTW 🤘 Much more concise

Source/Plugins/Core/com.equella.core/js/__stories__/search/SearchResult.stories.tsx
```diff
@@ -15,131 +15,64 @@
...
+  <SearchResult {...args} />
 );
+BasicSearchResult.args = {
+  item: {
+    ...mockData.basicSearchObj,
+  },
+  highlights: [],
+};
+
+export const AttachmentSearchResult: Story<SearchResultProps> = (args) => (
```
SammyIsConfused, line 40: Is it possible to type `args`?

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/components/SearchResult.test.tsx
```diff
@@ -29,7 +29,11 @@ describe("<SearchResult/>", () => {
     return render(
       //This needs to be wrapped inside a BrowserRouter, to prevent an `Invariant failed: You should not use <Link> outside a <Router>` error  because of the <Link/> tag within SearchResult
       <BrowserRouter>
-        <SearchResult {...itemResult} key={itemResult.uuid} />
+        <SearchResult
+          key={itemResult.uuid}
+          item={{ ...itemResult }}
```
SammyIsConfused, line 34: If you are using spread here with no modifiers, surely you can just use it directly? 
```suggestion
          item={ itemResult }
```

Source/Plugins/Core/com.equella.core/js/tsrc/util/TextUtils.ts
```diff
@@ -0,0 +1,78 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/**
+ * Based on work from com.tle.web.sections.render.TextUtils
+ */
+
+const removeNonWordCharacters = (highlight: string): string =>
```
SammyIsConfused, line 22: Do we need tests for the functions in this file?

SammyIsConfused (approved):
Looks good! Just a handful of queries.

Source/Plugins/Core/com.equella.core/js/__stories__/search/SearchResult.stories.tsx
edalex-ian: It's possible, but kind of redundant and doesn't seem the idiomatic way. I guess that's also because the type of `args` is that of the `Story` generic and would just look like:

```typescript
const AttachmentSearchResult: Story<SearchResultProps> = (args: SearchResultProps) => (
```

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/components/SearchResult.test.tsx
edalex-ian: hmmmm... :thinking: 

I originally saw this as a copy of the object, but it was more so spreading it so that it's populate the old props. So yes, with the new props I could happily just drop it in. I'll tweak it.

Source/Plugins/Core/com.equella.core/js/tsrc/util/TextUtils.ts
edalex-ian: There is some coverage of this due to the tests in `TextUtils.test.ts`. True it's not exhaustive, but seeing this was a conversion of the old I think it'll do for now.

If there were test for the Java, I would've converted this too. But to derive all the cases this was attempting to address - without tests or doco - was not justified effort.

SammyIsConfused:
Thanks for fixing that up! GTG from me.

Source/Plugins/Core/com.equella.core/js/tsrc/util/TextUtils.ts
```diff
@@ -0,0 +1,78 @@
...
+  highlights: string[],
+  cssClass: string
+): string => {
+  const highlightsRegex = highlightsAsRegex(highlights);
+  if (highlights.length < 1) {
+    // Nothing to highlight
+    return text;
+  }
+
+  const re = new RegExp("^(.*?)\\b(" + highlightsRegex + ")\\b(.*)$", "is");
```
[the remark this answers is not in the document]
edalex-ian: Yeah, fair point. I've copied this across from the Java side (which had no doco) and so I've put it as was. But indeed, as I normally tell others, if you're going to have a regex, then you should document it as per best practice.

I'll work on breaking it down and documenting it. :+1:

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/util/TextUtils.test.ts
```diff
@@ -0,0 +1,39 @@
...
+  it.each([
+    [
+      "The quick brown fox jumps over the lazy dog",
+      ["The", "fox", "dog"],
+      `<span class="${className}">The</span> quick brown <span class="${className}">fox</span> jumps over <span class="${className}">the</span> lazy <span class="${className}">dog</span>`,
+    ],
+    [
+      "The life of kelpies explained",
+      ["kelp*"],
+      `The life of <span class="${className}">kelpies</span> explained`,
```
PenghaiZhang, line 31: If my understanding is right, I won't see 
```html
<span class="${className}">knives</span>
```
if I search `knife*` ? oEQ's highlight approach does not include any stemming analyzer.

edalex-ian: That is correct. But that's also not something new due to this PR - i.e. this PR maintains the status quo. This is indeed one of the issues in oEQ not using the Lucene highlighting and instead doing it's own after the fact.

PR #2299, opened 2020-09-17
EBP: Make room for Find Uses/Add To External System dropdown

SammyIsConfused:
##### Checklist

##### Description of change
#1983 
The bottom of the dropdown used to fall off the edge of the card. 

image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
+++ b/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
@@ -431,6 +431,7 @@ select[multiple="multiple"] {
   @include typography-body-1;
   padding-left: 8px;
   min-width: 204px;
+  float: none;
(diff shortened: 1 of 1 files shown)


edalex-ian (commented):
I'm concerned this fix could be a bit brittle. What was the root cause of the issue - as by default you'd expect the parent element to expand to hold the children. Are we battling some other rule?

And for others benefit, here is what was happening before this PR:

image

SammyIsConfused:
> I'm concerned this fix could be a bit brittle. What was the root cause of the issue - as by default you'd expect the parent element to expand to hold the children. Are we battling some other rule?
> 
> And for others benefit, here is what was happening before this PR:


@edalex-ian You were right; it was a `float:left;` on the dropdowns.

edalex-ian (approved):
Good find. :+1:

Issue #796, opened 2019-03-27
New UI: User should be able to easily tell who they are logged in as


mrblippy:
**Is your feature request related to a problem? Please describe.**
Currently the only way to tell who is logged into openEquella in the new ui is to click on the user icon, and going to the My Preferences page
![dashboard page](https://user-images.githubusercontent.com/4625498/55045020-8944a200-5090-11e9-9b07-8faff9b75be6.jpg)

 
**Describe the solution you'd like**
Since we are using material ui it makes sense to do something similar to what is done in the google apps. Where the name appears as a tooltip when you hover over the icon
![google tooltip](https://user-images.githubusercontent.com/4625498/55045354-9e6e0080-5091-11e9-8acb-a2549feec1ad.png)

edalex-ian:
Definitely in and lookin' good. :+1:

PR #2307, opened 2020-09-21
Add a tooltip to show the username on the account icon

SammyIsConfused:
##### Checklist

##### Description of change
#796 
Change the tooltip on the account icon in the top right of a page to show the user's username instead. Also, refactor away from currentuser.ts to use the oeq-ts-rest-api.
image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/api/currentuser.ts
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/api/currentuser.ts
@@ -31,6 +31,7 @@ export interface UserData {
   guest: boolean;
   autoLoggedIn: boolean;
   prefsEditable: boolean;
+  username: string;
   counts?: {
     tasks: number;
     notifications: number;
@@ -40,6 +41,7 @@ export interface UserData {
 
 export const guestUser: UserData = {
   id: "guest",
+  username: "guest",
   guest: true,
   autoLoggedIn: false,
   prefsEditable: false,
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/mainui/Template.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/mainui/Template.tsx
@@ -46,11 +46,15 @@ import {
 import luxonUtils from "@date-io/luxon";
 import { makeStyles, Theme } from "@material-ui/core/styles";
 import { languageStrings } from "../util/langstrings";
-import { guestUser, MenuItem as MI, UserData } from "../api/currentuser";
 import MessageInfo from "../components/MessageInfo";
 import { Link } from "react-router-dom";
 import { LocationDescriptor } from "history";
 import { routes } from "./routes";
+import type {
+  CurrentUserDetails,
+  MenuItem as MI,
+} from "@openequella/rest-api-client/dist/LegacyContent";
+import { guestUser } from "../legacycontent/LegacyContent";
 
 export type MenuMode = "HIDDEN" | "COLLAPSED" | "FULL";
 export type FullscreenMode = "YES" | "YES_WITH_TOOLBAR" | "NO";
@@ -76,7 +80,7 @@ export interface TemplateProps {
   hideAppBar?: boolean;
   menuMode?: MenuMode;
   disableNotifications?: boolean;
-  currentUser?: UserData;
+  currentUser?: CurrentUserDetails;
   /* Extra meta tags */
   metaTags?: string;
 }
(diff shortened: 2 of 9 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/api/currentuser.ts
```diff
@@ -31,6 +31,7 @@ export interface UserData {
   guest: boolean;
   autoLoggedIn: boolean;
   prefsEditable: boolean;
+  username: string;
```
edalex-ian, line 34: Aaahhhh, this reminds me.

We need to drop out `currentuser.ts` and use the support in oeq-ts-rest-api. So, this PR seems as good as any.

Could you please remove `currentuser.ts` and move the code to use getCurrentUserDetails (and supporting interfaces) @ https://github.com/openequella/openEQUELLA/blob/develop/oeq-ts-rest-api/src/LegacyContent.ts#L38

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/Template.tsx
```diff
@@ -512,7 +512,7 @@ export const Template = React.memo(function Template({
                 topBarString.notifications
               )}
             </Hidden>
-            <Tooltip title={strings.menu.title}>
+            <Tooltip title={currentUser?.username}>
```
edalex-ian, line 515: Could be good to also have a fallback here - just so we never have an empty "bubble" of a tooltip. Maybe "Username unknown". :thinking:

edalex-ian (changes requested):
Good stuff @SammyIsConfused :+1: 

But it has reminded me of some outstanding refactoring that needs to be done. So let's squeeze that into this PR.

edalex-ian (changes requested):
Looks good @SammyIsConfused 

Just two things:

1. Please stick with the convention for importing `@openequella`; and
2. With your renamed `UserDetailsModule.ts` could we please go with `UserModule.ts`. This will align with what's been done in `component/new-search-page` with the [UserModule.ts](https://github.com/openequella/openEQUELLA/blob/component/new-search-page/Source/Plugins/Core/com.equella.core/js/tsrc/modules/UserModule.ts) there. Sure it'll create a bit of a conflict, but it'll highlight it should sit alongside it.

ChristianMurphy (approved):

Source/Plugins/Core/com.equella.core/js/tsrc/util/langstrings.ts
```diff
@@ -237,6 +237,7 @@ export const languageStrings = {
       title: "My Account",
       logout: "Logout",
       prefs: "My preferences",
+      usernameUnknown: "Username unknown",
```
PenghaiZhang, line 240: As we have a language string change log now, do we need to put this new string into the log  when component is merged to develop?

Source/Plugins/Core/com.equella.core/js/tsrc/modules/UserDetailsModule.ts
```diff
@@ -15,41 +15,15 @@
...
-export function getCurrentUser(): Promise<UserData> {
-  return Axios.get<UserData>("api/content/currentuser").then((r) => r.data);
-}
+/**
+ * Gets the current user's info from the server as OEQ.LegacyContent.CurrentUserDetails.
+ */
+export const getCurrentUserDetails = () =>
+  OEQ.LegacyContent.getCurrentUserDetails(API_BASE_URL).then(
+    (result: OEQ.LegacyContent.CurrentUserDetails) => {
+      return result;
```
PenghaiZhang, line 27: Can simplify the callback to `(result: OEQ.LegacyContent.CurrentUserDetails) => result` to save 2 lines :smile:

PenghaiZhang (approved):
Looks good to me Sam! Thanks.

Source/Plugins/Core/com.equella.core/js/tsrc/util/langstrings.ts
edalex-ian: You raise a good point there @PenghaiZhang 

I was only thinking though to do that with the ResourceCentre ones for now - i.e. the old properly released ones. It feels like all these (`langstrings.ts`) are essentially 'new' for 2020.2. But after that, then we'd want to document these too me thinks.

What do you think?

PenghaiZhang: That's a good idea! I agree.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/UserDetailsModule.ts
SammyIsConfused: Done!

edalex-ian (approved):
Aw man, looks like the hooks failed on you again. I wonder what that is.

But, aside from that, GTG. :+1: :wink:

PR #2308, opened 2020-09-21
Bugfix/search page acl issues

PenghaiZhang:
#1306 

### Checklist

- [ ] tests are included

##### Description of change

At least three ACL issues were found during the work of creating functional tests. They are: 
1. without `EDIT_SEARCH_SETTINGS`,  403 is returned when calling `api/settings/search` and 
`api/settings/facetedsearch/classification`.
2. if `SEARCH_COLLECTION` is granted on Collection level, front-end infinitely calls `api/collection`  because the resumption token is totally wrong; and
3. without `COMMENT_VIEW_ITEM`, 403 is returned when calling `api/search2`.

So there are three fixes in this PR: 
1. Remove the ACL checks for the GET endpoint of Search settings and Classifications.
2. Fix the logic of generating a resumption token in `PagedResult.scala`.
3. Perform ACL check for the comment one, and make the number of comments optional.

It might be easier to review if follow the commit history. The first and second commit is for the first and second fix respectively, and the remaining commits are for the third fix.


Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/entity/PagedResults.scala
```diff
@@ -58,53 +58,52 @@ object PagedResults {
...
         len: Int,
         offset: Int,
-        tried: Int,
-        vec: Vector[(BE, Boolean, Set[String])]): (Int, Vector[(BE, Boolean, Set[String])]) = {
-      if (len <= 0 || tried >= MaxEntities) (offset, vec)
-      else {
-        val amountToTry = if (tried == 0) len * 2 else MaxEntities - tried
-        val nextLot = res.getEntityService.query(
-          new EnumerateOptions(q,
-                               offset,
```
PenghaiZhang, line 72: Using `offset` here causes incorrect data such as duplicated data. For example, when offset is 0 and length is 10, 17 entities are found then 10 entities are included in the result. However, these 10 entities are not necessarily from No.1 to No.10.  No.2 entity could be filter out due to no permission and No.11 is included in the result. As a result, in the next call (offset is 10 now), No.11 entity is included in the result again.

PenghaiZhang:
I think the hardest one to review is definitely `PagedResult` so please grab me whenever needed!

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/entity/PagedResults.scala
```diff
@@ -58,53 +58,55 @@ object PagedResults {
...
+    // A list of Collections which users have permissions to access.
+    val availableEntities = baseEntities.asScala.collect {
+      case be if !LegacyGuice.aclManager.filterNonGrantedPrivileges(be, privilege.asJava).isEmpty =>
+        be
+    }
+    // Return a map where the key is a base entity and the value is a another map for the entity's ACLs.
+    // Depending on the value of offset, some available entities are not added into the map.
+    def getPrivilegeMap(offset: Int): Map[BE, java.util.Map[String, java.lang.Boolean]] =
+      LegacyGuice.aclManager
+        .getPrivilegesForObjects(allReqPriv.asJavaCollection, availableEntities.drop(offset).asJava)
```
PenghaiZhang, line 73: I wonder if peope are confused about the use of `drop` so let me give an example.
Say there are 15 available entities. The initial token is `0:10`, which means offset is 0 and length is 10. So the first response include the first 10 entities.  And then, the secdon token must be `10:10`. We no longer need the first 10 entities, so we drop them and only return 11th - 15th entities.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
```diff
@@ -197,7 +196,7 @@ object SearchHelper {
       createdDate = bean.getCreatedDate,
       modifiedDate = bean.getModifiedDate,
       collectionId = bean.getCollection.getUuid,
-      commentCount,
+      commentCount = getItemComments(key).map(_.size),
```
edalex-ian, line 199: Strange that to get a count of the number of comments, we first go and retrieve **all** the comments from the DB. This is unnecessary load on the DB, normally you'd just do a count query. I note however that the Service currently lacks such a method. Might be good to keep the new methods you have - as we'll need them later - but maybe we should do this properly with a DB call.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/entity/PagedResults.scala
```diff
@@ -58,53 +58,55 @@ object PagedResults {
 
     val allReqPriv = if (full) forFull ++ privilege else privilege
 
+    // All Collections, including dynamic Collections.
+    val baseEntities = res.getEntityService.query(
+      new EnumerateOptions(q, -1, MaxEntities, system, if (includeDisabled) null else false))
```
edalex-ian, line 63: As discussed, there's an issue here to resolve before I can review the Collections part of this PR.

edalex-ian (changes requested):
Heya @PenghaiZhang 

I'm happy with the removal of the permissions checks side of this fix; but I think we can do the comments count better, and there's an issue with the PagedResult update.

BTY.

PenghaiZhang:
Screenshots for the fixed resumption tokens.

Logged in as administrator: 
image
image

Logged in as a user who is not accessible to some Collections:

image

image

Source/Plugins/Core/com.equella.core/src/com/tle/core/item/standard/service/impl/ItemCommentServiceImpl.java
```diff
@@ -85,6 +89,27 @@ public Comment getComment(Item item, String commentUuid) {
...
+    }
+    return null;
+  }
+
+  @Override
+  @Transactional
+  public int getCommentCountWithACLCheck(ItemKey itemId) {
+    Item item = itemDao.getExistingItem(itemId);
+    if (canViewComment(item)) {
+      return itemDao.getCommentCounts(Collections.singletonList(item)).get(0);
```
PenghaiZhang, line 108: I did not create a new method for comment count. Instead, I reused `itemDao.getCommentCounts` like how old search page uses it.

edalex-ian: Good find, I see it's doing a HQL `count()`, so that's perfect. :+1:

Source/Plugins/Core/com.equella.core/src/com/tle/core/item/standard/service/impl/ItemCommentServiceImpl.java
```diff
@@ -159,6 +184,11 @@ AddCommentOperation comment(
     ReassignOperation reassign(
         @Assisted("fromUserId") String fromUserId, @Assisted("toUserId") String toUserId);
   }
+
+  private boolean canViewComment(Item item) {
+    Set<String> privileges = tleAclManager.filterNonGrantedPrivileges(item, COMMENT_VIEW_ITEM);
+    return !privileges.isEmpty();
+  }
```
edalex-ian, line 191: I'm just thinking. Both the methods above which call this are almost the same. And indeed, if you change the above two methods to instead return `Optional<List<Comment>>` and `Optional<Integer>` you could then possibly change this method to take a lambda which receives an `Item` and the returns an `Optional` (returning `Optional.empty()` if **can't** view comments). And then the other two methods could look something like (the very rough code):

```java
@Override
@Transactional
public Optional<Integer> getCommentCountWithACLCheck(ItemKey itemId) {
  return canViewComment( itemId, (item) -> Optional.of(
    itemDao.getCommentCounts(Collections.singletonList(item)).get(0)
  ));
}
```

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/entity/PagedResults.scala
```diff
@@ -58,59 +58,79 @@ object PagedResults {
...
+        .toList
+
+    def getPrivilegeMap(entities: List[BE]): Map[BE, java.util.Map[String, java.lang.Boolean]] =
+      LegacyGuice.aclManager
+        .getPrivilegesForObjects(allReqPriv.asJavaCollection, entities.asJava)
+        .asScala
+        .toMap
+
+    val available: Int = {
+      val entities = getBaseEntities(0, -1)
```
edalex-ian, line 74: We may want to make a comment here to highlight we're retrieving all. But this is done in a considered way that the targetted BaseEntities are all small datasets.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/entity/PagedResults.scala
```diff
@@ -58,59 +58,79 @@ object PagedResults {
...
-          (offset, vec)
-        } else {
-          val privMap = LegacyGuice.aclManager
-            .getPrivilegesForObjects(allReqPriv.asJavaCollection, nextLot)
-            .asScala
-          object ExtraPrivs {
-            def unapply(be: BE): Option[(BE, Set[String])] = privMap.get(be).map { p =>
-              (be, allReqPriv & p.asScala.keySet)
-            }
+      val baseEntities = getBaseEntities(offset, MaxEntities)
```
edalex-ian, line 89: I find it kinda wierd how we're relying on always getting `MaxEntities` (rather than say `length`) and then hoping we can prune it down. Rather then getting `length` entities, and topping it up with more queries as needed.

Feels like there's a chance that `MaxEntities` could potentially fail to have enough, and then we'd fall short. But maybe that's an extreme edge case. Or... Maybe I'm also mis-understanding something here. (I don't find `collectMore` that readable.)

edalex-ian (approved):
I'm thinking this is good to go. Maybe just a comment to add, and if you like a slight reworking of the comment retrieval and counting with Optionals and lambdas. But.. Up to you.

ChristianMurphy (approved):

autotest/OldTests/src/test/java/io/github/openequella/rest/BaseEntityAPITest.java
```diff
@@ -0,0 +1,63 @@
...
+import org.codehaus.jackson.JsonNode;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+public class BaseEntityAPITest extends AbstractRestApiTest {
+  private static final String COLLECTION_API_ENDPOINT =
+      TEST_CONFIG.getInstitutionUrl() + "api/collection";
+
+  @DataProvider(name = "initialResumptionTokens")
+  public static Object[][] tokens() {
```
edalex-ian, line 19: Just something to ponder...

You can make this a bit stricter on typing with:

```java
  @DataProvider(name = "initialResumptionTokens")
  public static Iterator<String> tokens() {
    return Arrays.asList({"0:5"}, {"0:10"}, {"0:20"}).iterator();
  }
```

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/entity/PagedResults.scala
```diff
@@ -58,61 +56,70 @@ object PagedResults {
...
+        }
+
+        // Filter entities by permissions.
+        val filteredEntities = filterByPermission(entities)
+
+        results ++= filteredEntities.take(filteredEntityQuota)
+
+        // If there is no enough quota for all filtered entities, find out index of the
+        // entity which is the last one in the result list. Then, plus the index to offset.
+        // If quota is enough then simply plus the length of entities to offset.
```
edalex-ian, line 96:
```suggestion
        // The offset for subsequent calls needs to take into account the potential
        // skipping of entities due to filtering by ACLs.
        // If there is not enough quota after filtering entities, calculate the offset
        // for the next call by adding the index of the last entity in the result list
        // to the current offset.
        // If there is enough for the quota, then simply return the current offset plus
        // the length of retrieved entities.
```

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/entity/PagedResults.scala
```diff
@@ -58,61 +56,70 @@ object PagedResults {
...
     pb.setStart(firstOffset)
     pb.setLength(results.length)
-    pb.setAvailable(available)
-    // Include resumption token if there are items which can be retrieved in next request.
-    if (results.length + firstOffset < available)
-      pb.setResumptionToken(s"${firstOffset + results.length}:${_length}")
-    pb.setResults(results.map {
-      case (be, canFull, privs) => addPrivs(privs, res.serialize(be, null, canFull))
+    pb.setAvailable(getAvailable)
+    // When there are no enough entities retrieved, do not return a resumption token.
```
edalex-ian, line 117:
```suggestion
    // When the number of results returned are less than asked for, we have reached the end.
    // So do not return a resumption token.
```

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/entity/PagedResults.scala
```diff
@@ -58,61 +56,70 @@ object PagedResults {
...
+    if (results.length == _length)
+      pb.setResumptionToken(s"${nextOffset}:${_length}")
+    // todo: Support getting full information.
```
edalex-ian, line 120: Is this still TODO?

edalex-ian (approved):
Looking good @PenghaiZhang . Just a couple of comment tweaks, and a suggestion around DataProviders.

But one big thing... You've still a TODO in there. Does that still need todoing?

autotest/OldTests/src/test/java/io/github/openequella/rest/BaseEntityAPITest.java
PenghaiZhang: We probably need to update TESTNG to change the type of `DataProvider`.

Source/Plugins/Core/com.equella.core/src/com/tle/core/portal/dao/PortletDaoImpl.java
```diff
@@ -158,10 +158,6 @@ public Object doInHibernate(Session session) throws HibernateException {
                           hql.append(callback.getAdditionalWhere());
                         }
 
-                        if (callback != null && callback.getOrderBy() != null) {
```
PenghaiZhang, line 161: Here is why PortletTest failed. The `callback` passed in this method is essentially a `PagedListCallback` so an `order by` will be appended to the HQL, which resulted in Postgresl error `[42803]: column "be.id" must appear in the GROUP BY clause or be used in an aggregate function Position: 42`  .

However, this method is used to count the number of Portlets. There is no point to do a sorting. So I believe we should remove this if statement. One good example is method `createEnumerateQuery` from `GenericInstitionalDaoImpl`.

autotest/OldTests/src/test/java/io/github/openequella/rest/BaseEntityAPITest.java
edalex-ian: Yeah, we did have a Renovate PR there for a bit, but it got closed. #2022

PenghaiZhang:
ApidocsTest failed. I will restart.

Source/Plugins/Core/com.equella.core/src/com/tle/core/portal/dao/PortletDaoImpl.java
edalex-ian: I agree.

This is odd it's doing an order for a count. I wonder if it's a case of copy pasta. :thinking:

autotest/OldTests/src/test/java/io/github/openequella/rest/BaseEntityAPITest.java
```diff
@@ -49,6 +51,27 @@ public void getEntities(String initialToken) throws IOException {
...
+    queryParams[2] = new NameValuePair("q", "Basic");
+    method.setQueryString(queryParams);
+
+    int statusCode = makeClientRequest(method);
+    assertEquals(statusCode, HttpStatus.SC_OK);
+
+    JsonNode results = getResultList(mapper.readTree(method.getResponseBody()));
+
+    // There are two Collections named 'Basic xxx'. Full information of the first one
+    // is accessible whereas that of the second is not accessible.
```
edalex-ian, line 69: Maybe expand this out then to highlight what the significance of then checking for `security` is. (As personally, I'm not sure what the point is...)

autotest/OldTests/src/test/java/io/github/openequella/rest/BaseEntityAPITest.java
PenghaiZhang: I see that PR failed on scala check, but I can't remember why we did not give Renovate some help.

autotest/OldTests/src/test/java/io/github/openequella/rest/BaseEntityAPITest.java
```diff
@@ -60,4 +83,8 @@ private String getTokenFromResponse(JsonNode result) {
     }
     return null;
   }
+
+  private JsonNode getResultList(JsonNode result) {
```
edalex-ian, line 87: You should update `getResultLength` to also use this new method.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/entity/PagedResults.scala
```diff
@@ -76,10 +76,10 @@ object PagedResults {
       filterByPermission(entities).length
     }
 
-    def collectMore(length: Int, initialOffset: Int): (Int, List[BE]) = {
-      val results: ListBuffer[BE] = ListBuffer()
-      var offset                  = initialOffset
-      var filteredEntityQuota     = length // Indicate how many entities can be put in the result list.
+    def collectMore(length: Int, initialOffset: Int): (Int, List[(BE, Boolean, Set[String])]) = {
+      val results: ListBuffer[(BE, Boolean, Set[String])] = ListBuffer()
```
edalex-ian, line 80: The addition of this tuple (`(BE, Boolean, Set[String])`) is getting a bit obtuse. Especially when the return of the method is now a tuple which includes a `List` which contains a tuple which also includes a `Set`. (I think that mouthful highlights the situation.)

I wonder if a case class may be in order for the inner tuple of `(BE, Boolean, Set[String])`. :thinking:

edalex-ian (approved):
Probably pretty good to go. Just maybe a couple of things to tidy up. But if you can do at least the Test tweak, then you should be good to merge.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/entity/PagedResults.scala
PenghaiZhang: It's a very good idea to use a case class here!

edalex-ian:
:crossed_fingers: for a green build :crossed_fingers:

autotest/Tests/tests/rest/institution/itemdefinition/25/f5db7438-a367-4c3e-85b9-ab67be09d5e6.xml
```diff
@@ -0,0 +1,114 @@
...
+            </strings>
+          </bundleTitle>
+        </com.tle.beans.entity.itemdef.SummarySectionsConfig>
+      </configList>
+      <hideOwner>false</hideOwner>
+      <hideCollaborators>false</hideCollaborators>
+    </itemSummarySections>
+  </slow>
+  <denyDirectContribution>false</denyDirectContribution>
+</com.tle.beans.entity.itemdef.ItemDefinition>
```
ChristianMurphy, line 114:
```suggestion
</com.tle.beans.entity.itemdef.ItemDefinition>

```

ChristianMurphy (approved):

autotest/Tests/tests/rest/institution/itemdefinition/25/f5db7438-a367-4c3e-85b9-ab67be09d5e6.xml
PenghaiZhang: Hi Christian, this xml file is from an Institution export and in the format oEQ expects.

PR #2326, opened 2020-09-25
Bugfix/search page view attachment acl issue

PenghaiZhang:
#1306 

##### Checklist

- [ ] tests are included

##### Description of change

Currently, new search page throws an error if users have no permissions to view attachments.
At the begining I thought the required permission would be `VIEW_ATTACHMENTS`, but after some digging I realised the required permission is `VIEW_ITEM`. 

So here are a list of changes to review:

1. We need to tell the ItemBean serialiser what permissions to check. Without doing this, the error still occurs even though the permission is given. And then, the serialiser will generate a list of attachment beans or null, depending on permission checks.

2. Change the type of attachments to optional in both back-end and front-end.

3. In component `SearchResult`, give a default value to attachments if it's undefined.

In the below example,  `VIEW_ITEM` of the first item is revoked whereas that of the second item is granted.
 
image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResult.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResult.tsx
@@ -108,7 +108,7 @@ export default function SearchResult({
     modifiedDate,
     status,
     displayOptions,
-    attachments,
+    attachments = [],
     keywordFoundInAttachment,
   },
   highlights,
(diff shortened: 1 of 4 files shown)


edalex-ian:
The build failures here are across two known flaky tests - ApiDocs and HierarchyTopics. But I'll restart both to be sure.

edalex-ian (approved):
LGTM. :+1:

SammyIsConfused (approved):

PR #2359, opened 2020-10-02
EBP: Improve URL attachment dialog styles

SammyIsConfused:
##### Checklist

##### Description of change
#1983 
Turns out the styles applied to the file attachment dialog in #2255 didn't apply in the URL attachment dialog, and so this PR applies the same styling to the URL dialog.
Before change:
image

After change:
image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
+++ b/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
@@ -71,7 +71,7 @@ $doubleMargin: 16px;
   -moz-border-radius: 4px;
 }
 
-@mixin editDialogCommonStyles {
+@mixin editAttachmentDialogStyles {
   height: 100%;
   box-shadow: none;
   padding: 0;
@@ -4160,7 +4160,7 @@ a.modal-control.modal-save {
 .fileHandler {
   .modal-content .modal-content-background .modal-content-inner .float-left {
     width: 100%;
-    @include editDialogCommonStyles;
+    @include editAttachmentDialogStyles;
   }
 
   .modal-content-inner {
@@ -4171,7 +4171,7 @@ a.modal-control.modal-save {
 
 .urlHandler {
   .modal-content .modal-content-background .modal-content-inner .float-left {
-    @include editDialogCommonStyles;
+    @include editAttachmentDialogStyles;
   }
 
   .modal-content-inner {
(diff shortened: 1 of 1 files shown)


mrblippy (commented):
I guess my only query would be the blank space between the two panels. Is that also how file attachment dialog looks?

SammyIsConfused:
> I guess my only query would be the blank space between the two panels. Is that also how file attachment dialog looks?

Not quite, but the structure of the dialog is different here. If I were to make it look identical, it would break the other page of the dialog (the initial dialog you input the URL in)

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -4159,40 +4152,56 @@ a.modal-control.modal-save {
...
+    flex-direction: row-reverse;
+  }
+}
+
+.urlHandler {
+  .modal-content .modal-content-background .modal-content-inner .float-left {
+    height: 100%;
+    box-shadow: none;
+    padding: 0;
+  }
```
edalex-ian, line 4173: It'd seem here that the intention is to match that of the same classes when under `.fileHandler`. As a result, a mixin should be used to keep things DRY and avoid copy pasta.

edalex-ian (changes requested):
Looks good other than the duplicated block. Please address, and then it should be fine to go.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -71,6 +71,12 @@ $doubleMargin: 16px;
   -moz-border-radius: 4px;
 }
 
+@mixin editDialogCommonStyles {
```
edalex-ian, line 74: A query on the name here. `editDialog` seems very generic, is that the intention - i.e. that it could be used on any of the dialogs for editing? Or, should it be slightly more specific like `editAttachmentDialog`?

edalex-ian (approved):
LGTM, but just a query on the naming of the new mixin.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
SammyIsConfused: It should mainly have been for editing attachments, yeah. I've renamed to `editAttachmentDialogStyles`.

PR #2390, opened 2020-10-12
EBP: Fix styles for "Add MIME Type" button

SammyIsConfused:
##### Checklist

##### Description of change
#1983 
There was no width set on this button, which caused it to look broken. 
As was used in other places, I simply added the col2width variable. 
Broken:
image
Fixed:
image


edalex-ian (changes requested):
Definitely an improvement, but it looks like setting a `width` alone is insufficient, as now the text has excess padding to the right.

edalex-ian:
Ahhhhh, I see this is consistent with the other buttons done via EBP. We can't match exactly what we've done in New UI land with actual icons. Ah well.

edalex-ian (approved):
GTG. :+1:

PR #2434, opened 2020-10-19
Show Item's star ratings in new search page

PenghaiZhang:
##### Checklist

- [x] commit message follows [commit guidelines][]1
- [x] tests are included

##### Description of change

The changes included in this PR are: 

1. Back-end change to support returning star rating from `search2`.
2. REST Module change - adding star ratings to the type definition of `SearchResultItem`.
3. Component `SearchResult` change - showing stars
4. A test for this feature.

image


Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/model/SearchResultItem.scala
```diff
@@ -50,6 +50,7 @@ case class SearchResultItem(
     modifiedDate: Date,
     collectionId: String,
     commentCount: Option[Integer],
+    starRatings: Float,
```
PenghaiZhang, line 53: I have run `NewSearchPageTest` and I did not find any ACL issue with low privileged user so I think the type does not need to be `Option`. 

Also I did not find  any permission about viewing the rating.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResult.tsx
```diff
@@ -140,6 +144,26 @@ export default function SearchResult({
...
+     * For example, 3.33 will be 3.5 and 3.23 will be 3.
+     */
+    const showStar = () => {
+      const rating = Math.round(starRatings * 2) / 2;
+      return [1, 2, 3, 4, 5].map((i) => {
+        if (i <= rating) {
+          return <Star aria-label="full-star" key={i} />;
+        }
+
+        if (i - 0.5 === rating) {
```
ChristianMurphy, line 159: Will this work?
Floating point equality can have some quirks, but I think this will be okay?

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResult.tsx
```diff
@@ -140,6 +144,26 @@ export default function SearchResult({
...
     );
 
+    /**
+     * Generate star icons for Item's ratings.
+     * Each rating will be rounded to its nearest 0.5.
+     * For example, 3.33 will be 3.5 and 3.23 will be 3.
+     */
+    const showStar = () => {
+      const rating = Math.round(starRatings * 2) / 2;
+      return [1, 2, 3, 4, 5].map((i) => {
```
ChristianMurphy, line 154: neat way of preventing >5 stars :+1: :star2:

Source/Plugins/Core/com.equella.core/js/__mocks__/searchresult_mock_data.ts
```diff
@@ -140,6 +143,7 @@ const customMetaSearchObj: OEQ.Search.SearchResultItem = {
   modifiedDate: new Date("2020-05-26T12:45:06.857+10:00"),
   collectionId: "b28f1ffe-2008-4f5e-d559-83c8acd79316",
   commentCount: 0,
+  starRatings: -1,
```
ChristianMurphy, line 146: `-1` is this context means no rating?
would it make sense to not pass the attribute `starRatings` if there are not ratings?

ChristianMurphy (approved):

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResult.tsx
PenghaiZhang: Hi @ChristianMurphy , yes, it works in this case. I can also set a epsilon. Do you have any better recommendations for handling float numbers with TS

Source/Plugins/Core/com.equella.core/js/__mocks__/searchresult_mock_data.ts
PenghaiZhang: Yes, it means no rating.

I can make this field optional.  On server side, check if the value is -1 or not and return the value if it's not -1.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResult.tsx
ChristianMurphy: This could be fine.
The alternative would be to skip the `/ 2` above and use whole numbers/ints.
But since this works as is, let's leave it.

Source/Plugins/Core/com.equella.core/js/__mocks__/searchresult_mock_data.ts
ChristianMurphy: Okay,
is the reason behind this decision to avoid null values in the DB?
Or something else?

PenghaiZhang: Sorry I don't know the reason. This was implemented long long ago by someone else without explaining the reason. But since the type is float, I think the value defaults to 0.0f.

ChristianMurphy: ah, so this is an existing pattern, got it.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/model/SearchResultItem.scala
edalex-ian, line 53: Missing `@param` in above ScalaDoc.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResult.tsx
edalex-ian, line 159: Nit: Should be `else if`, but hey.

edalex-ian (approved):
I think this is all fine as is, but.... I do feel that the UI side should have been done as a stand-alone presentation component so that then you could just have:

```typescript
<StarRating rating={rating}/>
```

Then you could easily wrap that in a set of stories, and a parametrised test to validate the handling of supported ranges. Also with moving to a presentation component, we could also look at putting an `aria-label` on a top level component (or div) which could read something like: "Item rating: x stars".

What lead me to note this was the impure `showStar()` function which jumped out - but not unfamiliar, as we have a few of such.

If you're willing, would be great to see it reworked.

Source/Plugins/Core/com.equella.core/js/__stories__/components/StarRating.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { Meta, Story } from "@storybook/react";
+import * as React from "react";
+import { StarRating, StarRatingProps } from "../../tsrc/components/StarRating";
+
+export default {
+  title: "StarRating",
+  component: StarRating,
+} as Meta<StarRatingProps>;
+
+export const FiveStars: Story<StarRatingProps> = (args) => (
```
edalex-ian, line 27: Probably good to rename this something like 'HalfStarRating' - as that's the state I guess you're demonstrating here (that is, where one of the stars is only half).

Probably then good to also have stories to cover the examples of:

* No stars;
* Full stars; and
* Different number of stars.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/components/StarRating.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import "@testing-library/jest-dom/extend-expect";
+import { render } from "@testing-library/react";
+import * as React from "react";
+import { StarRating } from "../../../tsrc/components/StarRating";
+
+describe("<StarRating />", () => {
+  it.each([
```
edalex-ian, line 24: Either this (or additional tests) should be testing boundaries. That is:

* Testing for ratings of: 0, half max, max, max + 1;
* Testing for number of stars: 0, 5, lots; amd
* Testing your rounding logic - to make sure it goes up to half, down to whole, etc.

Being able to easily test all that (at speed) is the real benefit you get here of splitting it out. And combined with parametrised tests, you should be able to do that succinctly.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResult.tsx
```diff
@@ -236,6 +238,15 @@ export default function SearchResult({
             </Typography>
           </>
         )}
+
+        {starRatings >= 0 && (
+          <>
+            {metaDataDivider}
+            <div aria-label={`item rating: ${starRatings}`}>
```
edalex-ian, line 245: Should probably be a language string.

edalex-ian (commented):
Looking good @PenghaiZhang , just needs a bit more fleshing out of the stories and tests.

Source/Plugins/Core/com.equella.core/js/__stories__/components/StarRating.stories.tsx
```diff
@@ -0,0 +1,57 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { Meta, Story } from "@storybook/react";
+import * as React from "react";
+import { StarRating, StarRatingProps } from "../../tsrc/components/StarRating";
+
+export default {
+  title: "StarRating",
```
edalex-ian, line 23: Sorry, should of mentioned this before. Bit by bit I'm trying to keep these stories a bit more structured in the Storybook UI. As a result, could you please change this to:

```suggestion
  title: "component/StarRating",
```

edalex-ian (approved):
Looks good @PenghaiZhang . If you can just make that one small tweak, then when green feel free to merge.

PR #2436, opened 2020-10-19
Fix warnings from eslint

SammyIsConfused:
##### Checklist

##### Description of change
This PR is to address warnings that arise from the check:ts task. All have been addressed except those affected by react-hooks/exhaustive-deps rule - as they would require a larger refactor and investigation to fix, and it would mean a lot of re-testing. Best to do as part of another PR. 

- Lots of eqeqeq rule fixes - double equals comparisons that should be triple equals, bang equals should be bang double equals.
- A few link tag fixes - we had a couple of javascript:void(0) links which are not allowed. These were reworked with a href of "/#", and a preventDefault on the onClick. A smoke test shows this to be working. 
- There was one link tag needed rel="noreferrer".
- A couple of eval calls were excused from the linter - since they were in dangerous areas (LegacyContent, JQueryDiv) I didn't feel safe to rework them.
- A few tests didn't directly have expect calls - they were either in other functions or reliant on error out for a failure. I have reworked the jest/expect-expect rule such that any function that begins with the word expect will suffice. 
- There was a couple places in which an interface and a default export function had the same identifier. Reworked by prepending 'I' to the interface definition.
- In the Template, the lefthand menu links are populated dynamically. When they are first defined there is no content as a child of the a tag. 


Source/Plugins/Core/com.equella.core/js/tsrc/config.ts
```diff
@@ -18,11 +18,11 @@
 // MUST end in /
 //export const INST_URL = 'http://localhost:8080/reports/';
 
-interface Config {
+interface IConfig {
   baseUrl: string;
 }
 
-export const Config: Config = {
+export const Config: IConfig = {
```
ChristianMurphy, line 25: A couple questions

I'm not seeing a typescript ESLint rule that enforces an `I` prefix for interfaces.
https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin
How was this flagged by ESLint? :raised_eyebrow:

Also is this generally a pattern that will/should be adopted?
This is the only place I currently see it.

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContent.tsx
```diff
@@ -66,7 +66,7 @@ interface FormUpdate {
   partial: boolean;
 }
 
-interface LegacyContent {
+interface ILegacyContent {
```
ChristianMurphy, line 69: same as above

Source/Plugins/Core/com.equella.core/js/tsrc/settings/AdminDownloadDialog.tsx
```diff
@@ -44,7 +44,7 @@ export default function AdminDownloadDialog({
       <DialogContent>
         <DialogContentText>
           {text.introTextOne}
-          <a href={link} target="_blank">
+          <a href={link} target="_blank" rel="noreferrer">
```
ChristianMurphy, line 47: :+1: :star: thanks for adding this!

ChristianMurphy (approved):
Thanks @SammyIsConfused!

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContent.tsx
edalex-ian: So based on what this is an how this is used, it would be better called `LegacyPageContent`. Other alternatives would be suffixing with `Response` - as it is one of the responses the Legacy API can return when submitting a form.

Source/Plugins/Core/com.equella.core/js/tsrc/config.ts
edalex-ian: Yes, I'd prefer to avoid this appending of `I` and go with more deliberate naming. So I've attempted to offer alternatives for the two places.

Here is an interesting one, as the interface seems rather redundant. It's not exported or used anywhere other then immediately below it which ultimately is really determining the structure. It seems like unnecessary double handling.

Option 1 then, is simply get rid of the interface.

But if we really really want to keep it then have the `interface` called `Config` and rename the object to `AppConfig`.

SammyIsConfused: It didn't seem happy without an interface so I went with the Config and AppConfig approach.

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContent.tsx
SammyIsConfused: Went with LegacyContentResponse.

ChristianMurphy (approved):

ChristianMurphy (approved):
Thanks @SammyIsConfused! :bow: 
Looks like a couple merge conflicts popped up, once those are resolved, this LGTM.

ChristianMurphy (approved):

edalex-ian (approved):
LGTM @SammyIsConfused 

I'll organise for us to loop back with a few PRs for the various dependency lists. I see some are super simple - we just need to add them - thankfully, but others are where the code is not quite idiomatic and indeed could cause pain.

Merge on. :+1:

Issue #688, opened 2018-12-28
New UI implementation for Selection Session


edalex-ian:
Create a 'New UI' version of the Selection Session - including searching.

Original description:

> LMS Selection sessions pages need to be updated to the new search UI

cathfitz:
Still being worked on

PR #2437, opened 2020-10-19
Feature/integrate new search page in selection section

PenghaiZhang:
#1306

##### Checklist

- [ ] tests are included

##### Description of change

This PR integrates the new search UI into Selection Section.

The approach basically contains two steps:
1. Add new html files named `NewSearchPage.html`  and tsx files named `NewSearchPage.tsx`  under the js entry folders (`entrydev` and `entrybuild`),  and the tsx files incorporate the new search UI.

2. On server side, check if new search UI is enabled and if the request comes from Selection Session. If yes, then parse `NewSearchPage.html` and output the parsed contents to the section.
 
Also, I need to mention a couple things.
1. Selection Session has three layouts: COURSE, SKINNY and NORMAL.  Both COURSE and NORMAL can display new search UI depending on the setting. For SKINNY, I talked to Charlie about how to trigger this layout but we have not found much useful information. All I know so far is this layout could be associated with Advanced Search. So I need some help here.

2. CSS changes made in this PR also affect `Resource Selector` as it uses NORMAL layout.

Here are two videos for [COURSE](https://streamable.com/fu4vyf) and [NORMAL](https://streamable.com/jwxqh9)

#688


Source/Plugins/Core/com.equella.core/js/entrydev/NewSearchPage.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { BrowserRouter } from "react-router-dom";
+import { basePath } from "../tsrc/mainui/App";
+import SearchPage from "../tsrc/search/SearchPage";
+import { oeqTheme } from "../tsrc/theme";
+
+const searchPageProps = { updateTemplate: () => {} };
+const renderSearchPage = () => {
+  ReactDOM.render(
+    <BrowserRouter basename={basePath}>
+      <ThemeProvider theme={oeqTheme}>
```
PenghaiZhang, line 30: I guess we don't need to use `StylesProvider` in dev environment ?

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -4461,22 +4472,37 @@ img.warningimg {
...
 #selection-page {
-  margin: 0 auto;
-  width: 784px;
-  position: relative;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  // The width of this div is Full viewport width minus vertical scrollbar width.
+  // Ideally we should us JS to get scrollbar width.
+  width: 100vw;
```
PenghaiZhang, line 4480: This `100vw` will cause a problem - when there is a vertical scroll bar, the actual width of the page is bigger than 100vw, so a horizontal scroll bar gets displayed.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -4461,22 +4472,37 @@ img.warningimg {
 }
 
 #selection-page {
-  margin: 0 auto;
-  width: 784px;
-  position: relative;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  // The width of this div is Full viewport width minus vertical scrollbar width.
```
PenghaiZhang, line 4478: I just forgot to remove this comment.

edalex-ian:
Thanks for the two videos @PenghaiZhang 

Before digging into the code, I note one issue with formatting for the `NORMAL` mode. The alignment of the top edge of the cards in New Search vs Legacy UI:

2020-10-23-143815_576x220_scrot

Actually, I note the selection card is missing rounded corners... Is this the latest EBP from `develop` - post merge of component/new-search-page ?

edalex-ian:
As discussed... I think this PR has pushed us to the point where we finally need to resolve `js/entrybuild` and `js/entrydev`. So I propose that as part of this PR and your new files:

* Merge those two folders into one (maybe just go with `entrypoint` for now), and update `js/package.json` accordingly;
* Tidy the collection of files in there such that you end up with:
  * `index.html` and `index.ts`;
  * `SearchPage.html` and `SearchPage.ts` (was your new `NewSearchPage`);
  * `Settings.html` and `Settings.ts` (was `oldsettings`, but previous re-used `index.ts`);
  * `tsrc/mainui/index.ts` should have that default function take a string union argument along the lines of `"react-index" | "react-search" | "react-settings"` and this can be passed in in each of the above `.ts` files. Further, that same value can be used as the `div` `id` in each of the `.html` files;
  * Also, the JSX you've got in `NewSearchPage.tsx` should be over in `tsrc/mainui/index.tsx` which should then still only have one `render` call, however it should swap between the JSX based on the string union - (ideally using runtypes `match` to choose the correct one).

I'll now proceed to read on the rest of this.

Source/Plugins/Core/com.equella.core/js/entrydev/NewSearchPage.html
```diff
@@ -0,0 +1,3 @@
+<div id="newSearchPage">
```
mrblippy: So this file is only for displaying the New Search Page inside a selection session? If so, could the filename be modified to reflect this?

Source/Plugins/Core/com.equella.core/js/entrybuild/NewSearchPage.tsx
```diff
@@ -0,0 +1,45 @@
...
+import * as ReactDOM from "react-dom";
+import { BrowserRouter } from "react-router-dom";
+import { basePath } from "../tsrc/mainui/App";
+import SearchPage from "../tsrc/search/SearchPage";
+import { oeqTheme } from "../tsrc/theme";
+import { StylesProvider, createGenerateClassName } from "@material-ui/core";
+
+const searchPageProps = { updateTemplate: () => {} };
+const generateClassName = createGenerateClassName({
+  productionPrefix: "new-search-page",
```
edalex-ian, line 29: Would it be good to keep it shorter and go with something specific to oEQ. Maybe `oeq-nsp`. Then maybe we should also specify one for our normal full build - perhaps just `oeq`.

Source/Plugins/Core/com.equella.core/js/entrydev/NewSearchPage.tsx
edalex-ian: I think we should keep it the same for both.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -4461,22 +4472,37 @@ img.warningimg {
...
+  width: 100vw;
+  overflow-x: hidden;
 }
 
 #selection-content {
   background: url(../images/content-middle.gif) repeat-y 0px 0px;
-  position: relative;
-  width: 784px;
-  float: left;
+  width: 98%;
```
edalex-ian, line 4486: Why 98%? Could be handy to add a comment here.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -4461,22 +4472,37 @@ img.warningimg {
...
 }
 
 #selection-content #breadcrumb-inner {
   margin-left: 18px;
 }
 
+#selection-content #new-search-page {
+  margin: 0 15px 0 0;
+  width: 80%;
```
edalex-ian, line 4495: Why 98%? Could be handy to add a comment here.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -4461,22 +4472,37 @@ img.warningimg {
...
-  float: left;
+  width: 98%;
 }
 
 #selection-content #breadcrumb-inner {
   margin-left: 18px;
 }
 
+#selection-content #new-search-page {
+  margin: 0 15px 0 0;
```
edalex-ian, line 4494: I believe there is a variable along the lines of `$doubleMargin` which is 16px. Would that be appropriate here?

Source/Plugins/Core/com.equella.core/src/com/tle/web/search/base/AbstractRootSearchSection.java
```diff
@@ -104,6 +106,20 @@ protected TemplateResult getTemplateResult(RenderEventContext info) {
     templateResult.addNamedResult(
         OneColumnLayout.BODY, CombinedRenderer.combineMultipleResults(cssIncludes));
 
+    // If new search UI should be rendered, keep the right column content and override
```
edalex-ian, line 109: The below code is not immediately clear, as it seems it has gone more with a pattern of augmentation. I wonder if instead the above code can be broken into a few useful private methods, and then further up in this method the below `if` becomes part of an `if/else`.

Source/Plugins/Core/com.equella.core/src/com/tle/web/search/base/AbstractRootSearchSection.java
```diff
@@ -104,6 +106,20 @@ protected TemplateResult getTemplateResult(RenderEventContext info) {
...
+    // the body content with new search UI.
+    if (model.isNewSearchUIInNormalLayout()) {
+      TemplateRenderable rightColumnContent =
+          templateResult.getNamedResult(info, TwoColumnLayout.RIGHT);
```
edalex-ian, line 113: Could this fail to `getNamedResult`?

Source/Plugins/Core/com.equella.core/src/com/tle/web/search/base/AbstractRootSearchSection.java
```diff
@@ -150,5 +168,21 @@ public InfoBookmark getPermanentUrl() {
     public void setPermanentUrl(InfoBookmark permanentUrl) {
       this.permanentUrl = permanentUrl;
     }
+
+    public boolean isNewSearchUIInNormalLayout() {
```
edalex-ian, line 172: I reckon a bit of JavaDoc would go a long way here to help readers understand what you mean by 'Normal Layout'.

Source/Plugins/Core/com.equella.core/src/com/tle/web/searching/section/RootSearchSection.java
```diff
@@ -80,6 +85,21 @@ public SectionResult renderHtml(RenderEventContext context) {
...
+    // If this method is triggered from Selection Section, then check if Selection Section
+    // is in 'structured' mode. If yes, then render the new search page if it's enabled.
+    SelectionSession selectionSession = selectionService.getCurrentSession(context);
+    if (isNewSearchUIInSelectionSession(selectionSession)) {
+      SimpleSectionResult newSearchUIContent = RenderNewSearchPage.renderNewSearchPage(context);
+      // For 'structured' layout, output the new search UI content directly.
+      // For 'addOrSelected' layout, save the content in model.
+      if (selectionSession.getLayout() == Layout.COURSE) {
+        return newSearchUIContent;
+      } else {
```
edalex-ian, line 98: Should this be an `else if`, and then follow with an `else` to throw an `IllegalStateException` if we've ended up with an unexpected `Layout`? Or are there only two `Layouts`? (Maybe if there are only two, we might want to protect from the unknown future.

edalex-ian (changes requested):
Looks pretty good @PenghaiZhang 

Obviously I've requested a few things. But also, are we able to add tests due to the integ tester?

PenghaiZhang:
> Thanks for the two videos @PenghaiZhang
> 
> Before digging into the code, I note one issue with formatting for the `NORMAL` mode. The alignment of the top edge of the cards in New Search vs Legacy UI:
> 
> 2020-10-23-143815_576x220_scrot
> 
> Actually, I note the selection card is missing rounded corners... Is this the latest EBP from `develop` - post merge of component/new-search-page ?

Yes this PR definitely has latest EBP work from develop. This PR includes EBP for `COURSE` layout but not for `NORMAL`. I will add more EBP after addressing the `entrypoint` issue.

Source/Plugins/Core/com.equella.core/js/entrybuild/NewSearchPage.tsx
PenghaiZhang: The normal full build is not happy with `StylesProvider` due to `PrivateHiddenCss`. 
For example, without the provider, a style from `PrivateHiddenCss` is named `jss25` which is unique.
However, with the provider used and the prefix being `oeq`, this style is named `oeq1` which is duplicated. I expect it should be named `oeq25`.

Source/Plugins/Core/com.equella.core/js/entrydev/NewSearchPage.html
PenghaiZhang: Selection session and the Resource Selector are the two places where this file is used. And it can be used in other places as long as we need. And this file has been renamed to `SearchPage.html`.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
PenghaiZhang: I have changed this back to 100%

PenghaiZhang:
> Thanks for the two videos @PenghaiZhang
> 
> Before digging into the code, I note one issue with formatting for the `NORMAL` mode. The alignment of the top edge of the cards in New Search vs Legacy UI:
> 
> 2020-10-23-143815_576x220_scrot
> 
> Actually, I note the selection card is missing rounded corners... Is this the latest EBP from `develop` - post merge of component/new-search-page ?

These two style issues should be fixed in 3a76ecc0b149d3e0c1d493a8c51bb5df1ad9af0b

Source/Plugins/Core/com.equella.core/src/com/tle/web/search/base/AbstractRootSearchSection.java
PenghaiZhang: I think this method is redundant so I remove it.

Source/Plugins/Core/com.equella.core/src/com/tle/web/searching/section/RootSearchSection.java
PenghaiZhang: After more digging I realised that there is no need to check what layout is used. 
But to answer your question, there are three layouts: NORMAL, COURSE, SKINNY. So far, I have seen NORMAL and COURSE in selection session. Take Moodle as an example, if the `openEQUELLA action` is configured to use `selectOrAdd` then NORMAL is used, and if that's configured to use 'structured' then COURSE is used.

edalex-ian:
> 
> These two style issues should be fixed in [3a76ecc](https://github.com/openequella/openEQUELLA/commit/3a76ecc0b149d3e0c1d493a8c51bb5df1ad9af0b)
> 
> image

I note in the image that there is a styling mismatch between EBP and MUI in the amount of shadow for the 'Selection' card. However a mixin has definitely been setup to standardise that, so looks like we need a fix. (I guess I'll see it more when I jump into the review.)

The only other thing that now jumps out is the font styling differences in the 'Selection' card. But maybe that's beyond our limits.

Overall though, visually improved on the original. :+1:

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -4701,12 +4720,13 @@ div#sssh_finishedButton {
...
   box-shadow: 0 3px 6px rgba(0, 0, 0, 0.16), 0 3px 6px rgba(0, 0, 0, 0.23);
   background: none;
+  border-radius: 4px;
 }
 
 .box_head {
   background-color: white;
   border: none;
-
+  border-radius: 4px;
```
edalex-ian, line 4729: You probably want to look at using `curvedCorners` mixin whenever you find yourself doing this.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -4720,6 +4740,7 @@ div#sssh_finishedButton {
 .box_content {
   background: none;
   border: none;
+  border-radius: 4px;
```
edalex-ian, line 4743: As above, look at using `curvedCorners`.

edalex-ian (changes requested):
Looking good @PenghaiZhang . Pretty well there, but just need some tweaks.

Plus I think you've got a good sized conflict waiting with the work you've done in App etc against `develop`, so I think it'd be worthwhile for this foundational PR if you go that merged across and the conflicts resolved so that this PR aligns nicely.

Source/Plugins/Core/com.equella.core/js/package.json
```diff
@@ -9,20 +9,23 @@
     "dist": "target/resources/web/reactjs",
     "jest": {
       "verbose": true
-    }
+    },
+    "entryHtml": "entrypoint/*.html",
+    "entryScript": "entrypoint/scripts/*.js",
+    "entryTarget": "entrypoint/target"
```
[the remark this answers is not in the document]
PenghaiZhang: Since we moved 'uploadlist.js' from `entrypoint/target` to `target`, we need to run `npm run dev:build` first after we do a clean.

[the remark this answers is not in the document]
edalex-ian: You mean rather than just going straight to `npm run dev`?

Because in theory `sbt cleanrun` would also have already done what's needed. :thinking:

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -4511,7 +4531,6 @@ img.warningimg {
 
 .selectiondashboard #col2 {
   width: 368px;
-  margin-left: $doubleMargin;
```
edalex-ian, line 4514: Oooh, this looks like something recently added to fix EBP stuff for 2020.2. Removing it's not removing a fix is it?

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -4701,12 +4720,13 @@ div#sssh_finishedButton {
   border: none;
   box-shadow: 0 3px 6px rgba(0, 0, 0, 0.16), 0 3px 6px rgba(0, 0, 0, 0.23);
   background: none;
+  border-radius: 4px;
```
edalex-ian, line 4723: This change, combined with the above, looks like you're attempting to achieve the standard style defined in `buttonShadowBorder`. Should that be used instead? (And maybe still keeping the background declaration.)

Otherwise at the very least, it should be using `curvedCorners`.

edalex-ian (approved):
Looks good @PenghaiZhang . There is one more DRY fix needed in legacy.scss, but really, it's good to merge me thinks.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
PenghaiZhang: Yes `buttonShadowBorder` is good. Thanks for reminding!

Source/Plugins/Core/com.equella.core/src/com/tle/web/searching/selection/SearchSelectable.java
```diff
@@ -50,7 +50,7 @@ public SectionInfo createSectionInfo(SectionInfo info, SelectionSession session)
 
   @SuppressWarnings("nls")
   protected SectionInfo getSearchTree(SectionInfo info) {
-    return controller.createForward(info, "/searching.do");
+    return controller.createForward(info, "/selectoradd/searching.do");
```
mrblippy, line 53: I'm just wondering how this may inadvertently affect clients who are not using the new ui? 
I can't think of any specific examples, but it might be safe to check if the new UI is enabled before using this path?

PR #2452, opened 2020-10-23
Allow skipping tests in TestNG at runtime

PenghaiZhang:
#### Checklist

- [ ] tests are included

##### Description of change

Add the ability of skipping tests in TestNG at runtime based on the environment.

There are three steps done to support this.
1. Add a new annotation which is used to specify whether to skip a test or not.
2. In 'TestAnnotationTransformer', dynamically disable a test based on the annotation and environment.
3. Include the Transformer in `tesnng-codebuild.yaml `.

This PR also refactored the directory structure where `TestAnnotationTransformer` is sitting.

### DIFF
--- a/autotest/OldTests/src/test/java/io/github/openequella/search/NewSearchPageTest.java
+++ b/autotest/OldTests/src/test/java/io/github/openequella/search/NewSearchPageTest.java
@@ -69,10 +69,6 @@ public class NewSearchPageTest extends AbstractSessionTest {
   @SkipTest(skipOldUI = true)
   public void backToSearchPage() {
     context.getDriver().navigate().back();
-    // todo: remove this when we can skip this test suite in Old UI mode.
-    if (!searchPage.usingNewUI()) {
-      context.getDriver().navigate().refresh();
-    }
     searchPage.get();
     // Expect when going 'back' to the search page, the previous search
     // settings have been remembered.
--- a/autotest/OldTests/src/test/java/io/github/openequella/search/NewSearchPageTest.java
+++ b/autotest/OldTests/src/test/java/io/github/openequella/search/NewSearchPageTest.java
@@ -19,7 +19,7 @@ public class NewSearchPageTest extends AbstractSessionTest {
   }
 
   @Test(description = "open the new Search page and wait for initial search completed")
-  @NewUIOnly(true)
+  @NewUIOnly
   public void initialSearch() {
     searchPage = new NewSearchPage(context).load();
     // The initial search should return 16 items.
@@ -27,7 +27,7 @@ public class NewSearchPageTest extends AbstractSessionTest {
   }
 
   @Test(dependsOnMethods = "initialSearch", description = "Search with a query and refine controls")
-  @NewUIOnly(true)
+  @NewUIOnly
   public void searchByFilters() {
     searchPage.newSearch();
     // Search by Collections.
@@ -56,7 +56,7 @@ public class NewSearchPageTest extends AbstractSessionTest {
   }
 
   @Test(description = "open an item's summary page", dependsOnMethods = "searchByFilters")
-  @NewUIOnly(true)
+  @NewUIOnly
   public void openItemSummaryPage() {
     final String ITEM_TITLE = "Java (cloned)";
     SummaryPage summaryPage = searchPage.selectItem(ITEM_TITLE);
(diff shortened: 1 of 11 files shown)


autotest/OldTests/src/test/java/io/github/openequella/search/NewSearchPageTest.java
```diff
@@ -18,13 +19,15 @@ protected void prepareBrowserSession() {
   }
 
   @Test(description = "open the new Search page and wait for initial search completed")
+  @SkipTest(skipOldUI = true)
```
PenghaiZhang, line 22: I wonder if people want to add this annotation on class level. Adding it on method level gives more flexibility though.

edalex-ian: Depends. If this class is dedicated to newUI, I'd say put it at the class level.

autotest/OldTests/src/test/java/testng/TestAnnotationTransformer.java
```diff
@@ -0,0 +1,31 @@
...
+      annotation.setRetryAnalyzer(FailureRetryAnalyzer.class);
+    }
+
+    SkipTest skipTest = testMethod.getAnnotation(SkipTest.class);
+    // Read the configuration of using new UI or not from environment variable.
+    boolean isNewUIEnabled = Boolean.parseBoolean(System.getProperty(OLD_TEST_NEWUI));
+    // Skip tests that should not run against Old UI when CI is running in Old UI.
+    if (skipTest != null && skipTest.skipOldUI() && !isNewUIEnabled) {
+      annotation.setEnabled(false);
+    }
```
edalex-ian, line 29: Arguably these should be split out into two methods:

* `checkRetryAnnotation`; and
* `checkSkipTestAnnotation`.

autotest/OldTests/src/test/java/testng/annotation/SkipTest.java
```diff
@@ -0,0 +1,9 @@
+package testng.annotation;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
+public @interface SkipTest {
```
edalex-ian, line 7: Seeing this annotation is just about skipping base on old vs new, should we maybe go with naming like NewUiOnly? And have a simple `value` so that then you can have:

```java
@NewUiOnly(true)
@Test
class NewUiTests {
```

autotest/OldTests/testng-codebuild.yaml
```diff
@@ -1,6 +1,7 @@
 name: SingleSuite
 threadCount: 1
-
+listeners:
+  - testng.TestAnnotationTransformer
```
edalex-ian, line 4: Nice work dude! Shame our retry tests appear to have not been running on CI.

edalex-ian (changes requested):
Good stuff @PenghaiZhang , just a couple of tweaks. But after they're done, you're probably good to merge.

ChristianMurphy (approved):

PenghaiZhang:
611 tests were executed on Codebuild and GHA (New UI on). And 606 tests were executed on GHA (New UI off). So I believe the build results are correct.

autotest/OldTests/src/test/java/io/github/openequella/search/NewSearchPageTest.java
```diff
@@ -18,13 +19,15 @@ protected void prepareBrowserSession() {
   }
 
   @Test(description = "open the new Search page and wait for initial search completed")
+  @NewUIOnly(true)
```
ChristianMurphy, line 22: out of curiosity, why does this need to be passed `true`?
Wouldn't it default to true?

PenghaiZhang: Oh I thought it defaults to false :thinking:  I might be wrong.

ChristianMurphy: Okay, and why is the default false?
If it shouldn't run in `NewUIOnly`, wouldn't it not have the `NewUIOnly` annotation?
It seems like `false` would be the exception, not the rule?
Unless I'm missing some context here

PenghaiZhang: I think you are right!

ChristianMurphy (approved):

edalex-ian (approved):
Good stuff @PenghaiZhang :+1:

Issue #2478, opened 2020-11-02
Improved image scaling when setting a theme logo


edalex-ian:
**Is your feature request related to a problem? Please describe.**

In the new UI Theme editor, setting a logo always scales the logo to a fixed size of 230px by 32px with no regard to original images aspect ratio or otherwise.

**Describe the solution you'd like**

To improve flexibility, you should be able to upload any sized image, and it's aspect ratio is kept after scaling.

edalex-ian:
Heya @SammyIsConfused 

Could you test this one please?

SammyIsConfused:
Looks good to me. Aspect ratio is maintained, and width is confined to 230 pixels. I looked at some edge cases - eg a file that was 230 x 3000, an animated gif, a broken file. All were handled fine, where possible converted in to a crisp PNG logo file. Obviously the animated gif just took the first frame, but that frame was high quality and the correct ratio. The super long files with width 230 were of high quality and displayed in full. As a sanity check, I renamed a text file to a PNG and uploaded that, and it was caught by the error boundary. 

Files that are too wide can still be used but the conversion step adds some aliasing when re scaling which is to be expected. 

I'd say this issue is complete, @edalex-ian . :+1:

PR #2479, opened 2020-11-02
feat(server): Maintain logo aspect ratio

edalex-ian:
##### Checklist

##### Description of change

Just a quick one when I was exploring what we can now do with setting up a custom theme. This increases what images we can use for a logo.

Biggest issue was before if you image was not 36px high, it'd get squashed and stretched making most images unrecognisable. As you can see in the screenshot, now we can use nice square ones for example - as somewhat common logo format.

#2478

image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/util/langstrings.ts
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/util/langstrings.ts
@@ -267,7 +267,7 @@ export const languageStrings = {
       title: "Logo Settings",
       siteLogo: "Site Logo",
       siteLogoDescription:
-        "The main logo for the site, primarily displayed in the top left of pages. (Use a PNG file of 230x36 pixels for best results.)",
+        "The main logo for the site, primarily displayed in the top left of pages. (Use a PNG file with a maximum width of 230 pixels for best results.)",
     },
     errors: {
       invalidimagetitle: "Image Processing Error",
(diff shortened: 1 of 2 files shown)


ChristianMurphy (approved):
LGTM :+1: 
Though a more general question, why is ImageMagick used elsewhere while Graphics2D is used here?

SammyIsConfused (changes requested):
Love this. Was far too restrictive before. 
The only thing I'd say is worth including is that you'll want to rework the language string that displays on the Theme Editor page, which prescribes a ratio to use. (Use a PNG file of 230 x 36 for best results).
image

edalex-ian:
> LGTM +1
> Though a more general question, why is ImageMagick used elsewhere while Graphics2D is used here?

I'd say because ImageMagick is a lot more capable than the inbuilt java graphics libraries. ImageMagick is generally considered the most powerful widely available tool suite out there when it comes to automation of media manipulation (and interrogation). As a result it is a common pattern for servers to call out to it (either the CLI, or if native the libraries) to do image related activities.

So I've just made the assumption that's why oEQ uses it. But definitely before my time, and I've not looked closely as it how it's being used.

edalex-ian:
> The only thing I'd say is worth including is that you'll want to rework the language string that displays on the Theme Editor page, which prescribes a ratio to use. (Use a PNG file of 230 x 36 for best results).

What do you suggest. Just drop the bit in parens?

SammyIsConfused:
> 
> What do you suggest. Just drop the bit in parens?
Yup, that should do it. Unless your method requires PNG? I vaguely remember that being the only way to get transparency. Perhaps (Supports .png files)?

edalex-ian:
In the end reworked to advise on max width - so as to avoid scaling artefacts:

image

Will merge on green builds.

SammyIsConfused (approved):
Thanks for fixing that up. Good to go!

PR #2498, opened 2020-11-09
Fix view video attachments from New Search UI search results

edalex-ian:
##### Checklist

- [x] tests are included

##### Description of change

Unfortunately, unless the viewer for the video MIME types was configured as 'file' then the viewer would fail. By default, the viewer is the HTML 5 Viewer of oEQ, so it was attempting to use the resultant HTML from that page in a `<video>` tag.

To fix this, the path of the file for file attachments is also returned in the attachment data for search results. This can then be used by the front end to build a direct `file` URL when wanted. In the case of the viewer, this is pretty well always unless the viewer configuration is for download.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/modules/ViewerModule.test.ts
+++ b/Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/modules/ViewerModule.test.ts
@@ -16,7 +16,10 @@
  * limitations under the License.
  */
 import * as OEQ from "@openequella/rest-api-client";
-import { determineViewer } from "../../../tsrc/modules/ViewerModule";
+import {
+  determineAttachmentViewUrl,
+  determineViewer,
+} from "../../../tsrc/modules/ViewerModule";
 
 describe("determineViewer()", () => {
   const fileAttachmentType = "file";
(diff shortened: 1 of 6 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/modules/ViewerModule.ts
```diff
@@ -65,3 +66,24 @@ export const determineViewer = (
...
+ */
+export const determineAttachmentViewUrl = (
+  itemUuid: string,
+  itemVersion: number,
+  attachmentType: string,
+  viewUrl: string,
+  fileAttachmentPath?: string
+): string =>
+  attachmentType === "file"
+    ? `${AppConfig.baseUrl}file/${itemUuid}/${itemVersion}/${fileAttachmentPath}`
```
SammyIsConfused, line 88: nit: since the string "file" is checked on line 87 and then used again in line 88's string literal, perhaps it could be a constant.

SammyIsConfused (approved):
Looks good! Just one nitpick - feel free to ignore.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/modules/ViewerModule.test.ts
```diff
@@ -58,3 +61,25 @@ describe("determineViewer()", () => {
...
+    ).toEqual(viewUrl));
+
+  it("returns an oEQ 'file' URL for attachmentType 'file'", () =>
+    expect(
+      determineAttachmentViewUrl(
+        uuid,
+        version,
+        "file",
+        viewUrl,
+        "directory/file.txt"
```
PenghaiZhang, line 82: Just be curious here . Does `fileAttachment.getFilename` actually return a relative path ?  I wonder this because the file path in this test is "directory/file.txt".

PenghaiZhang (approved):
I think it's good to go.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/ViewerModule.ts
edalex-ian: Thanks @SammyIsConfused .

In this case this is more the value of string union, vs. just the structure of a URL. I wouldn't want to imply they're directly related.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/modules/ViewerModule.test.ts
edalex-ian: I wondered the same thing. So went through some manual testing where I uploaded a zip file with several directories and then extracted that into an item and discovered it is indeed the path. (Shame the naming etc. is unclear.)

PR #2504, opened 2020-11-12
Feature/navigate to item summary page

PenghaiZhang:
#688 

##### Checklist

- [ ] tests are included

##### Description of change

This PR focuses on the feature of navigating from new Search page to the ItemSummary page. 

In the context of Selection Session, the link to ItemSummary page requires some extra query params to help the Legacy server rendering. 

All required information is available in front-end now,  and we just need to add these query params in each Item title link.

Also, you will see `MUI Link` is used. This is probably due to two reasons: 
1. the mixed environment where new tsx component is used in Legacy content;
2. the only component that can be used when the ItemSummary  page URL is matched in the routes is `LegacyPage`. 

So I suggest we come back and address the problem when we have a new tsx component for ItemSummary page. 

A [video](https://streamable.com/glaepc) is available here.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/AppConfig.ts
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/AppConfig.ts
@@ -50,12 +50,8 @@ export interface RenderData {
 }
 
 declare const renderData: RenderData | undefined;
-export const getRenderData = (): RenderData | undefined => {
-  if (typeof renderData !== "undefined") {
-    return renderData;
-  }
-  return undefined;
-};
+export const getRenderData = (): RenderData | undefined =>
+  typeof renderData !== "undefined" ? renderData : undefined;
 
 export const AppConfig: Config = {
   baseUrl: document?.getElementsByTagName("base")[0]?.href ?? "",
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/AppConfig.ts
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/AppConfig.ts
@@ -50,7 +50,12 @@ export interface RenderData {
 }
 
 declare const renderData: RenderData | undefined;
-export const getRenderData = () => renderData;
+export const getRenderData = () => {
+  if (typeof renderData !== "undefined") {
+    return renderData;
+  }
+  return undefined;
+};
 
 export const AppConfig: Config = {
   baseUrl: document?.getElementsByTagName("base")[0]?.href ?? "",
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/AppConfig.ts
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/AppConfig.ts
@@ -22,6 +22,36 @@ interface Config {
   baseUrl: string;
 }
 
+/**
+ * Structure of Selection Session information which is used to help build
+ * components(e.g. SearchResult) in the context of Selection Session.
+ */
+export interface SelectionSessionInfo {
+  /**
+   * The ID of a Selection Session
+   */
+  stateId: string;
+  /**
+   * The ID of an LMS Integration
+   */
+  integId?: string;
+  /**
+   * The UI layout used in Selection Session
+   */
+  layout: "coursesearch" | "search" | "skinnysearch";
+}
+
+export interface RenderData {
+  baseResources: string;
+  newUI: boolean;
+  autotestMode: boolean;
+  newSearch: boolean;
+  selectionSessionInfo: SelectionSessionInfo | null;
+}
+
+declare const renderData: RenderData | undefined;
+export const getRenderData = () => renderData;
+
 export const AppConfig: Config = {
   baseUrl: document?.getElementsByTagName("base")[0]?.href ?? "",
 };
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/AppConfig.ts
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/AppConfig.ts
@@ -50,7 +50,7 @@ export interface RenderData {
 }
 
 declare const renderData: RenderData | undefined;
-export const getRenderData = () => {
+export const getRenderData = (): RenderData | undefined => {
   if (typeof renderData !== "undefined") {
     return renderData;
   }
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/components/RichTextEditor.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/components/RichTextEditor.tsx
@@ -18,7 +18,7 @@
 import * as React from "react";
 import { Editor } from "@tinymce/tinymce-react";
 import { AxiosPromise, AxiosResponse } from "axios";
-import { AppConfig } from "../AppConfig";
+import { AppConfig, getRenderData } from "../AppConfig";
 
 import "tinymce/tinymce";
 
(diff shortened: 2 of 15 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResult.tsx
```diff
@@ -366,11 +381,36 @@ export default function SearchResult({
...
+      if (indexNumber !== undefined && totalCount !== undefined) {
+        itemSummaryPageLink = itemSummaryPageLink.concat(
+          `&index=${indexNumber}&available=${totalCount}`
+        );
+      }
+
+      // integId can be null in 'Resource Selector'.
+      if (integId) {
+        itemSummaryPageLink = itemSummaryPageLink.concat(`&_int.id=${integId}`);
+      }
```
edalex-ian, line 401: This block of logic should really be out in the Module layer. It's more to do with the internal workings of oEQ as opposed to how we present a result.

```typescript
const buildSelectionSessionItemSummaryLink =
  ({ stateId, integId, layout }: SelectionSessionInfo,
  indexNumber?: number, totalCount?: number): string => {
    ...
}
```

If you had something like that, then this method could be more (declarative) like:

```typescript
const itemLink = () => {
  const itemTitle = name ? highlightField(name) : uuid;
  const basicLink = <Link to={routes.ViewItem.to(uuid, version)}>{itemTitle}</Link>;
  return renderData?.selectionSessionInfo ?
    (<MUILink
      href={buildSelectionSessionItemSummaryLink(renderData.selectionSessionInfo)}
      underline="none">
        {itemTitle}
    </MUILink>) :
    basicLink;
}
```

Added bonus, would be that you can then also easily unit test this.

edalex-ian (changes requested):
Looks good @PenghaiZhang, just needs a couple of tweaks me thing.

And also, I think it could be prudent to add a two tests:

1. The generation of the link URL; and
2. A test of the page that it correctly switches between the two types of links based on the presence of `selectSessionInfo`.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResult.tsx
```diff
@@ -112,12 +117,22 @@ export interface SearchResultProps {
...
    */
   item: OEQ.Search.SearchResultItem;
+  /**
+   * The index of an search result Item which does not take Pagination into account.
+   */
+  indexNumber: number;
+  /**
+   * The total number of available Items matching current Search options.
+   */
+  totalCount: number;
```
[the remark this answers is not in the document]
PenghaiZhang: Yes they will! And perhaps more importantly,  these two buttons are missing if ItemSummary page is opened from new Search page. But I already worked out why and how to fix this.

PenghaiZhang:
> Looks good @PenghaiZhang, just needs a couple of tweaks me thing.
> 
> And also, I think it could be prudent to add a two tests:
> 
> 1. The generation of the link URL; and
> 2. A test of the page that it correctly switches between the two types of links based on the presence of `selectSessionInfo`.

I have done the first test. For the second one, I think we should do it in the test for `SearchResult` ?

edalex-ian:
> I have done the first test. For the second one, I think we should do it in the test for `SearchResult` ?

Yup, that's what I had in mind. :)

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SelectionSessionModule.ts
```diff
@@ -0,0 +1,43 @@
...
+ * limitations under the License.
+ */
+import { AppConfig } from "../AppConfig";
+import { SelectionSessionInfo } from "../mainui";
+
+/**
+ * Build a Selection Session specific ItemSummary Link.
+ * @param stateId The ID of a Selection Session
+ * @param integId The ID of a LSM Integration
+ * @param layout The UI layout used in Selection Session
```
edalex-ian, line 25: Hmmm, TSDoc is a bit curious here, as really the param is a `SelectionSessionInfo` object and you'd refer there for the value of the inside params. :confused: 

Here's how it's done in JSDoc land... But no TSDoc / typedoc description is jumping out for me. :disappointed:

Source/Plugins/Core/com.equella.core/js/tsrc/AppConfig.ts
```diff
@@ -22,6 +22,41 @@ interface Config {
...
+export interface RenderData {
+  baseResources: string;
+  newUI: boolean;
+  autotestMode: boolean;
+  newSearch: boolean;
+  selectionSessionInfo: SelectionSessionInfo | null;
+}
+
+declare const renderData: RenderData | undefined;
+export const getRenderData = () => {
```
edalex-ian, line 53: Being an exported function, we should make sure we fully type it:

```suggestion
export const getRenderData = (): RenderData | undefined => {
```

Source/Plugins/Core/com.equella.core/js/tsrc/AppConfig.ts
```diff
@@ -22,6 +22,41 @@ interface Config {
...
+  if (typeof renderData !== "undefined") {
```
edalex-ian, line 54: Nit: Should this be more specific - should we be checking to see if `renderData` is an instance of `RenderData`. If so, maybe rework this function to a simple `() => match(...)`.

Source/Plugins/Core/com.equella.core/js/tsrc/components/RichTextEditor.tsx
```diff
@@ -59,9 +59,8 @@ import "tinymce/plugins/toc";
 import "tinymce/plugins/visualblocks";
 import "tinymce/plugins/visualchars";
 import "tinymce/plugins/wordcount";
-import type { RenderData } from "../mainui";
 
-declare const renderData: RenderData | undefined;
+const renderData = getRenderData();
```
edalex-ian, line 63: Noice! :+1:

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/index.tsx
```diff
@@ -20,32 +20,6 @@ import * as ReactDOM from "react-dom";
...
-  layout: string;
-}
-
-export interface RenderData {
-  baseResources: string;
-  newUI: boolean;
-  autotestMode: boolean;
-  newSearch: boolean;
-  selectionSessionInfo: SelectionSessionInfo | null;
-}
```
edalex-ian, line 48: Yup, I think this makes much more sense now in `AppConfig`. Good work. :+1:

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResult.tsx
```diff
@@ -137,7 +140,7 @@ export default function SearchResult({
     attachment: OEQ.Search.Attachment;
     viewerDetails?: OEQ.MimeType.MimeTypeViewerDetail;
   }
-
+  const renderData = getRenderData();
```
edalex-ian, line 143: Nit. Seeing you're only using in once place below, throwing in a variable is a bit redundant.

edalex-ian (approved):
GTG dude. I've dropped a few queries in there, and a couple of optional nit-picks. Other than that, some wording enhancements for the tests.

Once you've addressed what you think should be, go for a merge. ;)

Source/Plugins/Core/com.equella.core/js/tsrc/AppConfig.ts
PenghaiZhang: In order to use `match(...)`, we need to create a `runtype` for `RenderData` and one for `SelectionSessionInfo`. 
There are discussions around the feature of transforming a TS interface to a runtype and this feature is not supported yet.

https://github.com/pelotom/runtypes/issues/128

PenghaiZhang: I guess we have to resort to `typeof` to ensure `settings page` works fine in Old UI. This is because `renderData` does not exist in Old UI at all. So any use of it will trigger a `uncaught reference error`.

PR #2414, opened 2020-10-15
Feature/legacy search params

mrblippy:
##### Checklist

- [x] the [contributor license agreement][] is signed
- [x] commit message follows [commit guidelines][]
- [x] tests are included
- [x] screenshots are included showing significant UI changes

##### Description of change

#1306

This feature allows a Legacy /searching.do url and query parameters to be used on the new search page.
old to new search

Some bumps were experienced while writing an additional SearchPage test. From my commit message:
> * I also had to add the esModuleInterop property to tsconfig. Otherwise, the rendering of a SearchPage with dateRangeQuickModeEnabled in jest will fail. The solution was found at https://github.com/mui-org/material-ui-pickers/issues/805
> * I had to modify how the imports work in the jest setup. This came about as a result of adding the esModuleInterop

[contributor license agreement]: https://www.apereo.org/node/676
[commit guidelines]: https://chris.beams.io/posts/git-commit/

PR #2506, opened 2020-11-12
Feature/new search query params

mrblippy:
##### Checklist

- [x] tests are included

##### Description of change

#1306

This builds further upon the PR i did to support legacy search parameters (#2414)

Summary of work performed:
- Adding SearchModule function to serialize a searchOptions object to a JSON string, to be used in a query string. This is intended to be used to allow for a new Search Page user to get a link to copy their current search.
  - in order to cut down on characters, only the bare minimum collection and owner identifiers are serialized. Undefined properties are also excluded.
- Deserialisation of the query string back into a searchOptions object
- Updating the SearchPage to accomodate the fact that we now can support two types of linkable searches.

- A PR will be submitted in the near future to add a button on the SearchPage to obtain the linkable url

All feedback welcomed, I have been staring at this for too long 😑

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/modules/SearchModule.test.ts
+++ b/Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/modules/SearchModule.test.ts
@@ -17,15 +17,19 @@
  */
 import * as OEQ from "@openequella/rest-api-client";
 import { getCollectionMap } from "../../../__mocks__/getCollectionsResp";
+import { allSearchOptions } from "../../../__mocks__/searchOptions.mock";
 import { getSearchResult } from "../../../__mocks__/SearchResult.mock";
 import { users } from "../../../__mocks__/UserSearch.mock";
-import * as CollectionModule from "../../../tsrc/modules/CollectionsModule";
+import * as CollectionsModule from "../../../tsrc/modules/CollectionsModule";
 import type { SelectedCategories } from "../../../tsrc/modules/SearchFacetsModule";
 import * as SearchModule from "../../../tsrc/modules/SearchModule";
 import {
-  convertParamsToSearchOptions,
+  queryStringParamsToSearchOptions,
   DateRange,
   defaultSearchOptions,
+  generateQueryStringFromSearchOptions,
+  legacyQueryStringToSearchOptions,
+  newSearchQueryToSearchOptions,
   SearchOptions,
 } from "../../../tsrc/modules/SearchModule";
 import { SortOrder } from "../../../tsrc/modules/SearchSettingsModule";
@@ -116,10 +120,10 @@ describe("SearchModule", () => {
     expect(SearchModule.generateCategoryWhereQuery([])).toBeUndefined();
   });
 
-  describe("convertParamsToSearchOptions", () => {
+  describe("newSearchQueryToSearchOptions", () => {
     const mockedResolvedUser = jest.spyOn(UserModule, "resolveUsers");
     const mockedCollectionListSummary = jest.spyOn(
-      CollectionModule,
+      CollectionsModule,
       "collectionListSummary"
     );
 
@@ -127,15 +131,46 @@ describe("SearchModule", () => {
       jest.clearAllMocks();
     });
 
+    it("should convert query string to searchOptions", async () => {
+      mockedResolvedUser.mockResolvedValue(users);
+      mockedCollectionListSummary.mockResolvedValueOnce(getCollectionMap);
+      const longSearch =
+        '{"rowsPerPage":10,"currentPage":0,"sortOrder":"NAME","query":"test machine","rawMode":true,"status":["LIVE","REVIEW"],"searchAttachments":true,"selectedCategories":[{"id":766943,"categories":["Hobart"]},{"id":766944,"categories":["Some cool things"]}],"collections":[{"uuid":"8e3caf16-f3cb-b3dd-d403-e5eb8d545fff"},{"uuid":"8e3caf16-f3cb-b3dd-d403-e5eb8d545ffe"},{"uuid":"8e3caf16-f3cb-b3dd-d403-e5eb8d545ffg"},{"uuid":"8e3caf16-f3cb-b3dd-d403-e5eb8d545ffa"},{"uuid":"8e3caf16-f3cb-b3dd-d403-e5eb8d545ffb"}],"lastModifiedDateRange":{"start":"2020-05-26T03:24:00.889Z","end":"2020-05-27T03:24:00.889Z"},"owner":{"id":"680f5eb7-22e2-4ab6-bcea-25205165e36e"}}';
+      const convertedParamsPromise = await newSearchQueryToSearchOptions(
+        longSearch
+      );
+      expect(convertedParamsPromise).toEqual(allSearchOptions);
+    });
+  });
+
+  describe("convertParamsToSearchOptions", () => {
+    const mockLocation = {
+      pathname: "/search",
+      hash: "",
+      search: "",
+      state: "",
+    };
     it("should return undefined if no query string parameters are defined", async () => {
-      const convertedParamsPromise = await convertParamsToSearchOptions("");
+      const convertedParamsPromise = await queryStringParamsToSearchOptions(
+        mockLocation
+      );
       expect(convertedParamsPromise).toBeUndefined();
     });
+  });
+
+  describe("legacyQueryStringToSearchOptions", () => {
+    const mockedResolvedUser = jest.spyOn(UserModule, "resolveUsers");
+    const mockedCollectionListSummary = jest.spyOn(
+      CollectionsModule,
+      "collectionListSummary"
+    );
 
     it("should return default search for any parameters that aren't supported", async () => {
       const unsupportedQueryString = "?test=nothing&cool=beans";
       expect(
-        await convertParamsToSearchOptions(unsupportedQueryString)
+        await legacyQueryStringToSearchOptions(
+          new URLSearchParams(unsupportedQueryString)
+        )
       ).toEqual(defaultSearchOptions);
     });
 
@@ -147,8 +182,8 @@ describe("SearchModule", () => {
       const fullQueryString =
         "?in=C8e3caf16-f3cb-b3dd-d403-e5eb8d545fff&q=test&sort=datecreated&owner=680f5eb7-22e2-4ab6-bcea-25205165e36e&dp=1601510400000&dr=AFTER";
 
-      const convertedParamsPromise = await convertParamsToSearchOptions(
-        fullQueryString
+      const convertedParamsPromise = await legacyQueryStringToSearchOptions(
+        new URLSearchParams(fullQueryString)
       );
 
       const expectedSearchOptions: SearchOptions = {
@@ -209,8 +244,8 @@ describe("SearchModule", () => {
     ])(
       "converts legacy date range query params: %s to search options containing lastModifiedDateRange of %s",
       async (queryString, expectedRange) => {
-        const convertedSearchOptions = await convertParamsToSearchOptions(
-          queryString
+        const convertedSearchOptions = await legacyQueryStringToSearchOptions(
+          new URLSearchParams(queryString)
         );
         expect(convertedSearchOptions).toEqual({
           ...defaultSearchOptions,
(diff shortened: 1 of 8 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -184,6 +186,92 @@ export const generateCategoryWhereQuery = (
...
+
+/**
+ * A function that takes a JSON representation of a SearchOptions object, and converts it into an actual SearchOptions object.
+ * @param searchOptionsJSON a JSON representation of a SearchOptions object.
+ * @return searchOptions Search options selected on Search page.
+ */
+export const newSearchQueryToSearchOptions = async (
+  searchOptionsJson: string
+): Promise<SearchOptions> => {
+  const parsedOptions: SearchOptions = JSON.parse(
```
edalex-ian, line 245: As discussed with @mrblippy 

1. The type here is not `SearchOptions`, really it should be `unknown` and we validate after further work below;
2. For the code further down to rehydrate collections and owner (which had to be split out due to needing async calls), they should be pushed into their own functions for readability.
3. At the end of all this, we really need some validation to confirm we are actually returning `SearchOptions`.

Grab me at some point, and we'll work together to refactor this function.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -221,18 +309,35 @@ export const searchItems = ({
...
 };
 
+const getCollectionDetails = async (
+  collectionUuids: string[]
+): Promise<Collection[] | undefined> => {
+  const collectionList = await collectionListSummary([
+    OEQ.Acl.ACL_SEARCH_COLLECTION,
+  ]);
+  const filteredCollectionList = collectionList.filter((c) => {
+    return collectionUuids.includes(c.uuid);
```
edalex-ian, line 319: `return` is superfluous.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -221,18 +309,35 @@ export const searchItems = ({
...
+  });
+  return filteredCollectionList.length > 0
```
edalex-ian, line 321: I wonder if prior to this we want to log a warning or something if `filteredCollectionList.length !== collectionUuids.length`.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -221,18 +309,35 @@ export const searchItems = ({
...
+    ? filteredCollectionList
+    : defaultSearchOptions.collections;
```
edalex-ian, line 323: This function reads more generic, and so shouldn't be tied to a specific implementation detail. Therefore I think it should return `undefined` or `[]`. I think possibly `[]` because if a list has been the input, it'd seem the output should also be a list - even if empty due to issues.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -221,18 +309,35 @@ export const searchItems = ({
...
+};
+
+const getUserDetails = async (
+  userId: string
+): Promise<OEQ.UserQuery.UserDetails | undefined> => {
+  const userDetails = await resolveUsers([userId]);
+  return userDetails.length > 0 ? userDetails[0] : defaultSearchOptions.owner;
```
edalex-ian, line 330: hmmm, if we can end up with more than 1 user, we should probably at least log a warning, but maybe we need to throw an error. Just assuming to use the first one is possibly a bit risky. :thinking:

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -221,18 +309,35 @@ export const searchItems = ({
...
+};
+
 /**
  * A function that takes query string params from a shared searching.do URL and converts all applicable params to Search options
- * @param queryString query string params from a shared `searching.do` URL
- * @return SearchOptions object, or undefined if there were no query string parameters.
+ * @param location query string params from a shared `searching.do` URL
```
edalex-ian, line 335: Looks like this comment is completely outta wack. :confused:

edalex-ian (changes requested):
So most of the comments are about comments and the like.

But two significant ones in `SearchModule` around the serialisation and de-serialisation - which I appreciate are the two hard parts of this work. So just let me know when you're in a position to work through those tricky bits and we can do together.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/modules/SearchModule.test.ts
```diff
@@ -116,26 +120,57 @@ describe("SearchModule", () => {
...
 
     afterEach(() => {
       jest.clearAllMocks();
     });
 
+    it("should convert query string to searchOptions", async () => {
+      mockedResolvedUser.mockResolvedValue(users);
+      mockedCollectionListSummary.mockResolvedValueOnce(getCollectionMap);
+      const longSearch =
+        '{"rowsPerPage":10,"currentPage":0,"sortOrder":"NAME","query":"test machine","rawMode":true,"status":["LIVE","REVIEW"],"searchAttachments":true,"selectedCategories":[{"id":766943,"categories":["Hobart"]},{"id":766944,"categories":["Some cool things"]}],"collections":[{"uuid":"8e3caf16-f3cb-b3dd-d403-e5eb8d545fff"},{"uuid":"8e3caf16-f3cb-b3dd-d403-e5eb8d545ffe"},{"uuid":"8e3caf16-f3cb-b3dd-d403-e5eb8d545ffg"},{"uuid":"8e3caf16-f3cb-b3dd-d403-e5eb8d545ffa"},{"uuid":"8e3caf16-f3cb-b3dd-d403-e5eb8d545ffb"}],"lastModifiedDateRange":{"start":"2020-05-26T03:24:00.889Z","end":"2020-05-27T03:24:00.889Z"},"owner":{"id":"680f5eb7-22e2-4ab6-bcea-25205165e36e"}}';
```
edalex-ian, line 138: Just a thought on this test...

What is currently here is essentially a regression test. That is, it will work to ensure that if the format of the params change in the future, we'll need to deliberately address it.

However what we missing is a test which explicitly takes the output from the serialisation of params and feeds to back into the deserializer. That is, validating what we produce we can definitely consume at any point in time.

So you may need to add one more test.

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -123,7 +125,7 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
   const [searchSettings, setSearchSettings] = useState<SearchSettings>();
   const [classifications, setClassifications] = useState<Classification[]>([]);
 
-  const location = useLocation();
+  const location = useLocation<LocationState>();
```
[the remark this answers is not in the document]
mrblippy: I mistakenly left this in while still in development. I will remove it, as it has no reason to be here anymore

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/modules/SearchModule.test.ts
mrblippy: Good point, I hadn't thought of that case. It has been added

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -184,6 +186,92 @@ export const generateCategoryWhereQuery = (
...
+      (key: string, value: (object | Collection)[]) => {
+        return match(
+          [
+            Literal("collections"),
+            () => value.map((collection) => pick(collection, ["uuid"])),
+          ],
+          [Literal("owner"), () => pick(value, ["id"])],
+          [Unknown, () => value ?? undefined]
+        )(key);
+      }
```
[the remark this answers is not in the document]
mrblippy: As discussed in-person, runtypes won't match on regular Typescript types. I would need to define them again as Runtypes, which is probably overkill

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
mrblippy: Thank you very much for assisting me in refactoring this @edalex-ian, it was very much appreciated.

However, i wasn't able to fully guard against a extraneous properties being added during the JSON.parse, due to this issue currently in [the Runtypes library](https://github.com/pelotom/runtypes/issues/169) https://github.com/pelotom/runtypes/issues/169

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
mrblippy: I'm (very slowly) getting better at not doing this, but I still slip up sometimes. Thanks for picking up on it

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
mrblippy: it sure was, my late night typing was subpar at best in this case

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -298,11 +439,22 @@ export const convertParamsToSearchOptions = async (
     )(rangeType.toLowerCase() as RangeType);
   };
 
+  const getCollections = async (collectionId: string | undefined) => {
```
edalex-ian, line 442:
```suggestion
  const parseCollectionId = async (collectionId: string | undefined): Promise<Collection[]> => {
```

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -221,18 +366,35 @@ export const searchItems = ({
   return OEQ.Search.search(API_BASE_URL, searchParams);
 };
 
+const getCollectionDetails = async (
```
edalex-ian, line 369: As discussed, let's tweak this name a bit. At first I was going to suggest `getCollectionsByUuid` however this is just a find, as anything that doesn't match is silently skipped. So:

```suggestion
const findCollectionsByUuid = async (
```

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -221,18 +366,35 @@ export const searchItems = ({
...
+  ]);
+  const filteredCollectionList = collectionList.filter((c) =>
+    collectionUuids.includes(c.uuid)
+  );
+  return filteredCollectionList.length > 0 ? filteredCollectionList : undefined;
+};
+
+const getUserDetails = async (
+  userId: string
+): Promise<OEQ.UserQuery.UserDetails | undefined> => {
```
edalex-ian, line 383: Similar to above, this is probably more so a `find` rather than a get. But I do also wonder if `undefined` is actually ever returned. :thinking: 
```suggestion
const findUser = async (
  userId: string
): Promise<OEQ.UserQuery.UserDetails | undefined> => {
```

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -221,18 +366,35 @@ export const searchItems = ({
...
+  const userDetails = await resolveUsers([userId]);
+  if (userDetails.length > 1)
+    throw new Error(`More than one user was resolved for id: ${userId}`);
+  return userDetails[0];
```
edalex-ian, line 387: What if `userDetails` is empty?

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -63,6 +65,8 @@ import {
 } from "./components/SearchResultList";
 import StatusSelector from "./components/StatusSelector";
 
+// import LocationState = History.LocationState;
+
```
edalex-ian, line 69: Rather than comment out, did you mean to:

```suggestion
```

edalex-ian (changes requested):
Fair effort here @mrblippy 

The issues around types were monumental, and I'll be sure to send out an email to the dev list later covering them. I think it's as good as we can get at this point in time.

But there's a few other minor bits left to sort out. BTY.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
mrblippy: Calling `userDetails[0]` when it's empty will return undefined. So it won't cause any errors.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
mrblippy: Yeah, that function naming makes more sense. But I had to change the return type to be `Promise<Collection[] | undefined>`. Because it turns out that `defaultSearchOptions.collections` returns undefined

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
mrblippy: https://static.wikia.nocookie.net/simpsons/images/c/c3/D%27oh.jpg/revision/latest/top-crop/width/360/height/360?cb=20141226214635 

Yep, that's definitely what should've been done

edalex-ian (approved):
Looks good @mrblippy 

You've now arrived at the ready for merge stage. :wink:

PR #2526, opened 2020-11-18
Handle PDF thumbnailing safely

SammyIsConfused:
##### Checklist

##### Description of change
Thumb-nailing of PDFs is done by an external ImageMagick process. This is then delegated by ImageMagick to a program called Ghostscript. If uploading a broken PDF, Ghostscript will attempt indefinitely to "fix" the PDF, slamming the server CPU and affecting performance. If the process takes too long to finish it can cause issues with server stability. 

This PR solves the problem by adding a new config property to `plugins/com.tle.core.imageMagick/config.properties` called `thumbnailing.timeout`. This adds a timeout cap to the imageMagick process - if exceeded thumb-nailing for that attachment will error out and cancel, and will not be retried. When not set, the default is 20 seconds. It is possible to use the standard non-capped thumb-nailing process by setting it to 0. 

**Note:** Similar problems occur with indexing broken PDFs for freetext search using Tika - this will be addressed in another PR. 

#2527 


Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
```diff
@@ -114,6 +115,38 @@ public static ExecResult exec(String[] cmdarray, Map<String, String> additionalE
     }
   }
 
+  public static ExecResult execWithTimeLimit(long maxDurationInSeconds, String[] options) {
+    if (maxDurationInSeconds < 1L) {
+      LOGGER.info("maxDurationInSeconds not set. Using a regular non-timed process.");
```
edalex-ian, line 120: This will make for some rather verbose logs, and is really only relevant for debugging issues so

```suggestion
      LOGGER.debug("maxDurationInSeconds not set. Using a regular non-timed process.");
```

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
```diff
@@ -114,6 +115,38 @@ public static ExecResult exec(String[] cmdarray, Map<String, String> additionalE
...
+    }
+    return execWithTimeLimit(options, null, null, maxDurationInSeconds);
+  }
+
+  public static ExecResult execWithTimeLimit(
+      String[] cmdarray, Map<String, String> additionalEnv, File dir, long durationInSeconds) {
+    try {
+      final Triple<Process, StreamReader, StreamReader> cp =
+          createProcess(cmdarray, additionalEnv, dir);
+      LOGGER.info("Started timed process");
```
edalex-ian, line 131:
```suggestion
      LOGGER.debug("Started timed process");
```

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
```diff
@@ -114,6 +115,38 @@ public static ExecResult exec(String[] cmdarray, Map<String, String> additionalE
...
+      final Process proc = cp.getFirst();
+      final StreamReader stdOut = cp.getSecond();
+      final StreamReader stdErr = cp.getThird();
+      proc.waitFor(durationInSeconds, TimeUnit.SECONDS);
+      if (!stdErr.isFinished() || !stdOut.isFinished()) {
+        throw new InterruptedException();
+      }
+      LOGGER.info("Timed process finished");
```
edalex-ian, line 139:
```suggestion
      LOGGER.debug("Timed process finished");
```

Source/Plugins/Core/com.equella.core/src/com/tle/core/imagemagick/ImageMagickServiceImpl.java
```diff
@@ -204,12 +209,34 @@ public Dimension getImageDimensions(FileHandle handle, String filename) throws I
...
   @Override
   public Dimension getImageDimensions(File image) throws IOException {
-    ExecResult result =
-        ExecUtils.exec(
-            identifyExe.getAbsolutePath(),
-            "-format",
-            "%wx%h",
-            new String(image.getAbsolutePath().getBytes("UTF-8"), "UTF-8"));
+    ExecResult result;
+    if (image.getAbsolutePath().toLowerCase().contains(".pdf")) {
```
edalex-ian, line 213: Doesn't need to be this specific. Arguably any thumbnail process going rouge/forever should be hit on the head. I think the check should be removed, and the configuration property made more generic - e.g. `thumbnailing.timeout`.

That in turn raises the question as to whether all thumbnailing goes through here. And if not, should we really do the circuit breaker higher up.

Source/Plugins/Core/com.equella.core/src/com/tle/core/imagemagick/ImageMagickServiceImpl.java
```diff
@@ -204,12 +209,34 @@ public Dimension getImageDimensions(FileHandle handle, String filename) throws I
...
+              });
+    } else {
+      result =
+          ExecUtils.exec(
+              identifyExe.getAbsolutePath(),
+              "-format",
+              "%wx%h",
+              new String(
+                  image.getAbsolutePath().getBytes(StandardCharsets.UTF_8),
+                  StandardCharsets.UTF_8));
```
edalex-ian, line 238: This duplicates the above. As per above, this branch may be removed. But if not, this should be centralised in a shared String[] as used above.

Source/Plugins/Core/com.equella.core/src/com/tle/core/imagemagick/guice/ImageMagickModule.java
```diff
@@ -32,5 +32,6 @@ protected String getFilename() {
   @Override
   protected void configure() {
     bindProp("imageMagick.path");
+    bindInt("imageMagick.seconds.before.cancel.pdf.thumbnailing", 20);
```
edalex-ian, line 35: This should also be added to the example optional config properties file as way of documentation.

edalex-ian (changes requested):
Good stuff @SammyIsConfused , looks like this would solve the issue. But just a few things:

1. Use caution with logging at INFO and above - as it can render the logs quickly useless for typical (non-dev) use;
2. The approach should really be one which is along the lines of "thumbnailing is a non-critical task, and so should be constrained to x seconds per attachment" (i.e. regardless of type); and
3. If we add a new config param, we should ensure it's documented at least in an example config file.

Source/Plugins/Core/com.equella.core/src/com/tle/core/imagemagick/ImageMagickServiceImpl.java
SammyIsConfused: All of the ThumbnailGenerators (ImageThumbnailGenerator, PDFThumbnailGenerator, VideoThumbnailGenerator) pass through here. There is one exception - in the ImageThumbnailGenerator it checks if the file is a gif. If it is, it goes straight to conversion. All others hit here first. 
I could resolve this in a few different ways: 
- Surround the conversion step with a time limit too
- Move this up to generateThumbnailAdvanced - all thumbnailing operations use this. However - this would work quite a bit differently as it's not an external process. 
- Make the GIF check run getImageDimensions too
- Refactor this PDF validation check out into it's own operation that gets run before all else in thumbnailing. Essentially running the identify call again but this time right at the start, doing nothing with the output - just as a validation step. And then operate as normal. While this would add a new load, if the file is valid this will take a split second and neglible processing power.

SammyIsConfused: Went with the option to refactor this into it's own validation step.

PenghaiZhang:
Hi Sam, I just have one query. If I upload a broken PDF which thumbnailing exceeds the timeout, will I see the 500 error or anything else that tells me the timeout?

edalex-ian (approved):
LGTM @SammyIsConfused 

You'll just want to restart the failing build though to make sure it's green and good to merge.

SammyIsConfused:
> Hi Sam, I just have one query. If I upload a broken PDF which thumbnailing exceeds the timeout, will I see the 500 error or anything else that tells me the timeout?

There will be a TimeoutException in the Resource centre logs, but no indication on the user side from the browser that anything has happened.

SammyIsConfused:
MimeTypesTest failed in the old UI :thinking: I'll retry in case it was a fluke.

SammyIsConfused:
Okay this time around everything was fine except one of the Scalacheck tests in the new UI (Wizard file uploads.edit). So definitely both flakey.

SammyIsConfused:
That MimeTypesTest keeps tripping up. I've added a wait before a check it fails on, hopefully that helps. Can't see anything that ties that test to this branch though :thinking:

SammyIsConfused:
There we go, that seems to have done the trick! :confetti_ball: Merging.

PR #2547, opened 2020-11-24
Feature/select resource from new search page

PenghaiZhang:
#688 

##### Checklist

- [x] tests are included

##### Description of change

This PR includes much more code than I imaged.

Here is a list of what changes have been made: 

1. Functions made in 'SelectionSessionModule.ts' to support selecting resources from new search UI (e.g. API calls and DOM update).

2. Changes made in some legacy Java/JS files in order to ensure selecting/unselecting resources from either `structure` or `selectOrAdd` in both New and Old UI  works correctly.

3. Changes made in `SearchPage.tsx` to skip query string param conversion if Selection Session is open. 

4. New Jest tests added for `SearchResult` and `SelectionSessionModule`. 

Here are two videos.
[selectOrAdd](https://streamable.com/av1juz)
[structured](https://streamable.com/r5x591)


Source/Plugins/Core/com.equella.core/resources/web/scripts/attachments/attachments.js
```diff
@@ -156,7 +156,13 @@ var Attachments = (function()
 				CourseList.scrollToSelected();
 				CourseList.transfer($ul);
 			}
-			selectAllCallback(uuids, itemId, itemExtensionType);
+
+			// To support using the standard reload function for selecting all attachment,
```
PenghaiZhang, line 160: The format of these comments is broken here but looks good locally. :thinking:

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/modules/SelectionSessionModule.test.ts
```diff
@@ -16,44 +16,73 @@
...
 
 const basicSelectionSessionInfo: SelectionSessionInfo = {
   stateId: "1",
   layout: "coursesearch",
 };
+const withIntegId: SelectionSessionInfo = {
+  ...basicSelectionSessionInfo,
+  integId: "2",
+};
+const basicRenderData: RenderData = {
```
PenghaiZhang, line 34: This mocked renderData consts and functions appear in here and `SearchResult.test.ts`. Perhaps  we can centralise them in a utility file ?

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SelectionSessionModule.ts
```diff
@@ -32,6 +107,44 @@ const isSelectionSessionInfo = (
...
+};
+
+/**
+ * Send a POST request to submit selected resources.
+ *
+ * @param path URL of an endpoint for submitting selected resources
+ * @param data Payload of the request
+ * @param callback Function called when the request is successful
+ */
+const submitSelection = <T>(
```
PenghaiZhang, line 142: I did not put this function in the REST Module as we don't have REST endpoints for Selection Session yet. In addition, I am unsure how to write tests for this function. I can make a mocked `Axios.post` though and check if the callback has been called or not.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SelectionSessionModule.ts
```diff
@@ -42,19 +155,120 @@ export const buildSelectionSessionItemSummaryLink = (
...
+    event: serverSideEvent,
+  };
+
+  const postData: SelectionSessionPostData = {
+    event__: ["_slcl.reloadFolder"], // This refers to the method 'reloadFolder' defined in 'CourseListSection'.
+    eventp__0: [`${JSON.stringify(courseListUpdateData)}`],
+    ...getBasicPostData(),
+  };
+
+  return submitSelection(
```
edalex-ian, line 222: So is this:

```suggestion
  return submitSelection<void>(
```

Source/Plugins/Core/com.equella.core/resources/web/scripts/attachments/attachments.js
edalex-ian: I'd imagine it's tabs vs. spaces in this ol' stuff.

Source/Plugins/Core/com.equella.core/resources/web/scripts/courselist.js
```diff
@@ -184,29 +184,34 @@ var CourseList = {
...
-				var $newTree = $(newTree.html);
-				$tree.find('[data-folderid="' + folderId + '"]').replaceWith($newTree.find('[data-folderid="' + folderId + '"]'));
-				$(document).trigger('equella_courselistupdate');
+        CourseList.updateCourseList(result);
 			});
 		}, JSON.stringify(reloadData));
-	}
+	},
+
+  updateCourseList: function(result) {
```
edalex-ian, line 204: Was this pushed into a function to allow you to call from TS side?

Source/Plugins/Core/com.equella.core/resources/web/scripts/courselist.js
```diff
@@ -184,29 +184,34 @@ var CourseList = {
...
+    var $tree = $(".foldertree");
+    if (!$tree.length)
+    {
+      return;
+    }
+    var folderId = result.folderId;
+    var newTree = result.updates['courselistajax'];
+    var $newTree = $(newTree.html);
+    // Update the whole folder tree rather than a single node.
+    $tree.html($newTree.find(".foldertree"))
```
edalex-ian, line 214: This seems like a significant change and removal of an optimisation. Could you explain the backstory here please?

Source/Plugins/Core/com.equella.core/src/com/tle/web/itemlist/standard/AbstractItemlikeListAttachmentDisplaySection.java
```diff
@@ -396,7 +394,42 @@ public void selectAttachment(
...
+          selectionService.getSelectAttachmentHandler(info, viewableItem, attachmentUuid);
+      if (selectAttachmentHandler != null) {
+        selectAttachmentHandler.handleAttachmentSelection(
+            info, itemId, attachment, extensionType, true);
+      }
+    }
+  }
 
+  @EventHandlerMethod
+  public void selectAttachmentsFromNewSearch(
```
edalex-ian, line 416: Sorry, I've lost track of the threads. Where is this triggered from?

edalex-ian (changes requested):
Wowsers, there's a fair tangle here. Looking forward to the future where we can do the full New UI implementation of Selection sessions. But in the mean time, fair effort wrangling all that together! :+1: 

A couple of minor tweaks requested. And then probably also a desire to futher highlight all this AJAX interaction code is 'Legacy'.

There's a lot of building of data structures etc in SelectionSessionModule on the legacy side. Should we test that? Feels like we should, feels like it could break easily. :confused:

Source/Plugins/Core/com.equella.core/resources/web/scripts/courselist.js
PenghaiZhang: Yes it's. It was part of `updateTargetFolder`

Source/Plugins/Core/com.equella.core/resources/web/scripts/courselist.js
PenghaiZhang: With previous code, if we want to remove some selected resources in the dialog, when we close the dialog, we will only see that the currently selected course gets its UI updated. 

For example, course A has two resources and course B has one. And B is currently selected.

So in the course list we should see
```
A   (2)
B   (1)
```

And then we remove all selected resources and it's expected to see
```
A   (0)
B   (0)
```

But what is actually displayed is 
```
A   (2)
B   (0)
```

A does not get updated. If we navigate to a different page which results in a page refresh, we will see the correct UI.

PenghaiZhang: With previous code, if we want to remove some selected resources in the dialog, when we close the dialog, we will only see that the currently selected course gets its UI updated. 

For example, course A has two resources and course B has one. And B is currently selected.

So in the course list we should see
```
A   (2)
B   (1)
```

And then we remove all selected resources and it's expected to see
```
A   (0)
B   (0)
```

But what is actually displayed is 
```
A   (2)
B   (0)
```

A does not get updated. If we navigate to a different page which results in a page refresh, we will see the correct UI.

Source/Plugins/Core/com.equella.core/resources/web/scripts/courselist.js
PenghaiZhang: Yes, we use this function in `SelectionSessionModule.ts`

Source/Plugins/Core/com.equella.core/src/com/tle/web/itemlist/standard/AbstractItemlikeListAttachmentDisplaySection.java
PenghaiZhang: The only usage of this is in `SelectionSessionModule.ts`. There is an existing one named `selectAttachment`, and there are two key differences:

1. this one will not unselect an attachment if the button is clicked again.
2. this one supports selecting multiple attachments.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SelectionSessionModule.ts
PenghaiZhang: Good idea. I have renamed the module.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/components/SearchResult.test.tsx
```diff
@@ -172,19 +180,78 @@ describe("<SearchResult/>", () => {
...
+      summaryPage,
+      allAttachments,
+      attachment,
+    } = languageStrings.searchpage.selectResource;
+    const STRUCTURED = "structured";
+    const SELECT_OR_ADD = "selectOrAdd";
+    const selectForCourseFunc =
+      SelectionSessionModule.selectResourceForCourseList;
+    const selectForNonCourseFunc =
+      SelectionSessionModule.selectResourceForNonCourseList;
```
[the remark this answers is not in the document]
PenghaiZhang: Because there is an import of the entire module into a single variable so I was thinking if I should skip the duplicated imports.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/components/SearchResult.test.tsx
```diff
@@ -172,19 +180,78 @@ describe("<SearchResult/>", () => {
...
+          ...basicRenderData,
+          selectionSessionInfo: selectionSessionInfo,
+        });
+
+        const { queryByLabelText, getByLabelText } = await renderSearchResult(
+          mockData.attachSearchObj
+        );
+        expect(queryByLabelText(resourceType)).toBeInTheDocument();
+
+        fireEvent.click(getByLabelText(resourceType));
```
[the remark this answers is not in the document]
PenghaiZhang: Sorry for the confusion. Yes your guess is 100% correct.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/components/SearchResult.test.tsx
```diff
@@ -172,19 +180,78 @@ describe("<SearchResult/>", () => {
...
+      attachment,
+    } = languageStrings.searchpage.selectResource;
+    const STRUCTURED = "structured";
+    const SELECT_OR_ADD = "selectOrAdd";
+    const selectForCourseFunc =
+      SelectionSessionModule.selectResourceForCourseList;
+    const selectForNonCourseFunc =
+      SelectionSessionModule.selectResourceForNonCourseList;
+
+    it.each([
```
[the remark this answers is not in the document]
PenghaiZhang: What about adding a test to check these buttons disappear in non-Selection Sessions ?

Source/Plugins/Core/com.equella.core/resources/web/scripts/courselist.js
edalex-ian: Okay, but... Does that mean this is effectively fixing a bug in the Legacy mode too? Or more so, that bug was just not exhibited due to a server side re-render?

Source/Plugins/Core/com.equella.core/src/com/tle/web/itemlist/standard/AbstractItemlikeListAttachmentDisplaySection.java
edalex-ian: Ah! How it is triggered? Just but building up the correct events etc for the POST?

Source/Plugins/Core/com.equella.core/resources/web/scripts/courselist.js
PenghaiZhang: Yes, the server side re-render hides the bug for Legacy mode.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SelectionSessionModule.ts
PenghaiZhang: I think we should use `unknown`  as this type is for the parameter passed to the callback, which is `unknown` in this case.

```
 updateCourseList: (data: unknown) => void;
```

Source/Plugins/Core/com.equella.core/src/com/tle/web/itemlist/standard/AbstractItemlikeListAttachmentDisplaySection.java
PenghaiZhang: Yes, as long as we tell the server which section we need, which is `ilad` here. 
`ilad` is probably the short name of `AbstractItemlikeListAttachmentDisplaySection`

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/modules/LegacySelectionSessionModule.test.ts
```diff
@@ -0,0 +1,192 @@
...
+      updateMockGetRenderData(basicRenderData);
+      const data: SelectionSessionPostData = buildPostDataForStructured(
+        itemKey,
+        attachmentUUIDs
+      );
+      expect(data).toMatchObject(expectedPostData);
+    }
+  );
+
+  describe("buildPostDataForSelectOrAdd", () => {
```
edalex-ian, line 142: That's odd. Are you aware this is sitting in the `buildPostDataForStructured` describe block?

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/components/SearchResult.test.tsx
```diff
@@ -172,19 +191,97 @@ describe("<SearchResult/>", () => {
...
+    const makeSelection = (
+      selectorLabel: string,
+      queryByLabelText: (text: string) => HTMLElement | null,
+      getByLabelText: (text: string) => HTMLElement
+    ) => {
+      // First, make sure the selector control is active
+      expect(queryByLabelText(selectorLabel)).toBeInTheDocument();
+      // And then make a selection by clicking it
+      fireEvent.click(getByLabelText(selectorLabel));
+    };
```
edalex-ian, line 227: This could be simplified with:

```suggestion
    const makeSelection = (
      findSelector: (text: string) => HTMLElement | null,
    ) => {
      const selectorControl = findSelector();
      // First, make sure the selector control is active
      expect(selectorControl).toBeInTheDocument();
      // And then make a selection by clicking it
      fireEvent.click(selectorControl);
    };
```

And that way when you use it you'd only have to:

```typescript
makeSelection(() => queryByLabelText(resourceType));
```

edalex-ian (approved):
Good stuff dude. This is good to go, just two bits for the tests first would be nice. :grin: 

Feel free to merge when you're ready.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/components/SearchResult.test.tsx
PenghaiZhang: But I guess `fireEvent.click` will complain because `selectorControl` could be null.

PenghaiZhang: Yeah, here is the error: 

Argument of type 'HTMLElement | null' is not assignable to parameter of type 'Element | Node | Window | Document'.
      Type 'null' is not assignable to type 'Element | Node | Window | Document'.

What about we use a if statement and throw an error if `queryByLabelText` returns null ? We have used this approach for `querySelector` in a few places.

edalex-ian: Yup, I'd say that'd be the plan. But just wrap the fireEvent in the `if`.

Actually, this is probably a valid case of using the not null (`!`) postfix operator - as we definitely know (due to validation with `expect`) that this is non-null. :thinking:

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/modules/LegacySelectionSessionModule.test.ts
PenghaiZhang: oops I will move his describe out.

edalex-ian (approved):
Merge on. :+1:

PR #2551, opened 2020-11-25
Fix MimeTypeFilterDialog Story

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

This PR fixes the Storybook of component `MimeTypeFilterDialog`.

By doing this, I also refactor the props of the component to allow injecting a function which retrieves Mime Types.

This approach provides more flexibility and has been used in some components so I think we should use this approach for 
`MimeTypeFilterDialog`.

image

image

image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/searchfilter/SearchFilterSettingsPage.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/searchfilter/SearchFilterSettingsPage.tsx
@@ -37,6 +37,7 @@ import {
 import EditIcon from "@material-ui/icons/Edit";
 import DeleteIcon from "@material-ui/icons/Delete";
 import AddCircleIcon from "@material-ui/icons/AddCircle";
+import { getMIMETypesFromServer } from "../../../modules/MimeTypesModule";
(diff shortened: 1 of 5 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/searchfilter/SearchFilterSettingsPage.tsx
```diff
@@ -372,6 +373,7 @@ const SearchFilterPage = ({ updateTemplate }: TemplateUpdateProps) => {
         addOrUpdate={addOrUpdateMimeTypeFilter}
         mimeTypeFilter={selectedMimeTypeFilter}
         handleError={handleError}
+        mimeTypeSupplier={getMIMETypesFromServer}
```
edalex-ian, line 376: You'll note in the other injections of supplier functions, that we've set a default supplier on the component when deconstructing the props. That way you don't need to specify this here, only when you're doing tests/storybooks.

edalex-ian (approved):
Like your work @PenghaiZhang - including the injected supplier.

One thing you may want to consider, is using a default supplier (as per the comment). This would then maintain consistency.

But either-way, good to merge.

PR #2560, opened 2020-11-27
Don't show select summary button if it is disabled

PenghaiZhang:
#688 

##### Checklist

- [x] tests are included

##### Description of change

This PR is about whether show the `Select SummaryPage button` or not.

The approach include a few steps:

1. Due to the fact that the visibility of this button is controlled by either `oEQ setting` or `LMS oEQ plugin setting`, read and merge the two settings into one boolean flag.
2. Include the flag in `renderData.selectionSessionInfo`.
3. Front-end shows/hides the button based on the flag.

I think what is most important is the logic of how the two settings are merged.

I only uploaded two screenshots for when the LMS setting is `No restrictions`.

Screenshot from 2020-11-27 11-57-04
Screenshot from 2020-11-27 11-57-18


edalex-ian (approved):
LGTM. Merge on. :+1:

Issue #2530, opened 2020-11-18
PDF text extraction for indexing handled unsafely


SammyIsConfused:
**Describe the bug**
When a PDF file is contributed it is then processed. This includes text extraction, to be then indexed for freetext search. This extraction is done by Tika. 
If the PDF is "broken" for any reason (garbage in the metadata, broken XREF table, any number of other common PDF abnormalities) then the text extraction can stall. This can take a long time, or even hang indefinitely. When this happens it hammers the server CPU causing performance and stability issues. 

Even worse, when restarting the server the problem will happen again - as the BackgroundIndexer picks up the unfinished item and reattempts.

**To Reproduce**

1. Start a contribution to a collection with an attachment control. 
2. Upload a broken PDF as the attachment.
3. Publish the item.
4. Wait for the indexing process to start. 
5. Server will repeatedly send "Thread {threadName} indexing {filename} and has been for {time elapsed} seconds." to the resource centre logs. This time elapsed keeps increasing and increasing and server load becomes higher and higher. Note that server becomes slow.
7. If waiting long enough, the node may even die.

**Expected behavior**
The indexing process should error out safely and not attempt again, should a problem like this occur. There should be some way of checking a PDF for problems before indexing it - or at least having a configurable timeout to allow it to give up the attempt safely and automatically close the process.

Related to  #2527 - found at the same time with the same PDF. 

**Note:** A workaround is to go into MIME settings in the openEQUELLA, find application/pdf and disable text extraction for the server. Of course, this means that valid PDFs would not get indexed.

**Platform:**

- OpenEquella Version: 6.6 onwards

PR #2561, opened 2020-11-27
Make pdf extraction safe

SammyIsConfused:
##### Checklist

##### Description of change
#2530 
Text extraction for indexing on "broken" PDF files can cause performance and stability issues, slamming server CPU. This PR is to address this by adding a new column to the ATTACHMENT table. After a timeout (controlled by the existing property `textExtracter.parseDurationCap` in milliseconds in the `plugins/com.tle.core.freetext/optional.properties` file) PDF extraction will error out and a flag will be set to true on the attachment, errored_when_indexing. When the item is re-indexed for any reason, that attachment will be skipped for extraction. 

This will serve to prevent perpetual server load hammering by the TextExtracter and also highlight problem attachments in the database. An admin would be able to go into the ATTACHMENT table and check for any with `errored_while_indexing`. If there are any, those attachments can be followed up and checked for issues. 

This PR includes a migration to add the new column, `com.tle.core.institution.migration.v20202.addIndexingErrorColumnMigration` and set it to false by default. It has also been incorporated into Attachment serialization.


SammyIsConfused:
Missing license header. Fixing up now

edalex-ian:
Noticed that you re-requested reviews but then still had a few commits to do. So let us know when it's ready again.

SammyIsConfused:
Gah! MimeTypesTest flaked again. 
Looked into it and found a fix, I've put it as a separate PR so that this branch doesn't hold up others. 
I think this test failure can be safely ignored as the fix already exists. 
@edalex-ian This branch is now ready for review.

Source/Plugins/Core/com.equella.base/src/com/tle/beans/item/attachments/IAttachment.java
```diff
@@ -67,4 +67,8 @@
   boolean isRestricted();
 
   IItem<?> getItem();
+
+  boolean isErroredIndexing();
```
edalex-ian, line 71: Although not done historically, for the new stuff we add JavaDoc to help our future selves. Please add.

Source/Plugins/Core/com.equella.base/src/com/tle/web/api/item/interfaces/beans/AttachmentBean.java
```diff
@@ -65,4 +66,12 @@ public boolean isPreview() {
   public void setPreview(boolean preview) {
     this.preview = preview;
   }
+
+  public boolean isErroredIndexing() {
```
edalex-ian, line 70: Please JavaDoc - even if it's a `@see` to `IAttachment` matching method.

Source/Plugins/Core/com.equella.core/resources/lang/i18n-resource-centre.properties
```diff
@@ -80,6 +80,7 @@
 /com.tle.core.entity.services.migration.v20192.unknownuser.alluser=Create a new table for all users
 /com.tle.core.entity.services.migration.v20202.facetedsearch.classification=Create a new table for faceted search classification
 /com.tle.core.entity.services.migration.v20202.removelastknownuserconstraint=Remove last known user composite constraint (username and institution ID)
+/com.tle.core.entity.services.migration.v20202.indexing.errored=Add column to attachment table to allow indexer to skip individual attachments that have timed out.
```
edalex-ian, line 83: Be worthwhile to keep wording around it generic. That way if in the future we have other failure modes, we're all set.

```suggestion
/com.tle.core.entity.services.migration.v20202.indexing.errored=Add column to attachment table to allow indexer to skip individual attachments that have failed.
```

Source/Plugins/Core/com.equella.core/src/com/tle/core/freetext/extracter/standard/PdfExtracter.java
```diff
@@ -50,25 +56,34 @@ public boolean isSupportedByDefault(MimeEntry mimeEntry) {
...
-      throws IOException {
+      throws IOException, InterruptedException, ExecutionException, TimeoutException {
     WriteOutContentHandler wrapped = new WriteOutContentHandler(maxSize);
     ContentHandler handler = new CappedBodyContentHandler(wrapped, parseDuration);
-    try {
-      Metadata meta = new Metadata();
-      Parser parser = new AutoDetectParser(new TikaConfig(getClass().getClassLoader()));
-      parser.parse(input, handler, meta, new ParseContext());
+    Runnable runnableParse =
+        new Runnable() {
```
edalex-ian, line 63: Being in Java 8 land, best to now go with an anonymous lambda rather than an anonymous class and all its overhead.

Source/Plugins/Core/com.equella.core/src/com/tle/core/freetext/extracter/standard/PdfExtracter.java
```diff
@@ -50,25 +56,34 @@ public boolean isSupportedByDefault(MimeEntry mimeEntry) {
...
-      if (wrapped.isWriteLimitReached(t)) {
-        // keep going
-        LOGGER.debug("PDF size limit reached.  Indexing truncated text");
-        appendText(handler, outputText, maxSize);
-        return;
-      }
-      throw Throwables.propagate(t);
-    }
+              appendText(handler, outputText, maxSize);
+            } catch (Exception t) {
```
edalex-ian, line 72: It could be good to add a `catch` for `InterruptException` to capture when it's terminated early (or by other unforeseen interrupts). :thinking:

Source/Plugins/Core/com.equella.core/src/com/tle/core/freetext/extracter/standard/PdfExtracter.java
```diff
@@ -50,25 +56,34 @@ public boolean isSupportedByDefault(MimeEntry mimeEntry) {
...
+                return;
+              }
+              throw Throwables.propagate(t);
+            }
+          }
+        };
+    ExecutorService executor = Executors.newSingleThreadExecutor();
+    Future<?> future = executor.submit(runnableParse);
+    future.get(parseDuration, TimeUnit.MILLISECONDS);
+    executor.shutdownNow();
```
edalex-ian, line 86: Bringing in an `ExecutorService` etc. seems rather heavy handed for a single thread. How about just going with:

```java
Thread extractThread = new Thread(() -> { ... }, "ExtractThread");
extractThread.join(parseDuration);
extractThread.interrupt();
```

I'm assuming that `interrupt()` will be sufficient, as that's they typical recommendation for `ExecutorService.shutdownNow()`. But maybe you had issues?

edalex-ian (changes requested):
Good stuff @SammyIsConfused . Just needs a bit of JavaDoc and some Java 8 treatment. Other than that, maybe the new Thread can be simplified.

Also, being that it has a data migration, I'd be really keen to have @PenghaiZhang review before merge.

Source/Plugins/Core/com.equella.core/src/com/tle/core/freetext/extracter/standard/PdfExtracter.java
SammyIsConfused: I tried your way and it doesn't seem to work. The problem with interrupt is that it seems to simply set a flag that needs to be handled in the thread with isInterrupted(), which would be fine if we had a loop but we don't, it's parser.parse that is the cause of the issue.

edalex-ian: Cool, evidentially the implementation is doing more of a hard stop - which is typically advised against due to stray objects etc. But that is more so where there are multiple threads of the same code.

So in this case, you have cause for the increased complexity. Thanks for checking. :+1:

Source/Plugins/Core/com.equella.core/src/com/tle/core/freetext/extracter/standard/PdfExtracter.java
SammyIsConfused: The InterruptedException isn't actually thrown here - it's thrown by future.get. So I surrounded that with try/catch and logged the InterruptedException/TimeoutException/ExecutionException that can throw. Then I rethrew so that the indexerThread can catch it further up where it sets the erroredIndexing flag.

SammyIsConfused:
AdvancedScriptControlTest flaked. Rerunning.

Source/Plugins/Core/com.equella.base/src/com/tle/beans/item/attachments/Attachment.java
```diff
@@ -86,6 +86,9 @@
 
   private boolean restricted;
 
+  @Column(name = "errored_when_indexing")
+  private boolean erroredIndexing;
```
ChristianMurphy, line 90: :thinking: would this make sense as a state/enumeration?
something like
```
             state machine
+-----------------------------------------+
|                                         |
|                        +--------------+ |
|                        |              | |
|         +-----------+->+ indexed      | |
| added   |           |  |              | |
+-------->+ unindexed |  +--------------+ |
|         |           |  |              | |
|         +-----------+->+ index failed | |
|                        |              | |
|                        +--------------+ |
|                                         |
+-----------------------------------------+

```

new attachments start as unindexed, they can move to indexed, to index failed when processed.
A state system would also leave to door open for partially indexed content, or other possible states.

Though it may be overkill here :thinking: 
Thoughts? :thought_balloon:

Source/Plugins/Core/com.equella.base/src/com/tle/beans/item/attachments/Attachment.java
```diff
@@ -281,4 +284,12 @@ public void setRestricted(boolean restricted) {
   public String getAttachmentSignature() {
     return item.getIdString() + "/" + this.uuid + " [" + this.description + "]";
   }
+
+  public boolean isErroredIndexing() {
+    return erroredIndexing;
+  }
+
+  public void setErroredIndexing(boolean skipIndexing) {
```
ChristianMurphy, line 292: would it make sense for the param to also be `errorIndexing`?

Source/Plugins/Core/com.equella.base/src/com/tle/beans/item/attachments/Attachment.java
```diff
@@ -281,4 +284,12 @@ public void setRestricted(boolean restricted) {
   public String getAttachmentSignature() {
     return item.getIdString() + "/" + this.uuid + " [" + this.description + "]";
   }
+
+  public boolean isErroredIndexing() {
```
ChristianMurphy, line 288: Maybe `hasErroredIndexing` would read more naturally here? :thinking:

ChristianMurphy (commented):
Error recovery/management is a nice add, thanks @SammyIsConfused!

Source/Plugins/Core/com.equella.base/src/com/tle/beans/item/attachments/Attachment.java
edalex-ian: This is what we were talking about at the start @SammyIsConfused . But I believe you highlighted that unfortunately the tracking of this is split across levels of the entity hierarchy. I think the primary index tracking is done at the item level, yet what was being addressed and flagged here was the actions occurring at the item attachment level.

So now we can have an item which is 'indexed' (after previously being 'unindexed'), but potentially some attachments are flagged as having errored - avoiding excessive future indexing. Is that right?

Source/Plugins/Core/com.equella.base/src/com/tle/beans/item/attachments/Attachment.java
edalex-ian: Although I agree that it would read more natural. The Java standard for booleans on Beans (this being an `Entity` bean) is to  use the `is` prefix for boolean getters.

edalex-ian (approved):
Looking good @SammyIsConfused . Feel free to proceed when others happy.

I see there's a method param you missed renaming of which @ChristianMurphy , and I've also requested @PenghaiZhang to come and review - especially seeing you've got a migration in here.

Source/Plugins/Core/com.equella.core/src/com/tle/core/institution/migration/v20202/AddIndexingErrorColumnMigration.java
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  @Entity(name = "Attachment")
+  @AccessType("field")
+  public static class FakeAttachment {
+    @Id
+    @GeneratedValue(strategy = GenerationType.AUTO)
+    long id;
+
+    @Column boolean errored_when_indexing;
```
PenghaiZhang, line 81: If you do `nullable = false` in `Attachment.java`, you also need to do that here.

Source/Plugins/Core/com.equella.core/src/com/tle/core/institution/migration/v20202/AddIndexingErrorColumnMigration.java
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  @Override
+  protected int countDataMigrations(HibernateMigrationHelper helper, Session session) {
+    return 1;
+  }
+
+  @Override
+  protected List<String> getDropModifySql(HibernateMigrationHelper helper) {
+    return helper.getAddNotNullSQL("attachment", "errored_when_indexing");
```
PenghaiZhang, line 56: Hi Sam, I have two questions here. 
1. Do we need to drop column 'errored_when_indexing'  in any situation?  If yes, I think we should use
'getDropModifySql'.

2. I am not quite sure why use `getAddNotNullSQL`. If this column needs a not null constraint, you can try
`@Column(name = "errored_when_indexing", nullable = false)` in  `attachment.java`.

SammyIsConfused: Hey Penghai - if we wanted to make it backwards compatible, that would require removing the column. So I should fix that up.
The reason for using getAddNotNullSQL was simply that I was using AddInstitutionFileWarningColumn as a template - if there's a better way I'm happy to use that.

Source/Plugins/Core/com.equella.base/src/com/tle/beans/item/attachments/Attachment.java
SammyIsConfused: Well summed up Ian, yes that's correct. The index tracking is at the item level and doesn't keep info about specific attachments. This change means that an entire item can be indexed even if it has an attachment flagged as errored - that attachment will simply be skipped in future indexes to prevent performance problems. 

This also gives the ability to look at the database and get a list of all attachments that need fixing up to be indexed properly. `SELECT * FROM attachment WHERE errored_when_indexing=true` will point a DBA at exactly where the problems are. It would not have been possible to glean this info from the `item` table alone.

Source/Plugins/Core/com.equella.base/src/com/tle/beans/item/attachments/Attachment.java
SammyIsConfused: Well spotted, will fix that up. That was the old name of this field.

Source/Plugins/Core/com.equella.core/src/com/tle/core/institution/migration/v20202/AddIndexingErrorColumnMigration.java
SammyIsConfused: There's no need for the getDropModifySql, I was following another as a template.
Adding nullable=false to Attachment.java now.

ChristianMurphy (approved):

PenghaiZhang (approved):
Looks good to me!

Issue #2527, opened 2020-11-18
PDF thumbnailing handled unsafely


SammyIsConfused:
**Describe the bug**
When a PDF file is contributed it is sent to ImageMagick to generate a thumbnail. If the PDF is "broken" for any reason (garbage in the metadata, broken XREF table, any number of other common PDF abnormalities) then ImageMagick's PDF thumbnailing delegate GhostScript will try to "fix" the PDF. This can take a long time, or even hang indefinitely. When this happens it hammers the server CPU causing performance and stability issues.

**To Reproduce**

1. Start a contribution to a collection with an attachment control. 
2. Upload a broken PDF as the attachment.
3. Publish the item.
4. Wait for the thumbnailing process to start. 
5. Open Task Manager/System Monitor on the equella server. 
6. Find the `identify` and `gs` processes. Note that they are consuming a lot of CPU and do not close.
7. If waiting long enough, the node may even die and restart.

**Expected behavior**
The thumbnailing process should error out safely and not attempt again, should a problem like this occur. There should be some way of checking a PDF for problems before thumbnailing it - or at least having a configurable timeout to allow it to give up the attempt safely and automatically close the process.

**Platform:**

- OpenEquella Version: 6.6 onwards

PR #2566, opened 2020-11-29
Make sure that thumbnailing timeout kills both imagick and ghostscript

SammyIsConfused:
##### Checklist

##### Description of change
Related to #2527, the delegate `ghostscript` doesn't inherently close when imagemagick does. This PR ensures that when thumbnailing times out, any child processes of imagemagick are killed first. 

### DIFF
--- a/Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
+++ b/Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
@@ -23,7 +23,6 @@ import com.google.common.io.Closeables;
 import com.tle.common.Triple;
 import java.io.File;
 import java.io.IOException;
-import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Reader;
 import java.lang.reflect.Field;
@@ -56,67 +55,69 @@ public final class ExecUtils {
   }; //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
 
   /**
-   * For a given Process ID, kill any child processes and then kill the process.
-   * Works on Unix only, as it leverages pgrep and kill commands.
+   * For a given Process ID, kill any child processes and then kill the process. Works on Unix only,
+   * as it leverages pgrep and kill commands.
+   *
    * @param pid The Process for which to terminate including it's direct children.
    */
-    public static void killLinuxProcessTree(int pid) {
-      try {
-        if (pid > 0) {
-          // get child process PIDs as strings
-          String[] children = getChildUnixProcessPids(pid);
-          // Kill child process(es)
-            for (String child : children) {
-              sendSigKill(Integer.parseInt(child));
-            }
-          // kill process itself
-          sendSigKill(pid);
+  public static void killLinuxProcessTree(int pid) {
+    try {
+      if (pid > 0) {
+        // get child process PIDs as strings
+        String[] children = getChildUnixProcessPids(pid);
+        // Kill child process(es)
+        for (String child : children) {
+          sendSigKill(Integer.parseInt(child));
         }
-      } catch (Exception e) {
-        LOGGER.warn("Process Kill failed. Process may be left hanging",e);
+        // kill process itself
+        sendSigKill(pid);
       }
+    } catch (Exception e) {
+      LOGGER.warn("Process Kill failed. Process may be left hanging", e);
     }
+  }
 
   /**
-   * Runs pgrep -P for a given Process ID,
-   * to get a list of child processes as Process IDs.
+   * Runs pgrep -P for a given Process ID, to get a list of child processes as Process IDs.
+   *
    * @param pid The parent process ID to check for child processes.
    * @return String[] An array of process IDs for the children of pid. Can be empty.
    */
-    public static String[] getChildUnixProcessPids(int pid){
-      try {
-        Process getChildPid = Runtime.getRuntime().exec("pgrep -P " + pid);
-        getChildPid.waitFor();
-        StringBuilder childPid = new StringBuilder();
-        if (getChildPid.exitValue() == 0) {
-          CharStreams.copy(new InputStreamReader(getChildPid.getInputStream()), childPid);
-        }
-        getChildPid.destroy();
-        return childPid.toString().replaceAll("\n", "").split(" ");
-      }catch(Exception e){
-        LOGGER.error("Error getting child processes for: " + pid, e);
+  public static String[] getChildUnixProcessPids(int pid) {
+    try {
+      Process getChildPid = Runtime.getRuntime().exec("pgrep -P " + pid);
+      getChildPid.waitFor();
+      StringBuilder childPid = new StringBuilder();
+      if (getChildPid.exitValue() == 0) {
+        CharStreams.copy(new InputStreamReader(getChildPid.getInputStream()), childPid);
       }
-      return new String[]{};
+      getChildPid.destroy();
+      return childPid.toString().replaceAll("\n", "").split(" ");
+    } catch (Exception e) {
+      LOGGER.error("Error getting child processes for: " + pid, e);
     }
+    return new String[] {};
+  }
 
   /**
    * Creates a process which then sends a SIGKILL signal to a given process.
+   *
    * @param pid the Process ID of the process to kill.
    * @return the exitValue of the SIGKILL process (not the process being killed)
    */
-    public static int sendSigKill(int pid){
-      try {
-        Process sigKill = Runtime.getRuntime().exec("kill -9 " + pid);
-        sigKill.waitFor();
-        int returnValue = sigKill.exitValue();
-        sigKill.destroy();
-        return returnValue;
-      }catch(Exception e){
-        LOGGER.error("killing process " + pid + " failed.",e);
-      }
-      //shouldn't get here
-      return -1;
+  public static int sendSigKill(int pid) {
+    try {
+      Process sigKill = Runtime.getRuntime().exec("kill -9 " + pid);
+      sigKill.waitFor();
+      int returnValue = sigKill.exitValue();
+      sigKill.destroy();
+      return returnValue;
+    } catch (Exception e) {
+      LOGGER.error("killing process " + pid + " failed.", e);
     }
+    // shouldn't get here
+    return -1;
+  }
   /**
    * Gets the PID of a given process.
    *
@@ -224,10 +225,11 @@ public final class ExecUtils {
       proc.waitFor(durationInSeconds, TimeUnit.SECONDS);
       if (!stdErr.isFinished() || !stdOut.isFinished()) {
         String platform = determinePlatform();
-        if(platform.equals(PLATFORM_LINUX) || platform.equals(PLATFORM_LINUX64)) {
+        if (platform.equals(PLATFORM_LINUX) || platform.equals(PLATFORM_LINUX64)) {
           killLinuxProcessTree(pid);
-        }else{
-          LOGGER.debug("Platform not yet supported for process tree kill. Processes may be left hanging");
+        } else {
+          LOGGER.debug(
+              "Platform not yet supported for process tree kill. Processes may be left hanging");
         }
         throw new InterruptedException();
       }
--- a/Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
+++ b/Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
@@ -86,6 +86,10 @@ public final class ExecUtils {
       StringBuilder childPid = new StringBuilder();
       if (getChildPid.exitValue() == 0) {
         CharStreams.copy(new InputStreamReader(getChildPid.getInputStream()), childPid);
+      } else {
+        StringBuilder errorOutput = new StringBuilder();
+        CharStreams.copy(new InputStreamReader(getChildPid.getErrorStream()), errorOutput);
+        LOGGER.debug("getChildPid function did not run properly.\n" + errorOutput);
       }
       getChildPid.destroy();
       return childPid.toString().replaceAll("\n", "").split(" ");
@@ -106,6 +110,15 @@ public final class ExecUtils {
       Process sigKill = Runtime.getRuntime().exec("kill -9 " + pid);
       sigKill.waitFor();
       int returnValue = sigKill.exitValue();
+      if (sigKill.exitValue() == 0) {
+        StringBuilder successOutput = new StringBuilder();
+        CharStreams.copy(new InputStreamReader(sigKill.getInputStream()), successOutput);
+        LOGGER.debug("Output of kill function: " + successOutput);
+      } else {
+        StringBuilder errorOutput = new StringBuilder();
+        CharStreams.copy(new InputStreamReader(sigKill.getErrorStream()), errorOutput);
+        LOGGER.debug("kill function did not run properly.\n" + errorOutput);
+      }
       sigKill.destroy();
       return returnValue;
     } catch (IOException | InterruptedException e) {
--- a/Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
+++ b/Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
@@ -79,10 +79,11 @@ public final class ExecUtils {
    */
   public static Optional<int[]> getChildUnixProcessPids(int pid) {
     Optional<int[]> pids = Optional.empty();
+    StringBuilder childPid = new StringBuilder();
     try {
       Process getChildPid = Runtime.getRuntime().exec("pgrep -P " + pid);
       getChildPid.waitFor();
-      StringBuilder childPid = new StringBuilder();
+
       if (getChildPid.exitValue() == 0) {
         CharStreams.copy(new InputStreamReader(getChildPid.getInputStream()), childPid);
       } else {
--- a/Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
+++ b/Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
@@ -23,14 +23,17 @@ import com.google.common.io.Closeables;
 import com.tle.common.Triple;
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Reader;
+import java.lang.reflect.Field;
 import java.net.URL;
 import java.net.URLDecoder;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.StringTokenizer;
 import java.util.concurrent.TimeUnit;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
```diff
@@ -52,6 +55,71 @@
     "", ".exe", ".bat"
   }; //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
 
+  /**
+   * Creates a process which will then kill a given process and it's child processes on Linux.
+   * arg[0] should be the PID of the main process you want to terminate.
+   */
+  public static class LinuxKill {
+    public static void main(String[] args) {
```
edalex-ian, line 63: I gather this must be following an example, as there is no need for this setup here. It can just be a stand-alone static method like the others in this class. Probably with a signature like: `public static void killLinuxProcessTree(int pid)`

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
```diff
@@ -130,10 +198,12 @@ public static ExecResult execWithTimeLimit(
           createProcess(cmdarray, additionalEnv, dir);
       LOGGER.debug("Started timed process");
       final Process proc = cp.getFirst();
+      String pid = Long.toString(getPidOfProcess(proc).orElse(0L));
       final StreamReader stdOut = cp.getSecond();
       final StreamReader stdErr = cp.getThird();
       proc.waitFor(durationInSeconds, TimeUnit.SECONDS);
       if (!stdErr.isFinished() || !stdOut.isFinished()) {
+        LinuxKill.main(new String[] {pid});
```
edalex-ian, line 206: Should you check first you're running on linux?

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
```diff
@@ -52,6 +55,71 @@
...
+          // get child process PIDs as string
+          Process getChildPid = Runtime.getRuntime().exec("pgrep -P " + args[0]);
+          getChildPid.waitFor();
+          StringBuilder childPid = new StringBuilder();
+          if (getChildPid.exitValue() == 0) {
+            InputStream procIn = getChildPid.getInputStream();
+            int c = 0;
+            while ((c = procIn.read()) != -1) {
+              childPid.append((char) c);
+            }
```
edalex-ian, line 75: Couple of things here.

1. you may wish to consider using `CharStreams.copy(new InputStremReader(getChildPid.getInputStream(), childPid);`; and
2. What if you get multiple PIDs back? Are they find coming back with new lines and getting fed to the following `exec`?

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
```diff
@@ -52,6 +55,71 @@
...
+          }
+
+          // Kill child process(es)
+          Process childKillProc = Runtime.getRuntime().exec("kill -9 " + childPid);
+          childKillProc.waitFor();
+          if (childKillProc.exitValue() == 0) {
+            LOGGER.debug("Child processes terminated: " + childPid);
+          }
```
edalex-ian, line 83: So if it returns non-zero, it won't throw an exception (I gather). So the catch block wont be triggered and nothing will be logged (or noted).

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
```diff
@@ -52,6 +55,71 @@
...
+
+          // kill process itself
+          Process proc = Runtime.getRuntime().exec("kill -9 " + args[0]);
+          int exitVal = proc.waitFor();
+          LOGGER.debug("Exit value: " + exitVal); // often 1 under Ubuntu 12.10
```
edalex-ian, line 88: Your comment is confusing me. Above you check to see that `kill -9` returns zero. Then here you're saying it could return 1. Which is it?

Also, be good here to do the same as the above and check the result.

And arguably this code is the same as the above and should be in it's own method, perhaps: `static void int sendSigKill(int pid)`. Could encapsulate waiting and logging, but then ultimately still return the exit code.

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
```diff
@@ -52,6 +55,71 @@
...
+
+          proc.destroy();
+          childKillProc.destroy();
+          getChildPid.destroy();
+        }
+      } catch (Exception e) {
+        LOGGER.warn("Process Kill failed. Process may be left hanging");
```
edalex-ian, line 95: Be good to know why it failed. (Also, slight change in capitalisation.)

```suggestion
        LOGGER.warn("Process kill failed. Process may be left hanging. " + e.getMessage());
```

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
```diff
@@ -52,6 +55,71 @@
...
+
+    try {
+      if (p.getClass().getName().equals("java.lang.UNIXProcess")) {
+        Field f = p.getClass().getDeclaredField("pid");
+        f.setAccessible(true);
+        pid = Optional.of(f.getLong(p));
+        f.setAccessible(false);
+      }
+    } catch (Exception e) {
+      return pid;
```
edalex-ian, line 118: What exceptions get thrown here? Could be good to be more targetted, and to log what happened.

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
```diff
@@ -52,6 +55,71 @@
...
+    }
+    return pid;
```
edalex-ian, line 120: Seems awkward to have the two `return` statements right next to each other. Would a `finally` block perhaps be a better fit?

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
```diff
@@ -52,6 +55,71 @@
...
+          getChildPid.destroy();
+        }
+      } catch (Exception e) {
+        LOGGER.warn("Process Kill failed. Process may be left hanging");
+      }
+    }
+  }
+
+  /**
+   * Gets the PID of a given process.
```
edalex-ian, line 101:
```suggestion
   * Gets the process ID (PID) of a given *nix process.
```

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
```diff
@@ -52,6 +55,71 @@
...
+   * @param p The Process of which to get the PID.
+   * @return An Optional long. If not on Linux, or if the PID declared field is not available, the
+   *     value will be empty.
+   */
+  public static synchronized Optional<Long> getPidOfProcess(Process p) {
+    Optional<Long> pid = Optional.empty();
+
+    try {
+      if (p.getClass().getName().equals("java.lang.UNIXProcess")) {
+        Field f = p.getClass().getDeclaredField("pid");
```
edalex-ian, line 112: What's the difference with say Windows? As those processes also have IDs, is it simply a case of a different name for the field?

edalex-ian (changes requested):
Heya @SammyIsConfused 

Good work sticking through the difficulties here. A few things to tidy up and consider.

Obviously we're first focusing on Linux, but still not a bad idea to ponder the Windows situation too for any quick wins - like the process ID.

BTY.

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
SammyIsConfused: No unfortunately, not with Java 8. With 9, you can just `process.pid();` I've been looking into it, the ID is not a field in a Windows Process. Other solutions include: 

- Using JNA (com.sun.jna) https://stackoverflow.com/a/43426878 but that requires us to install new dependencies
- Calling `tasklist` (windows command to get all running process IDs) before starting the process and again after, and using the diff of the two to find the new PIDs (I don't like this idea, if another process just happened to start between the two checks that could be a false positive)

Once we have a Windows PID however, killing it and it's child processes is not too difficult with `taskkill`. But targeting the right process is where it gets tricky. 

Another complication is that I have no dev environment for Windows (testing installers yes, but debugging a dev environment no).  So making it work for Windows would probably take too long (at least for 2020.2).

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
SammyIsConfused: I tried that, but apparently the inspector has a problem with having a return within a finally block. 

>'return' inside 'finally' block 
 Inspection info: Reports return statements inside of finally blocks. While occasionally intended, such return statements may mask exceptions thrown, and tremendously complicate debugging.

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
SammyIsConfused: So that comment was left over from an [example](https://askubuntu.com/questions/222604/killing-processes-from-java) that I was editing to suit me. In my testing, it returns 0. But apparently it can return 1 according to the SO link, so I'm unsure. 
Either way I have refactored, and removed the comment.

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
edalex-ian: I see it's now refactored to simply use the default return with `pid`. Much better.

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
```diff
@@ -52,6 +54,104 @@
...
+      StringBuilder childPid = new StringBuilder();
+      if (getChildPid.exitValue() == 0) {
+        CharStreams.copy(new InputStreamReader(getChildPid.getInputStream()), childPid);
+      } else {
+        StringBuilder errorOutput = new StringBuilder();
+        CharStreams.copy(new InputStreamReader(getChildPid.getErrorStream()), errorOutput);
+        LOGGER.debug("getChildPid function did not run properly.\n" + errorOutput);
+      }
+      getChildPid.destroy();
+      return childPid.toString().replaceAll("\n", "").split(" ");
```
edalex-ian, line 95: IShould that be:

```suggestion
      return childPid.toString().replaceAll("\n", " ").split(" ");
```

SammyIsConfused: Ah yup, you're right. I had assumed that it prints them all on one line space-separated and then ends with \n. But testing it now I see that it has \n after each one.

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
```diff
@@ -52,6 +54,104 @@
...
   }; //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
 
+  /**
+   * For a given Process ID, kill any child processes and then kill the process. Works on Unix only,
+   * as it leverages pgrep and kill commands.
+   *
+   * @param pid The Process for which to terminate including it's direct children.
+   */
+  public static void killLinuxProcessTree(int pid) {
+    if (pid > 0) {
```
edalex-ian, line 64: Rather than silently doing nothing, how about invert this logic:

```java
if(pid < 1) {
  throw new IllegalArgumentException("Process ID should be greater than 0");
}
```

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
```diff
@@ -52,6 +54,104 @@
...
+      // get child process PIDs as strings
+      String[] children = getChildUnixProcessPids(pid);
+      // Kill child process(es)
+      for (String child : children) {
+        sendSigKill(Integer.parseInt(child));
+      }
```
edalex-ian, line 70: The way you're having to use this highlights a few issues with the below methods. Really, you'd expect the methods to support something like:

```java
getChildUnixProcessPids(pid)
  .ifPresent(pids ->
    sendSigKill(pids));
```

Comments which follow have this in mind.

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
```diff
@@ -52,6 +54,104 @@
...
+      // kill process itself
+      sendSigKill(pid);
+    }
+  }
+
+  /**
+   * Runs pgrep -P for a given Process ID, to get a list of child processes as Process IDs.
+   *
+   * @param pid The parent process ID to check for child processes.
+   * @return String[] An array of process IDs for the children of pid. Can be empty.
```
edalex-ian, line 80: If 'Can be empty' how about return an `Optional`? You'd then reap the benefits hoped for above, plus your method signature than flags this without the need for a comment.

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
```diff
@@ -52,6 +54,104 @@
...
+      StringBuilder childPid = new StringBuilder();
+      if (getChildPid.exitValue() == 0) {
+        CharStreams.copy(new InputStreamReader(getChildPid.getInputStream()), childPid);
+      } else {
+        StringBuilder errorOutput = new StringBuilder();
+        CharStreams.copy(new InputStreamReader(getChildPid.getErrorStream()), errorOutput);
+        LOGGER.debug("getChildPid function did not run properly.\n" + errorOutput);
+      }
+      getChildPid.destroy();
+      return childPid.toString().replaceAll("\n", " ").split(" ");
```
edalex-ian, line 95: It's expected that this method would return an array of numbers (PIDs) not strings. Keep in mind the above suggestion for an `Optional` it'd be good if you also streamed across the strings to transform them to Integers. Something like:

```java
Arrays.stream(childPid...split(" "))
  .stream()
  .mapToInt(s -> Integer.parseInt(s))
  .toArray();
```

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
```diff
@@ -52,6 +54,104 @@
...
+    return new String[] {};
+  }
+
+  /**
+   * Creates a process which then sends a SIGKILL signal to a given process.
+   *
+   * @param pid the Process ID of the process to kill.
+   * @return the exitValue of the SIGKILL process (not the process being killed)
+   */
+  public static int sendSigKill(int pid) {
```
edalex-ian, line 108: Seeing `kill` has a definition of:

```
kill [options] <pid> [...]
```

It'd make sense to allow this to support a list of PIDs too:

```suggestion
  public static int sendSigKill(int[] pid) {
```

That also then supports your use case of killing potentially multiple PIDs. If you did also want to support the form of taking a single PID (which I'm not sure you have a use for here), then you could add an overload that does the transform simply to an array and chain to this method.

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
```diff
@@ -52,6 +54,104 @@
...
+        CharStreams.copy(new InputStreamReader(sigKill.getErrorStream()), errorOutput);
+        LOGGER.debug("kill function did not run properly.\n" + errorOutput);
+      }
+      sigKill.destroy();
+      return returnValue;
+    } catch (IOException | InterruptedException e) {
+      LOGGER.error("killing process " + pid + " failed.", e);
+    }
+    // shouldn't get here
+    return -1;
```
edalex-ian, line 128: If you should get here, should you perhaps ensure that by adding a line to your above catch to throw or rethrow?

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
```diff
@@ -130,10 +230,18 @@ public static ExecResult execWithTimeLimit(
...
       final StreamReader stdOut = cp.getSecond();
       final StreamReader stdErr = cp.getThird();
       proc.waitFor(durationInSeconds, TimeUnit.SECONDS);
       if (!stdErr.isFinished() || !stdOut.isFinished()) {
+        String platform = determinePlatform();
+        if (platform.equals(PLATFORM_LINUX) || platform.equals(PLATFORM_LINUX64)) {
+          killLinuxProcessTree(pid);
+        } else {
+          LOGGER.debug(
+              "Platform not yet supported for process tree kill. Processes may be left hanging");
```
edalex-ian, line 243:
```suggestion
              "Platform (" + platform + ") does not supported process tree kill. Processes may be left hanging.");
```

edalex-ian (changes requested):
Great tidy up there @SammyIsConfused . Now that we can see the wood for the trees, unfortunately a few more things to sort.

But after that you should have a winner.

SammyIsConfused:
Should be good now @edalex-ian :+1:

SammyIsConfused:
Just snook in a merge of develop so that MimeTypesTest doesn't bite us, but it's good to go :+1:

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
```diff
@@ -52,6 +54,125 @@
...
+
+    // kill process itself, after all the child processes have been terminated
+    sendSigKill(pid);
+  }
+
+  /**
+   * Runs pgrep -P for a given Process ID, to get a list of child processes as Process IDs.
+   *
+   * @param pid The parent process ID to check for child processes.
+   * @return Optional<String[]> An array of process IDs for the children of pid.
```
edalex-ian, line 78:
```suggestion
   * @return Process IDs for the children of {@code pid}.
```

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
```diff
@@ -52,6 +54,125 @@
...
+      // convert string to array of ints
+      pids =
+          Optional.of(
+              Arrays.stream(childPid.toString().replaceAll("\n", " ").split(" "))
+                  .mapToInt(Integer::parseInt)
+                  .toArray());
+    } catch (IOException | InterruptedException e) {
+      LOGGER.error("Error getting child processes for: " + pid, e);
+    } catch (NumberFormatException e) {
+      LOGGER.error("Output of getChildPids command not parsable as integers", e);
```
edalex-ian, line 103: Being an `error` level log message, would be good if it was a more useful error message for an administrator.

```suggestion
      LOGGER.error("Unsupported output from 'pgrep -P'. Unable to terminate child processes.");
      LOGGER.error("'pgrep' output: " + childPid.toString());
      LOGGER.error("Parsing exception.", e);
```

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
```diff
@@ -52,6 +54,125 @@
...
+      sigKill.waitFor();
+      int returnValue = sigKill.exitValue();
+      if (sigKill.exitValue() == 0) {
+        StringBuilder successOutput = new StringBuilder();
+        CharStreams.copy(new InputStreamReader(sigKill.getInputStream()), successOutput);
+        LOGGER.debug("Output of kill function: " + successOutput);
+      } else {
+        StringBuilder errorOutput = new StringBuilder();
+        CharStreams.copy(new InputStreamReader(sigKill.getErrorStream()), errorOutput);
+        LOGGER.debug("kill function did not run properly.\n" + errorOutput);
```
edalex-ian, line 131: Be useful for administrators to know when this kind of thing is not functioning correctly.

```suggestion
        LOGGER.warn("Attempt to terminate processes failed (" + command + "):\n" + errorOutput);
```

edalex-ian (approved):
Good stuff @SammyIsConfused .

If you just want to tweak the comment and log messages, then merge on in. :+1:

SammyIsConfused:
That failure was caused by `childPid` not being accessible from line 104:
`      LOGGER.error("'pgrep' output: " + childPid.toString());`
Fix was to declare the StringBuilder childPid outside of the try/catch.

SammyIsConfused:
ViewerTest failed because Vimeo search returned nothing - probably caused by temporary issues on Vimeo's end. Rerunning.

PR #2546, opened 2020-11-24
EBP: Make buttons for image preview box visible

SammyIsConfused:
##### Checklist

- [x] the [contributor license agreement][] is signed
- [x] commit message follows [commit guidelines][]
- [x] screenshots are included showing significant UI changes

##### Description of change
For the Image Preview box, the close button and the left and right buttons were not visible. This is because they were referring to a fancybox link that is no longer accessible in EBP.  This PR replaces them with icons from Material Design, Close, ChevronLeft and ChevronRight. 
Before: 
image
After:
image

[contributor license agreement]: https://www.apereo.org/node/676
[commit guidelines]: https://chris.beams.io/posts/git-commit/

PR #2569, opened 2020-11-30
EBP Bugfix: Stop arrows from appearing on dialogs that don't use them

SammyIsConfused:
##### Checklist

##### Description of change
#2546 brought in a bug in which various fancybox dialogs that did not use arrows would start showing them. This is because in places where fancybox doesn't specifically set "display:none" inline it gets overridden by "display:flex" in legacy.scss. Thankfully, when the arrows are required "display:flex" is set inline to the tag by fancybox so we don't need to set it directly. 
Before: 
image
After:
image


edalex-ian (approved):
LGTM. :+1:

PR #2568, opened 2020-11-30
Make MimeTypesTest more robust

SammyIsConfused:
##### Checklist

- [x] the [contributor license agreement][] is signed
- [x] commit message follows [commit guidelines][]
- [x] tests are included

##### Description of change
MimeTypesTest has been giving us a lot of problems in GHA as of late. The problem was caused by a missing wait in getIcon() which was fixed in a previous PR. Turns out there were more missing waits after this happens, namely the one in 903461d before clicking restoreIcon `(id="mei_ri")`. I've also gone through MimeEditorPage.java adding missing waits before clicking and sending text to elements, for good measure (eb6be5e).
Before making those changes I was able to replicate the test failure locally by disabling caching and throttling the speed of my test browser and retrying a few times since it's intermittent. Testing with these changes I have yet to see it happen. Hopefully CI can verify that.    

[contributor license agreement]: https://www.apereo.org/node/676
[commit guidelines]: https://chris.beams.io/posts/git-commit/

PR #2571, opened 2020-11-30
Disable MimeTypesTest temporarily whilst it is worked on

SammyIsConfused:
##### Checklist

##### Description of change
Disabling this flakey test temporarily whilst it is worked on in #2568 so that it doesn't get in the way of other builds.

### DIFF
--- a/autotest/OldTests/src/test/java/com/tle/webtests/test/viewing/MimeTypesTest.java
+++ b/autotest/OldTests/src/test/java/com/tle/webtests/test/viewing/MimeTypesTest.java
@@ -87,7 +87,7 @@ public class MimeTypesTest extends AbstractCleanupTest {
     assertTrue(attachments.attachmentDetails(ATTACHMENT_NAME).contains("application/octet-stream"));
   }
 
-  @Test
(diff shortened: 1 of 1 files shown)


edalex-ian:
Considering it didn't used to be this flakey and it's passing on the more resourced Codebuild, would it perhaps be better to  annotate with retry?

SammyIsConfused:
> Considering it didn't used to be this flakey and it's passing on the more resourced Codebuild, would it perhaps be better to annotate with retry?

Sure we can, I've pushed that change now. I just wanted to be sure that this wouldn't get in the way of any other work in the final week, and since it fails quite commonly I wasn't sure that retrying would be enough.

edalex-ian:
Wow, well that provides some great insight into how bad it is if even with retries in both modes!

Yup, let's disable for now then, but please also include a comment next to the disable noting why this is being done.

SammyIsConfused:
Okay @edalex-ian this has turned up green. Go for merge?

edalex-ian:
Yup, let'sa do it. :grin:

PR #2577, opened 2020-12-02
Parse Youtube Date Modified dates properly

SammyIsConfused:
##### Checklist

- [x] tests are included

##### Description of change
The YouTube controls were not working properly. This is because once oEQ got to parsing the date modified string that was returned for a YouTube result it was found to be unparsable. It used to work properly but updates to the YouTube API involved changes to the date format. 
This PR addresses the issue by parsing the date properly, in the YoutubeHandler when running Youtube searches and displaying results, in the YoutubeAttachmentSerializer when saving attachments to the database, and YoutubeViewableResource when displaying the resource summary.
image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/web/controls/youtube/YoutubeAttachmentSerializer.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/web/controls/youtube/YoutubeAttachmentSerializer.java
@@ -30,6 +30,7 @@ import com.tle.web.api.item.equella.interfaces.beans.EquellaAttachmentBean;
 import java.time.Duration;
 import java.util.Date;
 import java.util.Map;
+import org.joda.time.format.ISODateTimeFormat;
 
 @Bind
 @Singleton
@@ -45,7 +46,10 @@ public class YoutubeAttachmentSerializer extends AbstractAttachmentSerializer {
     ybean.setThumbUrl((String) cattach.getData(YoutubeUtils.PROPERTY_THUMB_URL));
     ybean.setViewUrl((String) cattach.getData(YoutubeUtils.PROPERTY_PLAY_URL));
     ybean.setCustomParameters((String) cattach.getData(YoutubeUtils.PROPERTY_PARAMETERS));
-    final Long date = (Long) cattach.getData(YoutubeUtils.PROPERTY_DATE);
+    final Long date =
+        ISODateTimeFormat.dateTimeNoMillis()
+            .parseDateTime((String) cattach.getData(YoutubeUtils.PROPERTY_DATE))
+            .getMillis();
     if (date != null) {
       ybean.setUploadedDate(new Date(date));
     }
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/web/controls/youtube/YoutubeAttachmentSerializer.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/web/controls/youtube/YoutubeAttachmentSerializer.java
@@ -28,9 +28,7 @@ import com.tle.core.item.edit.ItemEditor;
 import com.tle.core.item.serializer.AbstractAttachmentSerializer;
 import com.tle.web.api.item.equella.interfaces.beans.EquellaAttachmentBean;
 import java.time.Duration;
-import java.util.Date;
 import java.util.Map;
-import java.util.Optional;
 
 @Bind
 @Singleton
@@ -46,9 +44,8 @@ public class YoutubeAttachmentSerializer extends AbstractAttachmentSerializer {
     ybean.setThumbUrl((String) cattach.getData(YoutubeUtils.PROPERTY_THUMB_URL));
     ybean.setViewUrl((String) cattach.getData(YoutubeUtils.PROPERTY_PLAY_URL));
     ybean.setCustomParameters((String) cattach.getData(YoutubeUtils.PROPERTY_PARAMETERS));
-    final Optional<Date> date =
-        YoutubeUtils.parseDateModifiedToDate((String) cattach.getData(YoutubeUtils.PROPERTY_DATE));
-    date.ifPresent(ybean::setUploadedDate);
+    YoutubeUtils.parseDateModifiedToDate(cattach.getData(YoutubeUtils.PROPERTY_DATE))
+        .ifPresent(ybean::setUploadedDate);
     ybean.setUploader((String) cattach.getData(YoutubeUtils.PROPERTY_AUTHOR));
     final Object durationData = cattach.getData(YoutubeUtils.PROPERTY_DURATION);
     String duration = null;
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/web/controls/youtube/YoutubeAttachmentSerializer.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/web/controls/youtube/YoutubeAttachmentSerializer.java
@@ -30,7 +30,7 @@ import com.tle.web.api.item.equella.interfaces.beans.EquellaAttachmentBean;
 import java.time.Duration;
 import java.util.Date;
 import java.util.Map;
-import org.joda.time.format.ISODateTimeFormat;
+import java.util.Optional;
 
 @Bind
 @Singleton
@@ -46,13 +46,9 @@ public class YoutubeAttachmentSerializer extends AbstractAttachmentSerializer {
     ybean.setThumbUrl((String) cattach.getData(YoutubeUtils.PROPERTY_THUMB_URL));
     ybean.setViewUrl((String) cattach.getData(YoutubeUtils.PROPERTY_PLAY_URL));
     ybean.setCustomParameters((String) cattach.getData(YoutubeUtils.PROPERTY_PARAMETERS));
-    final Long date =
-        ISODateTimeFormat.dateTimeNoMillis()
-            .parseDateTime((String) cattach.getData(YoutubeUtils.PROPERTY_DATE))
-            .getMillis();
-    if (date != null) {
-      ybean.setUploadedDate(new Date(date));
-    }
+    final Optional<Date> date =
+        YoutubeUtils.parseDateModifiedToDate((String) cattach.getData(YoutubeUtils.PROPERTY_DATE));
+    date.ifPresent(ybean::setUploadedDate);
     ybean.setUploader((String) cattach.getData(YoutubeUtils.PROPERTY_AUTHOR));
     final Object durationData = cattach.getData(YoutubeUtils.PROPERTY_DURATION);
     String duration = null;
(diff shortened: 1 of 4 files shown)


Source/Plugins/Core/com.equella.core/src/com/tle/web/controls/youtube/YoutubeAttachmentSerializer.java
```diff
@@ -45,7 +46,10 @@ public EquellaAttachmentBean serialize(Attachment attachment) {
     ybean.setThumbUrl((String) cattach.getData(YoutubeUtils.PROPERTY_THUMB_URL));
     ybean.setViewUrl((String) cattach.getData(YoutubeUtils.PROPERTY_PLAY_URL));
     ybean.setCustomParameters((String) cattach.getData(YoutubeUtils.PROPERTY_PARAMETERS));
-    final Long date = (Long) cattach.getData(YoutubeUtils.PROPERTY_DATE);
+    final Long date =
+        ISODateTimeFormat.dateTimeNoMillis()
+            .parseDateTime((String) cattach.getData(YoutubeUtils.PROPERTY_DATE))
+            .getMillis();
```
ChristianMurphy, line 52: Would this a be a place where Joda time, could be dropped in favor of `java.Time`?
https://docs.oracle.com/javase/8/docs/api/java/time/ZonedDateTime.html#parse-java.lang.CharSequence-
https://docs.oracle.com/javase/8/docs/api/java/time/chrono/ChronoZonedDateTime.html#toInstant--
https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html#toEpochMilli--

edalex-ian (changes requested):
As @ChristianMurphy mentioned, and as dicussed today, when in Java 8+ land best to move to Java 8 date/time classes.

Further to that though, it would be worthwhile having a shared/static `parseYouTubeDate(string date)` method to abstract the data specification and avoid the duplication of `ISODateTimeFormat.dateTimeNoMillis().parseDateTime(...)`. Obviously this would return the needed Java 8 date type which you can then simply call the equivalent of `.toDate()` and `.getMillis()` on in these three places.

SammyIsConfused:
Ready for review - looks like I had the wrong version of the commit hook formatter installed for 398d0cc.

Source/Plugins/Core/com.equella.core/src/com/tle/web/controls/youtube/YoutubeAttachmentSerializer.java
```diff
@@ -45,10 +46,9 @@ public EquellaAttachmentBean serialize(Attachment attachment) {
     ybean.setThumbUrl((String) cattach.getData(YoutubeUtils.PROPERTY_THUMB_URL));
     ybean.setViewUrl((String) cattach.getData(YoutubeUtils.PROPERTY_PLAY_URL));
     ybean.setCustomParameters((String) cattach.getData(YoutubeUtils.PROPERTY_PARAMETERS));
-    final Long date = (Long) cattach.getData(YoutubeUtils.PROPERTY_DATE);
-    if (date != null) {
-      ybean.setUploadedDate(new Date(date));
-    }
+    final Optional<Date> date =
+        YoutubeUtils.parseDateModifiedToDate((String) cattach.getData(YoutubeUtils.PROPERTY_DATE));
+    date.ifPresent(ybean::setUploadedDate);
```
edalex-ian, line 51: In this case, the intermediate `date` variable is rather redundant. I'd keep it nice and tidy and simply have:

```suggestion
    YoutubeUtils.parseDateModifiedToDate((String) cattach.getData(YoutubeUtils.PROPERTY_DATE))
      .ifPresent(ybean::setUploadedDate);
```

Source/Plugins/Core/com.equella.core/src/com/tle/web/controls/youtube/YoutubeHandler.java
```diff
@@ -286,9 +287,10 @@ public void registered(String id, SectionTree tree) {
 
               YoutubeResultOption result = new YoutubeResultOption(videoId);
               result.setAuthor(new KeyLabel(ADD_AUTHOR_LABEL, vidInfo.getChannelTitle()));
-              result.setDate(
-                  dateRendererFactory.createDateRenderer(new Date(vidInfo.getPublishedAt())));
-
+              Optional<Date> parsedDate =
+                  YoutubeUtils.parseDateModifiedToDate(vidInfo.getPublishedAt());
+              parsedDate.ifPresent(
+                  date -> result.setDate(dateRendererFactory.createDateRenderer(date)));
```
edalex-ian, line 293: Similar to above.

```suggestion
              YoutubeUtils.parseDateModifiedToDate(vidInfo.getPublishedAt())
                .ifPresent(
                  date -> result.setDate(dateRendererFactory.createDateRenderer(date)));
```

(Unless I'm missing them being used somewhere else.)

Source/Plugins/Core/com.equella.core/src/com/tle/web/controls/youtube/YoutubeViewableResource.java
```diff
@@ -188,11 +189,13 @@ public boolean isCustomThumb() {
     }
 
     // Uploaded
-    Long date = (Long) youTubeAttachment.getData(YoutubeUtils.PROPERTY_DATE);
-    if (date != null) {
-      commonDetails.add(
-          makeDetail(UPLOADED, dateRendererFactory.createDateRenderer(new Date(date))));
-    }
+    Optional<Date> date =
```
edalex-ian, line 192: As above. `date` redundant.

Source/Plugins/Core/com.equella.core/src/com/tle/web/controls/youtube/YoutubeUtils.java
```diff
@@ -55,6 +59,45 @@ public static String formatDuration(String duration) {
...
+  public static Optional<Long> parseDateModifiedToMillis(String date) {
+    Optional<Long> parsedDate;
+    try {
+      parsedDate =
+          Optional.of(
+              ZonedDateTime.parse(date, DateTimeFormatter.ISO_DATE_TIME)
+                  .toInstant()
+                  .toEpochMilli());
+    } catch (DateTimeParseException e) {
+      throw new RuntimeException("Unable to parse youtube date modified: " + date, e);
```
edalex-ian, line 80: Rewrapping into a generic exception can limit possibilities for callers. I'd say just drop the `try/catch` adn allow the DateTimeParseException to percolate. Or, if you feel it's handy to add the extra message, consider doing so but maybe go with a more meaningful exception - say, `IllegalArgumentException` or reuse `DateTimeParseException`.

edalex-ian (approved):
Looks good. Just keep in mind with your optional you don't need to be as verbose with intermediate variables (if they're not needed later). And also, be careful re-wrapping exceptions into generic ones, as callers can't then filter to that.

If you could please address some of those suggestions, then you're good to merge I'd say.

Source/Plugins/Core/com.equella.core/src/com/tle/web/controls/youtube/YoutubeUtils.java
SammyIsConfused: I think it's worth keeping the message. So I'll just throw a new DateTimeParseException with the message and the contents of the initial thrown one.

SammyIsConfused:
Took out the <p> tag but it turns out the formatter requires it, so I reverted a367caa.

SammyIsConfused:
So the reason for the failures on 68b8970 is because in the test institutions there are existing YouTube attachments that were created in the old v2 API implementation - and for those they are stored in epoch long format. af68070 fixes that by making the parse function accept and Object - if its a long we assume its an epoch, otherwise we assume its a ISO_DATE_TIME string.

Source/Plugins/Core/com.equella.core/src/com/tle/web/controls/youtube/YoutubeUtils.java
```diff
@@ -86,14 +87,24 @@ public static String formatDuration(String duration) {
...
-  public static Optional<Date> parseDateModifiedToDate(String date) {
-    Optional<Long> parsedLong = parseDateModifiedToMillis(date);
+  public static Optional<Date> parseDateModifiedToDate(Object date) {
     Optional<Date> parsedDate = Optional.empty();
-    if (parsedLong.isPresent()) {
-      parsedDate = Optional.of(new Date(parsedLong.get()));
+    if (date instanceof Long) {
+      // if its a long, assume its an epoch long
+      parsedDate = Optional.of(new Date((Long) date));
+    } else {
```
edalex-ian, line 102: I like the rework of this method - and supporting calls. But, why not do an `else if(date instanceof String)` for this block, and then have a final `else` block which throws an `IllegalArgumentException`?

Best to only assume if you have to.

edalex-ian (changes requested):

Source/Plugins/Core/com.equella.core/src/com/tle/web/controls/youtube/YoutubeUtils.java
SammyIsConfused: Fixed in 3be02e2.

Source/Plugins/Core/com.equella.core/src/com/tle/web/controls/youtube/YoutubeUtils.java
```diff
@@ -55,6 +59,60 @@ public static String formatDuration(String duration) {
...
+    Optional<Date> parsedDate = Optional.empty();
+    if (date instanceof Long) {
+      // if its a long, assume its an epoch long
+      parsedDate = Optional.of(new Date((Long) date));
+    } else if (date instanceof String) {
+      // assume its an ISO_DATE_TIME string
+      Optional<Long> parsedLong = parseDateModifiedToMillis((String) date);
+      if (parsedLong.isPresent()) {
+        parsedDate = Optional.of(new Date(parsedLong.get()));
+      }
```
edalex-ian, line 108: Is this not just something like:

```suggestion
      parsedDate = parseDateModifiedToMillis((String) date)
        map(millis -> new Date(millis));
```

I'm just freestyling there, but hopefully you get the idea.

edalex-ian (approved):
That all looks nice and tidy @SammyIsConfused and a great improvement over what was there.

One last lingering query. But after that, you'd be good to merge. :+1:

Source/Plugins/Core/com.equella.core/src/com/tle/web/controls/youtube/YoutubeUtils.java
SammyIsConfused: You were close! It has to be `flatMap`.

ChristianMurphy (approved):

Issue #2533, opened 2020-11-18
Enable the module list to appear in the right hand Blackboard LTI selection session


cbeach47:
**Is your feature request related to a problem? Please describe.**
With the modern Blackboard LTI integration , the course list no longer shows up on the right-hand side of the selection session. 

**Describe the solution you'd like**
This will require REST calls to the Bb site to build out the course list.

**Describe alternatives you've considered**
Blackboard doesn't send course lists with their LTI launches, thus the need for Bb REST calls.

cathfitz:
@cbeach47 has testing been completed for this issue? Can it now be closed?

cbeach47:
sry, yes

PR #2581, opened 2020-12-03
Enable course folder selection in Bb LTI flow

cbeach47:
History: #2533 

##### Checklist
- [x] commit message follows [commit guidelines][] - will squash the commits
- [N/A] tests are included - no Bb LTI / REST tests exist
- [x] documentation is changed or added - will be opening a PR against the doc site shortly

##### Description of change
The Bb LTI flow now offers the configuration to enable the right-hand course folder selection area.  This existed in the B2 / WS flows with Bb, and is now being recreated in the LTI / REST flows.

This also fixes some caching issues with the original LTI / REST flows.

Original (single folder) view:
image

The now enhanced multi-folder view :
image

This has only been confirmed to work in the original oEQ UX and the Bb original courses.  Bb ultra course appear to add the content (no errors), but the content does not show up in the lesson modules.

To enable (I will add this to the documentation):
* Set the LTI Tool Provider placement URL to:
```
{oEQ-URL}/ltisignon.do?action=structured&connector_uuid={backing-oeq-external-system-connector}
```
* Add the custom parameters:
```
course_id=@X@course.pk_string@X@
content_id=@X@content.id@X@
```


Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/impl/BlackboardRESTConnectorServiceImpl.java
```diff
@@ -299,46 +326,161 @@ private Content getContentBean(Connector connector, String courseID, String fold
...
+    for (ConnectorFolder folder : folders) {
+      ConnectorFolder foundFolder = findFolder(folder, folderId);
+      if (foundFolder != null) {
+        return foundFolder.getFolders();
+      }
+    }
+
+    // Cache of folders exist, but folderId was not found
+    return null;
+  }
```
edalex-ian, line 348: My initial reaction here is that it'd be great to see these use Java 8 features, and specifically `Stream`. However I appreciate it's calling a recursive function and so you want it to end execution on the first match (to avoid unnecessary tree walking).

But it would still be great if you could use the Java 8 `forEach` and perhaps an `Optional` around the search with `findFolder`. Just helps to make the code a bit more robust.

Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/impl/BlackboardRESTConnectorServiceImpl.java
```diff
@@ -299,46 +326,161 @@ private Content getContentBean(Connector connector, String courseID, String fold
...
+
+  /**
+   * Recursively searches for a folder with the given ID
+   *
+   * @param folder
+   * @param folderId
+   * @return null if not found, the children of the folder if found
```
edalex-ian, line 355: Rather than returning the ol' "`null` if not found", in Java 8 there's usefulness of `Optional` and all it brings to the calling code. And imagine how nice it'd read here:

```java
Optional<ConnectorFolder> possibleSubFolder = findFolder(subFolder, folderId);
if(possibleSubFolder.isPresent()) {
  return possibleSubFolder.get();
}
```

edalex-ian (commented):
Heya @cbeach47 

I think the main bits of feedback I have are:

1. It would be handy if you had some unit tests around your functions building of the folder structures - especially on the recursive ones. There are some scattered around the oEQ codebase, and it feels like this data could be well suited to it - especially if you pushed those methods into their own stand-alone class;
2. Also, it'd be great if you made use of the Java 8 `Stream` and `Iterable.forEach` as they provide for more robust and readable code. Especially the `parseFolders` (and probably `parseFolder`) could make great use of `Collection.stream()`. Further, there'd be some wins about the place for `Optional` too.
3. Lastly the suppressing of the `AuthenticationException`s comes across a bit odd and place of concern - could be worthwhile to at least log them?

I would've liked more time to dig in deeper, but alas it was not to be.

Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/impl/BlackboardRESTConnectorServiceImpl.java
cbeach47: good point, that is cleaner

cbeach47:
@edalex-ian  thanks for the review.

For # 1: I've added a couple unit tests to start out with.  I also pulled out the more utility methods into a separate class to put under unit tests.

For # 2: I went through and revised the various loops to leverage streams, and made more use of `Optional`.   I realized that we have two `Optional` classes - one from Google, and the other from `java.util` floating around in oEQ.  I'll add that to the next CDM to discuss alignment.  For now, I stuck with Google's `Optional` due to a quick usage search in the code base.

For # 3: It's a fair concern.  I've removed most of the uses of throwing `AuthenticationException` in favor of checking the the returned `Optional` object. 

I've also discovered a bug in the caching that stemmed (or was exacerbated) from my switch from user session caches to `ReplicatedCaches`.  I'm going to merge this PR once the CI goes green, and get a bugfix PR in early next week.

cbeach47:
FYI - for the docs - https://github.com/openequella/openequella.github.io/pull/76

PR #2582, opened 2020-12-03
Support Advanced/Remote searches in Selection Session

PenghaiZhang:
#688 

##### Checklist

- [ ] tests are included

##### Description of change

This PR addresses the issue that Remote search page and Advanced search page didn't display properly in Selection Session if they are open from new search UI.

There are three changes: 

1. Create a new TS component named 'OeqLink'. The reason is we have a few places where we need to switch between `React Route Link` and `MUI Link`, depending on whether the page is in Selection Session. 

2. For Advanced search requests coming from new search UI in Selection Session, create a new Legacy endpoint which is `advanced/searching.do`. To support this endpoint, also create a Section module and a Section. 

3. Make sure functions related to `Drag and Drop` are only called when Selection Session is in structured.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/modules/LegacySelectionSessionModule.ts
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/modules/LegacySelectionSessionModule.ts
@@ -216,6 +216,42 @@ export const buildSelectionSessionItemSummaryLink = (
 };
 
 /**
+ * Build a Selection Session specific Remote search Link. Recommended to first call `isSelectionSessionOpen()`
+ * before use.
+ * @param uuid The UUID of a Remote search
+ */
+export const buildSelectionSessionRemoteSearchLink = (uuid: string): string => {
+  const { stateId, integId } = getSelectionSessionInfo();
+  const remoteSearchLink = AppConfig.baseUrl.concat(
+    `access/z3950.do?.repository=${uuid}&uc=true&_sl.stateId=${stateId}`
+  );
+
+  if (integId) {
+    return remoteSearchLink.concat(`&_int.id=${integId}`);
+  }
+  return remoteSearchLink;
+};
+
+/**
+ * Build a Selection Session specific Advanced search Link. Recommended to first call `isSelectionSessionOpen()`
+ * before use.
+ * @param uuid The UUID of an Advanced search
+ */
+export const buildSelectionSessionAdvancedSearchLink = (
+  uuid: string
+): string => {
+  const { stateId, integId } = getSelectionSessionInfo();
+  const remoteSearchLink = AppConfig.baseUrl.concat(
+    `searching.do?in=P${uuid}&editquery=true&_sl.stateId=${stateId}`
+  );
+
+  if (integId) {
+    return remoteSearchLink.concat(`&_int.id=${integId}`);
+  }
+  return remoteSearchLink;
+};
+
+/**
  * Update the content of DIV "selection-summary". This function is primarily for
  * 'selectOrAdd' mode. In this mode, what is returned from 'searching.do' is an
  * object of 'LegacyContentResponse'.
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/modules/LegacySelectionSessionModule.ts
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/modules/LegacySelectionSessionModule.ts
@@ -16,7 +16,12 @@
  * limitations under the License.
  */
 import Axios from "axios";
-import { getBaseUrl, getRenderData, SelectionSessionInfo } from "../AppConfig";
+import {
+  API_BASE_URL,
+  getBaseUrl,
+  getRenderData,
+  SelectionSessionInfo,
+} from "../AppConfig";
 import { LegacyContentResponse } from "../legacycontent/LegacyContent";
 import { routes } from "../mainui/routes";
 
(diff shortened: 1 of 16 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/modules/LegacySelectionSessionModule.ts
```diff
@@ -215,6 +226,42 @@ export const buildSelectionSessionItemSummaryLink = (
...
+/**
+ * Build a Selection Session specific Remote search Link. Recommended to first call `isSelectionSessionOpen()`
+ * before use.
+ * @param uuid The UUID of a Remote search
+ */
+export const buildSelectionSessionRemoteSearchLink = (uuid: string): string => {
+  const { stateId, integId } = getSelectionSessionInfo();
+  const remoteSearchLink = AppConfig.baseUrl.concat(
+    `access/z3950.do?.repository=${uuid}&uc=true&_sl.stateId=${stateId}`
+  );
```
edalex-ian, line 238: To keep things DRY so as to ensure accuracy, this should more so wrap the route in `routes`, something like:

```suggestion
  const remoteSearchLink = AppConfig.baseUrl.concat(
    `${routes.RemoteSearch.to(uuid)}&_sl.stateId=${stateId}`
  );
```

The same suggestion applies for the below Advanced Search.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/LegacySelectionSessionModule.ts
```diff
@@ -215,6 +226,42 @@ export const buildSelectionSessionItemSummaryLink = (
...
+): string => {
+  const { stateId, integId } = getSelectionSessionInfo();
+  const advancedSearchLink = AppConfig.baseUrl.concat(
+    `advanced/searching.do?in=P${uuid}&editquery=true&_sl.stateId=${stateId}`
+  );
+
+  if (integId) {
+    return advancedSearchLink.concat(`&_int.id=${integId}`);
+  }
+  return advancedSearchLink;
```
edalex-ian, line 262: This body pretty well duplicates the above function. I'd suggest you have one function such as:

```typescript
const buildSelectionSessionLink = (routerPath: string) => {
  const { stateId, integId } = getSelectionSessionInfo();
  const selectionSessionLink = AppConfig.baseUrl.concat(
    `${routerPath}&_sl.stateId=${stateId}`
  );

  if (integId) {
    return selectionSessionLink.concat(`&_int.id=${integId}`);
  }
  return selectionSessionLink;
}
```

And this this function would simply be:

```typescript
export const buildSelectionSessionAdvancedSearchLink = (
  uuid: string
): string => buildSelectionSessionLink(routes.AdvancedSearch.to(uuid));
```

And you'd obviously do the same for `buildSelectionSessionRemoteSearchLink`. (And further down this review I've noted you'd probably want the same for `buildSelectionSessionItemSummaryLink`.)

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResult.tsx
```diff
@@ -482,11 +480,13 @@ export default function SearchResult({
...
-            <DragIndicatorIcon />
-          </IconButton>
-        </Grid>
+        {inStructured && (
+          <Grid item>
+            <IconButton>
+              <DragIndicatorIcon />
+            </IconButton>
+          </Grid>
+        )}
```
edalex-ian, line 489: With this block we now have further conditional logic where we're already inside a ternary. I think for readability sake, this `true` branch of the ternary should now be in it's own function. So then this value expression might end up looking like:

```typescript
const itemPrimaryContent =
  inSelectionSession && !isSelectSummaryButtonDisabled() ? (
    selectionSessionItemContent()
  ) : (
    itemLink()
  );  
```

Source/Plugins/Core/com.equella.core/src/com/tle/web/searching/section/RootAdvancedSearchSection.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  @Override
+  public void registered(String id, SectionTree tree) {
+    super.registered(id, tree);
+  }
+
+  @Override
+  public boolean useNewSearch() {
+    return false;
```
edalex-ian, line 76: Could be value here in a comment saying why you're setting this to 'false'.

Source/Plugins/Core/com.equella.core/src/com/tle/web/searching/section/RootSearchSection.java
```diff
@@ -70,6 +70,11 @@ protected String getSessionKey() {
     return SEARCH_SESSIONKEY;
   }
 
+  // For child sections that need to skip new Search UI, override this function and return false.
```
edalex-ian, line 73: Be good to do as JavaDoc.

Source/Plugins/Core/com.equella.core/src/com/tle/web/searching/section/RootSearchSection.java
```diff
@@ -70,6 +70,11 @@ protected String getSessionKey() {
...
+  public boolean useNewSearch() {
```
edalex-ian, line 74: Should this perhaps just be `protected`?

```suggestion
  protected boolean useNewSearch() {
```

It's not intended to be used a public interface, just for internal operations.

edalex-ian (changes requested):
Looks great @PenghaiZhang 

Just a few bits of code tidy-up (things obviously just overlooked while you were attempting to wrangle this beasty).

Source/Plugins/Core/com.equella.core/js/tsrc/modules/LegacySelectionSessionModule.ts
PenghaiZhang: There is one small difference. For the ItemSummary URL, it requires `&a=${layout}` whereas the remote one and advanced one do not need. But I think it's fine to add this query string to the remote one and advanced one.

PenghaiZhang:
Although the rest module test failed due to `psc-package`, Codebuild passed so merge the PR.

PR #2583, opened 2020-12-04
Don't display the Leave or Cancel dialog when saving a new item

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

The previous fix made in #1330 does not work in Selection Session because the ItemSummary page URL has some query params (e.g. `stateId=2`). 

So in this PR, I slightly modified the ItemSummary URL Regex to make sure it also works for above situation.

A video is available [here](https://streamable.com/rdrck4).


edalex-ian (approved):
Yeah, I'm happy to approve this in the current timeline.

Ideally though, there'd be proper breakdown and doco for that regex.

PR #2590, opened 2020-12-08
Address eslint warnings (not exhaustive/deps)

SammyIsConfused:
##### Checklist

##### Description of change
This PR aims to address all remaining eslint warnings except those that fall under `react-hooks/exhaustive-deps`. Since those will take a lot more investigation to fix without breaking functionality, they will be done separately.

- Add typing of `Meta<propsOfStorySubject>` to all storybook default exports. Added export to interfaces where necessary.
- Add name to the default export function in index.tsx
- Removed false import of DateRange Material UI icon from SearchModule.ts
- Renamed RangeType so it doesn't get reassigned

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
@@ -15,7 +15,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { DateRange } from "@material-ui/icons";
+
 import * as OEQ from "@openequella/rest-api-client";
 import { Location } from "history";
 import { pick } from "lodash";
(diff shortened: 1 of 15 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -408,14 +408,14 @@ export const legacyQueryStringToSearchOptions = async (
...
-  const RangeType = Union(
+  const RangeTypeLiterals = Union(
     Literal("between"),
     Literal("after"),
     Literal("before"),
     Literal("on")
   );
 
-  type RangeType = Static<typeof RangeType>;
+  type RangeType = Static<typeof RangeTypeLiterals>;
```
edalex-ian, line 418: Was that actually triggering a lint issue?

edalex-ian (approved):
Good stuff @SammyIsConfused 

And I've confirmed it's only dependency lint warnings left by looking at the check output: https://github.com/openequella/openEQUELLA/pull/2590/checks?check_run_id=1515195105#step:12:61

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
SammyIsConfused: Yup! It was getting assigned as a `const` on line 411 and then on line 418 the same identifier is declared as a `type`. It triggered off `@typescript-eslint/no-redeclare`. The other one in this file was `DateRange` which was declared as an interface, but there was an import for a Material UI Icon called `DateRange` in there too. 
image

edalex-ian: Okay, good to know. As it makes sense that you can do this, but I've always wondered what's the idiomatic approach. I'm happy to be guided by ESLint - they're a good egg. :wink:

PR #2591, opened 2020-12-08
EBP: Make sure Youtube dialog doesn't stretch offscreen

SammyIsConfused:
##### Checklist

##### Description of change
Bit of a quick win that fell out of work involved with #2577. I won't be adding additional EBP changes to this PR as this is not scheduled work.
When the YouTube search happens, all of the results get put into the dialog list. Without a max-height, this causes the dialog to stretch down which looks strange. 
- Added a max-height for the YouTube Dialog
- Set overflow-y to scroll
- Added standard box shadow and border radius to dialog
- Move dialog upwards to make room for more results

Before change: 
Peek 2020-12-08 16-35
After change: 
Peek 2020-12-15 10-15

### DIFF
--- a/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
+++ b/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
@@ -540,6 +540,11 @@ select[multiple="multiple"] {
   width: 200px;
 }
 
+#youtube-query ~ .modal-search-results {
+  max-height: 300px;
+  overflow-y: scroll;
+}
+
 .modal-search-result li {
   position: relative;
   margin: 0 0 10px 0;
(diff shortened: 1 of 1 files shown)


edalex-ian (commented):
Thanks for the improvement @SammyIsConfused , but I'm not sure about the small size of the dialog. The new size really only fits one at a time, and so makes it somewhat difficult to scroll results.

I feel it needs one or both of:

* something like `75vh`
* use `position: sticky` so that the top of the dialog scrolls to almost the top and then pauses while the rest then scrolls. :thinking:

SammyIsConfused:
> Thanks for the improvement @SammyIsConfused , but I'm not sure about the small size of the dialog. The new size really only fits one at a time, and so makes it somewhat difficult to scroll results.
> 
> I feel it needs one or both of:
> 
>     * something like `75vh`
> 
>     * use `position: sticky` so that the top of the dialog scrolls to almost the top and then pauses while the rest then scrolls. thinking

Used 75vh. Tried to get position: sticky to work but the header hierarchy of the dialog made it awkward to get working nicely. That said we at least keep the title of the dialog showing now.

edalex-ian (approved):
Very nice. I love the application of `buttonShadowBorder` too. :+1:

Issue #2587, opened 2020-12-07
Caching issues with Bb REST integration


cbeach47:
**Describe the bug**
While the Bb LTI / REST integration functionally works, there are race and timeout conditions due to the caching that should be resolved.

**To Reproduce**
Primarily:
1. If the Bb token becomes expired partway through the flow, the behavior needs to be defined better.  
2. If the course or folder cache becomes expired partway through, the integration doesn't refresh the cache - it just fails.

**Expected behavior**

The token and user_id cache flows have well defined behavior if they are expire or are past their oEQ ttl partway through the flow.

The course and course_folder cache flows will pull data from Bb if the cache has no values available.

cathfitz:
@cbeach47 Hi Chris. Is testing completed for this issue? Can we please close this ticket?

cbeach47:
sry, yes

PR #2592, opened 2020-12-09
Fix Bb LTI / REST caching and request all course folders

cbeach47:
History in #2587

##### Checklist

- [x] tests are included - N/A
- [x] screenshots are included showing significant UI changes - N/A
- [x] documentation is changed or added - N/A

##### Description of change

Caching adjustments:
* Leverage Bb REST API refresh tokens so the user does not have to log back into Bb during the same oEQ session
* User session caching now in place for auth token
* Auth caches now combined into a single cache of the entire `Token`
* `Course`s and `Folder`s requests are now setup to check the cache, and if none are available, pull a fresh set of results from Bb

Found a unrelated bug where only the first 200 content items from a Bb course were being returned.  Now pulls all pages of content results.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/BlackboardRESTConnectorService.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/BlackboardRESTConnectorService.java
@@ -24,8 +24,6 @@ import com.tle.core.connectors.blackboard.beans.Token;
 import com.tle.core.connectors.service.ConnectorRepositoryImplementation;
 
 public interface BlackboardRESTConnectorService extends ConnectorRepositoryImplementation {
-  // TODO may need more method sigs
-
   /**
    * Admin setup function
    *
@@ -47,14 +45,20 @@ public interface BlackboardRESTConnectorService extends ConnectorRepositoryImple
    * The connector object will need to store an encrypted admin token in the DB. Use this method to
    * encrypt the one returned from Blackboard.
    *
-   * @param data
-   * @return
+   * @param data content to encrypt
+   * @return encrypted content
    */
   String encrypt(String data);
 
   String decrypt(String encryptedData);
 
-  String encryptKeyAndSecret(Connector connector);
+  /**
+   * Creates a Base 64 encoded string of the connector's key and secret.
+   *
+   * @param connector used to launch this REST flow
+   * @return encoded string of connector key and secret
+   */
+  String buildBasicAuthorizationCredentials(Connector connector);
 
   void setAuth(Connector connector, Token token);
 
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/BlackboardRESTConnectorService.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/BlackboardRESTConnectorService.java
@@ -20,6 +20,7 @@ package com.tle.core.connectors.blackboard.service;
 
 import com.tle.annotation.Nullable;
 import com.tle.common.connectors.entity.Connector;
+import com.tle.core.connectors.blackboard.beans.Token;
 import com.tle.core.connectors.service.ConnectorRepositoryImplementation;
 
 public interface BlackboardRESTConnectorService extends ConnectorRepositoryImplementation {
@@ -46,16 +47,16 @@ public interface BlackboardRESTConnectorService extends ConnectorRepositoryImple
    * The connector object will need to store an encrypted admin token in the DB. Use this method to
    * encrypt the one returned from Blackboard.
    *
-   * @param token
+   * @param data
    * @return
    */
   String encrypt(String data);
 
   String decrypt(String encryptedData);
 
-  void setToken(Connector connector, String value);
+  String encryptKeyAndSecret(Connector connector);
 
-  void setUserId(Connector connector, String value);
+  void setAuth(Connector connector, Token token);
 
   void removeCachedCoursesForConnector(Connector connector);
 }
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/impl/BlackboardRESTConnectorServiceImpl.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/impl/BlackboardRESTConnectorServiceImpl.java
@@ -636,7 +636,7 @@ public class BlackboardRESTConnectorServiceImpl extends AbstractIntegrationConne
       LOGGER.error(exception.getMessage());
       // Likely an expired auth token that failed to refresh.  Guide the user to try again
       removeCachedValue(authCache, buildCacheKey(CACHE_ID_AUTH, connector));
-      new AuthenticationException("Unable to refresh auth token.  Please retry action.");
+      throw new AuthenticationException("Unable to refresh auth token.  Please retry action.");
     }
   }
 
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/impl/BlackboardRESTConnectorServiceImpl.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/impl/BlackboardRESTConnectorServiceImpl.java
@@ -34,7 +34,6 @@ import com.tle.annotation.Nullable;
 import com.tle.beans.item.IItem;
 import com.tle.beans.item.ViewableItemType;
 import com.tle.common.Check;
-import com.tle.common.Pair;
 import com.tle.common.PathUtils;
 import com.tle.common.connectors.ConnectorContent;
 import com.tle.common.connectors.ConnectorCourse;
@@ -54,6 +53,7 @@ import com.tle.core.connectors.service.AbstractIntegrationConnectorRespository;
 import com.tle.core.connectors.service.ConnectorRepositoryService;
 import com.tle.core.connectors.service.ConnectorService;
 import com.tle.core.connectors.utils.ConnectorEntityUtils;
+import com.tle.core.encryption.EncryptionService;
 import com.tle.core.guice.Bind;
 import com.tle.core.institution.InstitutionService;
 import com.tle.core.plugins.AbstractPluginService;
@@ -72,7 +72,6 @@ import java.io.Serializable;
 import java.net.URI;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
@@ -113,16 +112,15 @@ public class BlackboardRESTConnectorServiceImpl extends AbstractIntegrationConne
   @Inject private InstitutionService institutionService;
   @Inject private AuditLogService auditService;
   @Inject private UserSessionService userSessionService;
+  @Inject private EncryptionService encryptionService;
 
   private static final String CACHE_ID_COURSE = "BbRestCoursesByUser";
   private static final String CACHE_ID_COURSE_FOLDERS = "BbRestFoldersByUserAndCourse";
-  private static final String CACHE_ID_TOKEN = "BbRestTokensByUser";
-  private static final String CACHE_ID_USERID = "BbRestIdsByUser";
+  private static final String CACHE_ID_AUTH = "BbRestAuthByUser";
 
   private ReplicatedCache<ImmutableList<Course>> courseCache;
   private ReplicatedCache<ImmutableList<ConnectorFolder>> courseFoldersCache;
-  private ReplicatedCache<String> tokenCache;
-  private ReplicatedCache<String> uidCache;
+  private ReplicatedCache<Token> authCache;
 
   private static final ObjectMapper jsonMapper = new ObjectMapper();
   private static final ObjectMapper prettyJsonMapper = new ObjectMapper();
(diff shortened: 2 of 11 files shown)


ChristianMurphy:
https://github.com/openequella/openEQUELLA/pull/815#pullrequestreview-222997052 applies here as well.
Maybe with the Spring upgrade Lombok would work?

cbeach47:
It's a fair point.  Budget on this effort is tight, so I added #2593

ChristianMurphy (approved):

Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/impl/BlackboardRESTConnectorServiceImpl.java
```diff
@@ -346,13 +315,27 @@ private boolean isCoursesCached(Connector connector) {
...
+    Paging paging = contents.getPaging();
+
+    while (paging != null && paging.getNextPage() != null) {
+      contents =
+          sendBlackboardData(
+              connector, paging.getNextPage(), Contents.class, null, Request.Method.GET);
+      allContent.addAll(contents.getResults());
+
+      paging = contents.getPaging();
+    }
```
edalex-ian, line 334: Should this just be a do-while? Something like:

```java
    Optional<String> url = Optional
        .of(API_ROOT_V1 + "courses/" + courseId + "/contents?recursive=true");
    final List<Content> allContent = new ArrayList<>();
    do {
      Contents contents =
          sendBlackboardData(connector, url.get(), Contents.class, null, Request.Method.GET);
      allContent.addAll(contents.getResults());
      url = Optional.ofNullable(contents.getPaging()).map(Paging::getNextPage);
    } while (url.isPresent());
```

Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/impl/BlackboardRESTConnectorServiceImpl.java
```diff
@@ -606,6 +596,56 @@ public boolean supportsReverseSort() {
...
   }
 
+  private Token getUserSessionAuth() {
+    Optional<Token> auth =
+        userSessionService.getAttribute(BlackboardRESTConnectorConstants.USER_SESSION_AUTH_KEY);
+    if (!auth.isPresent()) {
+      throw new AuthenticationException(
+          "User authentication details are not available in the session");
+    }
+    return auth.get();
```
edalex-ian, line 606: Should be an orElseThrow:

```suggestion
    return userSessionService.getAttribute(BlackboardRESTConnectorConstants.USER_SESSION_AUTH_KEY)
      .orElseThrow(
        () -> new AuthenticationException("User authentication details are not available in the session"));
```

Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/impl/BlackboardRESTConnectorServiceImpl.java
```diff
@@ -606,6 +596,56 @@ public boolean supportsReverseSort() {
...
+  }
+
+  private void refreshBlackboardToken(Connector connector) {
+    Token expAuth = getUserSessionAuth();
+
+    // Make a reasonable effort to obtain the latest refresh token
+    Optional<Token> cachedAuth = getAuth(connector);
+    if (cachedAuth.isPresent()) {
+      expAuth = cachedAuth.get();
+    }
```
edalex-ian, line 624: Is this:

```suggestion
    Token expAuth = getAuth(connector)
      .orElse(() -> getUserSessionAuth());
```

Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/impl/BlackboardRESTConnectorServiceImpl.java
```diff
@@ -606,6 +596,56 @@ public boolean supportsReverseSort() {
...
+              path,
+              Token.class,
+              "grant_type=refresh_token",
+              Request.Method.POST,
+              false));
+    } catch (Exception exception) {
+      LOGGER.error(exception.getMessage());
+      // Likely an expired auth token that failed to refresh.  Guide the user to try again
+      removeCachedValue(authCache, buildCacheKey(CACHE_ID_AUTH, connector));
+      new AuthenticationException("Unable to refresh auth token.  Please retry action.");
```
edalex-ian, line 645: Is this missing a throw?

```suggestion
      throw new AuthenticationException("Unable to refresh auth token.  Please retry action.");
```

Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/impl/BlackboardRESTConnectorServiceImpl.java
```diff
@@ -730,34 +786,67 @@ private void setCachedCourses(Connector connector, ImmutableList<Course> courses
...
+
+    while (paging != null && paging.getNextPage() != null) {
+      courses =
+          sendBlackboardData(
+              connector, paging.getNextPage(), CoursesByUser.class, null, Request.Method.GET);
+      for (CourseByUser cbu : courses.getResults()) {
+        allCourses.add(cbu.getCourse());
+      }
+      paging = courses.getPaging();
+    }
```
edalex-ian, line 827: This looks like it should be a do-while like earlier up.

One thing also interesting here is the duplicated pattern: `sendBlackboardData` -> process output to create a `List` -> check for another page -> repeat. Feels like that should be a method which takes a lambda to support abstracting the commonality of processing/consuming `List` output.

edalex-ian (changes requested):
Heya Chris,

Interesting stuff. Was a tad hard to review as one minute I've paged in the design around auth tokens, then I need to swap pages to building up folder trees. It would've been really great if these were split in two, as then I could've reviewed them properly and thoroughly - and much quicker.

But in the end, the things that jumped out at me:

* Great to see some more JavaDoc being added, but need a bit more polish in some cases (especialy with `@param` and `@return` from the template added, but then ignored);
* The `sendBlackboardData` function really needs to be tidied up, and each addition to it (such as here) is definitely making it worse - and now it even has duplicate string comparisons to check current state/mode; and
* Two big blocks of code for getting lists should simply be do-while as opposed to while, and even then share the same algorithm which could be simply broken out.

It's really great to see the increased use of Optional and streams - and the continuation here. Thank you so much for taking that on board in the last review. Was a great value add and will definitely help us into the future.

cbeach47:
Hmmm - those commits assume java's Optional, and there is a mix of Java and Guava Optional in these classes.  I've put it on the CDM agenda for Jan'21, however, to clean up the code, and since it sounds like Guava recommends Java Optional where possible, I'll convert the Guava Optional to Java Optional early on, and see how that goes.

https://stackoverflow.com/questions/8534442/shortcut-for-guavas-optional-use-with-exceptions

Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/impl/BlackboardRESTConnectorServiceImpl.java
cbeach47: doh - yes it is.

Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/impl/BlackboardRESTConnectorServiceImpl.java
```diff
@@ -687,25 +727,38 @@ public String decrypt(String encryptedData) {
...
+  @Override
+  public String encryptKeyAndSecret(Connector connector) {
+    final String apiKey = connector.getAttribute(BlackboardRESTConnectorConstants.FIELD_API_KEY);
+    final String apiSecret =
+        encryptionService.decrypt(
+            connector.getAttribute(BlackboardRESTConnectorConstants.FIELD_API_SECRET));
+    return new Base64()
+        .encode((apiKey + ":" + apiSecret).getBytes())
+        .replace("\n", "")
+        .replace("\r", "");
```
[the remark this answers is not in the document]
cbeach47: I pulled that from existing code.  When I remove the 2 `replace` methods, Blackboard responds with `Invalid Credentials`, so I'll leave that logic as-is for now.

Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/impl/BlackboardRESTConnectorServiceImpl.java
```diff
@@ -138,10 +136,15 @@
...
-    uidCache = cacheService.getCache(CACHE_ID_USERID, 1000, 60, TimeUnit.MINUTES);
+    // These two caches can be easily rebuilt, and the original
+    //  data on the Bb servers can frequently change
+    courseCache = cacheService.getCache(CACHE_ID_COURSE, 100, 2, TimeUnit.MINUTES);
+    courseFoldersCache = cacheService.getCache(CACHE_ID_COURSE_FOLDERS, 100, 2, TimeUnit.MINUTES);
+    // Depending on usage, setting this values (size and ttl) too low will break the
+    //  FU / MER / AES flows. Setting it higher to be safe.  Bb tokens expire after 60
+    //  minutes, and since we use the refresh_token flows, refreshing a token is less
+    //  noticeable to the user then requesting a new access token
+    authCache = cacheService.getCache(CACHE_ID_AUTH, 1000, 90, TimeUnit.MINUTES);
```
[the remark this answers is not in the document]
cbeach47: The 3LO flow with Blackboard requires user interaction (or at least a full page redirect).  This disrupts the user experience, and at times, can make the user stop what they are doing and relaunch the 'Add to External System' or LTI Selection Session flows.  The theory with setting this to 90 minutes is to offer a cushion to those users using the flows for extended periods, who may work with the integration over 60 minutes.

With these latest fixes to (a) put the auth token in the user session and (b) leverage the refresh_token flow, I think we can safely bring the cache ttl to match Blackboard's expiry.  I could be convinced that we can bring the size and the ttl lower, to say, 200 and 30 minutes, if folks are concerned with the resources to maintain that cache.

Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/impl/BlackboardRESTConnectorServiceImpl.java
cbeach47: fixed the do-while.  Time is a bit tight, so I need to forego gathering in the list logic, though I like the idea.

Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/impl/BlackboardRESTConnectorServiceImpl.java
```diff
@@ -548,47 +531,54 @@ public boolean supportsReverseSort() {
       } else {
         body = "";
       }
-      request.setBody(body);
-      if (body.length() > 0) {
-        request.addHeader("Content-Type", "application/json");
+
+      if ("grant_type=refresh_token".equals(data)) {
```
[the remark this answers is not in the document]
cbeach47: I pulled out the preparation bits and cleaned it up - quite a bit easier to read now, though it could likely use future work, including pagination of results.

Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/BlackboardRESTConnectorService.java
```diff
@@ -46,16 +47,16 @@ String getAuthorisationUrl(
    * The connector object will need to store an encrypted admin token in the DB. Use this method to
    * encrypt the one returned from Blackboard.
    *
-   * @param token
+   * @param data
    * @return
```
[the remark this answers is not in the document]
cbeach47: I don't have time for a complete overhaul of the Blackboard code atm, but I'll keep that in mind to not leave javadoc skeletons around :+1:

cbeach47:
@edalex-ian - thanks for the review

Good point about the JavaDoc, I'll keep that in mind.

I responded inline about the other comments - overall they made sense, just running up against some time constraints, so I had to prioritize.  I believe this PR is now in good shape to merge once it goes green.  

I'd like to merge this tomorrow (my Friday).  Please let me know if there are any blocking concerns.

Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/impl/BlackboardRESTConnectorServiceImpl.java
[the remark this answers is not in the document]
edalex-ian: hmmm, not sure that addresses my concern. It seems we're holding onto an expired token for an additional 30mins. Surely sending an expired token would force the refresh anyway from the Bb side. :thinking:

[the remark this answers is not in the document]
edalex-ian: Ah hold on, I now see you've address that. :+1:

edalex-ian (approved):
Thanks @cbeach47 for making some of the improvements especially with your time constraints. I reckon what you have is GTG. :+1:

PR #2594, opened 2020-12-10
Properly check whether thumbnail generation is timeout or not.

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

Since the Java method `waitFor` returns a boolean to indicate whether the process is finished or not, we should use this boolean rather than using `StreamReader.isFinished()`.  With this change, I no longer have errors like `kill and pgrep calls` and `Error generating thumbnail` locally.

In terms of why `StreamReader.isFinished()` is not working reliably, I think we need to dig into its `run`.

### DIFF
--- a/Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
+++ b/Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
@@ -89,7 +89,7 @@ public final class ExecUtils {
       } else {
         StringBuilder errorOutput = new StringBuilder();
         CharStreams.copy(new InputStreamReader(getChildPid.getErrorStream()), errorOutput);
-        LOGGER.debug("getChildPid function did not run properly.\n" + errorOutput);
+        LOGGER.warn("getChildPid function did not run properly.\n" + errorOutput);
       }
       getChildPid.destroy();
 
@@ -261,8 +261,10 @@ public final class ExecUtils {
       int pid = getPidOfProcess(proc).orElse(0);
       final StreamReader stdOut = cp.getSecond();
       final StreamReader stdErr = cp.getThird();
-      boolean isFinished = proc.waitFor(durationInSeconds, TimeUnit.SECONDS);
-      if (!isFinished) {
+      final boolean timeout = !proc.waitFor(durationInSeconds, TimeUnit.SECONDS);
+      if (timeout) {
+        // If the process is not terminated before the given timeout is reached,
+        // kill the process and throw an InterruptedException.
         String platform = determinePlatform();
         if (platform.equals(PLATFORM_LINUX) || platform.equals(PLATFORM_LINUX64)) {
           killLinuxProcessTree(pid);
--- a/Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
+++ b/Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
@@ -92,12 +92,16 @@ public final class ExecUtils {
         LOGGER.debug("getChildPid function did not run properly.\n" + errorOutput);
       }
       getChildPid.destroy();
-      // convert string to array of ints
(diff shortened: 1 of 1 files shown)


ChristianMurphy (approved):

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
```diff
@@ -257,8 +261,8 @@ public static ExecResult execWithTimeLimit(
       int pid = getPidOfProcess(proc).orElse(0);
       final StreamReader stdOut = cp.getSecond();
       final StreamReader stdErr = cp.getThird();
-      proc.waitFor(durationInSeconds, TimeUnit.SECONDS);
-      if (!stdErr.isFinished() || !stdOut.isFinished()) {
+      boolean isFinished = proc.waitFor(durationInSeconds, TimeUnit.SECONDS);
```
SammyIsConfused, line 264: This could probably be used directly in the if, rather being set to a `boolean isFinished` since it never gets referred to again after that.

SammyIsConfused (approved):
Looks great Penghai, thanks for fixing this up. 
Just one tweak and then I'd say this is good to go.

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
```diff
@@ -92,12 +92,16 @@ public static void killLinuxProcessTree(int pid) {
...
       getChildPid.destroy();
-      // convert string to array of ints
-      pids =
-          Optional.of(
-              Arrays.stream(childPid.toString().replaceAll("\n", " ").split(" "))
-                  .mapToInt(Integer::parseInt)
-                  .toArray());
+
+      String childPidInfo = childPid.toString();
+      if (!childPidInfo.isEmpty()) {
```
edalex-ian, line 97: Why would it be empty? Is it simply a normal situation, in that there were no child PIDs? Or it is always due to an error situation from going into the `else` block on line 89?

I'm just wondering if this needs a companion `else` block to either log the situation, or to throw an exception. :thinking: Or if it's just normal - i.e. there may be no child PIDs sometimes - then this is just a guard against this processing. Maybe a comment in this case would highlight the assumption. :thinking:

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
edalex-ian: I see this as aiding doco. I would say though, doco could be further assisted with something completely underlining the intent:

```java
final boolean timeout = !proc.waitFor(...);
if (timeout) {
  // ...
}
```

edalex-ian (approved):
GTG. Just offering a few things for you to ponder.

Platform/Plugins/com.tle.platform.common/src/com/tle/common/util/ExecUtils.java
SammyIsConfused: There may be no child PIDs sometimes - in fact most cases. PDF extraction causes ImageMagick to open Ghostscript which is the child PID - but for case in which ImageMagick can handle things fine on its own, such as most images, there will be no delegate process opened.

PenghaiZhang: I feel we need to throw an runtime exception in the `else` block on line 89 if `getChildPid` fails for some unknown reasons.

PenghaiZhang: Ah perhaps it's better to do a LOGGER.error rather than throwing an exception.

SammyIsConfused: I agree on the LOGGER.warn.

Issue #1204, opened 2019-08-30
Error occurs when trying to save item as draft if Save and Continue is used


PenghaiZhang:
**Describe the bug**
When saving an item as a draft, an error occurs.

**To Reproduce**
1. Begin contributing an item
2. Fill some fields and click 'save and continue'
3. Click 'Save' and then click 'Save draft'

**Expected behavior**
The page should display properly

**Screenshots**
![Screenshot from 2019-08-30 16-59-44](https://user-images.githubusercontent.com/47203811/63999942-03012000-cb48-11e9-91d7-96480f51f588.png)
![Screenshot from 2019-08-30 16-59-56](https://user-images.githubusercontent.com/47203811/63999953-072d3d80-cb48-11e9-8437-10631a6e9012.png)

**Stacktrace**
None.

**Platform:**

- OpenEquella Version: 2019.1 / 2019.2 / maybe 2018.2
- OS: Ubuntu
- Browser Chrome

**Additional context**
None

cathfitz:
@PenghaiZhang interestingly, it still saves the draft items

PR #2599, opened 2020-12-14
Ensure navigating to ItemSummary page if an Item is saved as a draft

PenghaiZhang:
#1204

##### Checklist

- [ ] tests are included

##### Description of change

Previously, if users `save and continue` an Item and later save the item as a draft, a 500 error will occur. The reason is New UI failed to stay in the Wizard (I haven't dug into why it failed). But what we want is the page should navigate to ItemSummary page instead of staying in the Wizard if user save an Item as a draft.

So this PR's changes can ensure the navigation happens.


edalex-ian (approved):
Without knowing the full reasons of the logic (like you), I also say this looks GTG. :wink:

Issue #2605, opened 2020-12-14
SQL Exception on Advanced Searches in new UI


cbeach47:
**Describe the bug**
Clicking on an advanced search in the new UI fails with a SQLException.  Only happens in the new UI

**To Reproduce**

1. Turn on the new UI and the new search UI
2. Navigate to the search UI
3. click on an advanced search
4. note failure

**Expected behavior**
The advanced search widgets to show up.

**Screenshots**
N/A

**Stacktrace**
```
<tr><td>14:51:31,418</td><td>9088395d-6e66-4483-be06-b4ee75064d47</td><td><font color="#993300"><strong>ERROR</strong></font></td><td>RestEasyExceptionMapper</td>
<td>REST API error</td></tr>
<tr><td bgcolor="#993300" style="color:White; font-size : xx-small;" colspan="5">com.google.common.util.concurrent.UncheckedExecutionException: java.sql.SQLException: Connection is closed
<br>&nbsp;&nbsp;&nbsp;&nbsp;	at com.google.common.cache.LocalCache$LocalLoadingCache.getUnchecked(LocalCache.java:4806)
<br>&nbsp;&nbsp;&nbsp;&nbsp;	at com.tle.web.sections.registry.TreeRegistry.getTreeForPath(TreeRegistry.java:150)
<br>&nbsp;&nbsp;&nbsp;&nbsp;	at com.tle.web.sections.registry.TreeRegistry.getTreeForPath(TreeRegistry.java:93)
<br>&nbsp;&nbsp;&nbsp;&nbsp;	at com.tle.web.api.LegacyContentApi.withTreePath(LegacyContentApi.scala:311)
<br>&nbsp;&nbsp;&nbsp;&nbsp;	at com.tle.web.api.LegacyContentApi.submit(LegacyContentApi.scala:452)
<br>&nbsp;&nbsp;&nbsp;&nbsp;	at sun.reflect.GeneratedMethodAccessor569.invoke(Unknown Source)
<br>&nbsp;&nbsp;&nbsp;&nbsp;	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
<br>&nbsp;&nbsp;&nbsp;&nbsp;	at java.lang.reflect.Method.invoke(Method.java:498)
[...]

PR #2609, opened 2020-12-15
Fix the 500 error when navigating to Advance search from new search UI

PenghaiZhang:
#2605 

##### Checklist

- [ ] tests are included

##### Description of change

So I can reproduce this error by logging in as a normal user such as AutoTest. And this PR's changes can fix this issue although I am not entirely sure what is the matter with ACL for this bug. If I revoke ACL `SEARCH_POWER_SEARCH` then I just simply can't see this advanced search from the list, not to mention to open the advanced search page.

There are a few things worth to mention here.

1. if Advanced Search is open from new Search then the Section used is `RootAdvancedSearchSection`;

2. `RootAdvancedSearchSection` contains the node Section `SelectionSummarySection`;

3. `SelectionSummarySection` needs the UI Settings to determine what callback is used for the `Unselect all` button.

Step 3 is where the error gets caused.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/template/RenderNewTemplate.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/template/RenderNewTemplate.scala
@@ -104,8 +104,7 @@ object RenderNewTemplate {
     RunWithDB
       .executeIfInInstitution(UISettings.cachedUISettings)
       .getOrElse(UISettings.defaultSettings)
-      .newUI
-      .newSearch
+      .isNewSearchActive
   }
 
   // Check if New UI is being used, but there is no guarantee that New UI is enabled.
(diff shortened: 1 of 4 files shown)


edalex-ian (commented):
Hmmm, I'm a bit lost between your description and the diff - and even the initial issue.

I see the main change you've made is to make everything use `RenderNewTemplate`. I'm not sure why.

Further, you've had to modify `RenderNewTemplate.isNewSearchPageEnabled` to make this work. But again, what's the significance there.

Lastly, it seems this all refers are the Selection Section. Does that mean this is tied to Selection Sessions even though that has not been mentioned anywhere?

If you could fill in the blanks, that'd be great. Or, just grab me in the office tomorrow.

edalex-ian (approved):
Okay, after chatting with @PenghaiZhang , this all now makes sense. Merge on. :+1:

Issue #1750, opened 2020-05-21
Adding a favourite search gets 500 error returned


PenghaiZhang:
With New UI turned on, when add a favourite search through the dialog, pressing Enter to submit the request will result in 500 error whereas clicking the ADD button works fine.

1. Go to Search page, do a normal search and then click `Add search to favourites`
2. In the dialog,  press Enter to submit the request.
3. The page will display `500 : Internal Server Error`

**Expected behavior**
The message `Successfully added this search to your favourites` should be displayed.

**Screenshots**
![image](https://user-images.githubusercontent.com/47203811/82512113-7887a300-9b52-11ea-97f2-72f42cccc0f3.png)

**Stacktrace**
tbz6a12 00:31:17,091 ERROR [RestEasyExceptionMapper] REST API error
tbz6a12 java.lang.NullPointerException
tbz6a12 at com.tle.web.sections.js.generic.expression.FunctionCallExpression.getExpression(FunctionCallExpression.java:53)
tbz6a12 at com.tle.web.sections.js.generic.statement.FunctionCallStatement.getStatements(FunctionCallStatement.java:44)
tbz6a12 at com.tle.web.sections.js.generic.StatementHandler.getStatements(StatementHandler.java:74)
tbz6a12 at com.tle.web.sections.events.StandardRenderContext.bindHandler(StandardRenderContext.java:280)
tbz6a12 at com.tle.web.sections.SectionWriter.bindHandler(SectionWriter.java:499)
tbz6a12 at com.tle.web.sections.render.TagRenderer.processHandler(TagRenderer.java:348)
tbz6a12 at com.tle.web.sections.standard.renderers.AbstractComponentRenderer.processHandler(AbstractComponentRenderer.java:66)
tbz6a12 at com.tle.web.sections.standard.renderers.LinkRenderer.processHandler(LinkRenderer.java:88)
tbz6a12 at com.tle.web.sections.render.TagRenderer.processHandlers(TagRenderer.java:316)
tbz6a12 at com.tle.web.sections.render.TagRenderer.prepareAttributes(TagRenderer.java:297)
tbz6a12 at com.tle.web.sections.render.TagRenderer.realRender(TagRenderer.java:187)
tbz6a12 at com.tle.web.sections.standard.renderers.AbstractComponentRenderer.realRender(AbstractComponentRenderer.java:53)
[...]

PR #2611, opened 2020-12-16
Ensure forms submit properly in New UI

PenghaiZhang:
#1750 

##### Checklist

- [ ] tests are included

##### Description of change

In Old UI, pushing Enter on a textfield which is the only input of a form does not submit the form whereas it submits the form in New UI. This is because the Javascript controlling whether to submit a form does not get executed in New UI. 

I have tested two places: the dialog for adding a favourite search and dialog for adding a link attachment. Now let's see if we can have green builds.


edalex-ian (approved):
Looks good to me.

The only thing that stood on was the method name. Wondered whether it should be `addFormSubmitBinding`. But, I'll leave that up to you. :wink:

SammyIsConfused (approved):
Looks good to me!

PR #2614, opened 2020-12-17
Bugfix/search portlet do not support new search UI

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

Currently, when a search is triggered from quick search in Selection Session, the endpoint is not `selectoradd/searching.do`. This will result in showing the regular page in Selection Session. So we need to change the endpoint.

Also, in order to pass the search query to new search UI, we need to 
1. enable the feature of `converting query strings to search options` in Selection Session;
2. append the query to the URL.

When I was testing,  I noticed that `converting query strings to search options` is not quite compatible with `back to search page`. The issue is if back to a search URL which has query strings, the search option stored in browser history is ignored. 

So overall, this PR fixes above issue and make quick search work fine with new search UI.

[quick search in regular page](https://streamable.com/k6nqul)

[quick search in resource selector](https://streamable.com/0ckvtu)

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
@@ -56,7 +56,6 @@ import {
   SortOrder,
 } from "../modules/SearchSettingsModule";
 import {
-  isSelectionSessionOpen,
   prepareDraggable,
   buildSelectionSessionAdvancedSearchLink,
   buildSelectionSessionRemoteSearchLink,
@@ -158,10 +157,7 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
 
     Promise.all([
       getSearchSettingsFromServer(),
-      // Do not convert query string params to search options in Selection Session.
-      isSelectionSessionOpen()
-        ? Promise.resolve(undefined)
-        : queryStringParamsToSearchOptions(location),
+      queryStringParamsToSearchOptions(location),
     ])
       .then(([searchSettings, queryStringSearchOptions]) => {
         setSearchSettings(searchSettings);
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
@@ -158,7 +158,7 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
 
     Promise.all([
       getSearchSettingsFromServer(),
-      // If the search option already exists, don't do query string conversion.
+      // If the search options are available from browser history, ignore those in the query string.
       searchPageHistoryState
         ? Promise.resolve(undefined)
         : queryStringParamsToSearchOptions(location),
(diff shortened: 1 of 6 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -121,15 +120,16 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
     searchPageOptions: defaultSearchPageOptions,
     filterExpansion: false,
   };
-
+  const searchPageHistoryState: SearchPageHistoryState | undefined = history
+    .location.state as SearchPageHistoryState;
```
ChristianMurphy, line 124: do we know the state will be this shape for certain?
Would a runtype make sense here?

Source/Plugins/Core/com.equella.core/src/com/tle/web/searching/section/SearchQuerySection.java
```diff
@@ -623,6 +623,22 @@ private void doBasicSearch(SectionInfo info, String query) {
...
+        inSelectOrAdd
+            ? from.createForward(SearchSelectable.NEW_FORWARD_PATH)
+            : RootSearchSection.createForward(from);
+    String basicUrl = info.getPublicBookmark().getHref();
+
+    String queryString = ((InfoBookmark) info.getPublicBookmark()).getQuery();
+    if (queryString.isEmpty()) {
+      basicUrl = basicUrl + "?";
+    }
+    from.forwardToUrl(basicUrl + "&q=" + query);
```
edalex-ian, line 639: As discussed, probably want to rework to this rough idea - maybe need try/catch too.

```suggestion
    URL basicUrl = new URL(info.getPublicBookmark().getHref());
    
    from.forwardToURL(
      basicUrl.toString()
      + (basicUrl.getQuery().isEmpty ? "?" : "&")
      + "q=" + query));
```

edalex-ian (changes requested):
Looks good @PenghaiZhang , need to tidy up some comments, but there's also a bit of undesirable result in building up the query params where we could have `&?q=blah` so I've a suggestion for that.

Source/Plugins/Core/com.equella.core/src/com/tle/web/searching/section/SearchQuerySection.java
edalex-ian: Oops.... URL.getQuery() returns `null` if there is no query, so that should be `basicUrl.getQuery() == null`.

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
PenghaiZhang: Thanks Christian! 

Like your idea ! I will raise an Github issue for this and do it in next release :smile:

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -197,10 +197,19 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
...
+    // as the history data has higher priority. So the end result is wrong search options
+    // are displayed in the page.
+    if (isInitialSearch && searchPageOptions === defaultSearchPageOptions) {
+      return;
+    } else {
+      history.replace({
+        ...history.location,
+        state: { searchPageOptions, filterExpansion },
+      });
+    }
```
ChristianMurphy, line 212: with the return, else is not needed
```suggestion
      return;
    }
    history.replace({
      ...history.location,
      state: { searchPageOptions, filterExpansion },
    });
```

ChristianMurphy (approved):

Issue #2623, opened 2020-12-22
New Search UI shared search does not work


edalex-ian:
**Describe the bug**

When I click the share search button in the New Search UI and then copy the results into a new browser all it does it show an empty new search - none of my previous search inputs are utilised.

**To Reproduce**

Steps to reproduce the behaviour:

1. Ensure the New UI and New Search UI are enabled
2. Click on 'Search'
3. Enter a term in the query bar
4. Click the share button and wait for the notification that the search has been copied to your clipboard
5. Go to a new tab in your browser and paste into your address bar
6. Notice that you're simply at an empty search

**Expected behaviour**

Upon pasting into the address bar, you should arrive at a New Search UI page with the query bar having your query in there.

**Screenshots**

Video with narrative: 

https://user-images.githubusercontent.com/43919233/102855255-6f5c0d00-4478-11eb-9170-6fcc59dbdba3.mp4

**Platform:**

- OpenEquella Version: 2020.2.0
- OS: Debian
- Browser Firefox (and seen in Chrome by others)

**Additional context**

It would appear this is due to a last minute change to support Quick Search Portlet. Specifically [line 254 of SearchModule.ts](https://github.com/openequella/openEQUELLA/blob/753a3abb37dabc2c0615bcbdd48420d920590b41/Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts#L254). It's only looking for legacy search params, and if none are found it never gets down to line 261 where it does the conversion.

PR #2625, opened 2020-12-22
Refactor the logic of converting query params to SearchOptions.

PenghaiZhang:
#2623 

##### Checklist

- [ ] tests are included

##### Description of change

This PR refactors the logic of how to convert query params to SearchOptions.

The biggest change is: 
 If the new search page is open from `Quick Search`, the location pathname will be `searching.do` rather than `page/search`. This means we should call `newSearchQueryToSearchOptions` instead of `legacyQueryStringToSearchOptions`.

Here is a [video](https://streamable.com/tta47h) for the fix.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
@@ -247,18 +247,23 @@ export const generateCategoryWhereQuery = (
 export const queryStringParamsToSearchOptions = async (
   location: Location
 ): Promise<SearchOptions | undefined> => {
-  if (!location.search) return undefined;
-  const params = new URLSearchParams(location.search);
-
-  // If no query strings is of type LegacySearchParams, return undefined.
-  if (!Array.from(params.keys()).some((key) => LegacySearchParams.guard(key))) {
+  if (!location.search) {
     return undefined;
   }
-
-  if (location.pathname.endsWith("searching.do")) {
+  const params = new URLSearchParams(location.search);
+  const searchOptions = params.get("searchOptions");
+  // If SearchOptions is in query string, convert the string to a SearchOptions.
+  // Or, if query string contains 'LegacySearchParams', convert the string
+  // to a SearchOptions.
+  // Otherwise, return undefined.
+  if (searchOptions) {
+    return await newSearchQueryToSearchOptions(searchOptions);
+  } else if (
+    Array.from(params.keys()).some((key) => LegacySearchParams.guard(key))
+  ) {
     return await legacyQueryStringToSearchOptions(params);
   }
-  return await newSearchQueryToSearchOptions(params.get("searchOptions") ?? "");
+  return undefined;
 };
 
 /**
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
@@ -252,10 +252,11 @@ export const queryStringParamsToSearchOptions = async (
   }
   const params = new URLSearchParams(location.search);
   const searchOptions = params.get("searchOptions");
-  // If SearchOptions is in query string, convert the string to a SearchOptions.
-  // Or, if query string contains 'LegacySearchParams', convert the string
-  // to a SearchOptions.
-  // Otherwise, return undefined.
+
+  // If the query params contain `searchOptions` convert to `SearchOptions` with `newSearchQueryToSearchOptions`.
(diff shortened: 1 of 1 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -247,18 +247,23 @@ export const generateCategoryWhereQuery = (
...
-  if (!location.search) return undefined;
-  const params = new URLSearchParams(location.search);
-
-  // If no query strings is of type LegacySearchParams, return undefined.
-  if (!Array.from(params.keys()).some((key) => LegacySearchParams.guard(key))) {
+  if (!location.search) {
     return undefined;
   }
-
-  if (location.pathname.endsWith("searching.do")) {
```
PenghaiZhang, line 258: `location.pathname.endsWith("searching.do")` can't guarantee the shared link is generated from old UI now due to the support of `Quick Search`.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -247,18 +247,23 @@ export const generateCategoryWhereQuery = (
...
+  const params = new URLSearchParams(location.search);
+  const searchOptions = params.get("searchOptions");
```
edalex-ian, line 254: Seems rather verbose, would it be fine to have:

```suggestion
  const searchOptions = new URLSearchParams(location.search).get("searchOptions");
```

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -247,18 +247,23 @@ export const generateCategoryWhereQuery = (
...
+  // If SearchOptions is in query string, convert the string to a SearchOptions.
+  // Or, if query string contains 'LegacySearchParams', convert the string
+  // to a SearchOptions.
+  // Otherwise, return undefined.
```
edalex-ian, line 258:
```suggestion
  // If the query params contain `searchOptions` convert to `SearchOptions` with `newSearchQueryToSearchOptions`.
  // Else if the query params contain params from legacy `searching.do` (i.e. `LegacySearchParams`) then convert to
  //   `searchOptions` with `legacyQueryStringToSearchOptions`
  // For all else, return `undefined`.
```

edalex-ian (approved):
LGTM, just some possible wording changes.

One thing, your video didn't confirm the Old - > New search still works, and the Quick Search Portlet -> new still works. The code looks like they should, but just wanted to confirm you tested them.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
PenghaiZhang: Both `params`  and  'searchOptions' are needed in the below `if` block.

PenghaiZhang:
> LGTM, just some possible wording changes.
> 
> One thing, your video didn't confirm the Old - > New search still works, and the Quick Search Portlet -> new still works. The code looks like they should, but just wanted to confirm you tested them.

I have updated a new [video](https://streamable.com/cuviys) for `Old -> New`.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
edalex-ian: Aaaah yes, of course. I should've read ahead before commenting. ;)

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -247,18 +247,24 @@ export const generateCategoryWhereQuery = (
...
+  // For all else, return `undefined`.
+  if (searchOptions) {
+    return await newSearchQueryToSearchOptions(searchOptions);
+  } else if (
+    Array.from(params.keys()).some((key) => LegacySearchParams.guard(key))
+  ) {
     return await legacyQueryStringToSearchOptions(params);
   }
-  return await newSearchQueryToSearchOptions(params.get("searchOptions") ?? "");
+  return undefined;
```
ChristianMurphy, line 267: what does undefined do in this case?

PenghaiZhang: If this function returns undefined, it means new search page will use the default search options rather than the one stored in browser history or the one converted from URL query params.

You can see this logic in SearchPage.tsx line 168 to 181.

ChristianMurphy (approved):

PR #2633, opened 2021-01-04
Ensure the year range of Wizard Calendar starts from 1970

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

The start of Wizard Calendar year range is relatively 50 years back to current year, which means 1970 is no longer available on the pick list since 2021. Some tests therefore fail to select 1970. So this PR makes sure the year range start is fixed to 1970.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/resources/web/js/datepicker.js
+++ b/Source/Plugins/Core/com.equella.core/resources/web/js/datepicker.js
@@ -5,7 +5,7 @@
 
 $.datepicker.setDefaults({
 	showOn : 'both',
-	yearRange : "c-50:c+10",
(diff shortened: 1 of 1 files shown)


edalex-ian (approved):
LGTM. :+1: 

Yes, rather odd how that was initially done.

In some ways, you'd wonder if we should go back another 10 odd years to ensure it worked how it did from closer to the earlier days. But, for now at least fixing it on a year should be good.

Shame as discussed we'll also need to back port this to ensure previous builds/tests pass.

SammyIsConfused (approved):
So weird that it was like that in the first place!
LGTM. Thanks Penghai.

Issue #2262, opened 2020-09-11
React components get rendered twice


PenghaiZhang:
**Describe the bug**
If open a `tsx` page from a legacy page, the React component in this `tsx` page will get mounted twice, which resulted in React state update warings and duplicated API calls. This issue does not happen if users open a `tsx` page from another `tsx` page.

**To Reproduce**

1. With New UI on, open a legacy page such as `contribute`;
2. Open any `tsx` page such as `Search` or `Settings` ;
3. Observe the Browser's Console and Network tabs.

Please note that you WON''T see the warning if you open  `Settings`  because this page has special code to prevent this warning.  
Please search `cleanupTriggered` to get more details.

**Expected behavior**
The React component should be mounted once.

**Screenshots**
![image](https://user-images.githubusercontent.com/47203811/92839233-10bbd780-f423-11ea-8df5-f0e052b0e632.png)

![image](https://user-images.githubusercontent.com/47203811/92839291-1f09f380-f423-11ea-8e72-baa1acc7d242.png)

**Stacktrace**
If applicable, a stacktrace to help explain your problem.

**Platform:**

- OpenEquella Version: 2020.2
- OS: Ubuntu 18.04
- Browser Chrome

**Additional context**

PenghaiZhang:
The last `useEffect` in `LegacyPage` may cause this problem.

edalex-ian:
Did you say you already had a fix for this @PenghaiZhang ?

PenghaiZhang:
> Did you say you already had a fix for this @PenghaiZhang ?

Not yet. I worked out why this issues happens but I am not confident in my fix without enough knowledge of `full screen mode` in oEQ.

edalex-ian:
Okay, so we've been having a bit more fun with this one and we really need to get it sorted - but alas, with the release about to go out we have to just capture our findings. So here we go.

First, the effect which @PenghaiZhang references above is found on [line 150 of LegacyPage.tsx](https://github.com/openequella/openEQUELLA/blob/215f54953712e6a052d25916bd9f3516c5dcf843/Source/Plugins/Core/com.equella.core/js/tsrc/mainui/LegacyPage.tsx#L150) and looks like:

```typescript
    React.useEffect(
      () => () =>
        legacyContent.setLegacyContentProps((p) => ({ ...p, enabled: false })),
      []
    );
```

Another bit worth understanding is the relevant part of the component tree for two cases of interest.

[...]

PR #2638, opened 2021-01-05
Stop double renders when moving from legacy content to new UI

edalex-ian:
##### Checklist

##### Description of change

Here's a fun one to start the year. A bit of refactor around `LegacyContent` so as to address #2262.

Firstly, I want to highlight I did try and go for a minimal amount of change staying focused on the issue at hand. There are more things that seem obvious to change, however: 1. This is a critical area of complexity which we don't want to risk - and ideally we look forward to the day when it's removed; 2. With just this focused change, the delta is already big and complex enough.

Secondly, being a few months out from release, I hope that'll enable us to shake out any issues which a change like this could trigger.

Alright, now on to what I actually did - and some of the why.

The biggest fundamental change was to how the component tree now looks. Before we always had the `LegacyContent` component in play - probably what would now days be done as a Context. There are now two primary trees (key components only):

* New UI: `App` -> `IndexPage` -> `BrowserRouter` -> `Template` -> New UI Page component (e.g. `SearchPage`)
* Legacy UI: `App` -> `IndexPage` -> `BrowserRouter` -> `Template` -> **`LegacyContent`** -> [`LegacyForm`] -> `LegacyContentRenderer`

The significance here is the root, which used to always be (New and Legacy):

* `App` -> `IndexPage` -> **`LegacyContent`** -> `Template` -> `...`

As you can see, `LegacyContent` was always sitting there in the background. And then if you were viewing actual legacy UI then a `LegacyPage` would be instantiated to display the content. On removal it had an effect clean-up function that would reach all the way back to a state setter function for `LegacyContent` and could trigger full (seemingly unnecessary) re-renders.

With this change, `LegacyContent` is only instantiated if the route is determined to be for a legacy page. Further, it now is directly concerned with the display of the legacy content it retrieves - where as it used to rely on a `render` prop which accepted a function (declared up in `IndexPage`) to do the rendering. Which content it needs to retrieved is simply based on the `path` which is then passed in as a prop. Increasing alignment with the React practice of passing data down.

Doing so removed the need for `LegacyPage` which was really just a complicated mash of side effects to try and handle all this tangle of state of the content and clean-up when not in play. Instead the substance of displaying content could simply be moved into the return of `LegacyContent`.

One other thing I noted is an excessive use of React.memo in this area. I removed at least one, but arguably there's a couple more which could go.

Unfortunately there are no tests in this area - there never have been. But I have undertaken manual testing and things seem to still work - and maybe a bit faster smoother.

(I'll add some notes against interesting bits of the code.)


Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContent.tsx
```diff
@@ -139,18 +140,35 @@ function submitRequest(path: string, vals: StateData): Promise<SubmitResponse> {
...
+        ind < 0
+          ? { pathname: "/" + href, search: "" }
+          : {
+              pathname: "/" + href.substr(0, ind),
+              search: href.substr(ind),
+            };
+      setPreventNavigation(false);
+      redirect(redirloc);
+    }
+  };
```
edalex-ian, line 170: Move in from `LegacyPage`: https://github.com/openequella/openEQUELLA/pull/2638/files#diff-e6f896c91af5057bd41e024a75a45380e9a4f3724694a14fdd7fa07720cfaac8L105

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContent.tsx
```diff
@@ -271,23 +289,49 @@ export const LegacyContent = React.memo(function LegacyContent({
...
+    const { form } = content.html;
+    return content.noForm ? (
+      renderedContent
+    ) : (
+      <>
+        <LegacyForm state={content.state}>{renderedContent}</LegacyForm>
+        {form && HtmlParser(form)}
+      </>
+    );
+  };
```
edalex-ian, line 324: Logic which used to be the `render()` function passed in via the `render` prop from `IndexPage`.

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContent.tsx
```diff
@@ -271,23 +289,49 @@ export const LegacyContent = React.memo(function LegacyContent({
...
-  return render(enabled ? content : undefined);
+  return content ? (
+    renderPageContent(content)
+  ) : (
+    <Grid container direction="column" alignItems="center">
+      <Grid item>
+        <CircularProgress />
+      </Grid>
+    </Grid>
+  );
```
edalex-ian, line 334: As used to be provided by `LegacyPage`.

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/IndexPage.tsx
```diff
@@ -73,19 +66,6 @@ export default function IndexPage() {
...
-  ] = React.useState<LegacyContentProps>({
-    enabled: false,
-    pathname: "",
-    search: "",
-    locationKey: "",
-    userUpdated: refreshUser,
-    redirected: () => {},
-    onError: () => {},
-    render: () => <div />,
-  });
```
edalex-ian, line 88: The key source of trouble, and now no longer a concern of anything but `LegacyContent` and only ever properly passed in by props as and when needed.

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/IndexPage.tsx
```diff
@@ -263,41 +246,9 @@ export default function IndexPage() {
           setNavAwayCallback(undefined);
         }}
       />
-      <LegacyContent
-        {...legacyContentProps}
-        render={(content) => {
```
edalex-ian, line 268: The infamous `render` prop and its function.

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/Template.tsx
```diff
@@ -132,6 +133,30 @@ export function templateError(errorResponse: ErrorResponse): TemplateUpdate {
...
+  return {
+    title,
+    metaTags,
+    hideAppBar,
+    fullscreenMode: fullscreenMode as FullscreenMode,
+    menuMode: menuMode as MenuMode,
+    menuExtra,
+    children: undefined,
+  };
+}
```
edalex-ian, line 158: Brought in from `LegacyPage` at https://github.com/openequella/openEQUELLA/pull/2638/files#diff-e6f896c91af5057bd41e024a75a45380e9a4f3724694a14fdd7fa07720cfaac8L51

Made sense that this really should sit in `Template.tsx` alongside it's friends like `templateDefaults()` and `templateError()`.

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/routes.tsx
```diff
@@ -41,12 +40,11 @@ const ContentIndexSettings = React.lazy(
   () => import("../settings/Search/ContentIndexSettings")
 );
 
-export interface OEQRouteComponentProps<T = TemplateUpdateProps>
-  extends RouteComponentProps<T> {
```
edalex-ian, line 45: Extending `RouteComponentProps` seemed rather excessive (and unnecessary). Further, it actually caused issues in `mkRouteProps` in `IndexPage`. After review, it made sense to simplify this down to just what's needed.

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/routes.tsx
```diff
@@ -83,19 +81,21 @@ export const routes = {
   },
   SearchSettings: {
     path: "/page/searchsettings",
-    render: (p: OEQRouteComponentProps) => <SearchPageSettings {...p} />,
+    render: (p: BaseOEQRouteComponentProps) => <SearchPageSettings {...p} />,
```
edalex-ian, line 84: As a side note, I do wonder why these are all done with `render: ` rather than `component: `.

edalex-ian:
And there ends my annotations. :wink:

edalex-ian:
Okay, so a few things to fix here. In some cases (like ApiDocsTest) simple things like showing the error in native New UI now rather than simply wrapping legacy error pages.

Will switch to draft.

edalex-ian:
Okay, so I've further refactored `LegacyContent` to encapsulate the rendering and execution of legacy javascript into `LegacyContentRender`. Also other minor tweaks around how some errors are now displayed (and checked in tests) due to being in the New UI - rather than just wrapped legacy errors.

Still some issues remain, need an updated list of failures - as all but one test failed locally, but I think there may be others lurking.

edalex-ian:
Alright, getting closer:

```
[error] Failed: Total 614, Failed 4, Errors 0, Passed 609, Skipped 1
[error] Failed tests:
[error] 	com.tle.webtests.test.viewing.ItemSummaryTest
[error] 	com.tle.webtests.test.webservices.rest.NotificationsApiTest
[error] 	com.tle.webtests.test.admin.CustomLinksTest
[error] (OldTests / Test / test) sbt.TestsFailedException: Tests unsuccessful
```

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContent.tsx
```diff
@@ -220,9 +237,19 @@ export const LegacyContent = React.memo(function LegacyContent({
...
+      if (!form) {
+        onError(
+          generateFromError({
+            name: "stdSubmit Failure",
+            message:
+              "stdSubmit unable to proceed due to missing " + legacyFormId,
+          })
+        );
+      } else {
+        const vals = collectParams(form, command, [].slice.call(arguments, 1));
```
ChristianMurphy, line 250: destructuring may be a tad easier to read here, thoughts?

```suggestion
        const [_, ...args] = arguments;
        const vals = collectParams(form, command, args);
```

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContentRenderer.tsx
```diff
@@ -29,22 +31,57 @@ const useStyles = makeStyles((t) => ({
...
   const classes = useStyles();
 
-  const { body, crumbs, upperbody } = html;
+  // Effect responsible for the execution of the legacy scripts etc which historically were
+  // simply added at the end of the server-side HTML rendered. Some of the scripts would
+  // fail with `form` being `null` as the React rendering needs to be done before we kick
+  // these off.
+  React.useEffect(() => {
+    const isReady = (): boolean =>
+      (noForm ? document.getElementById(mainContentDivId) : getEqPageForm()) !==
```
ChristianMurphy, line 48: would `ref` be usable in place of `document.getElementById(mainContentDivId)`?
https://reactjs.org/docs/refs-and-the-dom.html

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/routes.tsx
ChristianMurphy: Good question, not sure.
Looks like `component` works with `react.lazy`, so I'd be fine either way https://reactjs.org/docs/code-splitting.html#route-based-code-splitting

ChristianMurphy (commented):
Thanks for taking to time to dig into this and refactor @edalex-ian!

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyForm.tsx
```diff
@@ -18,14 +18,25 @@
...
+export const legacyFormId = "eqpageForm";
+
+export const getEqPageForm = (): HTMLFormElement | null => {
+  const form: HTMLElement | null = document.getElementById(legacyFormId);
+  if (form && (form as HTMLFormElement).action !== undefined) {
+    return form as HTMLFormElement;
+  }
+
+  return null;
+};
```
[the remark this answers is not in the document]
edalex-ian: Maybe, I could try it. My main concern is that this code is ultimately code triggered by the legacy AJAX code via references to those functions placed in `window.EQ`. But worth a try.

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContent.tsx
edalex-ian: Ha, nice pick-up. When I relocated this code, the complexity did catch my attention, but I thought I'd leave as is in case it wasn't doing something 'special'. So I didn't stop and actually read what it was doing.

There is a bunch of code in here which still falls in that category, and I may come back to after all is working.

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContentRenderer.tsx
edalex-ian: Ah! Now this one I think I could definite get away with - we'll, other than needing to give consideration to the shared `getEqPageForm()`.

edalex-ian:
Alright, running through the build again overnight. In theory - unless my recent fixes have unintended effects - this should leave just one test to fix: `com.tle.webtests.test.admin.CustomLinksTest#testUploadIcon`

Unfortunately with that one I'm seeing some odd behaviour under Selenium vs. Manual.

edalex-ian:
Aww man, it broke an additional one (`RunReportTest`):

```
[error] Failed: Total 614, Failed 4, Errors 0, Passed 609, Skipped 1
[error] Failed tests:
[error] 	com.tle.webtests.test.reporting.RunReportTest
[error] 	com.tle.webtests.test.admin.CustomLinksTest
[error] (OldTests / Test / test) sbt.TestsFailedException: Tests unsuccessful
```

edalex-ian:
Okay, the failing of `CustomLinksTest` highlighted another fundamental issue and I believe that has now been solved - as well as further refinements/simplification of execution of scripts. So I'm now keen to see what the whole suite things of this.

But the `RunReportTest` is looking a little bit interesting. It seem (maybe) that the first time you go to a report historically something would trigger that report to be built - but that now is not happening. But if you go there more than once, or manually it does get built - so eventually (possibly just for a given day) - the tests then pass.

So, while awaiting this latest build I'll be trying to figure that out. :thinking:

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContentRenderer.tsx
edalex-ian: Actually ended up removing this code. This I over complicated it and in the end could just let React do it's thing seeing I'm in an effect. Hopefully that all works now. :crossed_fingers:

edalex-ian:
Gah, it's a total rats nest this LegacyContent stuf. Okay, so those changes have broken two different tests:

```
[error] Failed: Total 615, Failed 9, Errors 0, Passed 606
[error] Failed tests:
[error] 	com.tle.webtests.test.dashboard.PortalsTest
[error] 	com.tle.webtests.test.reporting.RunReportTest
[error] 	com.tle.webtests.test.myresources.favourites.ResourceSelectorFavouritesTest
[error] (OldTests / Test / test) sbt.TestsFailedException: Tests unsuccessful
```

And the `RunReportTest` is still problematic - and the solution is definitely not immediately obvious.

Work continues.

edalex-ian:
Thankfully the two new failing test suites seemed to have just been issues with the tests. So now just to try and figure out the `RunReportTest`.

edalex-ian:
Okay, starting to look promising:

```
[error] Failed: Total 615, Failed 4, Errors 0, Passed 610, Skipped 1
[error] Failed tests:
[error] 	com.tle.webtests.test.contribute.EquellaConnectorTest
[error] 	com.tle.webtests.test.reporting.RunReportTest
[error] (OldTests / Test / test) sbt.TestsFailedException: Tests unsuccessful
```

(`EquellaConnectorTest` is a known flakey.)

Source/Plugins/Core/com.equella.core/js/package.json
```diff
@@ -39,6 +39,7 @@
     "@openequella/rest-api-client": "file:../../../../../oeq-ts-rest-api",
     "@tinymce/tinymce-react": "3.6.1",
     "axios": "0.21.0",
+    "clsx": "1.1.1",
```
edalex-ian, line 42: Added to try and improve readability of classes in Template when swapping in and out fullscreen mode etc.

it's only tiny, at apparently only 228 bytes: https://www.npmjs.com/package/clsx

edalex-ian:
Okay, so issues with `com.tle.webtests.test.reporting.RunReportTest` highlighted there was still a significant architecture flaw. So I've now hit that and all this is now passing locally - as are some previously problematic ones.

So I'm expecting a green build... Let's see what reality things though.

:crossed_fingers: 

(If that happens, then it looks like I have some package.json conflicts etc to resolve.)

edalex-ian:
Hmm, for some reason the GHA build is not being triggered... Maybe due to the conflict. :disappointed:

edalex-ian:
Oooh, CodeBuild tells me all is well - just one known flakey:

```
com.tle.webtests.test.contribute.EquellaConnectorTest
```

But I really want to see a GHA result.

edalex-ian:
Gah! So GHA also says:

```
[error] Failed: Total 615, Failed 1, Errors 0, Passed 613, Skipped 1
[error] Failed tests:
[error] 	com.tle.webtests.test.contribute.EquellaConnectorTest
[error] (OldTests / Test / test) sbt.TestsFailedException: Tests unsuccessful
```

edalex-ian:
Ah okay, and now I see why the `EquellaConnectorTest` used to be flaky. I think the changes have speed things up enough though to make it always fail, so the test needed a fix similar to what I had done elsewhere in this PR.

So fingers crossed (:crossed_fingers:) that this then is the last.

edalex-ian:
Alrighty, finally ready for review. Sorry it's ended up rather large, but needs must in this case.

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContent.tsx
```diff
@@ -271,23 +302,36 @@ export const LegacyContent = React.memo(function LegacyContent({
...
-      updateStylesheets([]).then(deleteElements);
-    }
-  }, [enabled, pathname, search, locationKey]);
+    const params = new URLSearchParams(search);
+    const urlValues: { [index: string]: string[] } = {};
+    params.forEach((val, key) => {
+      const exVal = urlValues[key];
+      if (exVal) exVal.push(val);
+      else urlValues[key] = [val];
+    });
```
ChristianMurphy, line 311:
```suggestion
    const urlValues = params.reduce<Record<string, string[]>>(
      (accumulator, [key, value]) => ({
        ...accumulator,
        [key]: accumulator[key] ? accumulator[key].concat(value) : [value],
      }),
      {}
    );
```

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/Template.tsx
```diff
@@ -274,7 +300,7 @@ function useFullscreen({ fullscreenMode, hideAppBar }: useFullscreenProps) {
   return hideAppBar || modeIsFullscreen;
 }
 
-export const Template = React.memo(function Template({
+export const Template = function Template({
```
ChristianMurphy, line 303:
```suggestion
export function Template({
```

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/Template.tsx
```diff
@@ -549,25 +575,39 @@ export const Template = React.memo(function Template({
...
-            >
-              <MenuIcon />
-            </IconButton>
-          )}
-          {titleArea()}
-          {menuArea()}
-        </Toolbar>
-        {tabs}
-      </AppBar>
+  function renderError(error: ErrorResponse) {
```
ChristianMurphy, line 578:
```suggestion
  function RenderError({error}: {error: ErrorResponse}) {
```

would allow

```jsx
{errorResponse && renderError(errorResponse)}
```
to become
```jsx
{errorResponse && <RenderError error={errorResponse} />}
```

below

Source/Plugins/Core/com.equella.core/resources/web/scripts/cloud.js
```diff
@@ -1,15 +1,20 @@
-var Cloud = {
-  onSearch: function(cb, $span, searchingText) {
-    setTimeout(function() {
+const Cloud = {
+  onSearch: function (cb, $span, searchingText) {
+    if ($("#eqpageForm").size()) {
```
ChristianMurphy, line 3:
```suggestion
    if (document.getElementById("eqpageForm")) {
```

ChristianMurphy (approved):
Thanks @edalex-ian! :bow:

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/JQueryDiv.tsx
```diff
@@ -19,18 +19,27 @@ import * as React from "react";
...
+ * also used to support the various JQuery AJAX stuff partially due to the clean-up effect of
+ * emptying out the `div`.
+ *
+ * It would be nice if we could get away with a simpler:
+ *
+ * `<div {...withoutOthers} dangerouslySetInnerHTML={{ __html: html}}/>`
+ *
+ * However that does not work, using jQuery to set the content of the `div` and then especially
+ * clearing the `div` in the clean-up effect is required for some edge cases. Why, is not entirely
+ * clear.
```
SammyIsConfused, line 36:
```suggestion
 * Provides a means to create a `div` whose content is the raw HTML returned from the server.
 * Also used to support the various JQuery AJAX stuff partially due to the clean-up effect of
 * emptying out the `div`.
 *
 * It would be nice if we could get away with a simpler:
 *
 * `<div {...withoutOthers} dangerouslySetInnerHTML={{ __html: html}}/>`
 *
 * However that does not work, using jQuery to set the content of the `div` and then especially
 * clearing the `div` in the clean-up effect is required for some edge cases. Why, is not entirely
 * clear.
```

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContent.tsx
```diff
@@ -161,6 +178,10 @@ export const LegacyContent = React.memo(function LegacyContent({
     content: LegacyContentResponse,
     scrollTop: boolean
   ) {
+    // First clear the content. As odd as this may seem we need this to ensure
+    // legacy elements work correctly without React rendering/dom optimisations. *sad*
+    // - This mimicks the idea of where the web browser would've reloaded the page.
```
SammyIsConfused, line 183:
```suggestion
    // - This mimics the legacy functionality where the web browser would've reloaded the page.
```

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContentRenderer.tsx
```diff
@@ -29,22 +31,42 @@ const useStyles = makeStyles((t) => ({
...
   menuMode,
+  noForm,
   script,
+  state,
 }: PageContent) {
   const classes = useStyles();
 
-  const { body, crumbs, upperbody } = html;
+  // Effect responsible for the execution of the legacy scripts etc which historically were
+  // simply added at the end of the server-side HTML rendered.
```
SammyIsConfused, line 43:
```suggestion
  // simply added at the end of the server-side pre-rendered HTML.
```

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/IndexPage.tsx
```diff
@@ -193,7 +172,7 @@ export default function IndexPage() {
         currentParams.get("in")?.startsWith("P") ?? false;
       const advancedSearchRequested: boolean =
         currentPath.endsWith(oldSearchPagePath) && advancedSearchParamsPresent;
-      // TODO: Before we release 2020.2 this can be removed, as the 'newSearch' toggle wil be removed
+      // TODO: Before we release 2021.1 this can be removed, as the 'newSearch' toggle wil be removed
```
SammyIsConfused, line 175:
```suggestion
      // TODO: Before we release 2021.1 this can be removed, as the 'newSearch' toggle will be removed
```

autotest/Tests/src/main/java/com/tle/webtests/pageobject/settings/ScheduledTasksPage.java
```diff
@@ -71,4 +71,22 @@ public ScheduledTasksPage runSyncEquellaContent() {
   public ScheduledTasksPage runHarvesterProfiles() {
     return clickAndUpdate(runHarvesterProfilesLink);
   }
+
+  /**
+   * Used for clicking the simply links which simply trigger a server action, and then the page is
+   * (historically) re-rendered. However in New UI react detects there's nothing new in the DOM so
+   * doesn't change anything. So in that case we have to resort to sleepTime().
+   *
+   * @param element the link to click
```
SammyIsConfused, line 80:
```suggestion
   * Used for clicking links which simply trigger a server action, and then the page is
   * (historically) re-rendered. However in the New UI, React detects there's nothing new in the DOM and
   * doesn't change anything. So in that case, we have to resort to sleepTime() to make the test work.
   *
   * @param element the link to click
```

autotest/Tests/src/main/java/com/tle/webtests/pageobject/settings/ScheduledTasksPage.java
```diff
@@ -71,4 +71,22 @@ public ScheduledTasksPage runSyncEquellaContent() {
...
+   */
+  private void clickLink(WebElement element) {
+    if (context.getTestConfig().isNewUI()) {
+      // Due to react DOM rendering optimisations, there is nothing for us to detect so we have
```
SammyIsConfused, line 84:
```suggestion
      // Due to React DOM rendering optimisations, there is nothing for us to detect so we have
```

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/Template.tsx
```diff
@@ -588,30 +628,52 @@ export const Template = React.memo(function Template({
...
+
+  /**
+   * Defines the main layout of the page by structuring the above `layoutXyz` constants. This is
+   * done to enhance the readability of what the layout looks like, especially with regards to
+   * support for fullscreen mode.
+   *
+   * Originally fullscreen mode support was done with a simple ternary expression that would
+   * set the layout as `<main>children</main>` when in fullscreen, but otherwise build up a
+   * component tree similar to the below. However this meant `children` would be unmounted when
+   * changing modes triggering excessive re-rendering. This became an issue with `LegacyContent`
```
SammyIsConfused, line 657:
```suggestion
   * changing modes, triggering excessive re-rendering. This became an issue with `LegacyContent`
```

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/Template.tsx
```diff
@@ -588,30 +628,52 @@ export const Template = React.memo(function Template({
...
+    <div className={classes.footer}>{footer}</div>
+  );
+
+  // Simple wrapper for `clsx` for easy consideration of fullscreen mode.
+  const layoutClasses = (...classes: ClassValue[]): string =>
+    clsx(!fullScreen && classes);
+
+  /**
+   * Defines the main layout of the page by structuring the above `layoutXyz` constants. This is
+   * done to enhance the readability of what the layout looks like, especially with regards to
```
SammyIsConfused, line 651:
```suggestion
   * done to enhance the readability of the layout code, especially with regards to
```

SammyIsConfused (approved):
Looks good to me!
I focused on simple typos and grammar tweaks for this review as opposed to the actual code structure. Feel free to ignore my suggestions if you wish.

autotest/OldTests/src/test/java/com/tle/webtests/test/admin/ApidocsTest.java
```diff
@@ -79,10 +79,9 @@ public void testLoginWithAccess() {
   //    }
 
   private void assertAccessDenied(ErrorPage error) {
-
-    assertEquals(
-        error.getMainErrorMessage(testConfig.isNewUI()),
-        testConfig.isNewUI() ? "Forbidden" : "Access denied");
+    if (!testConfig.isNewUI()) {
```
PenghaiZhang, line 82: Is there any reason for skipping this assertion for new UI ?

PenghaiZhang (commented):
Thanks Ian! Very good improvements! 

As you have spent such a good amount of time on refactoring this, I guess you must have better understanding of things like `stdSubmit` and effect for `window["EQ"]`. Would you like to add some comments to briefly explain what they do ?

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/JQueryDiv.tsx
```diff
@@ -19,18 +19,27 @@ import * as React from "react";
...
+ */
+const JQueryDiv = React.memo(({ html, ...withoutOthers }: JQueryDivProps) => {
```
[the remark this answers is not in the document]
edalex-ian: Yes, I concur, but this was another case of just leaving as was.

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContent.tsx
```diff
@@ -100,16 +105,11 @@ export interface PageContent {
...
+export interface LegacyContentProps extends BaseOEQRouteComponentProps {
   pathname: string;
   search: string;
   locationKey?: string;
-  userUpdated: () => void;
-  redirected: (redir: { href: string; external: boolean }) => void;
-  onError: (cb: { error: ErrorResponse; fullScreen: boolean }) => void;
-
-  render(content: PageContent | undefined): React.ReactElement;
+  onError: (error: ErrorResponse) => void;
```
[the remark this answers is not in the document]
edalex-ian: This was from early on when I didn't fully understand what full-screen is. Now I most certainly do, I'll need to go back and revisit this.

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContent.tsx
```diff
@@ -161,6 +178,10 @@ export const LegacyContent = React.memo(function LegacyContent({
     content: LegacyContentResponse,
     scrollTop: boolean
   ) {
+    // First clear the content. As odd as this may seem we need this to ensure
+    // legacy elements work correctly without React rendering/dom optimisations. *sad*
+    // - This mimicks the idea of where the web browser would've reloaded the page.
+    setContent(undefined);
```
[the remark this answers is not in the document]
edalex-ian: Yes, I should revisit this I think. See if I can improve it.

But what happens if we don't do this you ask... I captured this in the matching commit: https://github.com/openequella/openEQUELLA/pull/2638/commits/c612d57907ed99ff61602141c80b3d9aaba99ddf

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContentRenderer.tsx
```diff
@@ -29,22 +31,42 @@ const useStyles = makeStyles((t) => ({
...
+  React.useEffect(() => {
+    // eslint-disable-next-line no-eval
+    if (script) window.eval(script);
```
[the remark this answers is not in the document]
edalex-ian: Yup, this code I tried various other approaches, but in the end I had to stick with that. And by "stick" this was the original implementation which used to be in JQueryDiv, but it made more sense to do the execution of related scripts etc in this component.

autotest/OldTests/src/test/java/com/tle/webtests/test/admin/ApidocsTest.java
edalex-ian: Yup, because with the updates the error is now properly displayed as a New UI (`Template`) error - not just a wrapped Legacy UI error page.

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/Template.tsx
edalex-ian: Ah yes, another oversight. Thank you.

edalex-ian:
@PenghaiZhang wrote:

> I guess you must have better understanding of things like `stdSubmit` and effect for `window["EQ"]`. Would you like to add some comments to briefly explain what they do ?

Sure thing! I definitely have been focused on these a bit.

The point of the whole `window["EQ"]` is to register a whole lot of callback (kinda) functions which are then used by the various legacy side JS files and server side generated JS for event handlers in the rendered HTML. So you will often see (in the callstacks) that the functions within `window["EQ"]` have been triggered by event handlers and/or anonymous functions in the stack with some reference to one of the legacy JS files.

One of the key things that needs to be done is to submit the hidden form within the legacy pages (which contain the current state) when triggering an oEQ AJAX 'event'. So this is exposed by the `window["EQ"].event` and `window["EQ"].eventnv` - 'nv' meaning no validate. Ultimately the return a callback that will then do an AJAX call which has the `event__` param and a command signifying the event to trigger (server side) and will also have all the state info pulled from the hidden form.

Hope that helps, if not, let me know and I'll further clarify.

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContentRenderer.tsx
```diff
@@ -29,22 +31,42 @@ const useStyles = makeStyles((t) => ({
...
+  React.useEffect(() => {
+    if (afterHtml) afterHtml();
+  }, [afterHtml]);
+
   const extraClass =
     !fullscreenMode && menuMode !== "HIDDEN" ? classes.noPadding : "";
+
   const mainContent = (
-    <div className={`content ${extraClass}`}>
+    <div id="oeqMainLegacyContent" className={`content ${extraClass}`}>
```
ChristianMurphy, line 57: the id `oeqMainLegacyContent` doesn't appear to be used?

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyForm.tsx
```diff
@@ -18,14 +18,25 @@
 import * as React from "react";
 import { ReactNode, Fragment } from "react";
 
+export const legacyFormId = "eqpageForm";
+
+export const getEqPageForm = (): HTMLFormElement | null => {
+  const form: HTMLElement | null = document.getElementById(legacyFormId);
+  if (form && (form as HTMLFormElement).action !== undefined) {
+    return form as HTMLFormElement;
+  }
```
ChristianMurphy, line 27: :thinking: it may be easier to avoid casting with `as` in here a few times using a user defined type guard https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/Template.tsx
```diff
@@ -549,25 +575,39 @@ export const Template = React.memo(function Template({
...
-        </Toolbar>
-        {tabs}
-      </AppBar>
+  function renderError(error: ErrorResponse) {
+    return (
+      <MessageInfo
+        open={errorOpen}
+        onClose={() => setErrorOpen(false)}
+        variant="error"
+        title={error.error_description ? error.error_description : error.error}
```
ChristianMurphy, line 584: either
```suggestion
        title={error.error_description ?? error.error}
```

or

```suggestion
        title={error?.error_description ?? error.error}
```

may work

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/Template.tsx
```diff
@@ -549,25 +575,39 @@ export const Template = React.memo(function Template({
...
+
+  const fullScreen = useFullscreen({ fullscreenMode, hideAppBar });
+
+  const layoutAppBar = !fullScreen && (
+    <AppBar className={classes.appBar}>
+      <Toolbar disableGutters>
+        {hasMenu && (
+          <IconButton
+            className={classes.navIconHide}
+            onClick={(_) => setNavMenuOpen(!navMenuOpen)}
```
ChristianMurphy, line 597:
```suggestion
            onClick={() => setNavMenuOpen(!navMenuOpen)}
```

ChristianMurphy (approved):

autotest/Tests/src/main/java/com/tle/webtests/pageobject/settings/ScheduledTasksPage.java
edalex-ian: No longer required, changes reverted.

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyForm.tsx
```diff
@@ -18,14 +18,25 @@
 import * as React from "react";
 import { ReactNode, Fragment } from "react";
 
+export const legacyFormId = "eqpageForm";
+
+export const getEqPageForm = (): HTMLFormElement | null => {
+  const form: HTMLElement | null = document.getElementById(legacyFormId);
```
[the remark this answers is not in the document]
edalex-ian: I had not, but now I have:

```typescript
export const getEqPageForm = (): HTMLFormElement | null =>
  document.querySelector<HTMLFormElement>(`#${legacyFormId}`);
```

Thanks @PenghaiZhang , it'll be in 46df1bf1e64db0c08f780fe712e264135bd230a9

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/IndexPage.tsx
```diff
@@ -203,13 +182,17 @@ export default function IndexPage() {
...
     };
 
-    const renderLegacyPage = (p: RouteComponentProps) => (
-      <LegacyPage
-        {...mkRouteProps(p)}
-        errorCallback={errorCallback}
-        legacyContent={{ content, setLegacyContentProps }}
-      />
-    );
+    const renderLegacyPage = (p: RouteComponentProps) => {
```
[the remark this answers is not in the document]
edalex-ian: Ah yes, thank you. Will be in commit 6f10d314dc49ec8ce7b29bc7fa20e8f86a2102c3

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContent.tsx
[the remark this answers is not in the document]
edalex-ian: Okay, rather than being lazy I brought in a new state flag `updatingContent` to control the switching between the content and the spinner. Quickly highlights that potentially this should head towards a reducer, but hell no in this context.

Change will be seen in commit: 2832404d8f916d5f266213d5dc32175b7b12a2f1

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyForm.tsx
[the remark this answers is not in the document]
edalex-ian: I've decide for now I'll let this be in the interest of trying to stay close to original code and minimising the cross-over of React into the AJAX legacy calls. Some attempts (like in `JQueryDiv`) have told me it's not worth the time - especially due to long term we still hope to see all this code gone.

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContent.tsx
edalex-ian: Yeah, nah, I'm not touching this bit. Why is it using the magical `arguments` variable anyway rather than just placing that in the function signature... Oh, probably because the callback function being returned is being called by the legacy AJAX code... Gah, I'm leaving this be on that premise alone.

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContent.tsx
```diff
@@ -139,18 +139,35 @@ function submitRequest(path: string, vals: StateData): Promise<SubmitResponse> {
...
+      const ind = href.indexOf("?");
+      const redirloc =
+        ind < 0
+          ? { pathname: "/" + href, search: "" }
+          : {
+              pathname: "/" + href.substr(0, ind),
+              search: href.substr(ind),
+            };
+      setPreventNavigation(false);
+      redirect(redirloc);
```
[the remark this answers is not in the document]
edalex-ian: This is the old code just relocated. Just to minimise changes in this PR - even though this change is super simple - I'm going to leave as is.

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContent.tsx
edalex-ian: Sam here, just relocated code. And as much as I love dropping in a reduce instead of a for loop, I will continue to restrain myself here in this PR.

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/Template.tsx
edalex-ian: Resolved in commit to be 0f2dd1c78c71bdd0767475d63b62e09b940071c4

Source/Plugins/Core/com.equella.core/resources/web/scripts/cloud.js
edalex-ian: Addressed in commit to be: 7ff156fb52c84233b33322cbe9c8acb6ab7eb105

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/Template.tsx
edalex-ian: Good plan. Actioned in commit ddc3a9769b2911ec53946999a4aac20c56343779

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyForm.tsx
edalex-ian: Reworked in 46df1bf1e64db0c08f780fe712e264135bd230a9

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/Template.tsx
edalex-ian: I'll leave as per is (existing code) for now, as I'm unclear as to if `error_description` is sometimes `""` which with the current code would then show the `error`.

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/Template.tsx
edalex-ian: Ah yup, some relocated code, but a quick tweak for the above is in commit to be ac1d9c053e57a12ebeec95d460ca2e661273b5fd

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContentRenderer.tsx
edalex-ian: Ah yes, thank you - left over from one of many different attempts. Fix will be found in future commit of 74573d9e3887ca3b0af114e428dabc4348ee6c73

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContent.tsx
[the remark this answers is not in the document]
edalex-ian: Okay, so I've gone back to the old code and am working through what it used to do. For reference, above where @PenghaiZhang  described what it used to do, this can be found in (now removed) LegacyPage.tsx @ https://github.com/openequella/openEQUELLA/blob/develop/Source/Plugins/Core/com.equella.core/js/tsrc/mainui/LegacyPage.tsx#L126

And the code looks like:

```typescript
    const onError = React.useCallback(
      (err: { error: ErrorResponse; fullScreen: boolean }) => {
        const { error, fullScreen } = err;
        if (fullScreen) {
          errorCallback(error);
        } else {
          updateTemplate(templateError(error));
        }
      },
      [updateTemplate, errorCallback]
    );
```

It's not immediately obvious why in full screen mode (i.e. No app bar or menu) only the error callback should be called; as opposed to otherwise (full screen `false` or `undefined` - the later being used explicitly in some calls) a whole `updateTemplate` should be triggered - which also includes setting full screen mode in some cases. :confused:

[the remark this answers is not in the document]
edalex-ian: Ahhh, rightoh, I think this makes sense. Keeping in mind that `errorCallback` looked like:

```typescript
  const errorCallback = React.useCallback((err) => {
    errorShowing.current = true;
    setTemplateProps((p) => ({ ...p, fullscreenMode: undefined }));
    setFullPageError(err);
  }, []);
```

We can then make the following statements about the previous `onError`:

- if in full screen mode
  - first swap back to non-full screen mode
  - then set the error to be displayed (via `ErrorPage`)
- else
  - leave full screen mode as is - assumed to be non-full screen mode
  - and simply update the error message in the `Template` which will trigger a re-render of the error (displayed in an `ErrorPage`)

That is, if there is an error, always display it within the template with AppBar and Menu displayed - probably so that you still have navigation options to move about and try something again.

Seems fair, I'll make sure the new code matches that idea.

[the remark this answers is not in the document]
edalex-ian: Ahh, seemed I missed a few nuances there. Previous comment updated.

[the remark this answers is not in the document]
edalex-ian: Gah! Everytime I think I've got it there's another nuance. But... Ultimately it seems the only place this is used is for calls to `submitCurrentForm` in `LegacyContent.tsx`. And specifically `window.EQ.postAjax` calls it with `fullscreen` `false` so as to obviously trigger nice toast error messages for AJAX calls, yet the main effect for page content has `fullscreen` as `true` so that when fetching a page (or resulting redirect) a fullscreen/page error is displayed.

Got it... I think...

edalex-ian:
Phew, finally all done and another green build to boot.

Thank you all for taking the time to review, I know it was a bit of a beasty (sorry about that). I'll merge it in now so that we can start doing some hands on (manual) testing.

Thanks. :bow:

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/routes.tsx
edalex-ian: Hmm, okay, might do as a quick refactor on the side after the merge.

PR #2639, opened 2021-01-08
EBP: Add max-height to folderlist in selection session

SammyIsConfused:
##### Checklist

##### Description of change
In a selection session, when the number of selection destinations exceeds 10, you can't scroll down far enough to see the lowest entries unless you scroll down to the bottom of the openEQUELLA search page. Adding max height to the list means that if it is too long it will have a scroll bar.

Before:
https://user-images.githubusercontent.com/24543345/103978903-aea49080-51d0-11eb-919d-5d0c6caf38aa.mp4

After:
https://user-images.githubusercontent.com/24543345/103978970-ced44f80-51d0-11eb-8325-9411d28ba6d0.mp4

### DIFF
--- a/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
+++ b/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
@@ -4958,7 +4958,7 @@ div#sssh_finishedButton {
   @include typography-subtitle-2;
   float: left;
   width: 100%;
-  max-height: 500px;
+  max-height: 60vh;
 }
 
 .foldertree {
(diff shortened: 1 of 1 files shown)


Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -4958,6 +4958,7 @@ div#sssh_finishedButton {
   @include typography-subtitle-2;
   float: left;
   width: 100%;
+  max-height: 500px;
```
edalex-ian, line 4961: Think we might need to use something relative to support different display resolutions/size.

edalex-ian (approved):
Cool, that makes more sense. And I assume it's still looks the same as in the video and is fine at zoom etc?

SammyIsConfused:
> Cool, that makes more sense. And I assume it's still looks the same as in the video and is fine at zoom etc?

Yup! That's correct.

PR #2640, opened 2021-01-12
Make sure that the Collection Selector doesn't overlap the Selection Panel in a Selection Session

SammyIsConfused:
##### Checklist

##### Description of change
When in a selection session (as the selection panel takes up room) or on a small screen (there's less room to begin with), the Refine Search Panel has less width. So the collection selector when getting populated with chips can grow too big and force the panel to widen which can cause overlap with the selection panel or the side of the screen. 

This is addressed in this PR by changing the ratio of search panel width to refine panel width from 9:3 to 8:4. 

This change also limits the width of the chips so they can't take up more than 50 percent of the collection selector's total width.

In order to not lose information with the now truncated collection names, I have added in a tooltip for these chips which displays the full name.

Before: 
https://user-images.githubusercontent.com/24543345/104384442-942e3680-5585-11eb-99b9-f65bd1247fde.mp4

After:
https://user-images.githubusercontent.com/24543345/104986406-14153e80-5a67-11eb-95ff-47af1b393312.mp4

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/search/components/CollectionSelector.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/search/components/CollectionSelector.tsx
@@ -26,7 +26,6 @@ import {
   Collection,
   collectionListSummary,
 } from "../../modules/CollectionsModule";
-import { isSelectionSessionOpen } from "../../modules/LegacySelectionSessionModule";
 import { languageStrings } from "../../util/langstrings";
 import useError from "../../util/useError";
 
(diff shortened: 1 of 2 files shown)


ChristianMurphy (approved):

edalex-ian:
Interesting fix here @SammyIsConfused . It actually highlighted to me that the issue is not really Selection Session specific.

If a user is say on a lower resolution - even not that low (say 1200px wide) - they will have issues. Example:

https://user-images.githubusercontent.com/43919233/104658670-71cc2280-5717-11eb-991e-690dbdf3793b.mp4

So I think the solution we need is general, therefore:

1. You possibly don't need to check for `isSelectionSession`; and
2. The hardcoded width may more so want to be based on the initial rendered width of the control.

In some ways, I'm kind of surprised the that Material UI components don't have something built in to support this. :thinking:

SammyIsConfused:
Ready for review now @edalex-ian :+1: Settled on adjusting the width ratio of the search and refine panels, and limiting the size of the now tooltipped chips.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/CollectionSelector.tsx
```diff
@@ -67,38 +67,54 @@ export const CollectionSelector = ({
...
-      getOptionSelected={(collection, selected) =>
-        selected.uuid === collection.uuid
-      }
-      renderOption={(collection, { selected }) => (
-        <>
-          <Checkbox
-            icon={<CheckBoxOutlineBlankIcon fontSize="small" />}
-            checkedIcon={<CheckBoxIcon fontSize="small" />}
-            checked={selected}
+    <div style={{ width: "100%" }}>
```
edalex-ian, line 70: Hmmm, we aim to do layout stuff with `Grid` or at least flexbox. Did you try that here?

edalex-ian (commented):
Looks good thanks @SammyIsConfused , but could you please confirm / explain the need for that `div` and the use of `maxWidth`?

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/CollectionSelector.tsx
SammyIsConfused: I did and it wasn't happy, but then I changed tack. This was accidentally left over. I've tested it without the div and it works the same. I have removed the div.

edalex-ian (approved):
LGTM! :+1:

ChristianMurphy (approved):

PR #2687, opened 2021-01-21
refactor(front-end): Use 'component' for routes

edalex-ian:
##### Checklist

##### Description of change

Previous use of render was unnecessarily verbose. Briefly discussed in #2638 .

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/mainui/routes.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/mainui/routes.tsx
@@ -81,21 +81,19 @@ export const routes = {
   },
   SearchSettings: {
     path: "/page/searchsettings",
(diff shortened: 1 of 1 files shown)


ChristianMurphy (approved):
LGTM :+1: 
The only gotcha with `component` is

> When you use component (instead of render or children, below) the router uses React.createElement to create a new React element from the given component. That means if you provide an inline function to the component prop, you would create a new component every render. This results in the existing component unmounting and the new component mounting instead of just updating the existing component. When using an inline function for inline rendering, use the render or the children prop (below).

https://reactrouter.com/web/api/Route/component
which shouldn't be an issue here.

edalex-ian:
Yeah, it's an interesting one, as this is only an abstraction of a later `render` at https://github.com/openequella/openEQUELLA/blob/develop/Source/Plugins/Core/com.equella.core/js/tsrc/mainui/IndexPage.tsx#L124

But anyway, it's slightly tidier. :man_shrugging:

PR #2699, opened 2021-01-25
Further refactor to address react-hooks/exhaustive-deps

edalex-ian:
##### Checklist

##### Description of change

Alright, yet another batch of addressing `react-hooks/exhaustive-deps` ESLint warnings. This leaves us with only 5 which are all focused around `SearchPage` (including `DateModifiedSelector` as that needs `SearchPage` fixed first).

Components addressed:

- `SearchPageSettings`
- `ContentIndexSettings`
- `SchemaSelector`
- `SchemaNodeSelector`
- `SearchResult`
- `IndexPage`
- `Template`
- `LegacyContent`

In `LegacyContent` I took a pragmatic approach and resorted to suppressing the 2/3 warnings. This is because the amount of effort was not justified when: 1. we plan to get rid of this long term; and 2. a ton of effort has just been expended on it to get it in better shape. (It feels like it would require treatment similar to what will be needed for `SearchPage`.)

I did observe other oddities - like error handling/wrapping in `SearchPageSettings`/`ContentIndexSettings` however these are planned to be addressed later. So overall, I stayed pretty close to task of just refactoring to address hook dependency issues - even when moving code around I tried to mostly leave as is - e.g. code in the new `MainMenu.tsx`.

While this is up for review, I'll now branch off branch and start work on `SearchPage`. I believe it's state and complexity is well past the point where a reducer should be employed, and now that we're at the start of the release it should be a good time to try. This will also ease the management of the dependency and attempted flow of effects. If successful, then that'll bring us to 0 warnings and we can swap this over to an error on failure.

That work will be in a future PR - combing here I think would've made things unmanageable.

I have done some manual testing, as well as executing the core Jest test suite and things have looked good. But I look forward to seeing what CI thinks.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/mainui/IndexPage.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/mainui/IndexPage.tsx
@@ -34,7 +34,12 @@ import { getCurrentUserDetails } from "../modules/UserModule";
 import { basePath } from "./App";
 import ErrorPage from "./ErrorPage";
 import { defaultNavMessage, NavAwayDialog } from "./PreventNavigation";
-import { BaseOEQRouteComponentProps, routes } from "./routes";
+import {
+  BaseOEQRouteComponentProps,
+  isNewUIRoute,
+  OEQRouteNewUI,
+  routes,
+} from "./routes";
 import { Template, TemplateProps, TemplateUpdate } from "./Template";
 
 const SearchPage = React.lazy(() => import("../search/SearchPage"));
@@ -109,25 +114,27 @@ export default function IndexPage() {
 
   const newUIRoutes = React.useMemo(
     () =>
-      Object.keys(routes).map((key, ind) => {
-        const oeqRoute = routes[key];
-        return (
-          (oeqRoute.component || oeqRoute.render) && (
-            <Route
-              key={ind}
-              exact={oeqRoute.exact}
-              path={oeqRoute.path}
-              render={(p) => {
-                const oeqProps = mkRouteProps(p);
-                if (oeqRoute.component) {
-                  return <oeqRoute.component {...oeqProps} />;
-                }
-                return oeqRoute.render?.(oeqProps);
-              }}
-            />
-          )
-        );
-      }),
+      Object.keys(routes)
+        .map<OEQRouteNewUI | undefined>((name) => {
+          // @ts-ignore:  Element implicitly has an 'any' type because expression of type 'string' can't be used to index type 'Routes'
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          const maybeRoute: any = routes[name];
+          return isNewUIRoute(maybeRoute) ? maybeRoute : undefined;
+        })
+        .filter((maybeRoute): maybeRoute is OEQRouteNewUI => !!maybeRoute)
+        .map((oeqRoute: OEQRouteNewUI, ind) => (
+          <Route
+            key={ind}
+            path={oeqRoute.path}
+            render={(p) => {
+              const oeqProps = mkRouteProps(p);
+              if (oeqRoute.component) {
+                return <oeqRoute.component {...oeqProps} />;
+              }
+              return oeqRoute.render?.(oeqProps);
+            }}
+          />
+        )),
     [mkRouteProps]
   );
 
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/mainui/IndexPage.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/mainui/IndexPage.tsx
@@ -34,7 +34,7 @@ import { getCurrentUserDetails } from "../modules/UserModule";
 import { basePath } from "./App";
 import ErrorPage from "./ErrorPage";
 import { defaultNavMessage, NavAwayDialog } from "./PreventNavigation";
-import { BaseOEQRouteComponentProps, OEQRoute, routes } from "./routes";
+import { BaseOEQRouteComponentProps, routes } from "./routes";
 import { Template, TemplateProps, TemplateUpdate } from "./Template";
 
 const SearchPage = React.lazy(() => import("../search/SearchPage"));
@@ -58,7 +58,7 @@ export default function IndexPage() {
     getCurrentUserDetails().then(setCurrentUser);
   }, []);
 
-  React.useEffect(() => refreshUser(), []);
+  React.useEffect(() => refreshUser(), [refreshUser]);
 
   const [navAwayCallback, setNavAwayCallback] = React.useState<{
     message: string;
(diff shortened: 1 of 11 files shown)


edalex-ian:
Gah! We've lost the type checking we used to have down in the core module with `build`, and the `test` only validates types for source that is actually under test.

1. We need to address that, but
2. For now I need to enhance my IntelliJ run target

I'll post a comment once I've addressed the type errors.

edalex-ian:
Well there we go, that took quite a bit of an update to the routes and they're now rather explicit and verbose. Might be okay... Regardless, the type checks should now pass. So looking for a green build. :crossed_fingers:

edalex-ian:
Excellent, all green. Please review. :grin:

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/MainMenu.tsx
```diff
@@ -0,0 +1,97 @@
...
+} from "@material-ui/core";
+import * as OEQ from "@openequella/rest-api-client";
+import * as React from "react";
+import { Link } from "react-router-dom";
+import { useStyles } from "./Template";
+
+interface MainMenuProps {
+  menuGroups: Array<Array<OEQ.LegacyContent.MenuItem>>;
+  onClickNavItem: () => void;
+}
```
ChristianMurphy, line 36: tsdoc comments for attributes could be helpful

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/MainMenu.tsx
```diff
@@ -0,0 +1,97 @@
...
+        }
+      />
+    </ListItem>
+  );
+
+  return (
+    <div id="menulinks">
+      {menuGroups.map((group, ind) => (
+        <React.Fragment key={ind}>
+          {ind > 0 && <Divider />}
```
ChristianMurphy, line 89:
```suggestion
      {menuGroups.map((group, index) => (
        <React.Fragment key={ind}>
          {index > 0 && <Divider />}
```

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/MainMenu.tsx
edalex-ian: Ah yes, added in fe1cb8d

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/MainMenu.tsx
edalex-ian: Sure thing: 91502f5

ChristianMurphy (approved):

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/ContentIndexSettings.tsx
```diff
@@ -70,28 +73,28 @@ function ContentIndexSettings({ updateTemplate }: TemplateUpdateProps) {
...
+    }
+  }, [error, updateTemplate]);
 
   function handleSubmitButton() {
     saveSearchSettingsToServer(searchSettings)
       .then(() => setShowSuccess(true))
-      .catch((error: TemplateUpdate) => handleError(error))
-      .finally(() => getSettings());
+      .catch((error: TemplateUpdate) => setError(() => error))
+      .finally(() => setLoadSettings(!loadSettings));
```
PenghaiZhang, line 97: I have two things unclear here. 
1. I can't remember why we need to get the settings again after saving.  I notice in other search setting pages we do similar things. Is it just to ensure UI is consistent with the configurations ? 

2. The effect depending on `loadSettings` will get settings from server regardless whether `loadSettings` is true or false. I image when it is false it indicates that there is no need to load settings.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/SearchPageSettings.tsx
```diff
@@ -89,22 +91,25 @@ function SearchPageSettings({ updateTemplate }: TemplateUpdateProps) {
           setInitialCloudSettings(settings);
         })
       )
-      .catch((error) => handleError(error));
-  }
+      .catch((error) => setError(() => error));
```
PenghaiZhang, line 94:
```suggestion
      .catch((error: TemplateUpdate) => setError(() => error));
```

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/IndexPage.tsx
```diff
@@ -129,10 +134,9 @@ export default function IndexPage() {
               return oeqRoute.render?.(oeqProps);
             }}
           />
-        )
-      );
-    });
-  }, [refreshUser]);
+        )),
+    [mkRouteProps]
```
PenghaiZhang, line 138: Looks like we can move `mkRouteProps` into this hook ?

PenghaiZhang (commented):
Thanks Ian! Looks good to me! 
I have left some questions.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/ContentIndexSettings.tsx
```diff
@@ -70,28 +73,28 @@ function ContentIndexSettings({ updateTemplate }: TemplateUpdateProps) {
...
-  function getSettings() {
+  React.useEffect(() => {
     getSearchSettingsFromServer()
       .then((settings: SearchSettings) => {
         setSearchSettings(settings);
         setInitialSearchSettings(settings);
       })
-      .catch((error) => handleError(error));
-  }
+      .catch((error) => setError(() => error));
```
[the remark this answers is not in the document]
edalex-ian: Yup, you are correct - right on the money. Good observation.

`getSearchSettingsFromServer` and related need refactoring, but I left as is here with the knowledge that we have that planned. (That is, the refactoring of remaining Axios calls out into the REST module.) Once that occurs we'll be more consistent and can use the `useError` hook as I did with the others.

Also, I'm still hoping that we improve that even further in the future with a error dedicated Context and ErrorBoundary. So all this is baby steps, with a longer term vision.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/ContentIndexSettings.tsx
edalex-ian: 1. I believe this is done to ensure we accurately represent what is on the server - debatable as to whether needed, but left things like this as is for this PR; and
2. Keep in mind here `loadSettings` is more a toggle switch - the idea simply being that it's value is changed when the settings should be loaded (hence the use of `setLoadSettings(!loadSettings)` rather than `setLoadSettings(true)` or `setLoadSettings(false)`). So values of `true` or `false` are meaningless - only that they changed is meaningful. On initial effect creation settings will be loaded, and then subsequent any time the value is changed (via calls such as `setLoadSettings(!loadSettings)`) the settings get reloaded.

Source/Plugins/Core/com.equella.core/js/tsrc/mainui/IndexPage.tsx
edalex-ian: Unfortunately not, used twice down in `routeSwitch`.

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/SearchPageSettings.tsx
edalex-ian: Could do, but if you don't mind terribly I might leave this as is for now - in lieu of it being reworked when we refactor the Axios side of it (not to mention the general error handling).

Source/Plugins/Core/com.equella.core/js/tsrc/settings/Search/ContentIndexSettings.tsx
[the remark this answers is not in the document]
edalex-ian: > 

Could do, but if you don't mind terribly I might leave this as is for now - in lieu of it being reworked when we refactor the Axios side of it (not to mention the general error handling).

PenghaiZhang (approved):

PenghaiZhang:
Just restarted Codebuild due to `browser may die`.

edalex-ian:
> Just restarted Codebuild due to `browser may die`.

Thank you.

Interesting, I'd already restarted it once due to that. Grrr, good ol' CodeBuild.

PR #2705, opened 2021-01-27
Update 'SearchOptionsRunTypes' for  'lastModifiedDateRange' 

PenghaiZhang:
#2623 

##### Checklist

- [ ] tests are included

##### Description of change

At the moment,  the Runtypes used for 'lastModifiedDateRange'  is `Record`, which means if either the start or end of a date range is undefined the Runtypes guard for `DehydratedSearchOptionsRunTypes` will fail. As a result, the feature of shared search does not work.

So in this PR I change the Runtypes used for 'lastModifiedDateRange' to `Partial`.


edalex-ian (commented):
Looks good @PenghaiZhang . But... Seeing this was to address a bug that was only found through manual testing, does that mean we're missing a test? Or is it too hard to write a test for?

PenghaiZhang:
> Looks good @PenghaiZhang . But... Seeing this was to address a bug that was only found through manual testing, does that mean we're missing a test? Or is it too hard to write a test for?

I think we have one test for this feature but the test is too simple. I will try to rewrite the test.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/modules/SearchModule.test.ts
```diff
@@ -144,6 +147,65 @@ describe("SearchModule", () => {
       expect(convertedParamsPromise).toEqual(allSearchOptions);
     });
 
+    const emptyStartDateQueryString =
```
edalex-ian, line 150: Maybe I'm confused, but shouldn't this be called `emptyEndDateQueryString` and the one below `emptyStartDateQueryString`?

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/modules/SearchModule.test.ts
```diff
@@ -144,6 +147,65 @@ describe("SearchModule", () => {
...
+      '{"lastModifiedDateRange":{"start":"2020-05-26T03:24:00.889Z"}}';
+    const expectedEmptyStartDateSearchOptions: SearchOptions = {
```
edalex-ian, line 152: Similar to above comment.

edalex-ian (commented):
Ummm, looks good but did you get the variable names around the wrong way? Or have I misread?

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/modules/SearchModule.test.ts
PenghaiZhang: My bad. I will change the names.

edalex-ian (approved):

PR #2706, opened 2021-01-28
Move the Axios usage for 'heartbeat' to REST module.

PenghaiZhang:
##### Checklist

- [x] tests are included

##### Description of change

Refactor the Axios usage in 'util/heartbeat.ts' to REST Module.


Source/Plugins/Core/com.equella.core/js/tsrc/modules/Heartbeat.ts
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Check oEQ server status by sending a heartbeat request.
+ * If the result is not equal to "OK", throw an error.
+ */
+export function checkHeartbeat() {
+  OEQ.Heartbeat.checkHeartbeat(API_BASE_URL)
+    .then((result: string) => {
+      if (result !== "OK") {
+        throw new Error(
```
ChristianMurphy, line 29: Since this appears to be a form of type check would it make sense to throw a `TypeError`?
```suggestion
        throw new TypeError(
```

oeq-ts-rest-api/src/Heartbeat.ts
```diff
@@ -0,0 +1,30 @@
...
+
+const HEART_BEAT_API_PATH = '/status/heartbeat';
+
+/**
+ * Retrieve oEQ server status by sending a heartbeat request.
+ * @param apiBasePath Base URI to the oEQ institution and API
+ */
+export const checkHeartbeat = (apiBasePath: string): Promise<string> =>
+  GET<string>(apiBasePath + HEART_BEAT_API_PATH, (data): data is string =>
+    is<string>(data)
```
ChristianMurphy, line 29: Above it appears that is should specifically be the literal `"OK"`, does `typescript-is` support checking literal types? Would it make sense to check if the value is `"OK"` at this level?

ChristianMurphy (approved):

Source/Plugins/Core/com.equella.core/js/tsrc/modules/Heartbeat.ts
edalex-ian: Hmmmm, this is not so much validating the type, it's more so validating a business layer response. Ideally it could probably return something like `OK` when all in order, or `UNAVAILABLE` if server is undergoing some kind of maintenance mode etc.

But what is also interesting is that currently the server (if it responds at all) will only ever respond with `OK` - so we could even choose to skip this for now.

oeq-ts-rest-api/src/Heartbeat.ts
edalex-ian: It would be a good idea to check it is the correct acceptable literals. Considering there is currently only the one (`OK`) we could probably skip `typescript-is` on this one (and just do a string comparison). However, nice to have the consistency if it works.

oeq-ts-rest-api/test/Heartbeat.test.ts
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import * as OEQ from '../src';
+import { checkHeartbeat } from '../src/Heartbeat';
+import * as TC from './TestConfig';
+
+beforeAll(() => OEQ.Auth.login(TC.API_PATH, TC.USERNAME, TC.PASSWORD));
+
+afterAll(() => OEQ.Auth.logout(TC.API_PATH, true));
```
edalex-ian, line 24: It's actually an unsecured endpoint, so can probably remove these.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/Heartbeat.ts
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+import { API_BASE_URL } from "../AppConfig";
+import * as OEQ from "@openequella/rest-api-client";
+
+/**
+ * Check oEQ server status by sending a heartbeat request.
+ * If the result is not equal to "OK", throw an error.
+ */
+export function checkHeartbeat() {
```
edalex-ian, line 25: For consistency of new stuff:

```suggestion
export const checkHeartbeat = (): void => {
```

edalex-ian (changes requested):
Looks good @PenghaiZhang , but if you could just make a coupla tweaks that'd be handy:

* Use arrow function syntax;
* Don't worry about checking for `OK` in the module layer; and
* In the REST module validate the allowed string literal.

(That may also mean tweaking the test - as it won't need to validate `OK` is returned, just that no exception is thrown.)

oeq-ts-rest-api/src/Heartbeat.ts
PenghaiZhang: I believe we can do `is<"OK">`

edalex-ian (approved):
GTG! :+1:

ChristianMurphy (approved):

PR #2713, opened 2021-02-02
Refactor of SearchPage to useReducer to address last set of ESLint react-hooks/exhaustive-deps

edalex-ian:
##### Checklist

- [x] tests are included - covered by existing tests

##### Description of change

Well, there we have it. This PR brings the count of `react-hooks/exhaustive-deps` warnings to _zero_!! :tada: 

To achieve that I've done two things:

1. Refactored `SearchPage` to `useReducer` for it's main state relating to searching; and
2. Removed a superfluous effect from `DateRangeSelector`.

**SearchPage**

Towards the conclusion of 2020.2 the amount of state and effects the `SearchPage` was looking after had made the component's function somewhat opaque, however there wasn't time for significant refactor. This PR now does that refactor with a pivot to `useReducer`.

Doing so has improved readability of the code, and  made it more explicit as to when searches are occurring - rather than depending on the chain of effects and their dependencies. I think it has also addressed a few cases of extraneous API calls.

In doing this I have also tried to balance which state to still manage with `useState` (such as `SearchPageOptions`) vs what to put in the state managed with `useReducer`. I left `SearchPageOptions` in `useState` as they also govern the surrounding/general state, and I also left the `SearchSettings` in `useState` as these are really a one off query from the server at the start.

Mostly, this all works just like it did before. However, there is one UX change which is when a new search is occurring the old results are removed - i.e. the spinner is not just overlayed a faded out set of previous results, they all go and an empty list is shown behind the spinner. This comes about from explicit specification of the state where I saw there can only be result items if the current state is `success` - i.e. don't show results when in `failed` state or `initialising` state. Maybe this is too logical though, and when in `searching` state the previous results (if any) should be kept. I'll leave that open for debate.

**DateRangeSelector**

All that occurred here was to move what is in the effect into an explicit handler. All still works as it used to, but arguably a bit better - as the missing effect dependency was indeed indicating an issue.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/components/DateRangeSelector.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/components/DateRangeSelector.tsx
@@ -15,8 +15,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import * as React from "react";
-import { makeStyles } from "@material-ui/core/styles";
+import LuxonUtils from "@date-io/luxon";
 import {
   FormControl,
   Grid,
@@ -24,16 +23,17 @@ import {
   MenuItem,
   Select,
 } from "@material-ui/core";
+import { makeStyles } from "@material-ui/core/styles";
 import {
   KeyboardDatePicker,
   MuiPickersUtilsProvider,
 } from "@material-ui/pickers";
-import type { DateRange } from "../modules/SearchModule";
-import SettingsToggleSwitch from "./SettingsToggleSwitch";
-import { ReactNode, useEffect, useRef, useState } from "react";
-import { languageStrings } from "../util/langstrings";
-import LuxonUtils from "@date-io/luxon";
 import { DateTime } from "luxon";
+import * as React from "react";
+import { ReactNode, useEffect, useState } from "react";
+import type { DateRange } from "../modules/SearchModule";
+import { languageStrings } from "../util/langstrings";
+import SettingsToggleSwitch from "./SettingsToggleSwitch";
 
 interface DatePickerProps {
   /**
@@ -128,18 +128,13 @@ export const DateRangeSelector = ({
   );
   const [showCalenderIcon, setShowCalenderIcon] = useState<boolean>(false);
 
-  const isInitialRender = useRef(true);
+  // Only when prop dateRange is cleared, clear stateDateRange as well.
   useEffect(() => {
-    if (isInitialRender.current) {
-      isInitialRender.current = false;
-      return;
-    }
-    // When state is cleared, do not proceed to avoid duplicated search.
-    if (!stateDateRange) {
-      return;
-    }
-    const start = stateDateRange?.start;
-    const end = stateDateRange?.end;
+    setStateDateRange(dateRange);
+  }, [dateRange]);
+
+  const handleDateRangeChange = (range: DateRange): void => {
+    const { start, end } = range;
     const isStartValid = start && DateTime.fromJSDate(start).isValid;
     const isEndValid = end && DateTime.fromJSDate(end).isValid;
 
@@ -154,15 +149,9 @@ export const DateRangeSelector = ({
       start && end ? isStartValid && isEndValid && start <= end : false;
     // Call onDateRangeChange for above four cases.
     if (openStart || openEnd || openRange || closedRange) {
-      onDateRangeChange(stateDateRange);
+      onDateRangeChange(range);
     }
-  }, [stateDateRange]);
-
-  // Only when prop dateRange is cleared, clear stateDateRange as well.
-  useEffect(() => {
-    setStateDateRange(dateRange);
-  }, [dateRange]);
-
+  };
   /**
    * Provide labels and values for options of pre-defined date ranges.
    */
@@ -238,6 +227,7 @@ export const DateRangeSelector = ({
   const handleQuickDateOptionChange = (option: string): void => {
     const dateRange = dateOptionToDateRangeConverter(option);
     setStateDateRange(dateRange);
+    handleDateRangeChange(dateRange);
   };
 
   const quickOptionSelector: ReactNode = (
@@ -304,12 +294,14 @@ export const DateRangeSelector = ({
             label={label}
             // When value is undefined use null instead so nothing is displayed in the TextField.
             value={value ?? null}
-            onChange={(newDate, _) =>
-              setStateDateRange({
+            onChange={(newDate, _) => {
+              const newRange = {
                 ...stateDateRange,
                 [field]: newDate?.toJSDate(),
-              })
-            }
+              };
+              setStateDateRange(newRange);
+              handleDateRangeChange(newRange);
+            }}
           />
         </Grid>
       );
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
@@ -290,7 +290,7 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
 
   // In Selection Session, once a new search result is returned, make each
   // new search result Item draggable. Could probably merge into 'searching'
-  // effect, however this is only required while selections sessions still
+  // effect, however this is only required while selection sessions still
   // involve legacy content.
   useEffect(() => {
     if (state.status === "success" && isSelectionSessionInStructured()) {
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
@@ -79,7 +79,7 @@ import {
 import StatusSelector from "./components/StatusSelector";
 
 // destructure strings import
-const searchStrings = languageStrings.searchpage;
+const { searchpage: searchStrings } = languageStrings;
 const {
   title: dateModifiedSelectorTitle,
   quickOptionDropdown,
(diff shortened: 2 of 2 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -132,132 +175,144 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
...
-        window.scrollTo(0, 0);
-      })
-      .catch(handleError)
-      .finally(() => setShowSpinner(false));
-  };
+  }, [state]);
 
   const handleSortOrderChanged = (order: Static<typeof SortOrder>) =>
-    setSearchPageOptions({ ...searchPageOptions, sortOrder: order });
+    search({ ...searchPageOptions, sortOrder: order });
```
edalex-ian, line 304: I'm quite happy how this is now so explicit. Before you had to know that any change to `searchPageOptions` would trigger a search and other related updates. Now if a search is to occur, the code explicitly says so.

Source/Plugins/Core/com.equella.core/js/tsrc/components/DateRangeSelector.tsx
```diff
@@ -304,12 +294,14 @@ export const DateRangeSelector = ({
             label={label}
             // When value is undefined use null instead so nothing is displayed in the TextField.
             value={value ?? null}
-            onChange={(newDate, _) =>
-              setStateDateRange({
+            onChange={(newDate, _) => {
```
SammyIsConfused, line 297: I'm curious as to whether the `, _` is required if not used. Especially since it's the last parameter in the list.

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -102,15 +110,50 @@ interface SearchPageHistoryState {
...
-  const { title: collectionSelectorTitle } = searchStrings.collectionSelector;
+type Action =
+  | { type: "init" }
+  | { type: "search"; options: SearchPageOptions }
+  | {
+      type: "search-complete";
+      result: OEQ.Search.SearchResult<OEQ.Search.SearchResultItem>;
+      classifications: Classification[];
+    }
+  | { type: "error"; cause: Error };
```
SammyIsConfused, line 121: a lot of these strings - search, search-complete, error, init, success, failure, searching, initialising etc - are used repeatedly in this file. Potentially worth some constants or enum?

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -132,132 +175,144 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
...
+              result: { ...result },
+              classifications: [...classifications],
+            });
+            // Update history
+            history.replace({
+              ...history.location,
+              state: { searchPageOptions: state.options, filterExpansion },
+            });
+            // scroll back up to the top of the page
+            window.scrollTo(0, 0);
```
SammyIsConfused, line 284: just an FYI, but this won't work on IE and Safari (Edge is fine from version 79 onwards)

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -132,132 +175,144 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
...
-      ...history.location,
-      state: { searchPageOptions, filterExpansion },
-    });
-  }, [filterExpansion, pagedSearchResult]);
+  }, [dispatch, filterExpansion, handleError, history, state]);
 
   // In Selection Session, once a new search result is returned, make each
-  // new search result Item draggable.
+  // new search result Item draggable. Could probably merge into 'searching'
+  // effect, however this is only required while selections sessions still
```
SammyIsConfused, line 293:
```suggestion
  // effect, however this is only required while selection sessions still
```

SammyIsConfused (approved):
Looks good! :+1:

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
SammyIsConfused: Never mind, I read it wrong. It won't work if you use it with ScrollToOptions as a parameter type. Using just an X Y is fine.

edalex-ian: All good, this is existing code anyway.

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
edalex-ian: Good observation, something I've pondered a lot and also discussed a fair bit with others.

Basically, this is a matter of two things:

1. Pretty well the idiomatic way to do things in JS land - and you'll see that extends to libraries like MUI; but
2. That's because here (TS land specifically) they're all type checked due to being string unions - and thereby don't suffer the issues of hardcoded strings in languages like C/C++/Java/Scala, and the IDEs are even able to assist.

So the plan is to go with the language, not against it.

Source/Plugins/Core/com.equella.core/js/tsrc/components/DateRangeSelector.tsx
edalex-ian: I almost removed it... but I've left it alone for two reasons:

1. It's existing code and completely unrelated to the change; and
2. There is value in the explicit declaration that the code is aware there's a second param but it's choosing not to use it.

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -78,6 +78,14 @@ import {
 } from "./components/SearchResultList";
 import StatusSelector from "./components/StatusSelector";
 
+// destructure strings import
+const searchStrings = languageStrings.searchpage;
```
ChristianMurphy, line 82: for consistency with line 87
```suggestion
const { searchpage: searchStrings } = languageStrings;
```

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -490,11 +538,38 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
     },
   ];
 
+  const renderClassifications = () => {
```
ChristianMurphy, line 541: Would it make sense for `renderClassifications` the function to become `RenderClassifications` a functional component?

ChristianMurphy (approved):

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
PenghaiZhang: What about keep the function and if statement here and move the MUI Grid to the component which takes `classifications` as props?

so it would become
```
const renderClassifications = () => {
if(...) {
return <RenderClassifications classification={...} />
}
return null
```

Can use a ternary operator instead of  if statement .

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -490,11 +538,38 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
...
+
+    return null;
+  };
+
   const {
     available: totalCount,
     highlight: highlights,
     results: searchResults,
-  } = pagedSearchResult;
+  } = state.status === "success" ? state.result : defaultPagedSearchResult;
```
PenghaiZhang, line 572: As discussed, there might be an issue about pagination.

PenghaiZhang (approved):
Thanks Ian! Looks good to me!

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
edalex-ian: Yes I did ponder this, but then wondered how much value I'd get with still needing the logic for the type narrowing external. I might have another quick play this morning. :+1:

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
edalex-ian: Yes, tis an interesting one. We have a few options here, but ultimately the issue with the results completely removing when you simply increase/decrease the page size is compounded by the existing functionality with the `window.scroll` that always takes you back to the top on new searches.

At least with the reducer in play, this will be much cleaner to resolve. So I'll raise a ticket and loop back around after this PR.

edalex-ian: #2716

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
edalex-ian: Done in a7fc6e5

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
edalex-ian: Applied in 60ac0e1

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
edalex-ian: Applied in 2bb8382

edalex-ian:
Awaiting a green build for merge.

ChristianMurphy (approved):

Issue #2629, opened 2020-12-22
New search UI: Implement skinny session for Brightspace's Insert Stuff option


cathfitz:
**Is your feature request related to a problem? Please describe.**
2020.2 does not include the implementation of the skinny selection session for the 'Insert Stuff' functionality in Brightspace.

PR #2717, opened 2021-02-03
Support skinny selection session 

PenghaiZhang:
#2629 

##### Checklist

- [x] tests are included

##### Description of change

This PR is to support using new search UI in Skinny Selection Session.
Changes can be basically categorised in three parts: 

1. EBP  changes to improve UI; and
2. Typescript side changes to make new search UI work properly in Skinny; and
3. Jest tests for above changes.

Here are two videos: [Moodle](https://streamable.com/3nwk95) and [Brightspace](https://streamable.com/4da7w2)

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContent.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContent.tsx
@@ -60,7 +60,7 @@ export const guestUser: OEQ.LegacyContent.CurrentUserDetails = {
   menuGroups: [],
 };
 
-interface ExternalRedirect {
+export interface ExternalRedirect {
   href: string;
 }
 
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContent.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContent.tsx
@@ -139,6 +139,12 @@ export function isChangeRoute(
   return (response as ChangeRoute).route !== undefined;
 }
 
+export function isExternalRedirect(
+  response: SubmitResponse
+): response is ExternalRedirect {
+  return (response as ExternalRedirect).href !== undefined;
+}
+
 function submitRequest(path: string, vals: StateData): Promise<SubmitResponse> {
   return Axios.post<SubmitResponse>(
     "api/content/submit" + encodeURI(path),
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContent.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContent.tsx
@@ -252,7 +252,7 @@ export const LegacyContent = React.memo(function LegacyContent({
             refreshUser();
           }
           redirected(content.route, false);
-        } else if (content.href) {
+        } else if (isExternalRedirect(content)) {
           redirected(content.href, true);
         }
       })
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/modules/LegacySelectionSessionModule.ts
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/modules/LegacySelectionSessionModule.ts
@@ -158,22 +158,16 @@ export const isSelectionSessionOpen = (): boolean =>
 /**
  * Return true if Selection Session is in 'Structured'.
  */
-export const isSelectionSessionInStructured = (): boolean => {
-  if (isSelectionSessionOpen()) {
-    return getSelectionSessionInfo().layout === "coursesearch";
-  }
-  return false;
-};
+export const isSelectionSessionInStructured = (): boolean =>
+  isSelectionSessionOpen() &&
+  getSelectionSessionInfo().layout === "coursesearch";
 
 /**
- * Return true if Selection Session is in 'Structured'.
+ * Return true if Selection Session is in 'Skinny'.
  */
-export const isSelectionSessionInSkinny = (): boolean => {
-  if (isSelectionSessionOpen()) {
-    return getSelectionSessionInfo().layout === "skinnysearch";
-  }
-  return false;
-};
+export const isSelectionSessionInSkinny = (): boolean =>
+  isSelectionSessionOpen() &&
+  getSelectionSessionInfo().layout === "skinnysearch";
 
 /**
  * Indicates if the Select Summary button is disabled or not.
@@ -422,15 +416,17 @@ export const selectResourceForSelectOrAdd = (
 /**
  * Select resources in 'skinny'. The approach is the same as 'selectResourceForSelectOrAdd'.
  * The response is an external href so use 'window.location' to redirect.
+ *
+ * @param itemKey A unique Item ID
+ * @param attachmentUUID A unique Attachment ID which can be undefined when no attachment is selected
  */
 export const selectResourceForSkinny = (
   itemKey: string,
-  attachmentUUIDs: string[]
+  attachmentUUID?: string
 ): Promise<void> => {
   const postData: SelectionSessionPostData = buildPostDataForSkinny(
     itemKey,
-    // Each selection In skinny can only have one attachment.
-    attachmentUUIDs.length > 0 ? attachmentUUIDs[0] : undefined
+    attachmentUUID
   );
 
   const callback = (response: SubmitResponse) => {
@@ -440,6 +436,7 @@ export const selectResourceForSkinny = (
       // script. Once the two steps are done, Selection session should be closed and the
       // iframe should return to whatever LMS page that Selection session was open from.
       $("#mainDiv").append(response.html["body"]);
+      // eslint-disable-next-line no-eval
       window.eval(response.script);
     } else if (isExternalRedirect(response)) {
       leaveSearchPage(response.href);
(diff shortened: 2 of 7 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/modules/LegacySelectionSessionModule.ts
```diff
@@ -389,6 +419,39 @@ export const selectResourceForNonCourseList = (
...
+    attachmentUUIDs.length > 0 ? attachmentUUIDs[0] : undefined
+  );
+
+  const callback = (response: SubmitResponse) => {
+    if (isPageContent(response)) {
+      // What's in 'html.body' is a form which should be submitted by the returned
+      // script. So the approach is to firstly add the form into DOM and then execute the
+      // script. Once the two steps are done, Selection session should be closed and the
+      // iframe should return to whatever LMS page that Selection session was open from.
+      $("#mainDiv").append(response.html["body"]);
```
PenghaiZhang, line 442: The reason for choosing `mainDiv` is it always exists in DOM.
Also, I did not add any check for 'html["body"]' and 'script' because they are provided by server.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/LegacySelectionSessionModule.ts
```diff
@@ -398,8 +461,20 @@ export const selectResource = (
...
+    ],
+    [
+      Literal("search"),
+      () => selectResourceForSelectOrAdd(itemKey, attachments),
+    ],
+    [
+      Literal("skinnysearch"),
+      () => selectResourceForSkinny(itemKey, attachments),
+    ]
+  );
```
PenghaiZhang, line 477: Do we want to add a `Unknown` here ?

Source/Plugins/Core/com.equella.core/js/tsrc/modules/LegacySelectionSessionModule.ts
```diff
@@ -162,6 +164,17 @@ export const isSelectionSessionInStructured = (): boolean => {
...
 };
+
+/**
+ * Return true if Selection Session is in 'Structured'.
+ */
+export const isSelectionSessionInSkinny = (): boolean => {
+  if (isSelectionSessionOpen()) {
+    return getSelectionSessionInfo().layout === "skinnysearch";
+  }
+  return false;
```
ChristianMurphy, line 175:
```suggestion
  return isSelectionSessionOpen() && getSelectionSessionInfo().layout === "skinnysearch";
```

Source/Plugins/Core/com.equella.core/js/tsrc/modules/LegacySelectionSessionModule.ts
ChristianMurphy: would it be possible to use `document.getElementById` here?

edalex-ian:
I notice on the brightspace video there is a _big_ delay between clicking the chosen attachment and anything happening (think it was > 5 seconds). I'd be keen to see what the old UI did for this, as it feels like we need some kind of spinner type indicator.

Could you please have a look at the old and see how it compares?

Source/Plugins/Core/com.equella.core/js/tsrc/legacycontent/LegacyContent.tsx
```diff
@@ -139,6 +139,12 @@ export function isChangeRoute(
   return (response as ChangeRoute).route !== undefined;
 }
 
+export function isExternalRedirect(
```
edalex-ian, line 142: :thinking: 

I wonder if this should then also be used down in `submitCurrentForm` rather than the plain old `else if (content.href)` as it sure would clarify things.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/LegacySelectionSessionModule.ts
```diff
@@ -340,6 +349,27 @@ export const buildPostDataForSelectOrAdd = (
         ...getBasicPostData(),
       };
 
+/**
+ * Build an object of SelectionSessionPostData for 'skinny'.
+ */
```
edalex-ian, line 354: Could be rather insightful here to document the two params, and especially comment on why sometimes there is an `attachmentUUID` and sometimes there's not.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/LegacySelectionSessionModule.ts
```diff
@@ -389,6 +419,39 @@ export const selectResourceForNonCourseList = (
...
+  );
+
+  const callback = (response: SubmitResponse) => {
+    if (isPageContent(response)) {
+      // What's in 'html.body' is a form which should be submitted by the returned
+      // script. So the approach is to firstly add the form into DOM and then execute the
+      // script. Once the two steps are done, Selection session should be closed and the
+      // iframe should return to whatever LMS page that Selection session was open from.
+      $("#mainDiv").append(response.html["body"]);
+      window.eval(response.script);
```
edalex-ian, line 443: Please add the ESLint suppression comment.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/LegacySelectionSessionModule.ts
edalex-ian: I think that'd be good too. Only use jQuery if we have to - such as the JQueryDiv.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/LegacySelectionSessionModule.ts
```diff
@@ -389,6 +419,39 @@ export const selectResourceForNonCourseList = (
...
+ * Select resources in 'skinny'. The approach is the same as 'selectResourceForSelectOrAdd'.
+ * The response is an external href so use 'window.location' to redirect.
+ */
+export const selectResourceForSkinny = (
+  itemKey: string,
+  attachmentUUIDs: string[]
+): Promise<void> => {
+  const postData: SelectionSessionPostData = buildPostDataForSkinny(
+    itemKey,
+    // Each selection In skinny can only have one attachment.
```
edalex-ian, line 432: Then why do you accept an array of UUIDs?

Source/Plugins/Core/com.equella.core/js/tsrc/modules/LegacySelectionSessionModule.ts
edalex-ian: If `getSelectionSessionInfo().layout` is not a string literal union, then yes.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/LegacySelectionSessionModule.ts
```diff
@@ -398,8 +461,20 @@ export const selectResource = (
...
+    [
+      Literal("skinnysearch"),
+      () => selectResourceForSkinny(itemKey, attachments),
+    ]
+  );
   const { layout } = getSelectionSessionInfo();
-  return layout === "coursesearch"
-    ? selectResourceForCourseList(itemKey, attachments)
-    : selectResourceForNonCourseList(itemKey, attachments);
+  return selectResource(layout);
```
edalex-ian, line 479: Keep in mind that the point of using `match` is so that we can step further away from imperative code. So for this function I'd really expect a simple:

```typescript
export const selectResource = (
  itemKey: string,
  attachments: string[]
): Promise<void> =>
  match([....])(getSelectionSessionInfo().layout);
```

edalex-ian (commented):
Looks good @PenghaiZhang . Just a couple tweaks and clarifications sought. :+1:

Source/Plugins/Core/com.equella.core/js/tsrc/modules/LegacySelectionSessionModule.ts
PenghaiZhang: Yes, it's :)

PenghaiZhang:
We have found another issue in Selection Session - the implementation of 'updateTemplate' is a `nop` so there is no actual error handling happening when errors occur. Perhaps we create another bug ticket and do that in another PR ?

PenghaiZhang:
> I notice on the brightspace video there is a _big_ delay between clicking the chosen attachment and anything happening (think it was > 5 seconds). I'd be keen to see what the old UI did for this, as it feels like we need some kind of spinner type indicator.
> 
> Could you please have a look at the old and see how it compares?

I have confirmed that there is nothing indicating a selection is in progress in old UI, too.

edalex-ian:
> We have found another issue in Selection Session - the implementation of 'updateTemplate' is a `nop` so there is no actual error handling happening when errors occur. Perhaps we create another bug ticket and do that in another PR ?

Good plan. :+1: Please do and reference to this PR.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/LegacySelectionSessionModule.ts
edalex-ian: I see you marked this as resolve, but it hasn't been changed. Was it missed, or is jQuery required, or...?

edalex-ian (approved):
I think it's GTG @PenghaiZhang .

Just two things though:

1. Yes please create a bug for error handling; and
2. You marked one item as resolved, but nothing had changed. :thinking:

Source/Plugins/Core/com.equella.core/js/tsrc/modules/LegacySelectionSessionModule.ts
PenghaiZhang: Ah I forgot this. It is in the last commit now.

PenghaiZhang:
Here is the ticket about error handling #2725 .

PR #2726, opened 2021-02-09
Feature - Search2 MIME type filtering support

edalex-ian:
##### Checklist

- [x] tests are included
- [x] documentation is changed or added (apidocs)

##### Description of change

Adds support for filtering by a list of MIME types. We need this for a few things, but my immediate concern is the image gallery.

image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
@@ -65,8 +65,7 @@ object SearchHelper {
     search.setUseServerTimeZone(true)
     search.setQuery(params.query)
     search.setOwner(params.owner)
-    search.setMimeTypes(
-      (if (params.mimeTypes.isEmpty) None else Some(params.mimeTypes.toList.asJava)).orNull)
+    search.setMimeTypes(params.mimeTypes.toList.asJava)
 
     val orderType =
       DefaultSearch.getOrderType(Option(params.order).map(_.toLowerCase).orNull, params.query)
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
@@ -18,10 +18,6 @@
 
 package com.tle.web.api.search
 
-import java.time.format.DateTimeParseException
-import java.time.{LocalDate, LocalDateTime, LocalTime, ZoneId}
-import java.util.Date
-
 import com.dytech.edge.exceptions.BadRequestException
 import com.tle.beans.entity.DynaCollection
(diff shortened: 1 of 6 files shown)


PenghaiZhang:
I noticed that on your screenshot you did a search with filtering by jpeg but the result had a `mp4` attachment.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
```diff
@@ -66,6 +65,8 @@ object SearchHelper {
     search.setUseServerTimeZone(true)
     search.setQuery(params.query)
     search.setOwner(params.owner)
+    search.setMimeTypes(
+      (if (params.mimeTypes.isEmpty) None else Some(params.mimeTypes.toList.asJava)).orNull)
```
PenghaiZhang, line 69: I was wondering if we can simplify above  code to
```scala
  search.setMimeTypes(params.mimeTypes)
```

I remember the reason we made it more complicated was to respect Scala coding pattern - avoid null as much as possible. But I might miss something else.

PenghaiZhang (approved):
Looks good to me Ian!

edalex-ian:
> I noticed that on your screenshot you did a search with filtering by jpeg but the result had a `mp4` attachment.

Yes, maybe not the best screenshot. That item has both videos and images, and it just happens that the top-most attachment is the video one (unfortunately).

So the way this filter works is it returns all items which have at least one attachment with the specified mime-type. And the search result still just returns the items details plus the list of all its attachments.

PenghaiZhang:
> > I noticed that on your screenshot you did a search with filtering by jpeg but the result had a `mp4` attachment.
> 
> Yes, maybe not the best screenshot. That item has both videos and images, and it just happens that the top-most attachment is the video one (unfortunately).
> 
> So the way this filter works is it returns all items which have at least one attachment with the specified mime-type. And the search result still just returns the items details plus the list of all its attachments.

Oh I see. That's all good.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
edalex-ian: Ah yes, here we definitely don't need that it seems. Thank you, addressed in 61703a7

edalex-ian:
Will merge on :green_circle:

edalex-ian:
Gah! That change breaks the `NewSearchPageTest` - even though it was fine on the `Search2ApiTest`. Will see what's happening - and why it could get past the `Search2ApiTest`.

edalex-ian:
Hmmm, I'm not so sure... Is this test a bit flakey... Will re-run while I play locally.

edalex-ian:
Looks like there was a smidge of flake in the `NewSearchPageTest`, so have tuned that up a bit. Will see if we can get that :green_circle: . :crossed_fingers:

Issue #2559, opened 2020-11-26
Convert Purescript Upload component to TypeScript


edalex-ian:
**Is your feature request related to a problem? Please describe.**
The community direction to remove Purescript from the project to assist with ongoing maintenance.

We need to convert the [file upload UI component](https://github.com/openequella/openEQUELLA/tree/develop/Source/Plugins/Core/com.equella.core/js/src/Uploads) from Purescript to TypeScript to assist with future maintenance. Plus we will most likely wish to re-use the resultant component for the work planned in 2021 for the balance of New UI work.

**Describe the solution you'd like**

Being a React component, the conversion should result in a new Upload component in `Source/Plugins/Core/com.equella.core/js/tsrc/components`.

Seeing it is a component which is consumed directly via the Legacy UI, the work may also include an additional bit of wrapping in App.tsx/index.ts etc.

I propose this is done as part of 2021.1.

PR #2735, opened 2021-02-17
Improve the tsx File Uploader

PenghaiZhang:
#2559 

##### Checklist

- [ ] tests are included

##### Description of change

This is the second PR for the tsx File Uploader. Some issues have been addressed.

1. If a file (e.g. zip) has child files, show child files with an indentation.
2. Make sure the upload percentage is NOT bigger than 100 when the file is quite small.
3. Fix a memory issue when upload large files.
4. More EBP work to improve the UI in both New and Old UI.

A screenshot for showing child files.
image

A video for EBP 

https://user-images.githubusercontent.com/47203811/108142592-2fcd3c80-711a-11eb-9061-368182f88c8e.mp4


Source/Plugins/Core/com.equella.core/js/tsrc/fileuploader/UploadInfo.tsx
```diff
@@ -43,8 +44,6 @@ export const UploadInfo = ({ file }: UploadInfoProps) => {
       </Grid>
     </Grid>
   ) : (
-    <Typography role="alert" color="error">
-      {failedReason}
-    </Typography>
+    <Fragment key={file.localId}>{errorMessage}</Fragment>
```
PenghaiZhang, line 47: The Typography is provided by `ListItemText` so there is no need to add one more Typography. 
However, if I simply return the error message, which is a string, this component can no longer be used as a react component (but we can still use it just as a function). 

I firstly added a `<>...</>` to avoid above issue, but es-lint rule `jsx-no-useless-fragment` complains.
So I went eslint website and found the correct approach:
```tsx
<Fragment key={key}>{something}</Fragment>
```

Source/Plugins/Core/com.equella.core/js/tsrc/fileuploader/FileUploaderActionLink.tsx
```diff
@@ -43,14 +43,14 @@ export const FileUploaderActionLink = ({
   id,
   className,
 }: FileActionLinkProps) => (
+  // This Link is used as a button but wo don't do ' component="button" ' because
```
ChristianMurphy, line 46:
```suggestion
  // This Link is used as a button but we don't use ' component="button" ' because
```

Source/Plugins/Core/com.equella.core/js/tsrc/fileuploader/FileUploaderActionLink.tsx
```diff
@@ -43,14 +43,14 @@ export const FileUploaderActionLink = ({
...
+  // this will break UI consistency (color & font size) in Old UI.
   <Link
     id={id}
     className={className}
-    component="button"
     onClick={onClick}
     title={text}
-    underline="none"
-    color="primary"
+    style={{ cursor: "pointer" }} // Add this style since this Link does not have 'href'.
```
ChristianMurphy, line 53: would it make sense for this to have `aria-role="button"` ?

Source/Plugins/Core/com.equella.core/js/tsrc/fileuploader/InlineFileUploader.tsx
```diff
@@ -131,11 +132,22 @@ export const InlineFileUploader = ({
...
-    entries.map((entry) => ({
-      fileEntry: entry,
-      status: "uploaded",
-    }))
+    entries.flatMap((entry) => {
+      const withChildren: UploadedFile[] = [];
+      return withChildren
+        .concat(initialiseEntry(entry, false))
+        .concat(entry.children.map((e) => initialiseEntry(e, true)));
+    })
```
edalex-ian, line 150: I wonder if we could make it simpler with just:

```suggestion
    entries.flatMap((entry) => [
      initialiseEntry(entry, false),
      ...(entry.children.map((e) => initialiseEntry(e, true)))
    ])
```

Source/Plugins/Core/com.equella.core/js/tsrc/fileuploader/UploadFileName.tsx
```diff
@@ -27,16 +27,28 @@ interface UploadFileNameProps {
...
-    <div>{fileName}</div>
-  );
+export const UploadFileName = ({
+  fileName,
+  link,
+  indented,
+}: UploadFileNameProps) => (
+  <div className={indented ? "indent" : ""}>
+    {link ? (
+      <Link href={link} target="_blank">
```
ChristianMurphy, line 47:
```suggestion
      <Link href={link} target="_blank" rel="noopener noreferrer">
```

Source/Plugins/Core/com.equella.core/js/tsrc/fileuploader/InlineFileUploader.tsx
```diff
@@ -275,36 +287,67 @@ export const InlineFileUploader = ({
...
+    ];
+  };
+
+  // Build an Icon button for adding resources. In Old UI, this is achieved by legacy CSS styles.
+  // In New UI, use component LabelledIconButton. The reason for using '<a>' for Old UI is because
+  // the legacy style 'add' only applies to '<a>'.
+  const AddResourceButton = () => {
+    const commonProps = {
+      id: `${ctrlId}_addLink`,
+      onClick: () => openDialog("", ""),
```
edalex-ian, line 328: `openDialog("", "")` reads as though someone forgot to put strings in here. I'm guess it's deliberate though.

As a result could you please document what is going on here - even reading the JSDoc for the prop this maps to provided no insight, so I'd have to meander around the code to figure out what this oddity is. (Probably highlights the `dialog` function needs reworking, however I'm guess this is just interfacing with legacy code.)

Source/Plugins/Core/com.equella.core/js/tsrc/fileuploader/UniversalFileUploader.tsx
```diff
@@ -176,6 +177,25 @@ export const UniversalFileUploader = ({
         },
   ];
 
+  const AddScrapBookButton = () => {
```
edalex-ian, line 180: May want to include the same commentary here as you had over in `InlineFileUploader` to clarify why this is required.

Source/Plugins/Core/com.equella.core/js/tsrc/fileuploader/UploadFileName.tsx
```diff
@@ -27,16 +27,28 @@ interface UploadFileNameProps {
    * The link to preview the file
    */
   link?: string;
+  /**
+   * True if an indentation is required
```
edalex-ian, line 31:
```suggestion
   * `true` if the entry should be indented.
```

Note that (unlike JavaDoc) JSDoc supported markdown, so best to use it when appropriate. (One day we may even generate the docs. :crossed_fingers: )

Source/Plugins/Core/com.equella.core/js/tsrc/fileuploader/UploadList.tsx
```diff
@@ -62,22 +61,50 @@ export const UploadList = ({
...
-        ) : (
-          <UploadInfo file={file} />
-        );
-        return (
-          <ListItem key={fileId} divider>
+
+        /**
+         * Build the content of each ListItem. Directly show the content for any UploadedFile that
+         * does not have any error message. For other cases, use ListItemText to wrap the content,
+         * that's, file name as primary text and other information as second text.
```
edalex-ian, line 68:
```suggestion
         * that's, file name as primary text and other information as secondary text.
```

Source/Plugins/Core/com.equella.core/js/tsrc/fileuploader/UploadList.tsx
```diff
@@ -62,22 +61,50 @@ export const UploadList = ({
...
+              indented={isUploadedFile(file) ? file.indented : false}
+              link={isUploadedFile(file) ? file.fileEntry.link : undefined}
+            />
+          );
+          const secondaryText = isUploadedFile(file) ? (
+            file.errorMessage
+          ) : (
+            <UploadInfo file={file} />
+          );
+          // Unset Legacy font size and margin styles.
```
edalex-ian, line 83: :thinking: 

Curious, is this actually classic legacy CSS that you're attempting to 'unset', or is it EBP?

Source/Plugins/Core/com.equella.core/js/tsrc/modules/FileUploaderModule.ts
```diff
@@ -66,6 +66,10 @@ export interface UploadedFile {
    * Text describing why fail to delete this file
    */
   errorMessage?: string;
+  /**
+   * Whether to show this file with an indentation. Typically, true for child AjaxFileEntry.
```
edalex-ian, line 70:
```suggestion
   * Whether to show this file indented. Typically, true for child AjaxFileEntry.
```

Source/Plugins/Core/com.equella.core/js/tsrc/modules/FileUploaderModule.ts
```diff
@@ -288,17 +292,20 @@ const doUpload = (
   const source = CancelToken.source();
   axiosSourceMap.set(uploadingFile.localId, source);
   const token = source.token;
-
-  const formData = new FormData();
-  formData.append("file", fileEntry);
-  return Axios.post<UpdateEntry | UploadFailed>(path, formData, {
+  return Axios.post<UpdateEntry | UploadFailed>(path, fileEntry, {
+    // IMPORTANT! Must specify the file type.
```
edalex-ian, line 296: If this is "IMPORTANT!" then we really should also say **why**.

edalex-ian (commented):
Looks good @PenghaiZhang . Just a few minor queries.

BTY.

Source/Plugins/Core/com.equella.core/js/tsrc/fileuploader/FileUploaderActionLink.tsx
PenghaiZhang: Good idea!

Source/Plugins/Core/com.equella.core/js/tsrc/fileuploader/InlineFileUploader.tsx
PenghaiZhang: This is a powerful legacy function. Whenever we want to open the `UniversalResourceDialog`, we call this function. Server will determine the content of the dialog, based on what parameters passed in. 

Here, because the first two parameters are empty strings the content is the resource adding interface. 

I have updated the JSDoc of this function.

Source/Plugins/Core/com.equella.core/js/tsrc/fileuploader/UploadFileName.tsx
PenghaiZhang: Good idea!! Thank you!

Source/Plugins/Core/com.equella.core/js/tsrc/modules/FileUploaderModule.ts
PenghaiZhang: What I found is if I don't specify the content type, it defaults to `application/x-www-form-urlencoded`. As a result, the inputstream of the request body will be consumed somewhere else and all saved files are actually empty files.

Source/Plugins/Core/com.equella.core/js/tsrc/fileuploader/UploadList.tsx
PenghaiZhang: Ah I think it's the EBP one.

PenghaiZhang: Do I need to update the comment ?

edalex-ian: Probably two things:

1. Update comment; but
2. Should we modify the EBP styling so it doesn't cause the issue?

Because the thing that comes to mind here, if someone wants to custom style and we do an `unset` I think we'll win no matter what.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/FileUploaderModule.ts
```diff
@@ -389,7 +391,7 @@ export const deleteUpload = (
       .catch((error: Error) => {
         onError({
           ...file,
-          errorMessage: `Fail to delete ${name} due to error: ${error.message}`,
+          errorMessage: `Failed to delete ${name} due to error: ${error.message}`,
```
edalex-ian, line 394: Ah! Shouldn't this be a language string?

edalex-ian:
Looks like that `aria-role` introduced a new ESLint warning. Will need to fix.

edalex-ian (approved):

ChristianMurphy (approved):

PR #2736, opened 2021-02-17
New Search UI: allow attachments contributed via Resource Selector to use the correct viewer on the search page

SammyIsConfused:
##### Checklist

- [x] tests are included

##### Description of change
When clicking attachment previews from the search page, if the attachment came from a non-integration selection session (Resource Attachment) then it wouldn't resolve the mimetype, and so the viewer would treat is as a default MIME type attachment and simply open in a new tab rather than a lightbox. 
To fix this, I added a case on the backend to handle mimetypes for resource attachments rather than just for file attachments. 
Since it is possible to select a resource attachment which points to a resource attachment which points to a resource attachment infinitely, we required a bit of recursion to drill down to the original attachment and get its mime type. 

### DIFF
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/core/item/dao/impl/AttachmentDaoImpl.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/core/item/dao/impl/AttachmentDaoImpl.java
@@ -125,4 +125,30 @@ public class AttachmentDaoImpl extends GenericDaoImpl<Attachment, Long> implemen
     // it's possible that value1 could be
     return attachments;
   }
+
+  @Override
+  public List<Attachment> findAllByUuid(String uuid) {
+    String[] names = new String[] {"institution", "uuid"};
+    Object[] values = new Object[] {CurrentInstitution.get(), uuid};
+    List<String> nameList = new ArrayList<String>(Arrays.asList(names));
+    List<Object> valueList = new ArrayList<Object>(Arrays.asList(values));
+
+    StringBuilder query = new StringBuilder();
+    query.append(
+        "SELECT a FROM Item i LEFT JOIN i.attachments a WHERE i.institution = :institution");
+    query.append(" AND a.uuid = :uuid");
+    List<Attachment> attachments =
+        (List<Attachment>)
+            getHibernateTemplate()
+                .findByNamedParam(
+                    query.toString(),
+                    nameList.toArray(new String[nameList.size()]),
+                    valueList.toArray());
+
+    return attachments;
+  }
+
+  public Attachment findByUuid(String uuid) {
+    return findAllByUuid(uuid).get(0);
+  }
 }
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/core/item/dao/impl/AttachmentDaoImpl.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/core/item/dao/impl/AttachmentDaoImpl.java
@@ -34,8 +34,7 @@ import java.util.Arrays;
 import java.util.List;
 import javax.inject.Singleton;
 import org.hibernate.Criteria;
-import org.hibernate.HibernateException;
-import org.hibernate.Session;
+import org.hibernate.criterion.DetachedCriteria;
 import org.hibernate.criterion.Restrictions;
 
 @SuppressWarnings("nls")
@@ -131,10 +130,9 @@ public class AttachmentDaoImpl extends GenericDaoImpl<Attachment, Long> implemen
   }
 
   // Criteria for checking an attachments against institution of the item and uuid of the
-  // attachment.
-  private Criteria createUuidCriterion(Session session, String uuid) {
-    return session
-        .createCriteria(Attachment.class)
+  // attachment. Detached so that we don't need to create a Hibernate session until one is required.
+  private DetachedCriteria criteriaByUuid(String uuid) {
+    return DetachedCriteria.forClass(Attachment.class)
         .createAlias("item", "i")
         .add(Restrictions.eq("i.institution", CurrentInstitution.get()))
         .add(Restrictions.eq("uuid", uuid));
@@ -142,27 +140,11 @@ public class AttachmentDaoImpl extends GenericDaoImpl<Attachment, Long> implemen
 
   @Override
   public List<Attachment> findAllByUuid(String uuid) {
-    return (List<Attachment>)
-        getHibernateTemplate()
-            .execute(
-                new TLEHibernateCallback() {
-                  @Override
-                  public Object doInHibernate(Session session) throws HibernateException {
-                    return (List<Attachment>) createUuidCriterion(session, uuid).list();
-                  }
-                });
+    return (List<Attachment>) findByDetachedCriteria(Criteria::list, criteriaByUuid(uuid));
   }
 
   @Override
   public Attachment findByUuid(String uuid) {
-    return (Attachment)
-        getHibernateTemplate()
-            .execute(
-                new TLEHibernateCallback() {
-                  @Override
-                  public Object doInHibernate(Session session) throws HibernateException {
-                    return createUuidCriterion(session, uuid).uniqueResult();
-                  }
-                });
+    return (Attachment) findByDetachedCriteria(Criteria::uniqueResult, criteriaByUuid(uuid));
   }
 }
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/core/item/dao/impl/AttachmentDaoImpl.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/core/item/dao/impl/AttachmentDaoImpl.java
@@ -140,11 +140,11 @@ public class AttachmentDaoImpl extends GenericDaoImpl<Attachment, Long> implemen
 
   @Override
   public List<Attachment> findAllByUuid(String uuid) {
-    return (List<Attachment>) findByDetachedCriteria(Criteria::list, criteriaByUuid(uuid));
+    return (List<Attachment>) findByDetachedCriteria(criteriaByUuid(uuid), Criteria::list);
   }
 
   @Override
   public Attachment findByUuid(String uuid) {
-    return (Attachment) findByDetachedCriteria(Criteria::uniqueResult, criteriaByUuid(uuid));
+    return (Attachment) findByDetachedCriteria(criteriaByUuid(uuid), Criteria::uniqueResult);
   }
 }
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/core/item/dao/impl/AttachmentDaoImpl.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/core/item/dao/impl/AttachmentDaoImpl.java
@@ -129,7 +129,7 @@ public class AttachmentDaoImpl extends GenericDaoImpl<Attachment, Long> implemen
     return attachments;
   }
 
-  // Criteria for checking an attachments against institution of the item and uuid of the
+  // Criteria for checking attachments against institution of the item and UUID of the
   // attachment. Detached so that we don't need to create a Hibernate session until one is required.
   private DetachedCriteria criteriaByUuid(String uuid) {
     return DetachedCriteria.forClass(Attachment.class)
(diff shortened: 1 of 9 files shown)


SammyIsConfused:
Gah, build was decidedly very unhappy. Must have missed something. Putting back to draft and trying to iron out the kinks.

SammyIsConfused:
Turned out that only resource attachments of type 'a' (attachment). If such an attachment was pointing at a resource summary, it would throw a null pointer (hence all the failed tests). Locally this change seemed to fix the problems with all of the tests.

Source/Plugins/Core/com.equella.core/src/com/tle/core/mimetypes/MimeTypeServiceImpl.java
```diff
@@ -167,6 +171,23 @@ public String getMimeTypeForFilename(String filename) {
...
+    Attachment attachment = attachmentDao.findByCriteria(Restrictions.eq("uuid", attachmentUuid));
+    return getMimeEntryForAttachment(attachment);
+  }
+
+  public String getMimeTypeForResourceAttachmentBean(
+      ResourceAttachmentBean resourceAttachmentBean) {
+    switch (resourceAttachmentBean.getResourceType()) {
+      case 'a':
+        return getMimeTypeForAttachmentUuid(resourceAttachmentBean.getAttachmentUuid());
+      case 'p':
```
ChristianMurphy, line 184: what do `'a'` and `'p'` mean in this context?
Would it make sense for known types to be an enumeration?

SammyIsConfused:
Argh. So we're down from 33 tests to just 3. I'll have this working soon. 
Main reason we got a green build before is because my if check had a typo (was checking string against char when I should have been comparing strings).

edalex-ian:
> Argh. So we're down from 33 tests to just 3. I'll have this working soon.
> Main reason we got a green build before is because my if check had a typo (was checking string against char when I should have been comparing strings).

No probs. Let us know when it's green and I'll review then. :wink:

SammyIsConfused:
Okay so the problems are with Equella Web Pages and SCORM packages. Fixing.

SammyIsConfused:
so the issue was that the way I was recursing handled only single linked attachments - if it was an unzipped zip or scorm it would return multiple attachments which broke it. So I changed the code to only recurse if a single attachment is returned, otherwise we can let the original method handle it, as that handled zips/scorms correctly.

Source/Plugins/Core/com.equella.core/src/com/tle/core/mimetypes/MimeTypeServiceImpl.java
SammyIsConfused: 'a' is an Attachment. 'p' is a resource summary Path. I've found constants for them, and I'm switching them out.

SammyIsConfused:
@edalex-ian Ready for review now! 
Got the green build and then just added a couple commits to refer to constants rather than inline strings.

SammyIsConfused:
Okay my bad, that's definitely the last commit now.

ChristianMurphy (approved):

Source/Plugins/Core/com.equella.core/src/com/tle/core/mimetypes/MimeTypeServiceImpl.java
```diff
@@ -56,6 +61,8 @@
 import net.sf.json.JSONArray;
 import net.sf.json.JSONObject;
 import org.hibernate.Hibernate;
+import org.hibernate.criterion.Criterion;
+import org.hibernate.criterion.Restrictions;
```
edalex-ian, line 65: It would be good if building criteria and/or queries remained in the DAO - in this case the AttachmentDao. It seems you want two methods:

1. findByUuid
2. findAllByUrl

Note two things:

1. Typically you also want to include the institution ID in the query to leverage the index to minimise response time/load; and
2. Because the AttachmentDao is ultimately a GenericDao, for the first method - `findByUuid` - you should note you'd already be able to do `attachmentDao.findById(Attachment.class, id)`.

edalex-ian (changes requested):
Thanks for that fix @SammyIsConfused 

Could you please address a few things? (Most importantly data access.)

Source/Plugins/Core/com.equella.core/src/com/tle/core/mimetypes/MimeTypeServiceImpl.java
SammyIsConfused: So the url part is a weird part of the attachment class. It doesn't actually necessarily contain an URL, it depends on the attachment type. If it's a file attachment, this will be a filename, if its an openEQUELLA resource attachment, it will be an attachment Uuid, if it contains a link attachment it will be an URL. googlebooks, youtube, flickr and itunesu have this blank. And so we aren't finding by URL, we are finding by UUID again since we are within the context of a resource attachment.
So, the two methods we needed were findByUuid and findAllByUuid and one fell out of the other. I've also linked it to an institution.

SammyIsConfused:
AdvancedScriptControlTest - testStagingMove flaked in the new UI on GHA. I think this is fine, especially since Codebuild was fine. I'll restart GHA just in case.

SammyIsConfused:
Yay awesome! Build is green again. @edalex-ian we should be good to go.

ChristianMurphy (approved):

Source/Plugins/Core/com.equella.core/src/com/tle/core/item/dao/impl/AttachmentDaoImpl.java
```diff
@@ -125,4 +125,30 @@ public AttachmentDaoImpl() {
     // it's possible that value1 could be
     return attachments;
   }
+
+  @Override
+  public List<Attachment> findAllByUuid(String uuid) {
```
edalex-ian, line 130: Typically you only want to use HQL if you really have too (e.g. performance or complex cross entity queries or aggregations) - as it has the least amount of validation (and thereby the greatest chance of runtime issues). This query is simple, and so should just [use Criteria](https://docs.jboss.org/hibernate/stable/orm/userguide/html_single/Hibernate_User_Guide.html#criteria).

GenericDao provide a rather handy `findAllByCriteria` (as you were using previously) which you could end up having this whole method along the lines of simply returning the result from that - although seeing you're only passing in Criterion you may have issues querying the association for current institution.

But if that's the case, you still just quickly whip up your own `createCriteria` call.

(Although note, that now we're on current Hibernate the above link will guide you through using the advised createQuery etc from the JPA API.)

Source/Plugins/Core/com.equella.core/src/com/tle/core/item/dao/impl/AttachmentDaoImpl.java
```diff
@@ -125,4 +125,30 @@ public AttachmentDaoImpl() {
...
+                .findByNamedParam(
+                    query.toString(),
+                    nameList.toArray(new String[nameList.size()]),
+                    valueList.toArray());
+
+    return attachments;
+  }
+
+  public Attachment findByUuid(String uuid) {
+    return findAllByUuid(uuid).get(0);
```
edalex-ian, line 152: What if multiple are returned. Would that be an issue? If so, you may want to at a minimum use the `GenericDao`'s `uniqueResult` method.

If it doesn't matter, may be better to change this function to getFirstByUuid - but note this will (and is currently) non-deterministic for most DBs unless you add an order by.

edalex-ian (changes requested):
Heya Sam, good move in the right direction, but slightly incorrect direction with moving to HQL. Please address.

Source/Plugins/Core/com.equella.core/src/com/tle/core/item/dao/impl/AttachmentDaoImpl.java
SammyIsConfused: Yeah, it was linking back to the institution that caused me to do this, I was having trouble getting a Criterion with a left join to another table.

edalex-ian: Cool, yeah just focus on the relationships at the entity table - vs. how to do it with SQL/HQL - and feel free to create a detached criteria.

Source/Plugins/Core/com.equella.core/src/com/tle/core/item/dao/impl/AttachmentDaoImpl.java
SammyIsConfused: I've used unique result now. :+1:

Source/Plugins/Core/com.equella.core/src/com/tle/core/item/dao/impl/AttachmentDaoImpl.java
SammyIsConfused: I think I've got it now - I removed the HQL and created a function which returns a Criteria which checks against institution and then uuid, and then I use it with both methods findAllByUuid and findByUuid with .list() and .uniqueResult() respectively.

edalex-ian (commented):
Heya @SammyIsConfused ,

That's the way, almost there. Two things:

1. You've not got a lot of repeated boiler plate which could be reduced with the Java 8 Functional interfaces - I've popped a potential refactor in below; and
2. Shame to see it's still done with the old `createCriteria` rather then using the new recommended JPA API - any reason?

Potential refactor to reduce boiler plate:

```java
  private Object findByUuid(String uuid, Function<Criteria, Object> process) {
    return getHibernateTemplate().execute(new TLEHibernateCallback() {
      @Override
      public Object doInHibernate(Session session) throws HibernateException {
        return process.apply(session
          .createCriteria(Attachment.class)
          .createAlias("item", "i")
          .add(Restrictions.eq("i.institution", CurrentInstitution.get()))
          .add(Restrictions.eq("uuid", uuid)));
      }
    });
  }

  public List<Attachment> findAllByUuid(String uuid) {
    return (List<Attachment>) findByUuid(uuid, Criteria::list);
  }

  public Attachment findByUuid(String uuid) {
    return (Attachment) findByUuid(uuid, Criteria::uniqueResult);
  }
```

This could be further reworked so that it could be moved to `GenericDao` and instead leverage the generic setup down there and then be of use to others by accepting a Criteria rather than Criterion. But up to you.

SammyIsConfused:
@edalex-ian Refactored out to `GenericDao` with a new function `findByDetachedCriteria`, which takes a `DetachedCriteria` and also a `Function<Criteria, Object> process` to determine what it does with the output of the query. Then in the `AttachmentDaoImpl` I created a `DetachedQuery criteriaByUuid(String uuid)` with the institution and uuid check, which simplifies `findByUuid` and `findAllByUuid` to `findByDetachedCriteria(Criteria::list/uniqueResult, criteriaByUuid(uuid));`. 
I don't know if `findByDetachedCriteria` is as generic as it could be, it got tricky when the output could be `T` and also could be `List<T>`, but it works and now could definitely be useful elsewhere. 

If you think I've over-complicated things, I can go ahead and just use your suggested code directly without touching `GenericDao`.

SammyIsConfused:
Also - the main reason for not using the JPA API was simply that I've gone through several different ways to do the same thing in the course of this PR, Criteria is used a lot in openEQUELLA already and I didn't want to waste your review time reworking to a new API again. (Although that said the move to GenericDao is a bit of a change around anyway).

Source/Plugins/Core/com.equella.core/src/com/tle/core/item/dao/impl/AttachmentDaoImpl.java
```diff
@@ -125,4 +128,23 @@ public AttachmentDaoImpl() {
     // it's possible that value1 could be
     return attachments;
   }
+
+  // Criteria for checking an attachments against institution of the item and uuid of the
```
edalex-ian, line 132:
```suggestion
  // Criteria for checking attachments against the institution of the item and UUID of the
```

Source/Plugins/Core/com.equella.serverbase/src/com/tle/core/hibernate/dao/GenericDao.java
```diff
@@ -64,6 +66,8 @@
 
   List<T> findAllByCriteria(Criterion... criterion);
 
+  Object findByDetachedCriteria(
+      final Function<Criteria, Object> process, DetachedCriteria criteria);
```
edalex-ian, line 70: Typically I'd expect these two be flipped - i.e. the data you're operating on first, and the function last.

edalex-ian (approved):
Nice work @SammyIsConfused 

In future, we'll definitely want to try and make sure any new code is with the JPA API - as we don't want to add to the tech debt - but this time around I'm happy for this to proceed.

If you can make the tweak to the comment and the order of params I reckon merge on. :+1:

PR #2754, opened 2021-02-22
Fix another two File Uploader issues

PenghaiZhang:
#2559 

##### Checklist

- [ ] tests are included

##### Description of change

This PR has two fixes:
1. Making sure deleting an uploaded file also deletes the file's child files.
2. Making sure the Attachment number restriction counts for unzipped files.

https://user-images.githubusercontent.com/47203811/108668166-c80a5d80-752e-11eb-9e87-7b7853af3876.mp4

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/fileuploader/InlineFileUploader.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/fileuploader/InlineFileUploader.tsx
@@ -155,7 +155,7 @@ export const InlineFileUploader = ({
   );
   const [uploadingFiles, setUploadingFiles] = useState<UploadingFile[]>([]);
   const [attachmentCount, setAttachmentCount] = useState<number>(
-    entries.length
+    uploadedFiles.length
   );
   const [showDuplicateWarning, setShowDuplicateWarning] = useState<
     boolean | undefined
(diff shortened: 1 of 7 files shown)


ChristianMurphy (approved):

Source/Plugins/Core/com.equella.core/js/tsrc/fileuploader/InlineFileUploader.tsx
```diff
@@ -255,13 +255,21 @@ export const InlineFileUploader = ({
   const onReplace = (fileId: string) => openDialog(fileId, "");
 
   const onDelete = (file: UploadedFile) => {
-    const { id } = file.fileEntry;
+    const { id, children } = file.fileEntry;
     const onSuccessful = (displayWarningMessage = false) => {
-      setUploadedFiles(
-        deleteElement(uploadedFiles, generateUploadedFileComparator(id), 1)
+      // Child files must be deleted as well.
+      const remainingFiles = [id, ...children.map((child) => child.id)].reduce(
```
edalex-ian, line 261: I sat here looking at this `reduce` for awhile, and it feels like an anti-pattern. What you're really doing here is filtering, and as a result should really be just a `filter`. Roughed out like:

```typescript
const toRemove = [id, ...children.map((child) => child.id)];
const remainingFiles = uploadedFiles.filter( f => !(f.id in toRemove));
```

You get the idea - I've not checked any types etc here, but hey.

edalex-ian (changes requested):
As discussed, I think for readability sake we need to rework/replace the `reduce` block.

BTY.

edalex-ian (approved):

Issue #737, opened 2019-02-01
Complete new search page feature set


cathfitz:
Complete the new Search page by adding features currently available on the legacy search page (where logical) to create feature parity.

cathfitz:
Closing this ticket. There are a number of areas outstanding in 2020.2 such as gallery views, adding favourites and displaying MIME type filters which will be addressed in 2021.1. A new issue will be created for theses

PR #2755, opened 2021-02-25
feat(front-end): Add Module layer for Image Gallery searching (introducing fp-ts)

edalex-ian:
##### Checklist

- [x] tests are included

##### Description of change

The module layer to simply provide a `imageGallerySearch` function. The intention being the results from which can be mapped over to generate a gallery with possibly the MUI image grid.

The main work on this was really transforming the data from a normal search into that which would be useful for a gallery. This turned out somewhat tedious due to all the optionals in the bit I cared about. So I've finally taken the plunge and pulled in fp-ts so that I could have optionals (Option) and eithers (Either) - which also aligns nicely with the Scala work that is being done for all the new REST endpoints.

I've still _a lot_ to learn with fp-ts, however I believe what I've got is close to how it should be (while also using a minimal subset of the data types). I found that with the various type checking most code I wrote first go once getting a valid type check from the IDE - and mind you, it also highlighted several additional cases I needed to handle.

I tried to be generous with comments, and also extraneous type information, so I hope that helps.

I'm interested to hear what people think about the inclusion of fp-ts, as we obviously just removed the last part of core Purescript to ensure accessibility. So I don't want to regress on that standing. But I do feel this aligns reasonably with the back-end, and also still sits nicely in the land of Typescript.

If people require some additional reading, I can recommend:

* [Practical Guide to fp-ts (Part 1 through 6)](https://rlee.dev/writing/practical-guide-to-fp-ts-part-1)
* [fp-ts Do syntax](https://paulgray.net/do-syntax-in-typescript/)
  * Including full example at <https://paulgray.net/notes/do-validation/>
* [fp-ts Cheatsheet](Cheatsheet)
* And if you're looking for a general FP (vanilla JS focused) guide then you can't go past [Profession Frisby's Mostly Adequate Guide to Functional Programming](https://mostly-adequate.gitbook.io/mostly-adequate-guide/)
* And I'd be amiss to not mention the [fp-ts docs](https://gcanti.github.io/fp-ts/) - however they're rather lacking, so YMMV. (I do recommend perusing some of the external links though.)

#737 


ChristianMurphy (commented):
Thanks @edalex-ian!
It's interesting to see this in practice.

I'm definitely fond of fp-ts's ecosystem around data, [io-ts](https://github.com/gcanti/io-ts), [monacle-ts](https://github.com/gcanti/monocle-ts), [fast-check](https://github.com/dubzzz/fast-check). (or even better all three rolled together https://github.com/sledorze/morphic-ts)
It addresses a gap in JS/TS where data validation isn't always as simple as it should be.

But I'm a bit dubious about fp-ts for flow control, `Do`, `Task`, etc.
JavaScript has flow control operators and especially in newer versions with `.?`, `??`, etc the operators can be clear, concise, and mostly functional.
It feels like this is to get the syntax to be more PureScript like? We just got rid of PureScript, I'm not sure how I feel about bringing it back in a slightly different form.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/GallerySearchModule.ts
```diff
@@ -0,0 +1,337 @@
...
+import { buildFileAttachmentUrl } from "./AttachmentsModule";
+import { getImageMimeTypes } from "./MimeTypesModule";
+import { searchItems, SearchOptions } from "./SearchModule";
+
+/**
+ * The shape of data for entries to display in a Gallery, primarily consisting of links to access
+ * the asset and the associated MIME Type.
+ */
+export interface GalleryEntry {
+  mimeType: string;
```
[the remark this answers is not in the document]
edalex-ian: Yes, I probably should - I think I got a bit of tunnel vision on the tricky ones. :blush: 

Will do.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/GallerySearchModule.ts
```diff
@@ -0,0 +1,337 @@
...
+  pipe(
+    thumbnail,
+    O.fromNullable,
+    O.chain(
+      boolFold(
+        () => O.none,
+        () => O.some(true)
+      )
+    )
+  );
```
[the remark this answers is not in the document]
edalex-ian: It basically represents that it's optional as to whether there is a thumbnail. And by wrapping this in an optional it can then short circuit later operations/pipelines where a thumbnail is required. (Rather than having do an imperative if/else or ternary.)

Source/Plugins/Core/com.equella.core/js/tsrc/modules/GallerySearchModule.ts
```diff
@@ -0,0 +1,337 @@
...
+  type?: "thumbnail" | "preview"
+): string =>
+  pipe(
+    type,
+    O.fromNullable,
+    O.fold(
+      () => "",
+      () => `?gallery=${type}`
+    ),
+    (suffix: string) => attachment.links.thumbnail.concat(suffix)
```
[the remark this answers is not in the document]
edalex-ian: Yup, definitely a great example of excessive use.

I believe this came about from many iterations of reworking. It is important that where we can still use fairly declarative syntax (and indeed where suitable with the absence of things like Either and Option) we do. So the above suggestion there is bang on and I'll pull that in.

And this also acts as something which highlights the goodness of this _not_ being Purescript.

edalex-ian:
> Thanks @edalex-ian!
> It's interesting to see this in practice.
> 
> I'm definitely fond of fp-ts's ecosystem around data, [io-ts](https://github.com/gcanti/io-ts), [monacle-ts](https://github.com/gcanti/monocle-ts), [fast-check](https://github.com/dubzzz/fast-check). (or even better all three rolled together https://github.com/sledorze/morphic-ts)
> It addresses a gap in JS/TS where data validation isn't always as simple as it should be.
> 
> But I'm a bit dubious about fp-ts for flow control, `Do`, `Task`, etc.
> JavaScript has flow control operators and especially in newer versions with `.?`, `??`, etc the operators can be clear, concise, and mostly functional.
> It feels like this is to get the syntax to be more PureScript like? We just got rid of PureScript, I'm not sure how I feel about bringing it back in a slightly different form.

Heya @ChristianMurphy 

Thanks muchly for providing feedback.

Firstly, I believe this is still very different syntax to Purescript. There definitely are analogies, but that's more so just coming out of the FP side of things. As it turns out, as I was reading bits of fp-ts doco and examples there are lots of mention of Haskell and Purescript, however there is also mention of Scala - things like Do is very close to Haskell `do` but analogous to `for` comprehension in Scala.

Another thing though (and also highlighted in your review) is that unlike in Purescript we can still definitely (and should) use plain  Typescript where appropriate. We should aim to have nice clean simple pure functions (ideally with no required ADTs) feed into function compositions (with `pipe` or `flow`).

And although yes in modern JS we get various helpful operators, we still need to utilise imperative flow control - especially once you start getting `undefined` sprinkled around the place where you really don't want them. Plus you have to start using somewhat verbose/cumbersome type guards if you really want to do things properly. And that is the main driver for me in these (what should be) straight forward data transforms. (This is provided through the use of Either and Option and for involved data structures alongside Do.)

The ability to declaratively specify the new data structure without wandering down the lanes of imperative flow control.

As an aside though, I also feel it makes for code that is easier to comprehend and less prone to errors. When I mentioned most of this code worked first time (once I got past syntax errors), I was endorsing the benefits of the FP approach - not my ability to write code that works. It highlighted to me how good the guide rails of the type system was that a newbie could do that.

At this point though I'd also like to say that pragmatism is key here. I don't see us going and using this everywhere and I (currently) have no intention in going and wrapping all the async stuff in `TaskEither` etc. But for cases like this, I think it adds value. However I like to re-iterate that where suitable we still use TS.

Otherwise, we'd probably be better served with a pragmatic JS transpiled language like ReScript (which I've only recently learned ReasonML was renamed to).

Source/Plugins/Core/com.equella.core/js/tsrc/modules/GallerySearchModule.ts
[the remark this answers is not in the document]
doolse: WTF? Seriously 10 lines to do absoutely nothing. It should be O.Option<string> if anything. Why have you gotten rid of purescript to replace it with this rubbish? Seriously if you're going to using fp types you're going to need a real fp type system, use scalajs if you're comfortable with scala now. Otherwise just use ?? and .?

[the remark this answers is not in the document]
edalex-ian: Nah, this has already been re-written (removed). It was a bit of tunnel vision while still figuring out fp-ts.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/GallerySearchModule.ts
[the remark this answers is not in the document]
edalex-ian: As well as addressing this (and other excessive uses of boolean fold), I also combined some map operations and used `E.mapLeft` rather than excessive `E.fold`.

Thanks for your input on this.

ChristianMurphy:
> we can still definitely (and should) use plain Typescript where appropriate

Agreed

---

> The ability to declaratively specify the new data structure without wandering down the lanes of imperative flow control
> I also feel it makes for code that is easier to comprehend and less prone to errors

I'm not saying that these two are mutually exclusive.
But scala/haskell style control flow are not common in JS/TS at this time.

It is moving in that direction, more declarative flows are in the works, notably: https://github.com/tc39/proposal-pipeline-operator which has WIP/in-consideration TS support https://github.com/microsoft/TypeScript/pull/38305, hopefully this and other more declarative function styles will become more mainstream in the language, because `Do` and `pipe` both feel a bit awkward from a pure JS/TS perspective.

> Otherwise, we'd probably be better served with a pragmatic JS transpiled language like ReScript (which I've only recently learned ReasonML was renamed to).

Please not another compile to JS language which needs a [bridge to use any JS library](https://rescript-lang.org/docs/manual/latest/import-from-export-to-js).
Whether it's PureScript, ReScript, Scala.js, Elm, or \<insert new shiny functional language here\>.
If it can't leverage https://www.npmjs.com directly, it will probably cause more pain than it's worth.

edalex-ian:
> I'm not saying that these two are mutually exclusive.
> But scala/haskell style control flow are not common in JS/TS at this time.
>
> It is moving in that direction, more declarative flows are in the works, notably: https://github.com/tc39/proposal-pipeline-operator which has WIP/in-consideration TS support [microsoft/TypeScript#38305](https://github.com/microsoft/TypeScript/pull/38305), hopefully this and other more declarative function styles will become more mainstream in the language, because `Do` and `pipe` both feel a bit awkward from a pure JS/TS perspective.

Ah cool, that'd be exciting when that gets in. I look forward to the pattern matching and enums proposal too (I'm sure I saw them somewhere). 
 
> > Otherwise, we'd probably be better served with a pragmatic JS transpiled language like ReScript (which I've only recently learned ReasonML was renamed to).
> 
> Please not another compile to JS language which needs a [bridge to use any JS library](https://rescript-lang.org/docs/manual/latest/import-from-export-to-js).
> Whether it's PureScript, ReScript, Scala.js, Elm, or <insert new shiny functional language here>.
> If it can't leverage https://www.npmjs.com directly, it will probably cause more pain than it's worth.

Hahah, yes I totally agree. I was just saying, not suggesting we go that way. :wink:

Source/Plugins/Core/com.equella.core/js/tsrc/modules/AttachmentsModule.ts
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Build a direct URL to a file attachment.
+ *
+ * @param itemUuid The attachment item's UUID
+ * @param itemVersion The attachment item's version
+ * @param fileAttachmentPath The `filePath` provided by the server
+ */
+export const buildFileAttachmentUrl = (
+  itemUuid: string,
```
ChristianMurphy, line 28: not something for this PR, but a note which could be handy for the future.
the introduction of fp-ts could open the door for some more strict types.
For example Branded/Refinement types https://github.com/gcanti/io-ts/blob/master/index.md#branded-types--refinements
where the type itself could verify and note that it is a UUID.

ChristianMurphy (approved):
Thanks @edalex-ian!

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/modules/MimeTypesModule.test.ts
```diff
@@ -116,3 +120,26 @@ describe("getMimeTypeDefaultViewerDetails()", () => {
...
+
+describe("getImageMimeTypes()", () => {
+  const imagePrefix = "image/";
+  const mimeTypeEntries: OEQ.MimeType.MimeTypeEntry[] = [
+    "audio/ogg",
+    "image/jpeg",
+    "image/png",
+    "video/mp4",
+  ].map((s) => ({ mimeType: s, desc: `Testing MIME type ${s}` }));
+  const imageMimeTypeCount = mimeTypeEntries.filter((mte) =>
```
PenghaiZhang, line 132: Do we really need this ? Given that this array is so simple, it's quite obvious that the count is 2.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/GallerySearchModule.ts
```diff
@@ -0,0 +1,303 @@
...
+  itemUuid: string,
+  itemVersion: number,
+  attachments: O.Option<OEQ.Search.Attachment[]>
+): O.Option<GalleryEntry[]> =>
+  pipe(
+    attachments,
+    O.map(
+      flow(
+        A.map((attachment) =>
+          pipe(
```
PenghaiZhang, line 171: It's a bit hard to follow this big nested block. Some comments would be helpful!

Source/Plugins/Core/com.equella.core/js/tsrc/modules/GallerySearchModule.ts
```diff
@@ -0,0 +1,303 @@
...
+            buildGalleryEntry(itemUuid, itemVersion, attachment),
+            // Let's log any issues - just so we're aware of any odd data
+            E.mapLeft((error) => {
+              console.warn(error);
+              return error;
```
PenghaiZhang, line 176: Do we need to return the error ?

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -355,22 +355,24 @@ export const newSearchQueryToSearchOptions = async (
...
+    sortOrder,
+    collections,
+    rawMode,
+    lastModifiedDateRange,
+    owner,
+    status = liveStatuses,
+    searchAttachments,
+    selectedCategories,
+  }: SearchOptions,
+  mimeTypes?: string[]
```
PenghaiZhang, line 374: Why `mimeTypes` is not part of  `SearchOptions` ?

PenghaiZhang (commented):
Thanks Ian!!! Since we have walked through the major `fp-ts` code on Friday, I have some minor questions :)

Source/Plugins/Core/com.equella.core/js/tsrc/modules/AttachmentsModule.ts
edalex-ian: Yes, that'd be pretty neat. I've long wanted to see about tightening up the UUIDs we pass around the place. All I've done so far is place a type alias on them in the REST module and that at least still allows typescript-is to still parse them. But hmmm, we may have some further options now. :+1:

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/modules/GallerySearchModule.test.ts
```diff
@@ -0,0 +1,298 @@
...
+        ...attachment,
+        hasGeneratedThumb,
+      });
+      expectLeft(galleryEntry);
+    }
+  );
+});
+
+describe("buildGallerySearchResultItem", () => {
+  const searchItem: OEQ.Search.SearchResultItem = {
```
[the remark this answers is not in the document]
edalex-ian: I felt where possible it's handy to have the (smaller) test data right next to where it's being used. Only if it's bulky and cumbersome or used in more than one place do I feel the need to push out to a mock file.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/modules/GallerySearchModule.test.ts
```diff
@@ -0,0 +1,298 @@
...
+    links: {
+      view:
+        "https://example.com/inst/items/535e4e9b-4836-4011-8857-eb29260bf155/1/",
+      self:
+        "https://example.com/inst/api/item/535e4e9b-4836-4011-8857-eb29260bf155/1/",
+    },
+    isLatestVersion: true,
+  };
+
+  const passThroughAttachmentFilter: AttachmentFilter = (a) =>
```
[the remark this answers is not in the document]
edalex-ian: So `a`'s type is defined by `AttachmentFilter` and is `OEQ.Search.Attachment[]` so yes I could use `a.length` however I quite like the explicit readability of `isNonEmpty`. There can be no risk of accidental typo etc., so if you don't mind I might keep that around in this case.

But I can add the type definition to `a` if you like.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/modules/GallerySearchModule.test.ts
```diff
@@ -0,0 +1,298 @@
...
+    });
+    expectLeft(result);
+  });
+
+  // It would've been good to merge this with the above via an `it.each`, but unfortunately
+  // type issues were encountered so they're done long form.
+  it("fails if the attachments array is empty", () => {
+    const result = buildGallerySearchResultItem(passThroughAttachmentFilter)({
+      ...searchItem,
+      attachments: undefined,
```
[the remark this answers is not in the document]
edalex-ian: Whoops! This most certainly should be `[]`. The above test was the one for `undefined`.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/modules/GallerySearchModule.test.ts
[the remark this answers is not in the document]
PenghaiZhang: Yeah the readability is better. I know what the type it is so I don't need the type definition now :smile:

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/modules/MimeTypesModule.test.ts
edalex-ian: I guess we could. Potentially brittle if it ever changes - doubt it; more so I think I had plans of a rather set so probably just over-engineered. Will change to constant with comment.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/GallerySearchModule.ts
edalex-ian: Best to, as we're still in an Either and the intent here is just to log any issues. I acknowledge that down below we do then strip stuff, but that knowledge should not be coupled to here.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/GallerySearchModule.ts
edalex-ian: Hmmm, may help if I functionalise this a bit further. I'll have a play to see how I can improve the readability.

edalex-ian: As discussed, now refactored into smaller functions to compose together better.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/GallerySearchModule.ts
edalex-ian: Addressed also via the refactoring.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
edalex-ian: Yes, indeed, this has been done. :smile:

edalex-ian:
All feedback addressed. Will merge on green.

PenghaiZhang (approved):
Looks good to me!

PR #2780, opened 2021-03-02
Dedicate component for IconButton with Tooptip

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

In the previous work of adding/removing favourite items from Search page, the Heart icon was a nested component inside component SearchResult. This could result in some weird UX.

This PR adds a dedicate component for Icon Buttons that require a Tooltip.


Source/Plugins/Core/com.equella.core/js/tsrc/components/TooltipIconButton.tsx
```diff
@@ -0,0 +1,42 @@
...
+ */
+export const TooltipIconButton = ({
+  title,
+  onClick,
+  children,
+  size,
+}: TooltipIconButtonProps) => (
+  <Tooltip title={title}>
+    <IconButton onClick={onClick} aria-label={title} size={size}>
+      {children}
```
SammyIsConfused, line 39: Wondering if it's worth here having the icon itself as a parameter of TooltipIconButton, so that we don't need to pass in children. We could have a prop of type ReactElement that gets passed here and then you could use this like 
`<TooltipIconButton onClick={onClick} aria-label={title} size={size} icon={icon}/>`

SammyIsConfused (approved):
Looks good to me. Something to think about regarding whether or not we have children as a prop. Also - are there other tooltipped icon buttons that could be refactored to use this?

Source/Plugins/Core/com.equella.core/js/tsrc/components/TooltipIconButton.tsx
edalex-ian: I also though this could be good - and set `children: never` on the props. But then I thought,what if we need to pass in more than just an icon - say to do badged icons. :thinking:

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResult.tsx
```diff
@@ -303,34 +304,23 @@ export default function SearchResult({
...
-        ? favouriteItemStrings.title.remove
-        : favouriteItemStrings.title.add;
-      const icon = bookmarkId ? <FavoriteIcon /> : <FavoriteBorderIcon />;
-      const isAddedToFavourite = bookmarkId !== undefined;
-      const onConfirmProps = bookmarkId
-        ? favDialogConfirmToDelete
-        : favDialogConfirmToAdd;
+    const favouriteItemButtonText = bookmarkId
+      ? favouriteItemStrings.title.remove
+      : favouriteItemStrings.title.add;
```
edalex-ian, line 309: Rather than having these all disjointed, could be good to do something like:

```typescript
const favouriteItemButtonPops = {
  title: bookmarkId
      ? favouriteItemStrings.title.remove
      : favouriteItemStrings.title.add;
  onClick: () =>
      onFavouriteItem({
        isAddedToFavourite: bookmarkId !== undefined,
        isLatestVersion,
        onConfirmProps: bookmarkId
          ? favDialogConfirmToDelete
          : favDialogConfirmToAdd,
      });
}
```

So then below you can just:

```typescript
<TootipIconButton {...favouriteItemButtonText}>
  {bookmarkId ? (
    <FavoriteIcon />
  ) : (
    <FavoriteBorderIcon />
  )}
</TooltipIconButton>
```

And I think with that - and due to the introduction of the component - readability is maintained.

edalex-ian (commented):
Looks pretty good @PenghaiZhang . As (I think) @SammyIsConfused mentioned, would it be much effort to go and update a few other main places where this could be used?

Also, just a minor suggestion to consider.

Source/Plugins/Core/com.equella.core/js/tsrc/components/TooltipIconButton.tsx
PenghaiZhang: Why not use two of this component  if we need two icons ?

edalex-ian: Looks like for Badged ones (like `attachFileBadge()`) that's actually done outside the `<Tooltip>` anyways. :shrug:

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResult.tsx
PenghaiZhang: Good idea! I have refactored this bit.

PenghaiZhang:
> Looks pretty good @PenghaiZhang . As (I think) @SammyIsConfused mentioned, would it be much effort to go and update a few other main places where this could be used?
> 
> Also, just a minor suggestion to consider.

I have updated four places: one in `ResourceSelector`, two in `SearchResultList` and one in `Template`

edalex-ian (approved):
LGTM! :+1: 

(Well, other than the build being red.)

PenghaiZhang:
> LGTM! 
> 
> (Well, other than the build being red.)

It was because of old problem `Browser may have died`

edalex-ian:
Ah that ol' chestnut. Drats. :disappointed: 

I see it's running again, so :crossed_fingers:

PR #2781, opened 2021-03-02
Search page MIME type filter

PenghaiZhang:
##### Checklist

- [x] tests are included

##### Description of change

This PR adds a new tsx component `MimeTypeFilterSelector` and integrates it into `SearchPage`. The filter selector supports multiple selections and filtering options by keywords. Tests and Storybook are also included.

https://user-images.githubusercontent.com/47203811/109731702-e26bc780-7c0f-11eb-92fa-54aedd4222c7.mp4


Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
```diff
@@ -380,6 +388,31 @@ describe("Refine search by Owner", () => {
...
+    const filters = getMimeTypeFilters;
+    const page = await renderSearchPage();
+    userEvent.click(page.getByLabelText(helperText));
+    filters.forEach((filter) => {
+      userEvent.click(screen.getByText(filter.name));
+    });
+    await waitForSearch();
+    expect(SearchModule.searchItems).toHaveBeenLastCalledWith({
+      ...defaultSearchPageOptions,
+      mimeTypes: filters.flatMap((f) => f.mimeTypes),
```
PenghaiZhang, line 404: Please note that IDE is not happy with this `flatMap`. The error is : TS2339: Property 'flatMap' does not exist on type 'MimeTypeFilter[]'.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/MimeTypeFilterSelector.tsx
```diff
@@ -0,0 +1,89 @@
...
+  /**
+   * All configured MIME type filters.
+   */
+  filters: MimeTypeFilter[];
+}
+
+const { helperText } = languageStrings.searchpage.mimeTypeFilterSelector;
+/**
+ * This component displays a list of configured MIME type filters which can be selected to
+ * filter search results by Mime Types. It supports filtering options by typing keywords.
```
edalex-ian, line 45:
```suggestion
 * filter search results by MIME Types. It supports filtering options by typing keywords.
```

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/settings/SearchFilterResource.scala
```diff
@@ -135,6 +133,7 @@ class SearchFilterResource {
     responseContainer = "List"
   )
   def batchUpdate(searchFilters: Array[SearchFilter]): Response = {
+    searchPrivProvider.checkAuthorised()
```
edalex-ian, line 136: Oooh, nice pick-up. :+1:

edalex-ian (changes requested):
Looks good @PenghaiZhang , just two things:

1. It looks as though when sharing a search the MIME types - not the filter IDs - are being stored, but this'd cause issues then selecting the filters in the UI no?; and
2. I think we need to do something with the state for storing the MIME types from the server, and I'm leaning towards having one bit of state for Search Page Settings - as they all have the same lifecycle.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
```diff
@@ -98,6 +100,11 @@ const mockConvertParamsToSearchOptions = jest.spyOn(
   SearchModule,
   "queryStringParamsToSearchOptions"
 );
+
+const mockMimeTypeFilters = jest
+  .spyOn(SearchFilterSettingsModule, "getMimeTypeFiltersFromServer")
+  .mockResolvedValue(getMimeTypeFilters);
```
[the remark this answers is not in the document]
PenghaiZhang: What about moving it to the test case and then do `mockMimeTypeFilters.mockResolvedValueOnce(...);` there ?
`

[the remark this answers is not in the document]
edalex-ian: Yup, that's exactly what I was talking about - as that's what we do with the others. And depends on the test as to whether you do a `Once` or not.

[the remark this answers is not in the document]
PenghaiZhang: We need this to ensure `renderSearchPage` does not throw any error. Without this mock, we will set lots of `network error` as the test is calling the real implementation of `getMimeTypeFiltersFromServer`

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/modules/SearchModule.test.ts
```diff
@@ -153,7 +153,7 @@ describe("SearchModule", () => {
...
+        '{"rowsPerPage":10,"currentPage":0,"sortOrder":"NAME","query":"test machine","rawMode":true,"status":["LIVE","REVIEW"],"searchAttachments":true,"selectedCategories":[{"id":766943,"categories":["Hobart"]},{"id":766944,"categories":["Some cool things"]}],"collections":[{"uuid":"8e3caf16-f3cb-b3dd-d403-e5eb8d545fff"},{"uuid":"8e3caf16-f3cb-b3dd-d403-e5eb8d545ffe"},{"uuid":"8e3caf16-f3cb-b3dd-d403-e5eb8d545ffg"},{"uuid":"8e3caf16-f3cb-b3dd-d403-e5eb8d545ffa"},{"uuid":"8e3caf16-f3cb-b3dd-d403-e5eb8d545ffb"}],"lastModifiedDateRange":{"start":"2020-05-26T03:24:00.889Z","end":"2020-05-27T03:24:00.889Z"},"owner":{"id":"680f5eb7-22e2-4ab6-bcea-25205165e36e"}, "mimeTypes": ["Image/png"]}';
```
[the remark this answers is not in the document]
PenghaiZhang: I actually have another PR which has changes to support conversion between query string and search option. The key change is to convert to search page option.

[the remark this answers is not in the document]
PenghaiZhang: According to the implementation of `generateQueryStringFromSearchOptions`, I can confirm that both `mimeTypes` and `mimeTypeFilters` are included in the query string. And then looking to `newSearchQueryToSearchOptions`, although we explicitly return a `SearchOptions`, `mimeTypeFilters` is also returned. As a result, `SearchPage` knows what filters to select.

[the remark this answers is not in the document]
PenghaiZhang: Using the filter ID is also OK - this is how Old UI does.

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -201,6 +211,8 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
     searchSettings,
     setSearchSettings,
   ] = useState<OEQ.SearchSettings.Settings>();
+  const [mimeTypeFilters, setMimeTypeFilters] = useState<MimeTypeFilter[]>([]);
```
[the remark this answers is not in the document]
PenghaiZhang: The code you suggested is more consistent with back-end -  SearchFilter is an inner class of SearchSetting. 

So I will go this way.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
PenghaiZhang: I can add a `ts-ignore` to get IDE happy...

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
```diff
@@ -469,70 +457,56 @@ describe("Collapsible refine filter section", () => {
 });
 
 describe("Hide Refine Search controls", () => {
-  let page: RenderResult;
```
PenghaiZhang, line 439: While moving the test for MIME type filter selector visibility into this describe, I found tests in this describe can be simplified a lot...so I have made changes for these tests.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -202,6 +207,12 @@ export const defaultSearchOptions: SearchOptions = {
   mimeTypes: [],
 };
 
+export const defaultSearchPageOptions: SearchPageOptions = {
```
PenghaiZhang, line 210: This variable was originally defined inside `SearchPage`. The first change I did was moving it to outside `SearchPage` but still in that tsx file. Then when I ran tests, this variable caused some trouble as `defaultSearchOptions` became undefined. So I moved it to here and then all trouble was gone.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -267,15 +278,16 @@ export const queryStringParamsToSearchOptions = async (
...
   // Else if the query params contain params from legacy `searching.do` (i.e. `LegacySearchParams`) then convert to
   // `searchOptions` with `legacyQueryStringToSearchOptions`.
   // For all else, return `undefined`.
   if (searchOptions) {
-    return await newSearchQueryToSearchOptions(searchOptions);
+    return await newSearchQueryToSearchPageOptions(searchOptions);
   } else if (
     Array.from(params.keys()).some((key) => LegacySearchParams.guard(key))
   ) {
+    //todo: Support converting legacy query string to SearchPageOptions. Once this done, change the return type of this functio
```
PenghaiZhang, line 290: This will be done in my next PR .

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -324,16 +342,28 @@ const rehydrateOwner = async (
...
+const rehydrateMIMETypeFilter = async (
+  options: DehydratedSearchOptions
+): Promise<MimeTypeFilter[] | undefined> => {
+  const selectedFilters = options.mimeTypeFilters;
+  if (!selectedFilters) {
+    return undefined;
+  }
+  const filterIDs = selectedFilters.map((f) => f.id);
+  const mimeTypeFilters = await getMimeTypeFiltersFromServer();
+  return mimeTypeFilters.filter((f) => f.id && filterIDs.includes(f.id));
```
ChristianMurphy, line 354:
```suggestion
  return mimeTypeFilters.filter(({ id }) => id && filterIDs.includes(id));
```

edalex-ian (commented):
Good work on simplifying the hiding of selectors @PenghaiZhang :+1: 

All looks good, except for the change around `defaultSearchPageOptions`. I'm not sure I entirely follow the why or wherefore there - so I can't really suggest alternatives - but I am concerned with the coupling of the lower layers to the UI.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -34,9 +34,14 @@ import {
   Unknown,
 } from "runtypes";
 import { API_BASE_URL } from "../AppConfig";
+import { SearchPageOptions } from "../search/SearchPage";
```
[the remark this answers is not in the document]
PenghaiZhang: > Good work on simplifying the hiding of selectors @PenghaiZhang 
> 
> All looks good, except for the change around `defaultSearchPageOptions`. I'm not sure I entirely follow the why or wherefore there - so I can't really suggest alternatives - but I am concerned with the coupling of the lower layers to the UI.

The implementation of defaultSearchPageOptions is 
```
{
  ...defaultSearchOptions,
  dateRangeQuickModeEnabled: true,
  mimeTypeFilters: [],
}
```

The problem is in the context of tests, `defaultSearchOptions` is undefined if we do  `import defaultSearchPageOptions from SearchPage`.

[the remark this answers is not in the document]
PenghaiZhang: As discussed, let's move `mimeTypeFilters` to `SearchOptions`.

PenghaiZhang:
Since we have moved 'mimeTypeFilters' to 'SearchOptions', I found it much easier to support shared searches that have MIME types. So I also include the support of legacy shared searches in this PR.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -442,6 +438,13 @@ const findCollectionsByUuid = async (
   return filteredCollectionList.length > 0 ? filteredCollectionList : undefined;
 };
 
+const findMIMETypeFiltersById = async (
```
edalex-ian, line 441: I wonder if perhaps this function should be `memoized` to reduce the calls to `getMimeTypeFiltersFromServer()`. :thinking: We don't want to memoize `getMimeTypeFiltersFromServer` though as we need that for configuration on the settings page I think.

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -182,6 +177,11 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
   const location = useLocation();
 
   const [state, dispatch] = useReducer(reducer, { status: "initialising" });
+  const defaultSearchPageOptions: SearchPageOptions = {
+    ...defaultSearchOptions,
+    dateRangeQuickModeEnabled: true,
+    mimeTypeFilters: [],
```
edalex-ian, line 183: Should this be specified here or as part of `defaultSearchOptions`?

edalex-ian (approved):
I think it's now good to go, but just two queries for your consideration.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
PenghaiZhang: As discussed, we don't need to memorize this function since it only gets called once when opening a new browser tab with shared search link.

PR #2789, opened 2021-03-05
Update ci.yaml to use postgres 13.2

SammyIsConfused:
##### Checklist

##### Description of change
This PR is just to test how GHA goes running the test suite against PostgreSQL 13.2 (current latest version). Recently when working on something else, I installed openEQUELLA with this version and it seemed to work. 
If this  goes well, I can take it out of draft and we can have GHA test against latest postgres from now on. If it doesn't go well, then we know that there are changes that need to be made to support this version. 

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -188,7 +188,7 @@ jobs:
 
     services:
       postgres:
-        image: postgres:9.6
+        image: postgres:13.2
         env:
           POSTGRES_USER: equellatests
           POSTGRES_PASSWORD: password
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -314,7 +314,7 @@ jobs:
 
     services:
       postgres:
-        image: postgres:9.6
+        image: postgres:latest
         env:
           POSTGRES_USER: equellatests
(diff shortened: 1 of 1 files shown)


SammyIsConfused:
Flaked on GHA on PackageAndNavigationTest-modifyingNavigation in the new UI. Looks to be that openEQUELLA supports PostgreS 13.2 yay!
Rebuilding so we can get a fully green build.

edalex-ian:
I'm not sure we only want to run out build against the latest and greatest. Maybe we want to bracket the builds between the [oldest supported version (currently 9.6)](https://www.postgresql.org/support/versioning/) and the newest.

We could use the matrix to do this, but I'm concerned having a 2x2 (newui x pgsql) matrix is excessive for this. We'd probably be happy with:

- newui with `latest`
- newui with `9.6`
- legacyui with either (maybe lean on the old side)

This could be done with

```yaml
    strategy:
      matrix:
        newui: [true, false]
        pgsql-image: [latest, 9.6]
        exclude:
          - newui: false
            pgsql-image: latest
      fail-fast: false
```

edalex-ian (approved):

ChristianMurphy (approved):

PR #2798, opened 2021-03-10
Support updating classification by mime types

PenghaiZhang:
##### Checklist

- [x] tests are included

##### Description of change

To support the feature of filtering classifications by MIME types, this PR updates the classification endpoint and REST module. A Jest test is also added, which requires updating testing institution `facet`.


Source/Plugins/Core/com.equella.core/src/com/tle/web/api/search/EquellaSearchResource.java
```diff
@@ -147,6 +148,7 @@
           @QueryParam("advancedsearch")
           String advancedSearch,
       @ApiParam(value = "An ID (not a username) of a user", required = false) @QueryParam("owner")
-          String owner);
+          String owner,
+      @ApiParam("A list of MIME types") @QueryParam("mimeTypes") List<String> mimeTypes);
```
PenghaiZhang, line 152: I wonder if we still need to maintain this old search endpoint.

PenghaiZhang:
Front-end changes are quite simple so I wonder if we can include them in this PR. But I am also happy to open another PR.

edalex-ian (changes requested):
So there's no explicit change to the front-end, so how does this work? Is it because we were already simply passing the whole SearchOptions into the classifications search?

Other than that query, just a couple of things to address.

edalex-ian (approved):
GTG! :+1:

PR #2802, opened 2021-03-11
Support adding a favourite search

PenghaiZhang:
##### Checklist

- [x] tests are included

##### Description of change

This PR, I added a new REST endpoint to support adding a favourite search. I also updated the endpoint for favourite item (from `favourite` to `favourite/item`). Then, I added the support to the REST Module for the new endpoint.

During testing, I encountered the issue of viewing a favourite search in new Search UI. Since the fix seems easy to implement, I added the fix in this PR.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/favourite/FavouriteResource.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/favourite/FavouriteResource.scala
@@ -33,18 +33,24 @@ import javax.ws.rs.{DELETE, POST, Path, PathParam, Produces, QueryParam}
 import scala.collection.JavaConverters._
 
 /**
-  * Provide basic information of a favourite Item.
+  * Model class for Items to be saved to user's favourites.
   * @param keywords Tags of this Favourite Item
   * @param isAlwaysLatest Whether this Favourite Item uses latest Item version
   * @param itemID ID of the Item
   * @param bookmarkID ID of the related Bookmark
   */
-case class FavouriteItem(itemID: String,
-                         keywords: Array[String],
-                         isAlwaysLatest: Boolean,
-                         bookmarkID: Long)
+case class FavouriteItemModel(itemID: String,
+                              keywords: Array[String],
+                              isAlwaysLatest: Boolean,
+                              bookmarkID: Long)
 
-case class FavouriteSearchInfo(name: String, url: String)
+/**
+  * Model class for search definitions to be saved to user's favourites.
+  * @param id ID of a search definition. The value is None before the search definition persists to DB.
+  * @param name Name of a search definition.
+  * @param url Path to new Search UI, including all query strings.
+  */
+case class FavouriteSearchModel(id: Option[Long], name: String, url: String)
 
 @Path("favourite")
 @Produces(Array("application/json"))
@@ -57,8 +63,8 @@ class FavouriteResource {
   @Path("/item")
   @ApiOperation(value = "Add one Item to user's favourites",
                 notes = "This operation is essentially adding a new bookmark.",
-                response = classOf[FavouriteItem])
-  def addFavouriteItem(favouriteItem: FavouriteItem): Response = {
+                response = classOf[FavouriteItemModel])
+  def addFavouriteItem(favouriteItem: FavouriteItemModel): Response = {
     // ItemNotFoundException will be thrown by itemService if there is no Item matching this
     // item ID so we don't validate item ID here again.
     val item = itemService.get(new ItemId(favouriteItem.itemID))
@@ -67,7 +73,7 @@ class FavouriteResource {
     Response
       .status(Status.CREATED)
       .entity(
-        FavouriteItem(
+        FavouriteItemModel(
           newBookmark.getItem.getItemId.toString,
           newBookmark.getKeywords.asScala.toArray,
           newBookmark.isAlwaysLatest,
@@ -97,8 +103,8 @@ class FavouriteResource {
   @POST
   @Path("/search")
   @ApiOperation(value = "Add a search definition to user's search favourites",
-                response = classOf[FavouriteSearchInfo])
-  def addFavouriteSearch(searchInfo: FavouriteSearchInfo): Response = {
+                response = classOf[FavouriteSearchModel])
+  def addFavouriteSearch(searchInfo: FavouriteSearchModel): Response = {
     val favouriteSearch = new FavouriteSearch
     favouriteSearch.setName(searchInfo.name)
     favouriteSearch.setUrl(searchInfo.url)
@@ -109,7 +115,10 @@ class FavouriteResource {
 
     Response
       .status(Status.CREATED)
-      .entity(FavouriteSearchInfo(newFavouriteSearch.getName, newFavouriteSearch.getUrl))
+      .entity(
+        FavouriteSearchModel(Option(newFavouriteSearch.getId),
+                             newFavouriteSearch.getName,
+                             newFavouriteSearch.getUrl))
       .build()
   }
 }
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/favourite/FavouriteResource.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/favourite/FavouriteResource.scala
@@ -19,6 +19,10 @@
 package com.tle.web.api.favourite
 
 import com.tle.beans.item.ItemId
+import com.tle.common.institution.CurrentInstitution
+import com.tle.common.usermanagement.user.CurrentUser
+import com.tle.core.favourites.bean.FavouriteSearch
+import java.util.Date
 import com.tle.legacy.LegacyGuice
 import com.tle.web.api.ApiErrorResponse
 import io.swagger.annotations.{Api, ApiOperation, ApiParam}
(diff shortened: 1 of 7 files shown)


Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/favourite/FavouriteResource.scala
```diff
@@ -86,4 +91,15 @@ class FavouriteResource {
           .resourceNotFound(s"No Bookmark matching ID: ${id}")
     }
   }
+
+  @POST
+  @Path("/search")
+  @ApiOperation(value = "Add one search to user's favourites", response = classOf[FavouriteSearch])
+  def addFavouriteSearch(favouriteSearch: FavouriteSearch): Response = {
```
edalex-ian, line 98: Typically we want to avoid the direct use of the Entity class for the input and output of the REST endpoints. That way the implementation is not directly coupled to the data layer, we don't accidentally expose anything internal, and  we can add API doco (which in this case we currently completely lack).

autotest/OldTests/src/test/java/io/github/openequella/rest/FavouriteApiTest.java
```diff
@@ -41,11 +43,32 @@ public void testAddFavourite() throws IOException {
...
+    body.put("name", searchName);
+    body.put("url", searchPath);
+    method.setRequestEntity(new StringRequestEntity(body.toString(), "application/json", "UTF-8"));
+
+    assertEquals(HttpStatus.SC_CREATED, makeClientRequest(method));
+    JsonNode response = mapper.readTree(method.getResponseBody());
+    assertNotNull(response.get("id"));
+    assertNotNull(response.get("owner"));
+    assertNotNull(response.get("dateModified"));
+    assertNotNull(response.get("institution"));
```
edalex-ian, line 70: An example of value I'm not sure needs to be returned to a caller - considering all API calls already target an institution.

edalex-ian (commented):
Thanks @PenghaiZhang 

A few tweaks, but mainly I think we need to define a model for the REST call here rather than simply re-using the entity.

Source/Plugins/Core/com.equella.core/src/com/tle/core/favourites/service/FavouriteSearchServiceImpl.java
```diff
@@ -53,8 +54,9 @@
 
   @Override
   @Transactional
-  public void save(FavouriteSearch search) {
+  public FavouriteSearch save(FavouriteSearch search) {
     dao.save(search);
+    return search;
```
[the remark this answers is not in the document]
PenghaiZhang: Good idea. I can call `dao.getById(id) ` to get the new entry.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/favourite/FavouriteResource.scala
PenghaiZhang: As discussed, let'a have a case class which has two fields: name and url.

ChristianMurphy (approved):
TS code LTGM :+1:

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/favourite/FavouriteResource.scala
```diff
@@ -44,6 +44,8 @@ case class FavouriteItem(itemID: String,
                          isAlwaysLatest: Boolean,
                          bookmarkID: Long)
 
+case class FavouriteSearchInfo(name: String, url: String)
```
edalex-ian, line 47: A few things:

1. Needs some doco - including Swagger annotations;
2. How about including an Optional ID parameter to be used on response; and
3. 'Info' is rather generic (and almost superfluous), how about Model?

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/favourite/FavouriteResource.scala
```diff
@@ -94,12 +96,20 @@ class FavouriteResource {
...
+    favouriteSearch.setUrl(searchInfo.url)
     favouriteSearch.setInstitution(CurrentInstitution.get())
     favouriteSearch.setDateModified(new Date())
     favouriteSearch.setOwner(CurrentUser.getUserID)
     val newFavouriteSearch = LegacyGuice.favouriteSearchService.save(favouriteSearch)
-    Response.status(Status.CREATED).entity(newFavouriteSearch).build()
+
+    Response
+      .status(Status.CREATED)
+      .entity(FavouriteSearchInfo(newFavouriteSearch.getName, newFavouriteSearch.getUrl))
```
edalex-ian, line 112: As mentioned above, be most useful to at least return the resulting ID of the new entity/record. So if you update the model to hold that, that'd be great - otherwise most important would be to just return the ID (on the premise the contract is that whatever you ask for is created).

Source/Plugins/Core/com.equella.core/src/com/tle/core/favourites/service/FavouriteSearchServiceImpl.java
```diff
@@ -73,12 +74,20 @@ public void executeSearch(SectionInfo info, long id) {
...
+      // If new Search UI is disabled but the search was added from new Search UI, throw an error.
+      // Else do however it used to do.
+      if (url != null) {
+        if (RenderNewTemplate.isNewSearchPageEnabled()) {
+          // Remove the last '/' from 'CurrentInstitution.get().getUrl()' and then construct a full
+          // path.
+          String fullPath = StringUtils.removeEnd(CurrentInstitution.get().getUrl(), "/") + url;
+          info.forwardToUrl(fullPath);
+          return;
+        } else if (url.contains("/page/search")) {
```
edalex-ian, line 87: hmmm, this extra nesting made me zoom out and review the whole method. And it seems slightly odd.

My first suggestion to solve the nesting was going to be change the opening to:

```java
FavouriteSearch search = dao.getById(id);
if(search == null || search.url == null) {
  return
}
```

But what caught my attention is that at the end of this method, it seems that even if URL is null it will still do:

```java
SectionInfo forward = info.createForwardForUri(url);
info.forwardAsBookmark(forward);
```

It would seem wrong to forward to a null URI, but if this has been there all this time... :confused: 

So I don't know, but I am inclined to make the above change which will then reduce all this nesting and indeed simplify the logic statements.

What do you think though? Could it break a desire to forward to null as then it forwards back to the root of the institution on error?

edalex-ian (commented):
Cheers @PenghaiZhang , I think that's improvement there.

But sorry, I do have a few more queries and suggestions.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/favourite/FavouriteResource.scala
PenghaiZhang: Good ideas!!!

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/favourite/FavouriteResource.scala
PenghaiZhang: I have added the resulting ID to the response.

image

Source/Plugins/Core/com.equella.core/src/com/tle/core/favourites/service/FavouriteSearchServiceImpl.java
PenghaiZhang: There are two things I found.

Firstly,  the value of `url` should never be null as it's NOT nullable. This means when we save a null URL we will get a Hibernate error. I guess whoever wrote this code might think it's redundant to check whether url is null or not here.

Secondly, if `url` is an empty string or whatever path that does not have a Section handling it, we will see `no tree for id` error.

PenghaiZhang: But if `url`  somehow becomes null here, we will see NPE.

So in inclusion, I think we should do 
```
if(search == null || search.url == null) {
  return
}
```

PenghaiZhang: I think the best way is to throw an exception.
```
throw new SectionsRuntimeException("This favourite search is invalid.");
```

Source/Plugins/Core/com.equella.core/src/com/tle/core/favourites/service/FavouriteSearchServiceImpl.java
```diff
@@ -71,42 +72,43 @@ public void deleteById(long id) {
...
+    // Else do however it used to do.
+    if (RenderNewTemplate.isNewSearchPageEnabled()) {
+      // Remove the last '/' from 'CurrentInstitution.get().getUrl()' and then construct a full
+      // path.
+      String fullPath = StringUtils.removeEnd(CurrentInstitution.get().getUrl(), "/") + path;
+      info.forwardToUrl(fullPath);
+      return;
+    } else if (path.contains("/page/search")) {
+      throw new SectionsRuntimeException(
+          "This favourite search is only available in new Search Page");
```
edalex-ian, line 93:
```suggestion
          "This favourite search is only available in New UI mode.");
```

edalex-ian (approved):
Looks good @PenghaiZhang , and thanks for fixing up the old badness. :+1: 

Optionally, could you please apply the rewording?

But merge when ready.

PR #2803, opened 2021-03-12
Support restricted attachments properly in the new Search UI

SammyIsConfused:
##### Checklist

- [x] tests are included

##### Description of change
The new search UI doesn't handle restricted attachments properly - they will still appear in a search, although the file itself will be inaccessible. 
Added to the convertToAttachment function to filter out these attachments when the user does not have VIEW_RESTRICTED_ATTACHMENTS granted.

### DIFF
--- a/Source/Plugins/Core/com.equella.base/src/com/tle/web/api/item/interfaces/beans/AttachmentBean.java
+++ b/Source/Plugins/Core/com.equella.base/src/com/tle/web/api/item/interfaces/beans/AttachmentBean.java
@@ -31,6 +31,7 @@ public abstract class AttachmentBean extends AbstractExtendableBean {
   private String viewer;
   private boolean preview;
   private boolean erroredIndexing;
+  private boolean restricted;
 
   public String getUuid() {
     return uuid;
@@ -65,6 +66,14 @@ public abstract class AttachmentBean extends AbstractExtendableBean {
 
   public void setPreview(boolean preview) {
     this.preview = preview;
+  }
+
+  public void setRestricted(boolean restricted) {
+    this.restricted = restricted;
+  }
+
+  public boolean isRestricted() {
+    return restricted;
   }
 
   /**
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/components/ItemAttachmentLink.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/components/ItemAttachmentLink.tsx
@@ -19,6 +19,7 @@ import { Link } from "@material-ui/core";
 import * as React from "react";
 import { SyntheticEvent, useState } from "react";
 import { ViewerDefinition } from "../modules/ViewerModule";
+import { languageStrings } from "../util/langstrings";
 import Lightbox from "./Lightbox";
 
 export interface ItemAttachmentLinkProps {
@@ -57,6 +58,7 @@ const ItemAttachmentLink = ({
   viewerDetails: [viewer, url],
 }: ItemAttachmentLinkProps) => {
   const [showLightbox, setShowLightbox] = useState<boolean>(false);
+  const { attachmentLink } = languageStrings.searchpage.searchResult;
 
   const buildLightboxLink = (): JSX.Element => {
     if (!mimeType) {
@@ -68,7 +70,7 @@ const ItemAttachmentLink = ({
     return (
       <>
         <Link
-          aria-label="Attachment link"
+          aria-label={`${attachmentLink} ${description}`}
           component="button"
           onClick={(event: SyntheticEvent) => {
             setShowLightbox(!showLightbox);
@@ -95,7 +97,7 @@ const ItemAttachmentLink = ({
   ) : (
     // Lightbox viewer not specified, so go with the default of a simple link.
     <Link
-      aria-label="Attachment link"
+      aria-label={`${attachmentLink} ${description}`}
       href={url}
       target="_blank"
       rel="noreferrer"
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/components/ItemAttachmentLink.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/components/ItemAttachmentLink.tsx
@@ -68,6 +68,7 @@ const ItemAttachmentLink = ({
     return (
       <>
         <Link
+          aria-label="Attachment link"
           component="button"
           onClick={(event: SyntheticEvent) => {
             setShowLightbox(!showLightbox);
@@ -93,7 +94,12 @@ const ItemAttachmentLink = ({
     buildLightboxLink()
   ) : (
     // Lightbox viewer not specified, so go with the default of a simple link.
-    <Link href={url} target="_blank" rel="noreferrer">
+    <Link
+      aria-label="Attachment link"
+      href={url}
+      target="_blank"
+      rel="noreferrer"
+    >
       {children}
     </Link>
   );
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResult.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResult.tsx
@@ -433,11 +433,7 @@ export default function SearchResult({
                 viewerDetails?.viewerId
               )}
             >
-              <ListItemText
-                title={description}
-                color="primary"
-                primary={description}
-              />
+              <ListItemText color="primary" primary={description} />
             </ItemAttachmentLink>
             {inSelectionSession && (
               <ListItemSecondaryAction>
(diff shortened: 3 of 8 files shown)


SammyIsConfused:
Need to fix up NewSearchPageTest to account for this (was planning to add some tests for this anyway).

edalex-ian:
No probs, will wait until your finish that and request review before I circle back around.

SammyIsConfused:
Should be ready now @edalex-ian. The initial problem was that the `attachmentBeans.removeIf(a -> a.isRestricted)` was running even when `attachmentBeans` was null, which broke the test and threw a null pointer. Fixed that, and also added tests to verify that: 
- for a user that doesn't have VIEW_RESTRICTED_ATTACHMENTS the attachment doesn't appear in a search at all
- for a user that does have VIEW_RESTRICTED_ATTACHMENTS the attachment does appear in a search

SammyIsConfused:
No more edits - just fixed up the description of the test.

SammyIsConfused:
Weird, GHA doesn't want to spin up this time round :thinking:. I'll reword my test description again, see if that wakes it up.

SammyIsConfused:
Alright, no idea what happened there, but we have GHA again. Once again ready for review :+1:

SammyIsConfused:
Alright, no idea what happened there, but we have GHA again. Once again ready for review :+1:

Source/Plugins/Core/com.equella.base/src/com/tle/web/api/item/interfaces/beans/AttachmentBean.java
```diff
@@ -67,6 +68,14 @@ public void setPreview(boolean preview) {
     this.preview = preview;
   }
 
+  public void setRestricted(boolean restricted) {
```
PenghaiZhang, line 71: May I ask where this method gets called ?  I image it would be somewhere in `AttachmentSerializerProvider - serializeAttachment()` .

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
```diff
@@ -217,6 +219,12 @@ object SearchHelper {
     */
   def convertToAttachment(attachmentBeans: java.util.List[AttachmentBean],
                           itemKey: ItemIdKey): Option[List[SearchResultAttachment]] = {
+    // Filter out restricted attachments if user doesn't have permission to view them
+    if (CollectionUtils.isNotEmpty(attachmentBeans) && LegacyGuice.aclManager
+          .filterNonGrantedPrivileges(AttachmentConfigConstants.VIEW_RESTRICTED_ATTACHMENTS)
+          .isEmpty) {
+      attachmentBeans.removeIf(a => a.isRestricted)
```
PenghaiZhang, line 226: An alternative is changing the type of `attachmentBeans` to List[AttachmentBean], and then we can do
```
val beans = if(no permission) attachmentBeans.filter(a => !a.isRestricted) else attachmentBeans
Option(beans).map(...)
```

autotest/OldTests/src/test/java/io/github/openequella/search/NewSearchPageTest.java
```diff
@@ -88,4 +88,33 @@ public void searchWithLessACLS() {
     searchPage.selectCollection("Hardware");
     searchPage.waitForSearchCompleted(7);
   }
+
+  @Test(
+      description =
+          "Search with two differently privileged users when there is a restricted attachment",
+      dependsOnMethods = "searchWithLessACLS")
```
PenghaiZhang, line 95: If this test case does not really depend on `searchWithLessACLS`, I think it's better to remove this dependency.

PenghaiZhang (commented):
Looks good Sam. Just have a few questions.

Source/Plugins/Core/com.equella.base/src/com/tle/web/api/item/interfaces/beans/AttachmentBean.java
SammyIsConfused: That's correct, yeah. It iwas initially added to EquellaAttachmentBean, but we use one level up at AttachmentBean for this, so I added it. In the debugger I could already see this value in AttachmentBean, we just didn't have the getter and setter to access it.

SammyIsConfused:
Passes locally but fails on CI :thinking: weird. I'll have another look this morning.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
SammyIsConfused: As discussed in the office, I moved the filtering inside the Option so that we could get rid of the null check and use it as a Scala list.

autotest/OldTests/src/test/java/io/github/openequella/search/NewSearchPageTest.java
SammyIsConfused: Agreed. Removed the dep.

SammyIsConfused:
Codebuild failed to pull down package.json dependencies which is a bit weird. Not seen this before. Retrying build. 
image

Source/Plugins/Core/com.equella.core/js/tsrc/components/ItemAttachmentLink.tsx
```diff
@@ -93,7 +94,12 @@ const ItemAttachmentLink = ({
     buildLightboxLink()
   ) : (
     // Lightbox viewer not specified, so go with the default of a simple link.
-    <Link href={url} target="_blank" rel="noreferrer">
+    <Link
+      aria-label="Attachment link"
```
edalex-ian, line 98: This should still be language stringed. Also, can be useful to keep them meaningful if there's lots. To do that here, I suggest you append the `description` when available. And seeing this is duplicated, probably throw it into a constant.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResult.tsx
```diff
@@ -433,7 +433,11 @@ export default function SearchResult({
                 viewerDetails?.viewerId
               )}
             >
-              <ListItemText color="primary" primary={description} />
+              <ListItemText
+                title={description}
```
edalex-ian, line 437: Why's this added?

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
```diff
@@ -220,6 +222,12 @@ object SearchHelper {
     Option(attachmentBeans).map(
       beans =>
         beans.asScala
+        // Filter out restricted attachments if the user does not have permissions to view them
+          .filter(
+            a =>
+              !a.isRestricted || !LegacyGuice.aclManager
+                .filterNonGrantedPrivileges(AttachmentConfigConstants.VIEW_RESTRICTED_ATTACHMENTS)
+                .isEmpty)
```
edalex-ian, line 230: It seems excessive to call `LegacyGuice.aclManager                .filterNonGrantedPrivileges(AttachmentConfigConstants.VIEW_RESTRICTED_ATTACHMENTS).isEmpty` for each attachment when it's not related _to_ the attachment. How about just throw it in a constant and then reference here?

(Added bonus that it'll then also have a name which is more readable.)

autotest/OldTests/src/test/java/io/github/openequella/pages/search/NewSearchPage.java
```diff
@@ -188,6 +188,41 @@ public void selectOwner(String ownerName) {
...
 
+  /**
+   * Returns a By for the aria-label attribute of an attachment link in a search result, given the
+   * attachment title.
+   *
+   * @param attachmentText The text to search against to find the name of the attachment within
+   *     search results.
+   * @return a By for the attachmentText.
+   */
+  private By xPathForAttachmentText(String attachmentText) {
```
edalex-ian, line 199: Bit of an odd name there, would expect something along the lines of:

```suggestion
  private By attachmentLinkByText(String text) {
```

edalex-ian (commented):
Looks good @SammyIsConfused , just a couple of queries and suggestions.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResult.tsx
SammyIsConfused: Left over from a previous xpath query. Sorry about that. Removed.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
SammyIsConfused: As discussed, moved to using lazy val instead, so this permissions check only gets called if a restricted attachment is part of this item.

edalex-ian (approved):
Looks good @SammyIsConfused ! Merge on. :+1:

PR #2819, opened 2021-03-17
Front-end support for adding favourite search from New Search UI

PenghaiZhang:
##### Checklist

- [x] tests are included

##### Description of change

This PR is all about the UI of saving a search to user's favourites.

Firstly, a new dialog named `FavouriteSearchDialog` was created and integrated into `SearchPage` .

Secondly, a Heart Icon Button was added to `SearchResultList`. By clicking this button, the new dialog should appear.  

Thirdly, I changed how we control the snack bar so that once a search is successfully saved, the snack bar pops up.

Lastly, Jest tests and Storybooks were added.

https://user-images.githubusercontent.com/47203811/111399616-d0168100-8719-11eb-9962-d66d3f7a993e.mp4


Source/Plugins/Core/com.equella.core/js/tsrc/modules/FavouriteModule.ts
```diff
@@ -47,3 +47,20 @@ export const addFavouriteItem = (
...
+  name: string,
+  url: string
+): Promise<OEQ.Favourite.FavouriteSearchModel> =>
+  OEQ.Favourite.addFavouriteSearch(API_BASE_URL, {
+    name,
+    url,
+  }).then(
+    (newFavouriteSearch: OEQ.Favourite.FavouriteSearchModel) =>
+      newFavouriteSearch
+  );
```
ChristianMurphy, line 66: This `then` appears to be a no-operation return?
Would:

```suggestion
  });
```

work?

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/components/FavouriteSearchDialog.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { render, screen } from "@testing-library/react";
+import "@testing-library/jest-dom/extend-expect";
+import userEvent from "@testing-library/user-event";
+import * as React from "react";
+import { act } from "react-dom/test-utils";
```
ChristianMurphy, line 22: it may make sense to import this from testing library react https://testing-library.com/docs/react-testing-library/api#act

ChristianMurphy (commented):
General design question, could the positioning of the favorite over the results, rather than over the search filers cause confusion?
Could it be interpreted as saving the results as showing as a collection, rather than saving the filters the created the results?

Source/Plugins/Core/com.equella.core/js/tsrc/modules/FavouriteModule.ts
PenghaiZhang: Yes! Thanks.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/components/FavouriteSearchDialog.test.tsx
PenghaiZhang: Yes we should import this from `@testing-library/react` !

PenghaiZhang:
> General design question, could the positioning of the favorite over the results, rather than over the search filers cause confusion?
> Could it be interpreted as saving the results as showing as a collection, rather than saving the filters the created the results?

I am not quite sure. @edalex-ian @cathfitz  What do you think  ?

edalex-ian:
> > General design question, could the positioning of the favorite over the results, rather than over the search filers cause confusion?
> > Could it be interpreted as saving the results as showing as a collection, rather than saving the filters the created the results?
> 
> I am not quite sure. @edalex-ian @cathfitz What do you think ?

I think it's a developing area. We discussed this a little bit and the main thing is that the favouriting of a search is a similar action to the sharing of a search (the icon next to it). However soon we'll be adding another button which is to download the search results to a CSV file.

That may potentially further muddy the waters, but also end up with a significant number of buttons so at the point we may need to look at this closer and come up with an improved UI.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
```diff
@@ -879,3 +887,46 @@ describe("Add and remove favourite Item,", () => {
...
+    userEvent.click(heartIcon);
+
+    const dialog = page.getByRole("dialog");
+    const searchNameInput = queryMuiTextField(
+      dialog,
+      languageStrings.searchpage.favouriteSearch.text
+    );
+    if (!searchNameInput) {
+      throw new Error("Failed to find TextField in FavouriteSearchDialog.");
+    }
```
edalex-ian, line 910: This is calling out that we need a `getMuiTextField`. I think this should be refactored over into the `MuiQueries`.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
```diff
@@ -879,3 +887,46 @@ describe("Add and remove favourite Item,", () => {
...
+    userEvent.type(searchNameInput, "test");
+
+    const confirmButton = dialog.querySelector(
+      "#confirm-dialog-confirm-button"
+    );
+    if (!confirmButton) {
+      throw new Error(
+        "Failed to find the Confirm button in FavouriteSearchDialog."
+      );
+    }
```
edalex-ian, line 920: Should this really just be `expect(confirmButton).toBeInDocument()`?

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/components/FavouriteSearchDialog.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+    const dialog = screen.getByRole("dialog");
+    const confirmButton = queryMuiButtonByText(
+      dialog,
+      languageStrings.common.action.ok
+    );
+    if (!confirmButton) {
+      throw new Error(
+        "Failed to find the Confirm button in FavouriteSearchDialog"
+      );
+    }
```
edalex-ian, line 48: As mentioned for the text field query, looks like this should be refactored over into `MuiQueries` as `getMuiButtonByText`.

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -199,10 +209,10 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
...
       defaultSearchPageHistory.filterExpansion
   );
-  const [
-    showSearchCopiedSnackBar,
-    setShowSearchCopiedSnackBar,
-  ] = useState<boolean>(false);
+  const [messageInfo, setMessageInfo] = useState<{
+    open: boolean;
+    message: string;
+  }>(defaultMessageInfo);
```
edalex-ian, line 215: I wonder if you need this much complexity. I wonder if all you need is the message itself.

Then as far as the open prop down in MessageInfo that could just be based on the presence of a message. And so onClose would simply clear the message back to `""` or `undefined`.

If that was done, this is then simple `snackBarMessage, setSnackBarMessage`.

What do you think?

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -775,8 +821,11 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
           {renderSidePanel()}
         </Drawer>
       </Hidden>
-      {favouriteDialogProps.open && (
-        <FavouriteItemDialog {...favouriteDialogProps} />
+      {favouriteItemDialogProps.open && (
+        <FavouriteItemDialog {...favouriteItemDialogProps} />
+      )}
+      {favouriteSearchDialogProps.open && (
+        <FavouriteSearchDialog {...favouriteSearchDialogProps} />
```
edalex-ian, line 828: Now with two dialogs and two bits of state to manage them - and ideally they're mutually exclusive (we never really want more than one dialog displayed) - and the knowledge there'll be possibly more (say, download to CSV maybe). Should we push this into a SearchPageDialogs component and perhaps use a reducer in there?

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/FavouriteSearchDialog.tsx
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Provide a Dialog where a search definition can be added to user's favourites.
+ */
+export const FavouriteSearchDialog = ({
+  open,
+  onConfirm,
+  closeDialog,
+}: FavouriteSearchDialogProps) => {
+  const [searchName, setSearchName] = useState<string>("");
+  const confirmHandler = () => onConfirm(searchName).finally(closeDialog);
```
edalex-ian, line 57: Should we have a `catch` here and handle any error? Arguably if there is an error, could be good to keep the dialog displayed so they can try again....

edalex-ian (changes requested):
Looks good @PenghaiZhang , but I feel a few pieces of refactoring are called for - with moving some get queries into MuiQueries and better managing the proliferation of dialogs and their state on the SearchPage.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/search/SearchPage.test.tsx
PenghaiZhang: Yes, but I changed it to `expect(confirmButton).not.toBeNull();`

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
PenghaiZhang: Yes this is definitely better. I have done this way for the Lightbox but forgot here..

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
PenghaiZhang: Good idea! I have created a new component `SearchPageDialog` which is the entry to other dialogs.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/FavouriteSearchDialog.tsx
PenghaiZhang: When an error is triggered, previous search result gets cleared. Would it look a bit strange to keep the dialog ? 
image

edalex-ian: The idea would be to keep the dialog up and have it report the error of the activity it just tried. (And possibly have the 'ok' button change to 'retry' or something...)

But as highlighted in the above screenshot, the error is being handled with the default `handleError` snackbar, AND all the other dialog in oEQ kind of work in the fashion of closing regardless. So... I'm happy to leave as is.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchPageDialog.tsx
```diff
@@ -0,0 +1,80 @@
...
+    | {
+        type: "search";
+        props: FavouriteSearchDialogSpecificProps;
+      };
+}
+
+/**
+ * Provide an entry to different Dialogs rendered in Search page.
+ */
+export const SearchPageDialog = ({
```
edalex-ian, line 55: Seeing this is not just for a single dialog, I think that needs to be highlighted. So its name could be along the lines of:

* `SearchPageDialogs` or
* `SearchPageDialogsSwitch` or
* `SearchPageDialogsController`

or something else along those lines. :thinking:

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchPageDialog.tsx
```diff
@@ -0,0 +1,80 @@
...
+  closeDialog: () => void;
+}
+
+export interface SearchPageDialogProps extends SearchPageDialogCommonProps {
+  /**
+   * Additional props required by a certain Dialog to be rendered in Search page.
+   */
+  additionalDialogProps?:
+    | {
+        type: "item";
```
edalex-ian, line 43: Seeing there could be other non-favourite related dialogs, how about a bit of simple prefix namespacing for these discriminators, like:

* `fav-item`; and
* `fav-search`.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchPageDialog.tsx
```diff
@@ -0,0 +1,80 @@
...
+        props: FavouriteItemDialogSpecificProps;
+      }
+    | {
+        type: "search";
+        props: FavouriteSearchDialogSpecificProps;
+      };
+}
+
+/**
+ * Provide an entry to different Dialogs rendered in Search page.
```
edalex-ian, line 53:
```suggestion
 * Manage the various dialogs which can be displayed on the search page, including ensuring only one is displayed at a time.
```

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchPageDialog.tsx
```diff
@@ -0,0 +1,80 @@
...
+    case "search":
+      return (
+        <FavouriteSearchDialog
+          open={open}
+          closeDialog={closeDialog}
+          {...additionalDialogProps.props}
+        />
+      );
+    default:
+      throw new Error("Unexpected Dialog type");
```
edalex-ian, line 78:
```suggestion
      throw new TypeError(`Unexpected Dialog type: [${additionalDialogProps?.type}]`);
```

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchPageDialog.tsx
```diff
@@ -0,0 +1,80 @@
...
+   * Fired when the dialog is closed
+   */
+  closeDialog: () => void;
+}
+
+export interface SearchPageDialogProps extends SearchPageDialogCommonProps {
+  /**
+   * Additional props required by a certain Dialog to be rendered in Search page.
+   */
+  additionalDialogProps?:
```
edalex-ian, line 41: If they're _required_ why are the optional?

edalex-ian (commented):
Nice work on the new component @PenghaiZhang . Just some queries and suggestions around it, then should be good to go. :+1:

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchPageDialog.tsx
PenghaiZhang: I will go with `SearchPageDialogsSwitch`

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchPageDialog.tsx
PenghaiZhang: Because when we are not going to show a dialog, we don't need this props. Or maybe another question: what would be the default value of this props which is used in the first rendering of search page?

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchPageDialog.tsx
PenghaiZhang: Actually, the type of `additionalDialogProps?.type` is `never` here because `type` must be either `fav-item` or `fav-search`. 

To make this working, I have added a third string literal `unknown`, which is also the default value as this props.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchPageDialog.tsx
PenghaiZhang: As mentioned above, since I use `unknown` as the default value, this property is not optional now.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchPageDialog.tsx
edalex-ian: Aaaah! That's right, of course, this is just for the exhaustive check. Yeah, don't add the additional literal, just use TypeError and keep the string as was.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchPageDialog.tsx
edalex-ian: I wonder if rather than `unknown` you should simply go with `none`? Mind you, now that's just sounding like an optional working around `undefined`.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchPageDialogsSwitch.tsx
```diff
@@ -0,0 +1,87 @@
...
+        type: "fav-item";
+        props: FavouriteItemDialogSpecificProps;
+      }
+    | {
+        type: "fav-search";
+        props: FavouriteSearchDialogSpecificProps;
+      }
+    | {
+        type: "unknown";
+      };
```
ChristianMurphy, line 53: does `unknown` mean it could have something, but we're not sure what it is?
Or does it mean there will not be additional props?

If the latter, would it make sense to rename it to `type: none`?

ChristianMurphy (approved):

edalex-ian (approved):
Nice refactor @PenghaiZhang . I think it's there, let's merge it. :+1:

PR #2820, opened 2021-03-19
Allow login notice editor to upload images

SammyIsConfused:
##### Checklist

##### Description of change
The folder (loginnoticeimages/) was missing from the getMimeType call which meant that uploading images to the login notice editor would error out. Added this to the image path for that call to fix it. 
I am open to spending time to add a test for this, currently we have a test for using an image URL but not uploading direct, which is how this went unnoticed. 
#2833 
Before change:

https://user-images.githubusercontent.com/24543345/111723724-23263a80-88b8-11eb-8e17-75e65dfc053f.mp4

After change:

https://user-images.githubusercontent.com/24543345/111723998-92039380-88b8-11eb-8598-d3bac016e6b9.mp4

### DIFF
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/loginnotice/impl/LoginNoticeServiceImpl.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/loginnotice/impl/LoginNoticeServiceImpl.java
@@ -169,7 +169,8 @@ public class LoginNoticeServiceImpl implements LoginNoticeService {
   public String getMimeType(String name) throws IOException {
     CustomisationFile customisationFile = new CustomisationFile();
     if (fileSystemService.fileExists(customisationFile, LOGIN_NOTICE_IMAGE_FOLDER_NAME + name)) {
-      return fileSystemService.getMimeType(customisationFile, name);
+      return fileSystemService.getMimeType(
+          customisationFile, LOGIN_NOTICE_IMAGE_FOLDER_NAME + name);
     }
     return MediaType.MEDIA_TYPE_WILDCARD;
   }
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/loginnotice/impl/LoginNoticeServiceImpl.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/loginnotice/impl/LoginNoticeServiceImpl.java
@@ -153,13 +153,13 @@ public class LoginNoticeServiceImpl implements LoginNoticeService {
     String nameWithoutExtension = FilenameUtils.removeExtension(name);
     String extension = '.' + FilenameUtils.getExtension(name);
     if (fileSystemService.fileExists(customisationFile, getLoginNoticeImageFileName(name))) {
-      int i = 1;
-      while (fileSystemService.fileExists(
-          customisationFile,
-          getLoginNoticeImageFileName(nameWithoutExtension + '_' + i + extension))) {
-        i++;
-      }
-      return nameWithoutExtension + '_' + i + extension;
+      int i = 0;
+      String uniqueFileName;
+      do {
+        uniqueFileName = (nameWithoutExtension + '_' + (++i) + extension);
+      } while (fileSystemService.fileExists(
+          customisationFile, getLoginNoticeImageFileName(uniqueFileName)));
+      return uniqueFileName;
     }
     return name;
   }
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/loginnotice/impl/LoginNoticeServiceImpl.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/loginnotice/impl/LoginNoticeServiceImpl.java
@@ -152,9 +152,7 @@ public class LoginNoticeServiceImpl implements LoginNoticeService {
     CustomisationFile customisationFile = new CustomisationFile();
     String nameWithoutExtension = FilenameUtils.removeExtension(name);
     String extension = '.' + FilenameUtils.getExtension(name);
-    String concatenatedName = nameWithoutExtension + extension;
-    if (fileSystemService.fileExists(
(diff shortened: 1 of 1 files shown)


Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/loginnotice/impl/LoginNoticeServiceImpl.java
```diff
@@ -169,7 +169,8 @@ private String iterateImageNameIfDuplicateExists(String name) {
   public String getMimeType(String name) throws IOException {
     CustomisationFile customisationFile = new CustomisationFile();
     if (fileSystemService.fileExists(customisationFile, LOGIN_NOTICE_IMAGE_FOLDER_NAME + name)) {
-      return fileSystemService.getMimeType(customisationFile, name);
+      return fileSystemService.getMimeType(
+          customisationFile, LOGIN_NOTICE_IMAGE_FOLDER_NAME + name);
```
edalex-ian, line 173: To avoid this kind of issue again, it'd make sense to create a constant up between 170 and 171 such as:

```java
final String fileName = LOGIN_NOTICE_IMAGE_FOLDER_NAME + name;
```

Then you can use that in both places.

edalex-ian (approved):
Looks good, but now that you have a duplicate string construction - and seeing that was also the source of the problem - it'd make sense to drop that in a constant and use it instead.

But other than that, GTG.

Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/loginnotice/impl/LoginNoticeServiceImpl.java
SammyIsConfused: There's actually a few other places in this file that have this, outside this method. So I'll make a method along the lines of `private String getLoginNoticeImageFileName(String fileName);` that does this and apply it wherever we have a `LOGIN_NOTICE_IMAGE_FOLDER_NAME + name` in this file.

Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/loginnotice/impl/LoginNoticeServiceImpl.java
```diff
@@ -144,42 +144,42 @@ public String uploadPreLoginNoticeImage(InputStream imageFile, String name) thro
...
-        customisationFile, LOGIN_NOTICE_IMAGE_FOLDER_NAME + nameToUse, imageFile, false);
+        customisationFile, getLoginNoticeImageFileName(nameToUse), imageFile, false);
     return nameToUse;
   }
 
   private String iterateImageNameIfDuplicateExists(String name) {
     CustomisationFile customisationFile = new CustomisationFile();
     String nameWithoutExtension = FilenameUtils.removeExtension(name);
     String extension = '.' + FilenameUtils.getExtension(name);
+    String concatenatedName = nameWithoutExtension + extension;
```
edalex-ian, line 155: This reads rather odd. It's like we have the name, then get the prefix, get the suffix (extension), then join them back together and call it some thing else - even though wouldn't that just be `name`?

edalex-ian (commented):
Good move @SammyIsConfused , but... There seems to be some odd logic there. :thinking:

Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/loginnotice/impl/LoginNoticeServiceImpl.java
SammyIsConfused: So this function is what iterates image names if an existing one is in the filestore. We need the `nameWithoutExtension` and the `extension` so that we can `nameWithoutExtension + '_' + i + extension` when naming the new file. 
We can just use `name` in all but the lines that deal with `i`.  Fixing that up.

Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/loginnotice/impl/LoginNoticeServiceImpl.java
```diff
@@ -144,41 +144,40 @@ public String uploadPreLoginNoticeImage(InputStream imageFile, String name) thro
...
-        customisationFile, LOGIN_NOTICE_IMAGE_FOLDER_NAME + nameWithoutExtension + extension)) {
+    if (fileSystemService.fileExists(customisationFile, getLoginNoticeImageFileName(name))) {
       int i = 1;
       while (fileSystemService.fileExists(
           customisationFile,
-          LOGIN_NOTICE_IMAGE_FOLDER_NAME + nameWithoutExtension + '_' + i + extension)) {
+          getLoginNoticeImageFileName(nameWithoutExtension + '_' + i + extension))) {
         i++;
       }
       return nameWithoutExtension + '_' + i + extension;
```
edalex-ian, line 162: Thanks for the tidy-up, but unfortunately that now highlights duplication of the filename format. Could the contents of this `if` block be:

```java
int i = 0;
String uniqueFilename;
do {
  uniqueFilename = getLoginNoticeImageFileName(nameWithoutExtension + '_' + (++i) + extension);
} while(fileSystemService.fileExists(customisationFile, uniqueFilename));
return uniqueFilename;
```

edalex-ian (commented):
Almost there @SammyIsConfused . It's definitely looking more robust than before, however just one more bit of duplication is highlighted.

Source/Plugins/Core/com.equella.core/src/com/tle/core/settings/loginnotice/impl/LoginNoticeServiceImpl.java
SammyIsConfused: Close - I just needed to move the `getLoginNoticeImageFileName` call into the `fileExists` - as thats the only place that needs the surrounding folder. The name itself is just a name. 
That's done now.

edalex-ian (approved):
Good stuff, ship it! :+1:

PR #2822, opened 2021-03-19
View previous or next attachment in Lightbox

PenghaiZhang:
##### Checklist

- [x] tests are included

##### Description of change

This PR adds the support of viewing previous or next attachment in the Lightbox.

The approach is to update the LightboxProps to take two more handlers which are fired when the two arrow icon buttons are clicked.

The two handlers are built in component 'ItemAttachmentLink` whose props is also updated to take a list of Lightbox viewable attachments.

I also updated the code of determining which viewer to be used for each attachment in `SearchResult`.

https://user-images.githubusercontent.com/47203811/111734235-d0f01400-88cd-11eb-8253-b866fdc29589.mp4


Source/Plugins/Core/com.equella.core/js/tsrc/components/Lightbox.tsx
```diff
@@ -106,17 +152,19 @@ const Lightbox = ({ mimeType, onClose, open, src, title }: LightboxProps) => {
     </Card>
   );
 
+  // Build an nested component which gets unmounted and mounted again when Lightbox re-renders.
+  // This ensures when navigate to previous/next image, the current image disappears immediately.
+  const LightBoxImage = () => (
```
edalex-ian, line 157: I wonder if instead o needing this, you could simply get away with having the below `const content` in state, and then you have an handleOnNext and handleOnPrev which clears `content` before they call the `onNext` and `onPrev`. That way the functioning is explicit, and you wouldn't need the comment. :thinking:

edalex-ian (commented):
I'm not sure on this one @PenghaiZhang .

I like what you've done with the `Lightbox` component, however I'm not so sure the approach for `ItemAttachmentLink` and `SearchResult` is quite the way to go. With seeing what you need to hold onto, `ItemAttachmentLink` has gone from concerned with a single attachment to a multi-headed hydra.

It feels as though we need to split the attachment handling in general out of `SearchResult` and possibly introduce a `SearchResultAttachmentsList` component which could then handle all this kind of thing.

:thinking: 

Let's discuss further.

Source/Plugins/Core/com.equella.core/js/tsrc/components/Lightbox.tsx
PenghaiZhang: This is really a good idea! To do this, I need one more `useEffect` to update the content.

Source/Plugins/Core/com.equella.core/js/tsrc/components/Lightbox.tsx
```diff
@@ -96,57 +129,93 @@ const Lightbox = ({ mimeType, onClose, open, src, title }: LightboxProps) => {
...
-    <Card>
-      <CardContent>
-        <Typography variant="h5" component="h2">
-          {labelUnsupportedContent}
-        </Typography>
-      </CardContent>
-    </Card>
-  );
+  const [content, setContent] = useState<ReactNode | undefined>();
+  const [lightBoxConfig, setLightBoxConfig] = useState<LightboxConfig>(config);
```
edalex-ian, line 133: Just one thought (but no action required) is that these are closely related and would make a good candidate for a reducer. But seeing they're also simple, I can see the pros of leaving them just like this - and hope there is no increase in complexity.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResultAttachmentsList.tsx
```diff
@@ -0,0 +1,342 @@
...
+  item: OEQ.Search.SearchResultItem;
+  getViewerDetails: (
+    mimeType: string
+  ) => Promise<OEQ.MimeType.MimeTypeViewerDetail>;
+  /**
+   * Error handler for standard management of errors during processing - especially comms errors.
+   */
+  handleError: (error: Error) => void;
+}
+export const SearchResultAttachmentsList = ({
```
edalex-ian, line 103: Please have a blank line around definitions of functions.

```suggestion

export const SearchResultAttachmentsList = ({
```

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResultAttachmentsList.tsx
```diff
@@ -0,0 +1,342 @@
...
+  const handleAttachmentPanelClick = (event: SyntheticEvent) => {
+    /** prevents the SearchResult onClick from firing when attachment panel is clicked */
+    event.stopPropagation();
+    setAttachExpanded(!attachExpanded);
+  };
+
+  const lightboxAttachments = attachmentsAndViewers.filter(
+    (av) => av.viewer[0] === "lightbox"
+  );
+  const buildLightboxNavigationHandler = (
```
edalex-ian, line 209: Please have a blank line around definitions of functions.

```suggestion

  const buildLightboxNavigationHandler = (
```

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResultAttachmentsList.tsx
```diff
@@ -0,0 +1,342 @@
...
+  };
+
+  const handleAttachmentPanelClick = (event: SyntheticEvent) => {
+    /** prevents the SearchResult onClick from firing when attachment panel is clicked */
+    event.stopPropagation();
+    setAttachExpanded(!attachExpanded);
+  };
+
+  const lightboxAttachments = attachmentsAndViewers.filter(
+    (av) => av.viewer[0] === "lightbox"
```
edalex-ian, line 207: Readability could improve with de-structuring.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResultAttachmentsList.tsx
```diff
@@ -0,0 +1,342 @@
...
+  );
+  const buildLightboxNavigationHandler = (
+    anotherAttachmentIndex: number
+  ): (() => LightboxConfig) | undefined => {
+    if (
+      anotherAttachmentIndex > -1 &&
+      anotherAttachmentIndex < lightboxAttachments.length
+    ) {
+      const previousAttachment = lightboxAttachments[anotherAttachmentIndex];
```
edalex-ian, line 216: De-structuring here with renaming and defaults will help below readability, and give you straight up variables named `src`, `title` and `mimeType`.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResultAttachmentsList.tsx
```diff
@@ -0,0 +1,342 @@
...
+    /** prevents the SearchResult onClick from firing when attachment panel is clicked */
+    event.stopPropagation();
+    setAttachExpanded(!attachExpanded);
+  };
+
+  const lightboxAttachments = attachmentsAndViewers.filter(
+    (av) => av.viewer[0] === "lightbox"
+  );
+  const buildLightboxNavigationHandler = (
+    anotherAttachmentIndex: number
```
edalex-ian, line 210: The naming of this variable is unclear. Please rename or add some TSDoc. :thinking:

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResultAttachmentsList.tsx
```diff
@@ -0,0 +1,342 @@
...
+            {inStructured ? <DragIndicatorIcon /> : <InsertDriveFile />}
+          </ListItemIcon>
+          <ItemAttachmentLink
+            selectedAttachment={attachmentAndViewer}
+            // Need the index in 'lightboxAttachments' to make sure the next or previous attachment must be viewable in Lightbox.
+            onPrevious={buildLightboxNavigationHandler(
+              lightboxAttachments.findIndex((v) => v.attachment.id === id) - 1
+            )}
+            onNext={buildLightboxNavigationHandler(
+              lightboxAttachments.findIndex((v) => v.attachment.id === id) + 1
```
edalex-ian, line 254: To keep it DRY, how about doing this just up above the return statement (line 234) and dropping it in a constant like `lightboxAttachmentIndex`?

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResultAttachmentsList.tsx
```diff
@@ -0,0 +1,342 @@
...
+  const lightboxAttachments = attachmentsAndViewers.filter(
+    (av) => av.viewer[0] === "lightbox"
+  );
+  const buildLightboxNavigationHandler = (
+    anotherAttachmentIndex: number
+  ): (() => LightboxConfig) | undefined => {
+    if (
+      anotherAttachmentIndex > -1 &&
+      anotherAttachmentIndex < lightboxAttachments.length
+    ) {
```
edalex-ian, line 215: This could be simplified with the help of `fp-ts` `Array` by something like:

    if(isSome(A.lookup(anotherAttachmentIndex, lightboxAttachments)))

or perhaps more readable with:

    if(pipe(lightboxAttachments, A.lookup(anotherAttachmentIndex), isSome))

edalex-ian (changes requested):
Great refactor @PenghaiZhang .

Almost halved SearchResult. Nice work. :+1: 

Few bits of feedback, but really just tidy up bits now.

Thanks!

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResultAttachmentsList.tsx
PenghaiZhang: With fp-ts I don't even need this if :smile:

Source/Plugins/Core/com.equella.core/js/tsrc/components/Lightbox.tsx
PenghaiZhang: Yeah I image the reducer type would have 'init', 'onNext' and 'onPrevious' :thinking:

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/components/Lightbox.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+      config: LightboxConfig,
+      handler: jest.Mock,
+      buttonText: string
+    ) => {
+      const { getByLabelText } = render(
+        <Lightbox onClose={jest.fn()} open config={config} />
+      );
+
+      userEvent.click(getByLabelText(buttonText));
+      expect(handler).toHaveBeenCalledTimes(1);
```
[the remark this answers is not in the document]
PenghaiZhang: Good idea!!!

Source/Plugins/Core/com.equella.core/js/tsrc/components/ItemAttachmentLink.tsx
```diff
@@ -18,30 +18,28 @@
...
 
 export interface ItemAttachmentLinkProps {
   /**
    * Content to be surrounded by the link.
    */
   children: React.ReactNode;
   /**
-   * Optional description that will be passed to chosen viewer - e.g. The viewer provided by
-   * the `<Lightbox/>` component.
+   * The attachment to be viewed in the LightBox.
```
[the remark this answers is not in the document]
PenghaiZhang: OK I will reword this.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResultAttachmentsList.tsx
```diff
@@ -176,11 +220,85 @@ export const SearchResultAttachmentsList = ({
...
+        pipe(
+          lightboxAttachments,
+          A.lookup(attachmentIndex),
+          O.fold(
+            () => undefined,
+            (av) => {
+              const {
+                attachment: { description, mimeType },
+                viewerDefinition: [viewer, viewUrl],
+              } = av;
```
edalex-ian, line 247: How about just repacing `(av) => {` with the de-structure?

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResultAttachmentsList.tsx
```diff
@@ -176,11 +220,85 @@ export const SearchResultAttachmentsList = ({
...
+                  src: viewUrl,
+                  title: description,
+                  mimeType: mimeType ?? "",
+                  onNext: buildLightboxNavigationHandler(attachmentIndex + 1),
+                  onPrevious: buildLightboxNavigationHandler(
+                    attachmentIndex - 1
+                  ),
+                });
+              }
+              throw new TypeError("Unexpected viewer configuration");
```
edalex-ian, line 259:
```suggestion
              throw new TypeError(`Unexpected viewer configuration: ${viewer}`);
```

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResultAttachmentsList.tsx
```diff
@@ -176,11 +220,85 @@ export const SearchResultAttachmentsList = ({
...
+
+      const lightboxAttachments = attachmentsAndViewerDefinitions.filter(
+        ({ viewerDefinition: [viewer] }) => viewer === "lightbox"
+      );
+
+      /**
+       * Build a function to handler navigation between Lightbox attachments.
+       * @param attachmentIndex Index of the attachment to be viewed
+       */
+      const buildLightboxNavigationHandler = (
```
edalex-ian, line 235: Just like `transform()` it could be good to move this function outside of this self calling async function to try and keep the final logic succinct.

As it is, this whole effect is now around 150 lines long. You almost have to wonder if it'd be better to move this two functions (`transform` and `buildLightboxNavigationHandler`) out into the `ViewerModule` (may need a new name for `transform`) - perhaps there may be somewhere better.... Because as it stands, they're both stand-alone and not dependent on anything specific to this component, and then things may be more readable.

What do you think?

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResultAttachmentsList.tsx
```diff
@@ -0,0 +1,429 @@
...
+  viewerType: "lightbox";
+  /**
+   * Config required by Lightbox.
+   */
+  config: LightboxConfig;
+}
+
+export type ViewerConfig = ViewerLinkConfig | ViewerLightboxConfig;
+
+export const isViewerLightConfig = (
```
edalex-ian, line 109:
```suggestion
export const isViewerLightboxConfig = (
```

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResultAttachmentsList.tsx
```diff
@@ -0,0 +1,429 @@
...
+export interface AttachmentAndViewerConfig {
+  /**
+   * The details of an attachment.
+   */
+  attachment: OEQ.Search.Attachment;
+  /**
+   * Viewer configuration for the attachment.
+   */
+  viewerConfig: ViewerConfig;
+}
```
edalex-ian, line 133: I can't help and wonder whether the above interfaces and functions belong here in this component, or more so generically over in the ViewerModule. I'm referring to:

* ViewerLinkConfig
* ViewerLightboxConfig
* ViewerConfig
* isViewerLightConfig
* AttachmentAndViewerDefinition
* AttachmentAndViewerConfig

As they're not part of the props for the component, and although they're building up those types possibly in the effect, is that kind of data processing etc more so what the Module layer should be looking after? Then we try and keep components simple and focused on display.

edalex-ian (commented):
Heya @PenghaiZhang 

I think you're there, and the layers are almost correct. But now with it all tidy with the abstractions etc it highlights there's probably one last step, and that looks like a whole bulk of interfaces and functions should now be moved over into ViewerModule and leave the components to just focus on the display.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResultAttachmentsList.tsx
PenghaiZhang: I agree. But when I tried to move `transform` I found this function is slightly coupled to UI layer such as the usage of 'handleError'.  So I only moved part of it to ViewerModule.

edalex-ian (approved):
Thanks for your perseverance @PenghaiZhang . I reckon it's GTG! :+1:

Issue #2617, opened 2020-12-21
Attachment Preview does not support images contributed via Resource Selector in new Search UI


cathfitz:
**Describe the bug**
Images contributed via the Resource Selector do not use the new Lightbox preview but instead launch into new tab. However the Old UI shows them with the normal fancy box setup.

This seems due to them having an attachment type of custom/resource rather than file .

**To Reproduce**

1. In the new search UI, go to a collection that has an oEQ resource selector and select some images
2. Save the item
3. Note that a thumbnail is not generated for the attachments

**Expected behavior**
Thumbnails are generated and dispaly on the search page

- OpenEquella Version: [2020.2 - New search UI]
- OS: [e.g. Windows 10, MacOS 10.13]
- Browser [e.g. Chrome, Firefox]

**Additional context**

SammyIsConfused:
Fixed.

PR #2840, opened 2021-03-24
Show thumbnails for Resource Selector attachments in the new search UI

SammyIsConfused:
##### Checklist

##### Description of change
#2617
Turned out that the thumbnails were generated after all, the new search UI frontend simply didn't handle them. 
Added custom/resource to OEQThumb's handling. 
There is an edge case, if a custom/resource points to an Item Summary we need a default icon for that as the existing one is very old. Added it as the Web icon.
I also discovered that Kaltura attachments weren't showing icons either so I added that too - and also included a default in case there are any left not handled.

Before:
image

After:
image


Source/Plugins/Core/com.equella.core/js/tsrc/components/OEQThumb.tsx
```diff
@@ -126,9 +127,19 @@ export default function OEQThumb({
     case "html":
       oeqThumb = <WebIcon {...generalThumbStyles} />;
       break;
+    case "custom/resource":
+      if (mimeType === "equella/item") {
+        // resource summaries can be pointed to by a Resource Selector.
+        oeqThumb = <Web {...generalThumbStyles} />;
```
edalex-ian, line 133: After looking at this code, and thinking about the Legacy UI, shouldn't this either:

1. Simply use the placeholder icon as per up on line 79 - which is the default for an item with no attachments; or
2. Possibly check to see if the item it points to has attachments and use those (or seeing that can be recursive, send that across from the server)?

Source/Plugins/Core/com.equella.core/js/tsrc/components/OEQThumb.tsx
```diff
@@ -126,9 +127,19 @@ export default function OEQThumb({
...
+      } else {
+        oeqThumb = oeqProvidedThumb;
+      }
+      break;
     case "custom/flickr":
     case "custom/youtube":
+    case "custom/kaltura":
     case "custom/googlebook":
+    default:
```
edalex-ian, line 142: I'm not sure adding this `default` is correct. This assumes that for any type not listed, then it will have an oEQ provided thumbnail. Where as before this change, it worked such that if we don't explicitly know about the type then use the `defaultThumb` (`oeqThumb` is set to it before this switch statement).

I believe that yes, most of the time there will be an oEQ Provided Thumb, but the placeholders will not be something that stylistically we want to use.

edalex-ian (commented):
Some queries for you Sam.

Source/Plugins/Core/com.equella.core/js/tsrc/components/OEQThumb.tsx
SammyIsConfused: 1. Is that quite right though? There is an attachment, it's just that attachment is another item. 
2. So that seems to be what the old UI does for equella/item attachments - it goes into that item and uses whatever its thumbnail is - though there is an equella/item specific very old thumbnail that I'm not sure where it gets used. But if I don't have this `if (mimeType === "equella/item")` check, it will use the oEQ provided thumb, which in the case of "equella/item" uses this. 
image

edalex-ian: > There is an attachment, it's just that attachment is another item.

But does that item have an attachment. I think that then addresses your second response too as to where that old default thumb is used. Have you tried having an item which has an attachment to `equella/item` which has no attachments itself?

Source/Plugins/Core/com.equella.core/js/tsrc/components/OEQThumb.tsx
PenghaiZhang: Perhaps it is better to do nothing but `break` in the default case? That way,  any type not listed will use `defautThumb`.

SammyIsConfused: I have gone with removing the default, so that we use defaultThumb when there is no handled attachment type.

Source/Plugins/Core/com.equella.core/js/tsrc/components/OEQThumb.tsx
SammyIsConfused: Okay, apologies. I went down a bit of a rabbit hole with this one. 
In the old UI search, if a Resource Attachment points at an item summary, there is no icon. 
image

What didn't make sense to me on this, is that from the resource summary attachment preview it *does* resolve the icon of the item, in search it doesn't show. 
image

Which is why I was having trouble finding the original item's thumbnail - the old search doesn't handle that, only the attachment preview from item summary page does. Confusing. That's another bug in itself. We know it's possible to reach the correct icon, as the attachment preview shows it, but neither old or new search handles it. 

Following your query on an attachment to an attachmentless `equella/item`, it uses the legacy icon that was coming up in the new search. 

image

Since the old UI search never showed an icon for `equella/item`, its safe to say we can just show a default Web icon, to replace the legacy icon. 

I did attempt to determine why the attachment preview can achieve this and the search can't - it would be good to fix it so that all thumbnail views are at least consistent, but it seemed like a bit of a can o' worms, and I'm unsure it's worth it at this stage of the release.

edalex-ian (approved):
Looks good @SammyIsConfused .

One optional change is to just go a ternary rather than that if/else statement.

But really, GTG. :+1:

SammyIsConfused:
> Looks good @SammyIsConfused .
> 
> One optional change is to just go a ternary rather than that if/else statement.
> 
> But really, GTG. +1

Cool, thanks. I've updated to ternary operator, and will merge on green.

PenghaiZhang (approved):
LGTM.

PR #2841, opened 2021-03-25
feat(rest-api): Add support for 'musts' to search2

edalex-ian:
##### Checklist

- [x] tests are included

##### Description of change

Add support for oEQ's internal search service's concept of 'musts'. This essentially provides support for lucene queries at the index field level.

This is needed so that we can implement the Video gallery where suitable items are searched for with `videothumb:true`.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
@@ -98,11 +98,8 @@ object SearchHelper {
     }
     search.setFreeTextQuery(freeTextQuery)
 
-    handleMusts(params.musts) match {
-      case Some(ms) =>
-        ms foreach {
-          case (field, value) => search.addMust(field, value.asJavaCollection)
-        }
+    handleMusts(params.musts) foreach {
+      case (field, value) => search.addMust(field, value.asJavaCollection)
     }
 
     search
@@ -186,10 +183,10 @@ object SearchHelper {
     * to the existing entry(key).
     *
     * @param musts a list of colon delimited strings to be split
-    * @return the processed strings, ready for calls into `DefaultSearch.addMusts` or None if there
+    * @return the processed strings, ready for calls into `DefaultSearch.addMusts` or throws if there
     *         was an issue processing the strings
     */
-  def handleMusts(musts: Array[String]): Option[Map[String, List[String]]] = {
+  def handleMusts(musts: Array[String]): Map[String, List[String]] = {
     val delimiter         = ':'
     val oneOrMoreNonDelim = s"([^$delimiter]+)"
     val mustExprFormat    = s"$oneOrMoreNonDelim$delimiter$oneOrMoreNonDelim".r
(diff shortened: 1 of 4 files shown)


Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
```diff
@@ -173,6 +180,35 @@ object SearchHelper {
...
+    */
+  def handleMusts(musts: Array[String]): Option[Map[String, List[String]]] = {
+    val delimiter         = ':'
+    val oneOrMoreNonDelim = s"([^$delimiter]+)"
+    val mustExprFormat    = s"$oneOrMoreNonDelim$delimiter$oneOrMoreNonDelim".r
+
+    def valid = (xs: Array[String]) => xs.forall(s => s.matches(mustExprFormat.regex))
+
+    if (valid(musts)) {
+      Option(musts.foldLeft(Map[String, List[String]]())((result, mustExpr) => {
```
PenghaiZhang, line 200: I think there are too many parenthesis here.  I would like to create a val for the initial map
```
val initialMap = Map[String, List[String]]()
Option(musts.foldLeft(initialMap)((result, mustExpr) 

```

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
```diff
@@ -173,6 +180,35 @@ object SearchHelper {
...
+        val mustExprFormat(k, v) = mustExpr
```
PenghaiZhang, line 201: I I thought the regex would return `(k, d, v)` where d is the delimiter. :thinking:

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
```diff
@@ -98,6 +98,13 @@ object SearchHelper {
     }
     search.setFreeTextQuery(freeTextQuery)
 
+    handleMusts(params.musts) match {
+      case Some(ms) =>
+        ms foreach {
+          case (field, value) => search.addMust(field, value.asJavaCollection)
+        }
+    }
+
```
PenghaiZhang, line 107: What about use foreach if we don't care `None`

```
handleMusts(params.musts).foreach(
ms => ms foreach {
    case (field, value) => search.addMust(field, value.asJavaCollection)
   }

)
```

PenghaiZhang (commented):
Looks good to me Ian! Just some queries.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
```diff
@@ -98,6 +98,13 @@ object SearchHelper {
     }
     search.setFreeTextQuery(freeTextQuery)
 
+    handleMusts(params.musts) match {
```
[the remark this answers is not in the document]
edalex-ian: You highlight a good point. There no longer is None as I realised our error handling is more about throwing an exception. So I can probably remove the Option.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
edalex-ian: Yes, indeed. As per above, I think this can be totally changed up.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
edalex-ian: Yeah, that's fair. Can do.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
edalex-ian: There's no group parens around the delimiter. So there's only two groups. And that's one value of using groups, to just grab what you care about - vs just raw tokenization.

edalex-ian:
Thanks @PenghaiZhang , all feedback reduced - and code better for it. Please re-review. :bow:

PenghaiZhang (approved):
LGTM!!

PR #2842, opened 2021-03-25
Hide menu panel in selection sessions and items viewed from integration

SammyIsConfused:
##### Checklist

- [x] videos are included showing significant UI changes

##### Description of change
We were defaulting to not full screen, which became a problem in selection sessions as it would take a short while to update based on width and height of the viewport. Now the default is false *unless* in a selection session or integration item view (integ/gen item link) where it is true. 

This involved adding to the renderData information passed from RenderNewTemplate a new variable viewedFromIntegration, which depends on a check as to whether the request is an /integ/gen item link. 
Before:
https://user-images.githubusercontent.com/24543345/112430075-c03e1300-8d91-11eb-979f-08c494bc6a8e.mp4

After:
https://user-images.githubusercontent.com/24543345/112429361-a6e89700-8d90-11eb-839f-def89e8e41ee.mp4

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/AppConfig.ts
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/AppConfig.ts
@@ -51,6 +51,7 @@ export interface RenderData {
   autotestMode: boolean;
   newSearch: boolean;
   selectionSessionInfo: SelectionSessionInfo | null;
+  viewedFromIntegration: boolean;
 }
 
 declare const renderData: RenderData | undefined;
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/AppConfig.ts
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/AppConfig.ts
@@ -46,11 +46,29 @@ export interface SelectionSessionInfo {
 }
 
 export interface RenderData {
+  /**
+   * Path to base resources for this institution.
+   */
   baseResources: string;
+  /**
+   * True if in the new UI.
+   */
   newUI: boolean;
+  /**
+   * True if server is running in autotestMode (VM argument -Dequella.autotest=true)
+   */
   autotestMode: boolean;
+  /**
+   * True if using the new Search UI.
+   */
   newSearch: boolean;
+  /**
+   * If in a Selection Session, this will be populated with info about the session. If not, this will be null.
+   */
   selectionSessionInfo: SelectionSessionInfo | null;
+  /**
+   * True if this is an item viewed via integration (integ/gen item link)
+   */
   viewedFromIntegration: boolean;
 }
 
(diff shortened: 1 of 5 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/AppConfig.ts
```diff
@@ -51,6 +51,7 @@ export interface RenderData {
   autotestMode: boolean;
   newSearch: boolean;
   selectionSessionInfo: SelectionSessionInfo | null;
+  viewedFromIntegration: boolean;
```
edalex-ian, line 54: Please add TSDoc.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/LegacySelectionSessionModule.ts
```diff
@@ -155,6 +155,14 @@ const isSelectionSessionInfo = (
 export const isSelectionSessionOpen = (): boolean =>
   isSelectionSessionInfo(getRenderData()?.selectionSessionInfo);
 
+/**
+ * Returns true if this is an /integ/gen item link - that is, a resource summary viewed via an integration.
+ */
+export const isItemViewedFromIntegration = (): boolean => {
+  const viewedFromIntegration = getRenderData()?.viewedFromIntegration;
+  return typeof viewedFromIntegration == "boolean" && viewedFromIntegration;
```
edalex-ian, line 163: I was going to comment on the type check needing triple eq, but then that made me wonder, why are you checking the type and not just doing a truthy check?

Maybe if it was just about wrangling around the possibly undef, then this function could just be:

```typescript
export const isItemViewedFromIntegration = (): boolean =>
  getRenderData()?.viewedFromIntegration ?? false;
```

Or if you want more concision:

```typescript
export const isItemViewedFromIntegration = (): boolean =>
  !!(getRenderData()?.viewedFromIntegration);
```

(Personally, I think I prefer the readability of the nullish coalescing.)

edalex-ian (changes requested):
Thanks @SammyIsConfused , just needs a comment and possibly a refinement of some code.

BTY

Source/Plugins/Core/com.equella.core/js/tsrc/modules/LegacySelectionSessionModule.ts
SammyIsConfused: The main reason I did it this way was for consistency with other functions in the same file - example, 
```
const isSelectionSessionInfo = (
  data: unknown
): data is { [K in keyof SelectionSessionInfo]: unknown } =>
  typeof data === "object" && data !== null && "stateId" in data;
```

I've simplified my function using nullish coalescing now.

edalex-ian: Yeah I was wondering if that was why - as I looked through the rest of the file. That example you provide is a little more special though.

That's for updating. :+1:

Source/Plugins/Core/com.equella.core/js/tsrc/AppConfig.ts
edalex-ian: Wow, thanks for doing all the others while at it. :+1: :star:

edalex-ian (approved):
Thanks @SammyIsConfused - and thanks for the updates. :+1:

PR #2849, opened 2021-04-01
Handle long strings without breaking layout

SammyIsConfused:
##### Checklist

##### Description of change
#2850
This problem was discovered as a support issue. So various strings can break the layout if they are too long, both in the legacy EBP UI and the new search UI. 
In the Template.tsx file, I added a global rule to all `a` and `p` tags to `overflow-wrap:anywhere`, which means they can simply break to a new line rather than extending the width of their containers.
In the `legacy.scss` I handle the same issue in contribution wizards and in item summaries. 
I was able to confirm that this handles well in the main UI and in a selection session.

Before change:

https://user-images.githubusercontent.com/24543345/113230769-7602e780-92e5-11eb-8549-0c58196c9771.mp4

After change:

https://user-images.githubusercontent.com/24543345/113230778-7c915f00-92e5-11eb-9689-947b43c2d863.mp4

### DIFF
--- a/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
+++ b/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
@@ -1237,6 +1237,16 @@ div.area {
   float: none;
 }
 
+.itemsummary-layout #col1 {
+  //handle long strings  in item summaries without breaking the layout
+  overflow-wrap: anywhere;
+}
+
+.wizard-layout #col1 {
+  //handle wizard control names and descriptions without breaking the layout
+  max-width:85%;
+}
+
 .area h3 {
   @include typography-headline-6;
   color: $primaryTextColor;
--- a/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
+++ b/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
@@ -1244,7 +1244,7 @@ div.area {
 
 .wizard-layout #col1 {
   //handle wizard control names and descriptions without breaking the layout
-  max-width:85%;
+  max-width: 85%;
 }
 
 .area h3 {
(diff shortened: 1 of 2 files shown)


Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -1237,6 +1237,16 @@ div.area {
...
 }
 
+.itemsummary-layout #col1 {
+  //handle long strings  in item summaries without breaking the layout
+  overflow-wrap: anywhere;
+}
+
+.wizard-layout #col1 {
+  //handle wizard control names and descriptions without breaking the layout
+  max-width: 85%;
```
PenghaiZhang, line 1247: I wonder if there is any chance we can use calc or grid instead of max-width. But I think this will requires a lot more changes.

I know we sometimes need these magic numbers to support Old UI, so I am fine with this number.

PenghaiZhang (commented):
Thanks Sam! Looks good to me. I noticed that you have made changes for the Wizard layout. Have you also checked how it looks like in Selection Session?

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
SammyIsConfused: Hi Penghai - yes, I checked selection sessions (though I will double check skinny sessions tomorrow, thats one area I could be more thorough). 
I'll see if I can break down the number into something more meaningful tomorrow morning.

edalex-ian (approved):
LGTM @SammyIsConfused  - pending your discussions/confirmations outstanding with @PenghaiZhang :+1:

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
SammyIsConfused: Alrighty, I found something more meaningful to use. `#col2` (the save panel) is hardcoded to `221px` wide. So we can save that to a variable `$savePanelWidth` and use that here as `calc(100% - #{$savePanelWidth})`

PenghaiZhang (approved):
LGTM.

PR #2852, opened 2021-04-01
Add support to the GallerySearchModule for Video Gallery - and inital basic integration

edalex-ian:
##### Checklist

- [x] tests are included

##### Description of change

This adds the transformation support to the `GallerySearchModule` to support Video Searches. As well as that, it then adds a function to do the `videoGallerySearch()`. Then, it was only a small bit of extra work to plug it in, so basic gallery view of videos is now functional (but need to add lightbox support for YouTube still - that's a future PR).

While at it, reworked the the Jest checks for `E.Left` and `E.Right` by setting up some custom Jest Matchers (`FpTsMatchers.ts`).

There was also a refactor around GalleryEntry so that it was no longer so Image specific.

Below video shows a quick look at the video gallery in action - noting the missing support for YouTube (coming soon).

https://user-images.githubusercontent.com/43919233/113254567-7bc2f200-9312-11eb-9b1b-81726bfe898e.mp4

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/FpTsMatchers.ts
+++ b/Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/FpTsMatchers.ts
@@ -1,0 +1,76 @@
+/*
+ * Licensed to The Apereo Foundation under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * The Apereo Foundation licenses this file to you under the Apache License,
+ * Version 2.0, (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import * as E from "fp-ts/Either";
+
+expect.extend({
+  /**
+   * An expect matcher for Jest to confirm the response is a E.Left - typically to confirm an
+   * error result.
+   *
+   * @param received The E.Either to validate
+   */
+  toBeLeft: (received: E.Either<any, any>): jest.CustomMatcherResult =>
+    E.isLeft(received)
+      ? {
+          pass: true,
+          message: () => `expected ${received} not to be E.Left`,
+        }
+      : {
+          pass: false,
+          message: () => `expected ${received} to be E.Left`,
+        },
+
+  /**
+   * An expect matcher for Jest to confirm the response is an E.Right - typically to confirm an
+   * success result.
+   *
+   * @param received The E.Either to validate
+   */
+  toBeRight: (received: E.Either<any, any>): jest.CustomMatcherResult =>
+    E.isRight(received)
+      ? {
+          pass: true,
+          message: () => `expected ${received} not to be E.Right`,
+        }
+      : {
+          pass: false,
+          message: () => `expected E.Right but got E.Left(${received.left})`,
+        },
+});
+
+declare global {
+  namespace jest {
+    interface Matchers<R> {
+      toBeLeft(): jest.CustomMatcherResult;
+      toBeRight(): jest.CustomMatcherResult;
+    }
+  }
+}
+
+/**
+ * An expect matcher for Jest to confirm the response is an E.Right - typically to confirm an
+ * success result. The value is also returned, albeit (unfortunately) unioned with an undefined. It
+ * is considered safe to access the value via a bang - .e.g `const value = expectRight(either)!;`
+ *
+ * @param either The E.Either to validate
+ * @return The value contained in the E.Right
+ */
+export const expectRight = <V>(either: E.Either<unknown, V>): V | undefined => {
+  expect(either).toBeRight();
+  return E.isRight(either) ? either.right : undefined;
+};
--- a/Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/FpTsMatchers.ts
+++ b/Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/FpTsMatchers.ts
@@ -64,9 +64,9 @@ declare global {
 }
 
 /**
- * An expect matcher for Jest to confirm the response is an E.Right - typically to confirm an
- * success result. The value is also returned, albeit (unfortunately) unioned with an undefined. It
- * is considered safe to access the value via a bang - .e.g `const value = expectRight(either)!;`
+ * An expect matcher for Jest to confirm the response is an E.Right - typically to confirm a
+ * success result. The value is also returned, albeit (unfortunately) unioned with an `undefined`.
+ * It is considered safe to access the value via a bang - .e.g `const value = expectRight(either)!;`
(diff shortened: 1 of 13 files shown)


Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/FpTsMatchers.ts
```diff
@@ -0,0 +1,77 @@
...
+  }
+}
+
+/**
+ * An expect matcher for Jest to confirm the response is an E.Right - typically to confirm an
+ * success result. The value is also returned, albeit (unfortunately) unioned with an undefined. It
+ * is considered safe to access the value via a bang - .e.g `const value = expectRight(either)!;`
+ *
+ * @param either The E.Either to validate
+ * @return The value contained in the E.Right
```
SammyIsConfused, line 72:
```suggestion
 * An expect matcher for Jest to confirm the response is an E.Right - typically to confirm a
 * success result. The value is also returned, albeit (unfortunately) unioned with an undefined. It
 * is considered safe to access the value via a bang - .e.g `const value = expectRight(either)!;`
 *
 * @param either The E.Either to validate
 * @return The value contained in the E.Right
```

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/modules/GallerySearchModule.test.ts
```diff
@@ -297,3 +316,16 @@ describe("imageGallerySearch", () => {
...
+
+describe("videoGallerySearch", () => {
+  it("returns a list of GallerySearchItems containing only videos", async () => {
+    mockSearchItems.mockResolvedValue(basicVideoSearchResponse);
+
+    const result = await videoGallerySearch(defaultSearchOptions);
+    expectValidGalleryResult(
+      result,
+      basicVideoSearchResponse.length,
+      (s) => s.startsWith("video") || s === CustomMimeTypes.YOUTUBE
```
SammyIsConfused, line 328: Does the video gallery accept openEQUELLA Resource attachments pointing at videos? 
I'd assume they have a type of "custom/resource", so they would be omitted.

SammyIsConfused (approved):
LGTM! Just a query regarding custom/resource and a slight comment tweak.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/FpTsMatchers.ts
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import * as E from "fp-ts/Either";
+
+expect.extend({
```
PenghaiZhang, line 20: May I ask if we can do in the jest setup file rather than here?  If yes, I guess we can then call
`expect.toBeRight()` in other jest tests ?

PenghaiZhang (approved):
Thanks Ian! Looks good to me.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/modules/GallerySearchModule.test.ts
edalex-ian: Really not sure. If so, we'd have to look at in the future - probably when/before we look at Kaltura support etc. At the moment so as to get broadest support possibly we've narrowed the scope to just local videos (but direct in this case) and YouTube.

Be interesting to test against to find out at some point.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/GallerySearchModule.ts
```diff
@@ -90,9 +165,67 @@ const warnOnLeft = E.mapLeft((left) => {
...
 const thumbnailLink = (
-  attachment: OEQ.Search.Attachment,
-  type?: "thumbnail" | "preview"
-): string => attachment.links.thumbnail.concat(type ? `?gallery=${type}` : "");
+  { attachmentType, id, links }: OEQ.Search.Attachment,
+  type: "small" | "large"
+): E.Either<string, string> => {
+  const isSmall = type === "small";
+  const fileThumbnailLink = (basePath: string): string =>
+    basePath + "?gallery=" + (isSmall ? "gallery" : "preview");
```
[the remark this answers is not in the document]
edalex-ian: In this case I prefer it in its individual parts, as it highlights them - especially seeing one is logic etc.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/GallerySearchModule.ts
```diff
@@ -77,11 +83,80 @@ export interface GallerySearchResultItem
...
+ * other types build a custom `openequella/*` MIME type.
+ *
+ * @param attachment target from which to derive the MIME type
+ */
+const mimeType = ({
+  id,
+  mimeType,
+  attachmentType,
+}: OEQ.Search.Attachment): E.Either<string, string> =>
+  pipe(
```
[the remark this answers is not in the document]
edalex-ian: Ah, this was a deliberate approach. I actually find with the large `match` calls it can sometimes become oblique as to what is happening. So I was rather tickled here that we can now make it so clear that we're piping that value into the match and its resulting machinations. Felt like a match (pardon the pun) made in heaven. :wink:

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/FpTsMatchers.ts
edalex-ian: You could do it in the Jest setup file, but then that'd pollute the matchers for all the (majority) of cases where you don't need this. Having it in its own file like this is the same approach as used for the additional matches provided when we use the React Testing Library for things like `inDocument` etc.

In that case it's achieve via:

    import "@testing-library/jest-dom/extend-expect";

Source/Plugins/Core/com.equella.core/js/tsrc/modules/GallerySearchModule.ts
```diff
@@ -77,11 +83,80 @@ export interface GallerySearchResultItem
...
+ *
+ * @param attachment validation candidate
+ */
+const validateAttachmentType = (
+  attachment: OEQ.Search.Attachment
+): E.Either<string, string> =>
+  pipe(
+    attachment.attachmentType,
+    E.fromPredicate(
+      (type) => [ATYPE_FILE, ATYPE_YOUTUBE].includes(type),
```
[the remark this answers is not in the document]
edalex-ian: Mostly correct. Just some detail - which you possibly implied: If the predicate (defined in the first function/argument) returns true then yes the string passed in will be returned contained in an `E.Right`, otherwise an `E.Left` will be returned with the contents resulting from the evaluation of the second function.

Source/Plugins/Core/com.equella.core/js/__tests__/tsrc/FpTsMatchers.ts
edalex-ian: Thank you, addressed in 1472129

edalex-ian:
Thank you both. All feedback responded to or addressed as required. Please let me know if there's still anything you'd like further addressed - or to discuss. Otherwise, on green build I'll merge.

Thanks. :bow:

SammyIsConfused (approved):

PR #2866, opened 2021-04-12
Replace 'raw search' with 'wildcard search'

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

This PR changes the current 'Raw search' label to be 'Wildcard search', and changes it to be enabled by default.

https://user-images.githubusercontent.com/47203811/114329261-34dec300-9b82-11eb-9709-d48ea76412dd.mp4


edalex-ian (changes requested):
That's definitely functionally what we're after @PenghaiZhang , however structurally I think we should make it so that the `SearchBar` only knows about 'wildcard mode' and so we change the props to also align with this and make no mention of 'raw mode' within `SearchBar`. That way it acts as a clean boundary between what we do at a UX level verses internal lower level implementation.

Admittedly, this will then also require updates to the integration points and tests, but I think this is worth that.

P.S. Make sure you also don't miss the `id="rawSearch"` too.

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -436,8 +436,9 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
       false
     );
 
-  const handleRawModeChanged = (rawMode: boolean) =>
-    search({ ...searchPageOptions, rawMode: rawMode });
+  const handleWildcardModeChanged = (wildcardMode: boolean) =>
+    // We use `rawMode` which is the logic NOT of `wildcardMode` at lower level.
```
edalex-ian, line 440: Sorry, I'm not sure what that is saying, but I propose:

```suggestion
    // `wildcardMode` is a presentation concept, in the lower levels its inverse is the value for `rawMode`
```

edalex-ian (approved):
Looks good! I think just need to tweak that one comment, then should be GTG.

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResultAttachmentsList.tsx
```diff
@@ -273,7 +273,7 @@ export const SearchResultAttachmentsList = ({
   );
 
   const accordionText = (
-    <Typography>
+    <Typography component="div">
```
ChristianMurphy, line 276: Does this explicitly need to be set as a `div`?

ChristianMurphy (approved):
Thanks @PenghaiZhang!

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchResultAttachmentsList.tsx
edalex-ian: I asked the same question. @PenghaiZhang highlighted he did this due to a new warning emitting from the jest tests due to Typography being a `p` and should not be nesting `div`s which the `<Chip>` is.

edalex-ian:
Now the builds are finally all green, merging on behalf of @PenghaiZhang

PR #2871, opened 2021-04-12
Add Download CSV button to Search page

PenghaiZhang:
##### Checklist

- [x] tests are included

##### Description of change

In this PR, I created a new ts component named `ExportSearchResultLink`. It is either a link or a icon, depending on whether the search result has been downloaded or not. This component is integrated in `SearchResultList`. I also refactored how to use `snackbar` in search page, to support showing a warning.

https://user-images.githubusercontent.com/47203811/114338155-cc014600-9b95-11eb-9093-ef812cd29edc.mp4

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/__mocks__/UserSearch.mock.ts
+++ b/Source/Plugins/Core/com.equella.core/js/__mocks__/UserSearch.mock.ts
@@ -53,6 +53,19 @@ export const users: OEQ.UserQuery.UserDetails[] = [
   },
 ];
 
+export const defaultCurrentUserDetails: OEQ.LegacyContent.CurrentUserDetails = {
+  id: "test",
+  username: "test",
+  firstName: "test",
+  lastName: "test",
+  accessibilityMode: true,
+  autoLoggedIn: false,
+  guest: false,
+  prefsEditable: true,
+  menuGroups: [],
+  canDownloadSearchResult: true,
+};
+
 /**
  * Helper function to inject into component for user retrieval.
  *
--- a/Source/Plugins/Core/com.equella.core/js/__mocks__/UserSearch.mock.ts
+++ b/Source/Plugins/Core/com.equella.core/js/__mocks__/UserSearch.mock.ts
@@ -53,19 +53,6 @@ export const users: OEQ.UserQuery.UserDetails[] = [
   },
 ];
 
-export const defaultCurrentUserDetails: OEQ.LegacyContent.CurrentUserDetails = {
-  id: "test",
-  username: "test",
-  firstName: "test",
-  lastName: "test",
-  accessibilityMode: true,
-  autoLoggedIn: false,
-  guest: false,
-  prefsEditable: true,
-  menuGroups: [],
-  canDownloadSearchResult: true,
-};
-
 /**
  * Helper function to inject into component for user retrieval.
  *
--- a/Source/Plugins/Core/com.equella.core/js/__stories__/search/ExportSearchResultLink.stories.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/__stories__/search/ExportSearchResultLink.stories.tsx
@@ -1,0 +1,48 @@
+/*
+ * Licensed to The Apereo Foundation under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * The Apereo Foundation licenses this file to you under the Apache License,
+ * Version 2.0, (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import * as React from "react";
+import { Meta, Story } from "@storybook/react";
+import {
+  ExportSearchResultLink,
+  ExportSearchResultLinkProps,
+} from "../../tsrc/search/components/ExportSearchResultLink";
+
+export default {
+  title: "Search/ExportSearchResultLink",
+  component: ExportSearchResultLink,
+  argTypes: {
+    onExport: { action: "on Export" },
+  },
+} as Meta<ExportSearchResultLinkProps>;
+
+export const ExportEnabled: Story<ExportSearchResultLinkProps> = (args) => (
+  <ExportSearchResultLink {...args} />
+);
+
+ExportEnabled.args = {
+  url: "test",
+  exportDisabled: false,
+};
+
+export const ExportCompleted: Story<ExportSearchResultLinkProps> = (args) => (
+  <ExportSearchResultLink {...args} />
+);
+ExportCompleted.args = {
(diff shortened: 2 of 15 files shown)


edalex-ian:
As discussed @PenghaiZhang , we need to make it so the button only shows for users who are able to use it. May need to add a flag to CurrentUserDetails to indicate if they have the ACL which allows them to do so.

Source/Plugins/Core/com.equella.core/js/__stories__/search/ExportSearchResultLink.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+  component: ExportSearchResultLink,
+  argTypes: {
+    onExport: { action: "on Export" },
+  },
+} as Meta<ExportSearchResultLinkProps>;
+
+export const ExportEnabled: Story<ExportSearchResultLinkProps> = (args) => (
+  <ExportSearchResultLink {...args} />
+);
+
```
edalex-ian, line 36: Typically (and as per lower down in this file) we don't typically have a blank line here.

```suggestion
```

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
```diff
@@ -255,6 +263,11 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
     [dispatch]
   );
 
+  // Allow exporting a search result when searchPageOptions gets changed.
+  useEffect(() => {
+    setExportDisabled(false);
+  }, [searchPageOptions]);
+
```
edalex-ian, line 270: Rather than creating another effect, why not tie this to the reducer lifecycle. That is, in the effect that is responsible for `state.status === "searching"`, why not `setExportDisabled(false)` after it sends the `dispatch` for `type: "search-complete"`?

That then means you can download search results only once you've successfully executed a search. (Which seems like what we're after.)

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/ExportSearchResultLink.tsx
```diff
@@ -0,0 +1,72 @@
...
+   */
+  url: string;
+  /**
+   * Handler fired to check an export. Return `false` to indicate an export is invalid.
+   */
+  onExport: () => boolean;
+  /**
+   * `true` if an export is not allowed.
+   */
+  exportDisabled: boolean;
```
edalex-ian, line 37: This is more about showing export is complete - a true display property. So should this be:

```suggestion
  /**
   * `true` to show a complete indicator and disable additional clicking.
   */
  exportComplete: boolean;
```

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/ExportSearchResultLink.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { TooltipIconButton } from "../../components/TooltipIconButton";
+import { languageStrings } from "../../util/langstrings";
+
+export interface ExportSearchResultLinkProps {
+  /**
+   * The full URL to download a search result
+   */
+  url: string;
+  /**
+   * Handler fired to check an export. Return `false` to indicate an export is invalid.
```
edalex-ian, line 31:
```suggestion
   * Handler fired before triggering an export. Return `false` to prevent an export being triggered.
```

oeq-ts-rest-api/src/Search.ts
```diff
@@ -385,6 +386,7 @@ const processSearchParams = (
   params ? { ...params, musts: processMusts(params.musts) } : undefined;
 
 const SEARCH2_API_PATH = '/search2';
+const EXPORT_PATH = '/search2/export';
```
edalex-ian, line 389:
```suggestion
const EXPORT_PATH = `${SEARCH2_API_PATH}/export`;
```

oeq-ts-rest-api/test/Search.test.ts
```diff
@@ -132,3 +132,17 @@ describe('Search for attachments', () => {
     }
   );
 });
+
+describe('Build URL for downloading search result', function () {
```
edalex-ian, line 136: Should this perhaps be:

```suggestion
describe('Exports search results for the specified search params', function () {
```

And then perhaps add another `it` which tests calling the generated URL?

edalex-ian (changes requested):
Heya @PenghaiZhang ,

Just a couple of minor tweaks required, plus obviously the ACL check work (which I know you're doing right now).

Cheers.

Source/Plugins/Core/com.equella.core/js/tsrc/search/SearchPage.tsx
PenghaiZhang: Ah yes here is indeed the best place for doing this.

oeq-ts-rest-api/test/Search.test.ts
PenghaiZhang: Good idea! A new test has been added.

PenghaiZhang:
Hi @edalex-ian , just remind you that there are commits made before your reviewed.

PenghaiZhang:
Here is an update video.
https://user-images.githubusercontent.com/47203811/114474814-5b603500-9c3a-11eb-8982-a24ebb2bf911.mp4

Source/Plugins/Core/com.equella.core/js/__mocks__/UserSearch.mock.ts
```diff
@@ -53,6 +53,19 @@ export const users: OEQ.UserQuery.UserDetails[] = [
   },
 ];
 
+export const defaultCurrentUserDetails: OEQ.LegacyContent.CurrentUserDetails = {
```
edalex-ian, line 56: Couple of things here:

1. This isn't anything to do with searching for users (UserSearch) it's to do with the UserModule so should perhaps be in UserModule.mock.ts; and
2. It's not so much a 'default' as it is the result of a call to getCurrentUser so perhaps the name should be getCurrentUserMock.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/LegacyContentApi.scala
```diff
@@ -344,6 +346,11 @@ class LegacyContentApi {
 
     val accessibilityMode = LegacyGuice.accessibilityModeService.isAccessibilityMode
 
+    val canDownloadSearchResult: Boolean = LegacyGuice.aclManager
+      .filterNonGrantedPrivileges(SecurityConstants.EXPORT_SEARCH_RESULT)
+      .asScala
+      .nonEmpty
```
edalex-ian, line 352: No action required here.

But it'd be good if one day we simply have a function such as `hasACL` in place of always repeating this block all over the place. Then we'd just have:

```scala
val canDownloadSearchResult: Boolean = hasACL(SecurityConstants.EXPORT_SEARCH_RESULT)
```

One day...

edalex-ian (approved):
Looks pretty well there @PenghaiZhang . If you can just tidy up the mock you should be fine to merge.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/LegacyContentApi.scala
PenghaiZhang: Yeah...there is a function called `hasPrivilege`  in `TLEAclManager`. When we implement `hasACL` in the future we can have a look this.

PR #2891, opened 2021-04-13
Ensure a long search params does not break audit log

PenghaiZhang:
##### Checklist

- [x] tests are included

##### Description of change

Previously, the search param was stored in column 'data2' which has a length limit. So when  search param was too long the audit log threw an error. 

This PR changes  'data2' to  'data4' which has no limit and saves search param in Json format.

image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchResource.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchResource.scala
@@ -18,6 +18,8 @@
 
 package com.tle.web.api.search
 
+import com.fasterxml.jackson.databind.json.JsonMapper
+import com.fasterxml.jackson.module.scala.DefaultScalaModule
 import com.tle.beans.entity.Schema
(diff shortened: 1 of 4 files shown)


SammyIsConfused (approved):
Looks great Penghai, thanks for fixing that up. 
The only thing I could think is that it's potentially worth adding to the tests to ensure it handles long strings, to prevent regression of this bug. But that's optional.

PenghaiZhang:
> Looks great Penghai, thanks for fixing that up.
> The only thing I could think is that it's potentially worth adding to the tests to ensure it handles long strings, to prevent regression of this bug. But that's optional.

Good idea. Added.

SammyIsConfused:
Thanks for adding the test - assuming green build, I think this is good to go :+1:

edalex-ian (approved):
LGTM.

(I'm assuming you also looked in the audit table to confirm this all got persisted nicely / as expected.)

ChristianMurphy (approved):

edalex-ian (approved):
LGTM! :+1:

PR #2892, opened 2021-04-14
Support display mode for Old UI saved or shared searches

PenghaiZhang:
##### Checklist

- [x] tests are included

##### Description of change

In Old UI, the query string for display mode is named `type`. And it has three values: 'standard', 'gallery' and 'video'. So when we convert a legacy query string to search options, we need to map above three values to the display modes defined in New UI. 

https://user-images.githubusercontent.com/47203811/114637376-89ae4500-9d0c-11eb-9feb-2ead2236e1a4.mp4


Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -543,6 +556,20 @@ export const legacyQueryStringToSearchOptions = async (
...
+      [Union(Literal("standard"), Undefined), (): DisplayMode => "list"],
+      [Literal("gallery"), (): DisplayMode => "gallery-image"],
+      [Literal("video"), (): DisplayMode => "gallery-video"],
+      [
+        Unknown,
+        () => {
+          throw new Error("Unknown Legacy display mode");
+        },
+      ]
+    )(getQueryParam("type"));
```
edalex-ian, line 571: This `legacyQueryStringToSearchOptions` function with all its nested types and functions is getting a bit unwieldy. But maybe that's for another day.

That did lead me though to wonder whether:

1. This really needs to be a function; and
2. If it does should it rather take an argument rather than relying on variables it has closed over.

If it does remain a function, I'd suggest:

```typescript
const getDisplayMode = (_: string): DisplayMode =>
  match(
    // .. your current matchers
  ); // Note that we just use the returned function (hence the arg of `_: string`)
```

But maybe you don't need the function at all and you can just call the match now and store in a constant. (And maybe use `pipe` to feed the `match`.)

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -110,6 +112,10 @@ export interface SearchOptions {
    * @see OEQ.Search.SearchParams for examples
    */
   musts?: OEQ.Search.Must[];
+  /**
+   * How to display the search results - also determines the type of results.
+   */
+  displayMode?: DisplayMode;
```
edalex-ian, line 118: Yeah, not sure about this, as technically it has nothing to do with a search, only how it is displayed (i.e. presentation) - as per its name.

Will discuss in person.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -30,10 +30,12 @@ import {
   Record,
   Static,
   String,
+  Undefined,
   Union,
   Unknown,
 } from "runtypes";
 import { API_BASE_URL } from "../AppConfig";
+import { DisplayMode } from "../search/components/DisplayModeSelector";
```
edalex-ian, line 38: Gah! We don't want the module layer coupled to the presentation layer.

Source/Plugins/Core/com.equella.core/js/tsrc/util/langstrings.ts
```diff
@@ -333,7 +333,7 @@ export const languageStrings = {
     shareSearchHelperText: "Copy search link to clipboard",
     shareSearchConfirmationText: "Search link saved to clipboard",
     displayModeSelector: {
-      title: "Display Mode",
+      title: "Display mode",
```
edalex-ian, line 336: @cathfitz will be happy. :wink: :+1:

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
```diff
@@ -185,6 +192,11 @@ const DehydratedSearchOptionsRunTypes = Partial({
   ),
   searchAttachments: Boolean,
   mimeTypeFilters: RuntypeArray(Record({ id: String })),
+  displayMode: Union(
+    Literal("list"),
+    Literal("gallery-image"),
+    Literal("gallery-video")
+  ),
```
edalex-ian, line 199: If we've got this runtype definition here, how about we export the static version and use that in the presentation layer - that way inverting the coupling I've commented on above.

edalex-ian (changes requested):
Okay, we need to discuss what's happened here with `DisplayMode` will grab you in person.

But the new `getDisplayMode` needs rework/reconsideration too.

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
PenghaiZhang: Yeah `pipe` is much better!

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
edalex-ian: As discussed, this highlights a slightly bigger issue that we'll address as part of the tech debt cycle at the start of 2021.2. We'll refactor SearchModule and relocate some items that should be over in UserModule, CollectionModule and MimeTypesModule as well as create a new SearchPageHelper which can look after all the wrangling of URLs etc unique to new and old SearchPages.

edalex-ian (approved):
Thanks @PenghaiZhang 

This aligns with where we currently are, and we'll do a refactor for 2021.2 for the structural issues. :+1:

Source/Plugins/Core/com.equella.core/js/tsrc/modules/SearchModule.ts
PenghaiZhang: As discussed, we will figure out this issue after we release 2021.1

ChristianMurphy (approved):

edalex-ian:
Gah! The GHA seems to have hung on one of the jobs (> 3hrs rather than 1.25hrs). I suggest you merge @PenghaiZhang , and if there is an issue we'll address as needed. :man_shrugging:

PR #2895, opened 2021-04-14
REST endpoint for export validation

PenghaiZhang:
##### Checklist

- [x] tests are included

##### Description of change

As discussed, we'd like to send a `HEAD` request to check whether an export is valid or not. So I added a new endpoint to support the HEAD request. In REST module, I firstly created new functions to execute a HEAD request to do the export validation.

Will re-target to the release branch.


Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/ExportCSVHelper.scala
```diff
@@ -203,14 +201,6 @@ object ExportCSVHelper {
     })
   }
 
-  def checkDownloadACL(): Unit = {
```
PenghaiZhang, line 206: I moved this to `SearchResource` as I remember that we want to throw these exceptions in presentation layer rather than a helper.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchResource.scala
```diff
@@ -118,6 +118,17 @@ class SearchResource {
 
     mapper.writeValueAsString(params)
   }
+
+  private def confirmExport(params: SearchParam): Unit = {
+    if (params.collections.length != 1) {
+      throw new BadRequestException("Download limited to one collection.")
+    }
```
edalex-ian, line 125: Probably also makes sense to do the validation of the collection in here too. That is, the code block:

```scala
    val schema: Schema = getSchemaFromCollection(collectionId) match {
      case Some(s) => s
      case None =>
        throw new NotFoundException(s"Failed to find Schema for Collection: $collectionId")
    }
```

And then you could return the `schema`. :thinking:

oeq-ts-rest-api/src/AxiosInstance.ts
```diff
@@ -60,6 +59,30 @@ export const GET = <T>(
...
+    .head(path, {
+      params: queryParams,
+      paramsSerializer: (params) => stringify(params),
+    })
+    .then(() => true)
+    .catch(catchHandler);
+
+/**
+ * Executes a HTTP PUT request for a given path.
+ * @param path The URL path for the target HEAD
```
edalex-ian, line 83:
```suggestion

 * @param path The URL path for the target HEAD
```

oeq-ts-rest-api/src/Search.ts
```diff
@@ -418,3 +418,13 @@ export const search = (
  */
 export const buildExportUrl = (apiBasePath: string, params: SearchParams) =>
   apiBasePath + EXPORT_PATH + '?' + stringify(params);
+
+/**
+ * Communicate with REST endpoint 'search2/export' to confirm if an export request is valid.
+ * @param apiBasePath Base URI to the oEQ institution and API.
```
edalex-ian, line 424:
```suggestion

 * @param apiBasePath Base URI to the oEQ institution and API.
```

edalex-ian (changes requested):
Heya @PenghaiZhang 

Basically all good, just two things:

1. I think the validation in the endpoint should also include getting the collection; and
2. We need two tests added to `io.github.openequella.rest.Search2ExportTest` to cover negative and positive cases.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchResource.scala
PenghaiZhang: Not a problem!

autotest/OldTests/src/test/java/io/github/openequella/rest/Search2ExportTest.java
```diff
@@ -83,15 +96,27 @@ public void exportWithLongQueryString() throws IOException {
...
             "LongStringb2be4e8e-a0d4-4e6a-b9ff-4c65a7c8024eb2be4e8e-a0d4-4e6a-b9ff-4c65a7c8024eb2be4e8e-a0d4-4e6a-b9ff-4c65a7c8024eb2be4e8e-a0d4-4e6a-b9ff-4c65a7c8024eb2be4e8e-a0d4-4e6a-b9ff-4c65a7c8024eb2be4e8e-a0d4-4e6a-b9ff-4c65a7c8024eb2be4e8e-a0d4-4e6a-b9ff-4c65a7c8024e");
-    HttpMethod method = buildExportRequest(queryStrings);
-    int statusCode = makeClientRequest(method);
-    assertEquals(statusCode, 200);
+
+    List<HttpMethod> requests = buildExportRequests(null);
+    for (HttpMethod request : requests) {
+      int statusCode = makeClientRequest(request);
+      assertEquals(statusCode, 200);
+    }
```
edalex-ian, line 104: These five lines are duplicated 3 times. Seems you simply need a wrapper function which would end up looking like:

    assertResponseCode(buildExportRequest(null), 200);

You would then just change the 200 to the 400 and the 403 in each case above - as well as using the different params to `buildExportRequests`.

edalex-ian (approved):
Thanks @PenghaiZhang , just need to tweak the duplication in the test and then you're probably good to merge when I'm away.

autotest/OldTests/src/test/java/io/github/openequella/rest/Search2ExportTest.java
PenghaiZhang: Good idea!!!

PR #2920, opened 2021-04-27
Use 'useReducer' to better handle the state of SearchBar

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

This PR fixes the debounce issue of SearchBar by using `useReducer`. Also, the usage of `debounce` has been moved to SearchPage as whether to debounce a search is what SearchPage cares.

https://user-images.githubusercontent.com/47203811/116189953-19bca780-a76d-11eb-9d0a-33576229dc13.mp4

### DIFF
--- a/Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchBar.tsx
+++ b/Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchBar.tsx
@@ -26,7 +26,7 @@ import {
 import { makeStyles } from "@material-ui/core/styles";
 import SearchIcon from "@material-ui/icons/Search";
 import * as React from "react";
-import { useCallback, useEffect, useReducer, useState } from "react";
+import { useCallback, useEffect, useReducer } from "react";
 import { languageStrings } from "../../util/langstrings";
 
 const useStyles = makeStyles({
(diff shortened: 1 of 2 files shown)


Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchBar.tsx
```diff
@@ -71,8 +68,31 @@ export interface SearchBarProps {
...
+    case "newSearch":
+      return { status: "init", query: action.initialQuery };
+    case "debouncedSearch":
+      return { status: "searching", query: action.debouncedQuery };
+    case "waitForNewQuery":
+      return { status: "done" };
+    default:
+      throw new TypeError("Unexpected action passed to reducer!");
+  }
+};
```
ChristianMurphy, line 92: really like the finite state machine approach here! :+1: 
It makes me wonder a bit if it would make sense to pull in something like https://xstate.js.org/docs or https://thisrobot.life ? or if that would be overkill? :thinking:

ChristianMurphy: Also would it make sense to invoke the search from inside the state machine?
Something along the lines of https://xstate.js.org/viz/?gist=f68893d849bd0eba2afd6a5a7a0769df

ChristianMurphy (approved):
Thanks @PenghaiZhang!

edalex-ian (changes requested):
Heya @PenghaiZhang 

As discussed the naming was a bit confusing and I also think we can get rid of the query in its own state and just use the reducer's state.

I've attached the diff that came out of our trials. Obviously you'll need to make sure any of it works first.

Cheers.

[searchbar.diff.txt](https://github.com/openequella/openEQUELLA/files/6388665/searchbar.diff.txt)

Source/Plugins/Core/com.equella.core/js/tsrc/search/components/SearchBar.tsx
edalex-ian: I like the look of XState, that'd be cool to take for a spin - some reasonable examples of React/TS too. It'd be interesting to redo the SearchPage with it.

PenghaiZhang:
> Heya @PenghaiZhang
> 
> As discussed the naming was a bit confusing and I also think we can get rid of the query in its own state and just use the reducer's state.
> 
> I've attached the diff that came out of our trials. Obviously you'll need to make sure any of it works first.
> 
> Cheers.
> 
> [searchbar.diff.txt](https://github.com/openequella/openEQUELLA/files/6388665/searchbar.diff.txt)

Thanks Ian for the diff file.  I have tested locally again and the SearchBar works well!

edalex-ian (approved):
Good stuff @PenghaiZhang , I reckon let's merge it in.

Maybe then if at some-point someone has time, we could see what things look like if we change this reducer and the SearchPage one to use XState. :thinking: (But definitely something separate to this PR.)

PenghaiZhang:
> Good stuff @PenghaiZhang , I reckon let's merge it in.
> 
> Maybe then if at some-point someone has time, we could see what things look like if we change this reducer and the SearchPage one to use XState.  (But definitely something separate to this PR.)

Good plan ! Maybe in the next tech debt ?

edalex-ian:
I think that's just flaky and could be merged.

PR #2972, opened 2021-05-07
Recreate AuditLogDao and drop AuditLogJavaDao

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

This PR recreates `AuditLogDao` and its implementation, and uses `AuditLogDao` in `AuditLogService` and `AuditLogConverter`.

During implementation, I found a bug where exported audit logs cannot be imported back. The reason is the folder structure of these logs are changed. So I also added a fix.

Lastly, I dropped two files: `AuditLogJavaDao.scala` and `AuditLogEntryXml.java`.


Source/Plugins/Core/com.equella.core/src/com/tle/core/auditlog/convert/AuditLogConverter.java
```diff
@@ -100,8 +104,22 @@ public void doExport(
...
+        List<AuditLogEntry> entries =
+            auditLogDao.findAllByCriteria(
+                null,
+                page * PER_XML_FILE,
+                PER_XML_FILE,
+                Restrictions.eq("institution", institution));
+
+        xmlHelper.writeXmlFile(auditFolder, page + ".xml", entries, xstream);
+        message.incrementCurrent();
+      }
```
edalex-ian, line 121: Could we go a little more declarative?

```suggestion
      IntStream.range(0,count / PER_XML_FILE + 1).forEach(page -> {
        List<AuditLogEntry> entries =
            auditLogDao.findAllByCriteria(
                null,
                page * PER_XML_FILE,
                PER_XML_FILE,
                Restrictions.eq("institution", institution));

        xmlHelper.writeXmlFile(auditFolder, page + ".xml", entries, xstream);
        message.incrementCurrent();
      }
```

Source/Plugins/Core/com.equella.core/src/com/tle/core/auditlog/convert/AuditLogConverter.java
```diff
@@ -100,8 +104,22 @@ public void doExport(
...
+    int count = (int) auditLogDao.countByCriteria(Restrictions.eq("institution", institution));
+    if (count > 0) {
+      // Read all audit logs by paging.
+      for (int page = 0; page < count / PER_XML_FILE + 1; page++) {
+        List<AuditLogEntry> entries =
+            auditLogDao.findAllByCriteria(
+                null,
+                page * PER_XML_FILE,
+                PER_XML_FILE,
+                Restrictions.eq("institution", institution));
```
edalex-ian, line 117: This is now the third time this Criterion is declared in this file. Perhaps there's a need in the DAO to create a method to return a detached criteria to filter by institution.

Source/Plugins/Core/com.equella.core/src/com/tle/core/auditlog/impl/AuditLogDaoImpl.java
```diff
@@ -0,0 +1,45 @@
...
+public class AuditLogDaoImpl extends AbstractAuditLogDaoImpl<AuditLogEntry> implements AuditLogDao {
+  public AuditLogDaoImpl() {
+    super(AuditLogEntry.class);
+  }
+
+  @Override
+  public void logWithRequest(AuditLogEntry entry, HttpServletRequest request) {
+    ObjectMapper mapper = new ObjectMapper();
+    ObjectNode node = mapper.createObjectNode();
+    node.put("referrer", request.getHeader("Referer"));
```
edalex-ian, line 40: Rather than just doing this freehand, would it make sense to create a simple nested class for these types of entries?

Source/Plugins/Core/com.equella.core/src/com/tle/core/auditlog/impl/AuditLogDaoImpl.java
```diff
@@ -0,0 +1,45 @@
...
+
+@Bind(AuditLogDao.class)
+@Singleton
+public class AuditLogDaoImpl extends AbstractAuditLogDaoImpl<AuditLogEntry> implements AuditLogDao {
+  public AuditLogDaoImpl() {
+    super(AuditLogEntry.class);
+  }
+
+  @Override
+  public void logWithRequest(AuditLogEntry entry, HttpServletRequest request) {
```
edalex-ian, line 37: I'm not convinced this should be here, shouldn't it be in the Service layer?

Source/Plugins/Core/com.equella.core/src/com/tle/core/institution/convert/XmlHelper.java
```diff
@@ -128,11 +130,27 @@ public void writeXmlFile(TemporaryFileHandle file, String path, Object obj) {
...
   public List<String> getXmlFileList(final TemporaryFileHandle folder) {
     if (isBucketed(folder)) {
-      return fileSystemService.grep(folder, "", "*/*.xml");
+      return fileSystemService.grep(folder, "", EXPORT_BUCKET_FILE_PATTERN);
     }
     return fileSystemService.grep(folder, "", "*.xml");
   }
 
+  /**
+   * Use the provided pattern to find XML files first. If none is found then find by standard
```
edalex-ian, line 139:
```suggestion
   * Use the provided pattern to find XML files first. If none are found then find by standard
```

Source/Plugins/Core/com.equella.core/src/com/tle/core/auditlog/impl/AuditLogServiceImpl.java
```diff
@@ -168,14 +179,21 @@ public void logContentViewed(
...
+            CurrentUser.getSessionID(),
+            category,
+            CONTENT_VIEWED_TYPE,
+            new Date(),
+            itemId.getUuid(),
+            Integer.toString(itemId.getVersion()),
+            contentType,
+            path,
+            CurrentInstitution.get(),
+            null);
```
edalex-ian, line 194: I was going to comment that you have three instances here where the last param to this take all constructor is `null`. How about having another constructor without it. HOWEVER...

I see that is meta which is being provided by `logWithRequest`. So perhaps that further supports the need for `logForRequest` in this layer.

edalex-ian (commented):
Looks mostly there @PenghaiZhang , but just a couple of suggestions for you.

Source/Plugins/Core/com.equella.core/src/com/tle/core/institution/convert/XmlHelper.java
```diff
@@ -128,11 +130,27 @@ public void writeXmlFile(TemporaryFileHandle file, String path, Object obj) {
...
+   * pattern.
+   *
+   * @param folder Folder where XML files are searched for.
```
SammyIsConfused, line 142:
```suggestion
   * @param folder Folder in which to search for XML files.
```

Source/Plugins/Core/com.equella.core/src/com/tle/core/auditlog/impl/AuditLogDaoImpl.java
SammyIsConfused, line 40: Is there a typo in this? 
"referrer" and "Referer" are spelled differently.

SammyIsConfused (commented):
Just a comment tweak and a query so far Penghai! Nice work.

Source/Plugins/Core/com.equella.core/src/com/tle/core/auditlog/convert/AuditLogConverter.java
```diff
@@ -100,8 +104,22 @@ public void doExport(
 
     // write out the format details
     xmlHelper.writeExportFormatXmlFile(auditFolder, true);
-    AuditLogJavaDao.writeExport(
-        auditFolder, PER_XML_FILE, institution, message, xmlHelper, xstream);
+
+    int count = (int) auditLogDao.countByCriteria(Restrictions.eq("institution", institution));
```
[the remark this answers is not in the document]
PenghaiZhang: I have added a new method named 'countByInstitution' in service.

Source/Plugins/Core/com.equella.core/src/com/tle/core/auditlog/convert/AuditLogConverter.java
PenghaiZhang: Not a problem!  I thought `IntStream` might be overkilled so I did not use it.

edalex-ian: On some level it is heavy handed, but seeing this loop is not the performance bottle neck in this block of code, I think the trade-off for robustness and readability is warranted.

Source/Plugins/Core/com.equella.core/src/com/tle/core/auditlog/impl/AuditLogDaoImpl.java
PenghaiZhang: Yes it should. I have moved it to the Service layer.

Source/Plugins/Core/com.equella.core/src/com/tle/core/auditlog/impl/AuditLogDaoImpl.java
PenghaiZhang: I think these two spelling are both right?  But I can change `referrer` to `referer` .

PenghaiZhang: I have created a simple nested class which has only one field `referer`, but I  wonder if we should have a class not bound to `referer` in case we need different fields in the future.

edalex-ian: I think it's all fine, as it captures the shape of `meta`data which is being captured for this audit record entry type. If in the future, we want different fields we modify it then and as needed.

Keep in mind though, that the new nested class should be specific to this audit entry - it's not a generic one to cater to all the different structures that might get dumped in `meta`.

PenghaiZhang: Yes I have added a very specific one named `AuditLogMetaReferer`.

Source/Plugins/Core/com.equella.core/src/com/tle/core/auditlog/impl/AuditLogServiceImpl.java
```diff
@@ -233,6 +237,42 @@ public void logSearchExport(String format, String searchParams) {
...
+      String d4,
+      HttpServletRequest request) {
+
+    ObjectMapper mapper = new ObjectMapper();
+    AuditLogMetaReferer referer = new AuditLogMetaReferer(request.getHeader("Referer"));
+    String meta = "";
+    try {
+      meta = mapper.writeValueAsString(referer);
+    } catch (JsonProcessingException e) {
+      LOGGER.error("Failed to build a JSON string for meta of audit log.");
```
edalex-ian, line 256: If we continue to create an audit entry here, it'll be malformed. Do we need a `RuntimeException`?

Source/Plugins/Core/com.equella.core/src/com/tle/core/auditlog/impl/AuditLogServiceImpl.java
```diff
@@ -284,9 +326,37 @@ public void setPluginService(PluginService pluginService) {
...
+  @Override
+  public int countByInstitution(Institution institution) {
+    return (int) dao.countByCriteria(Restrictions.eq("institution", institution));
+  }
+
+  @Override
+  public List<AuditLogEntry> findAllByInstitution(
+      Order order, int firstResult, int maxResults, Institution institution) {
+    return dao.findAllByCriteria(
+        order, firstResult, maxResults, Restrictions.eq("institution", institution));
```
edalex-ian, line 344: This `Restrictions.eq("institution", institution)` is duplicated and contains knowledge/logic (as simply as it is) how to restrict to an institution and a string literal (prone to human error - with no detection until runtime, but not even then). As mentioned before, it should be self contained in a DetachedCriteria - if needs be, have a builder function for it e.g. `DetachedCriteria restrictByInstitution(Institution institution)`.

That said, you may need a Criterion here, so perhaps a different function signature etc. But hopefully you see what I mean.

Source/Plugins/Core/com.equella.core/src/com/tle/core/auditlog/impl/AuditLogServiceImpl.java
```diff
@@ -284,9 +326,37 @@ public void setPluginService(PluginService pluginService) {
...
+  }
+
+  static class AuditLogMetaReferer {
```
edalex-ian, line 347: Going back to the discussion, you probably want to more so tie this to the log entry type, so maybe a name like `HttpRequestMeta`. If you do that, then you constructor could also be `public HttpRequestMeta(HttpRequest request)` and you can do the transformation there. This will also provide flexibility in the future and how it's used - i.e. it could also be nested in another Meta structure/class.

Please also add JavaDoc at the class level.

edalex-ian (commented):
Almost there @PenghaiZhang . Just a few last things.

Thanks! :+1:

Source/Plugins/Core/com.equella.core/src/com/tle/core/auditlog/impl/AuditLogServiceImpl.java
PenghaiZhang: This will require extra changes in some very low level classes in order to support `DetachedCriteria` -> `Criteria` . I think we need to talk tomorrow. :smile:

edalex-ian (approved):
Good stuff @PenghaiZhang . Merge on green. I see there's been a scalacheck failure, but that passed on the other two combinations so I'm calling flakey. :wink:

PR #3004, opened 2021-05-14
Feature/drop Scala 'SettingsDB'

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

The intention of this PR is to drop `SettingsDB.scala`.  But in the end, I also drop `ACLChecks.scala`.
Key changes are: 

1. Use `ConfigurationService` to handle UI Settings. You can see that `SettingsDB` essentially interacts with table `configuration_property` so I think reusing  `ConfigurationService` is a good choice. In addition, `ConfigurationService` has already implemented cache on Institution level. Although I am not 100% sure the differences between this cache and the SimpleDBA cache yet, the testing I have done proves they have no significant differences.

2. Added a new function `hasPrivilege` to replace `ensureEditSystem` defined in SettingsDB. This change results in changes in 
`CloudProviderApi.scala` and `SettingsResource`.

After above two changes there were no usage of SettingsDB. And I found there was only one usage of  `ACLChecks.scala` in `SettingsList.scala` so I also managed to remove that usage and then dropped ACLChecks.

I also tested all these changes in clustered environment as updating `ConfigurationProperty` creates an event `ConfigurationChangedEvent` and publishes to other nodes.

**ACL check for adding a Cloud provider**
Screenshot from 2021-05-14 10-46-37

**ACL check for editing UI settings**
Screenshot from 2021-05-14 10-47-09


Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/settings/SettingsList.scala
```diff
@@ -42,25 +42,9 @@ object CoreSettingsPage {
             nameKey: String,
             descKey: String,
             page: String,
-            editable: () => Boolean) =
-    SettingsPage(CoreStrings.lookup, id, group, nameKey, descKey, page, "web", editable)
-}
-
-object CoreSettingsRest {
```
PenghaiZhang, line 49: I am not sure the benefits of  having this object rather than reusing object `CoreSettingsPage`.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/settings/SettingsList.scala
```diff
@@ -77,13 +61,17 @@ object SettingsList {
...
     "ui",
     "ui",
     "uisettings.name",
     "uisettings.desc",
     "api/settings/ui",
-    AclChecks.filterNonGrantedPrivileges(Iterable("EDIT_SYSTEM_SETTINGS"), true))
+    () =>
+      LegacyGuice.aclManager.hasPrivilege(List(SecurityConstants.EDIT_SYSTEM_SETTINGS).asJava,
+                                          true),
+    "rest"
```
PenghaiZhang, line 73: I think it should be "web" but it has been "rest" for a while so I did not change it.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/settings/UISettings.scala
```diff
@@ -18,42 +18,39 @@
...
+import com.tle.common.institution.CurrentInstitution
+import io.circe.syntax._
+import com.tle.legacy.LegacyGuice
+import io.circe.generic.auto._
+import io.circe.parser.parse
 
 case class NewUISettings(enabled: Boolean, newSearch: Boolean = false)
 
 case class UISettings(newUI: NewUISettings) {
+  @JsonIgnore
```
PenghaiZhang, line 31: We should probably ignore this as we don't even have `newSearchActive` in the ts type for `UISettings` defined in REST module .

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/settings/UISettings.scala
```diff
@@ -18,42 +18,39 @@
...
-      uiSettingsCache.invalidate.apply()
-
-  def cachedUISettings: DB[Option[UISettings]] = uiSettingsCache.get.apply()
-}
-
-object UISettingsJava {
-  def getUISettings: UISettings = RunWithDB.executeWithHibernate {
-    UISettings.getUISettings.map(_.getOrElse(UISettings.defaultSettings))
+  def getUISettings: UISettings = {
+    Option(CurrentInstitution.get()) match {
```
PenghaiZhang, line 42: Previously, this function was mostly called within `executeIfInInstitution`. So I moved that logic to here.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/settings/UISettings.scala
edalex-ian: hmmm, this makes me wonder why we even have UISettings. It's curious that even the endpoints return and accept `UISettings` when they only operate on `NewUISettings`. Is this a case of YAGNI supporting the idea that in the future we might want other UI settings - not specific to NewUI?

Source/Plugins/Security/com.tle.core.security/src/com/tle/core/security/TLEAclManager.java
```diff
@@ -49,6 +49,15 @@
...
   boolean hasPrivilege(Object domainObj, Privilege... privilege);
 
+  /**
+   * Check if user has the supplied privileges for any object.
+   *
+   * @param privileges List of privileges to be checked
+   * @param includePossibleOwnerAcls true to include 'ownerAcl'.
+   * @return true if ANY of the supplied privileges are granted.
+   */
+  boolean hasPrivilege(Collection<String> privileges, boolean includePossibleOwnerAcls);
```
edalex-ian, line 59: Good stuff @PenghaiZhang ! :+1:

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/settings/SettingsList.scala
edalex-ian: Be great to know the point of this.... Wonder how/where web/rest is used.

edalex-ian (commented):
Looks pretty good @PenghaiZhang . Just wondering if maybe we can sneak in those two hasAcl and hasAclOrThrow functions. (Maybe even a sneaky refactor in the endpoint classes where they'd fit nicely. :thinking:

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/settings/SettingsList.scala
PenghaiZhang: It's used to determine what sort of links returned to front-end in `SettingsResource`. I also found something else that we need to discuss

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/settings/SettingsList.scala
PenghaiZhang: I believe we can drop this object.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/settings/UISettings.scala
PenghaiZhang: I am not aware of any UI settings that we would implement for Old UI. 

I can try to drop  `UISettings`.

PenghaiZhang: Since the old structure `{newUI: {enable: true, newSearch: true}}` already exists in database, 
a simple case class like this `case class UISettings(enabled: Boolean = false, newSearch: Boolean = false)` does not work due to the field `newUI`. Looks like need to do something like `get(key)` and `set(key, {enable: true, newSearch: true})`.

But that seems to make things worse.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/settings/SettingsApiHelper.scala
```diff
@@ -30,4 +33,12 @@ object SettingsApiHelper {
   def updateSettings[T <: ConfigurationProperties](settings: T): Unit = {
     LegacyGuice.configService.setProperties(settings)
   }
+
+  def ensureEditSystem(): Unit = {
+    val hasSettingsAcl =
+      LegacyGuice.aclManager.hasPrivilege(List(SecurityConstants.EDIT_SYSTEM_SETTINGS).asJava, true)
```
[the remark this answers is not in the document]
PenghaiZhang: Not a problem! I will go with a single ACL for now.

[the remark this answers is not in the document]
PenghaiZhang: Scala friendly reminds me that `Accessor-like named method has Unit result type ` for `hasAclOrThrow`. So how about we rename it to `checkAclOrThrow`?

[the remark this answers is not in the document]
PenghaiZhang: Well `checkAclOrThrow` does not sound quite right. Maybe we should go with `hasAclOrThrow`.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/security/ACLChecks.scala
```diff
@@ -0,0 +1,48 @@
...
+    LegacyGuice.aclManager.hasPrivilege(Set(privilege).asJava, includePossibleOwnerAcls)
+  }
+
+  /**
+    * Checks if the current user has the specified ACL or throws `PrivilegeRequiredException`.
+    *
+    * @param privilege Required ACL, typically defined as a constant in `com.tle.common.security.SecurityConstants`
+    * @param includePossibleOwnerAcls `true` to include possible owner ACLs
+    */
+  def hasAclOrThrow(privilege: String, includePossibleOwnerAcls: Boolean = false): Unit = {
```
edalex-ian, line 43: Why `includePossibleOwnerAcls: Boolean = false` here yet in `hasAcl` it's defaulting to `true`? Seeing `true` seems the common use, would seem to make sense `true` remains the default.

edalex-ian (commented):
Changes look mostly good with the `ACLChecks` object, however query around default param.

And I'm unclear if you're saying we actually have an issue with the UISettings setup. Could you please clarify?

Also, any chance we can sneak in using the new `hasAcl` or `hasAclOrThrow` in the other scala code which uses `LegacyGuice.aclManager`? For example:

- SearchMenuContributor
- LegacyContentApi
- SearchHelper
- SearchResource

There's a couple others, but then we need support for a Set of privs, so they can stay as is for now.

I know this is kind of outside the scope of this PR, but I really think it'd be worthwhile now with such a useful function vs. the long chains.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/security/ACLChecks.scala
PenghaiZhang: Sorry I probably accidentally change one of the default values. Actually, `false` seems to be used more.

PenghaiZhang:
> Changes look mostly good with the `ACLChecks` object, however query around default param.
> 
> And I'm unclear if you're saying we actually have an issue with the UISettings setup. Could you please clarify?
> 
> Also, any chance we can sneak in using the new `hasAcl` or `hasAclOrThrow` in the other scala code which uses `LegacyGuice.aclManager`? For example:
> 
> * SearchMenuContributor
> * LegacyContentApi
> * SearchHelper
> * SearchResource
> 
> There's a couple others, but then we need support for a Set of privs, so they can stay as is for now.
> 
> I know this is kind of outside the scope of this PR, but I really think it'd be worthwhile now with such a useful function vs. the long chains.

Apart from the support for a Set of privs, we also need to support checking ACL against an entity such as a Collection.

edalex-ian (approved):
GTG @PenghaiZhang :+1:

PR #3006, opened 2021-05-19
refactor(server): Replace decompiled MD5 implementation

edalex-ian:
##### Checklist

- [x] tests are included

##### Description of change

This class (Md5) was the source of the most warnings in the CheckStyle
report. When looking at the class it became obvious it was just a
decompiled MD5 implementation which is rather redundant when there are
many standard offerings out there now. Indeed, even across the oEQ base
although there are many uses of this, there are also uses of others.

So to at least tidy things up, this has been refactored. However
arguably a refactoring to consolidate hashing on a broader scale should
be undertaken. This might be easier now with this one tidied up.

(Added the unit test before the refactor so that I could validate there
was no change.)

### DIFF
--- a/Platform/Plugins/com.tle.platform.common/src/com/dytech/devlib/Md5.java
+++ b/Platform/Plugins/com.tle.platform.common/src/com/dytech/devlib/Md5.java
@@ -16,332 +16,134 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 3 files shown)


PenghaiZhang (approved):
Thanks Ian! Happy to see we advance 20 years from 2001 -> 2021.

PR #3011, opened 2021-05-20
Refactor `OAuthServerAccess.scala`

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

This PR replaces the Scala `CacheValue` with the Java one and replaces Scala `tokenQuery` with Java `ReplicatedCacheDao` in `OAuthServerAccess`.  The changes are mostly around the feature of Cloud provider. Due to my limited knowledge of Scala monad, I am honestly not 100% confident in whether the simplified code works as exactly the same as how those monads work. But I did run through the feature of Cloud provider locally and every things seem fine.

Unfortunately, we can't remove all the code related to the use of simpleDBA until we rewrite `CloudProviderDB`.

Register a CP.
image

Use a CP  Wizard control.
image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/oauthserver/OAuthServerAccess.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/oauthserver/OAuthServerAccess.scala
@@ -121,8 +121,10 @@ object OAuthServerAccess {
 
     case CloudProviderClient(clientId, _) =>
       // The value of CachedValue is encoded by Base64 so here we need to encode the client ID.
-      val encodedClientId = new Base64().encode(clientId.toString.getBytes(StandardCharsets.UTF_8))
-      Option(LegacyGuice.replicatedCacheDao.getByValue(CloudTokenCache, encodedClientId)) match {
+      val clientIdBytes = clientId.toString.getBytes(StandardCharsets.UTF_8)
+      Option(
+        LegacyGuice.replicatedCacheDao.getByValue(CloudTokenCache,
+                                                  new Base64().encode(clientIdBytes))) match {
         case Some(tokenEntry) =>
           CloudAuthToken(tokenEntry.getKey,
                          Option(tokenEntry.getTtl).map(_.toInstant).getOrElse(Instant.now))
@@ -133,7 +135,7 @@ object OAuthServerAccess {
             CloudTokenCache,
             tokenUuid,
             Date.from(expires),
-            clientId.toString.getBytes()
+            clientIdBytes
           )
           CloudAuthToken(tokenUuid, expires)
       }
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/oauthserver/OAuthServerAccess.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/oauthserver/OAuthServerAccess.scala
@@ -122,9 +122,7 @@ object OAuthServerAccess {
     case CloudProviderClient(clientId, _) =>
       // The value of CachedValue is encoded by Base64 so here we need to encode the client ID.
       val clientIdBytes = clientId.toString.getBytes(StandardCharsets.UTF_8)
-      Option(
-        LegacyGuice.replicatedCacheDao.getByValue(CloudTokenCache,
-                                                  new Base64().encode(clientIdBytes))) match {
+      Option(LegacyGuice.replicatedCacheDao.getByValue(CloudTokenCache, clientIdBytes)) match {
         case Some(tokenEntry) =>
           CloudAuthToken(tokenEntry.getKey,
                          Option(tokenEntry.getTtl).map(_.toInstant).getOrElse(Instant.now))
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/oauthserver/OAuthServerAccess.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/oauthserver/OAuthServerAccess.scala
@@ -18,22 +18,23 @@
 
 package com.tle.core.oauthserver
 
+import com.dytech.devlib.Base64
+import com.tle.common.institution.CurrentInstitution
 import java.time.{Duration, Instant}
 import java.util.{Date, UUID}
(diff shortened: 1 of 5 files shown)


Source/Plugins/Core/com.equella.serverbase/src/com/tle/core/replicatedcache/dao/ReplicatedCacheDao.java
```diff
@@ -26,6 +26,8 @@
 public interface ReplicatedCacheDao extends GenericDao<CachedValue, Long> {
   CachedValue get(String cacheId, String key);
 
+  CachedValue getByValue(String cacheId, String value);
```
edalex-ian, line 29: This should have symmetry with `put()` and so should be:

```suggestion
  CachedValue getByValue(String cacheId, byte[] value);
```

That also then means you can keep the knowledge around Base64 encoding encapsulated.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/oauthserver/OAuthServerAccess.scala
```diff
@@ -114,40 +113,31 @@ object OAuthServerAccess {
...
-                  .map(_ => CloudAuthToken(tokenUuid, expires))
-            }
-          } yield token
-        }.compile.lastOrError
+    case CloudProviderClient(clientId, _) =>
+      // The value of CachedValue is encoded by Base64 so here we need to encode the client ID.
+      val clientIdBytes = clientId.toString.getBytes(StandardCharsets.UTF_8)
+      Option(
+        LegacyGuice.replicatedCacheDao.getByValue(CloudTokenCache,
+                                                  new Base64().encode(clientIdBytes))) match {
```
edalex-ian, line 127: As mentioned in other comment, we don't want to be concerned with internal implementations of Base64 encoding here - that should be down in the `ReplcatedCacheDao`.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/oauthserver/OAuthServerAccess.scala
```diff
@@ -114,40 +113,31 @@ object OAuthServerAccess {
...
-                  }
-                  .map(_ => CloudAuthToken(tokenUuid, expires))
-            }
-          } yield token
-        }.compile.lastOrError
+    case CloudProviderClient(clientId, _) =>
+      // The value of CachedValue is encoded by Base64 so here we need to encode the client ID.
+      val clientIdBytes = clientId.toString.getBytes(StandardCharsets.UTF_8)
+      Option(
+        LegacyGuice.replicatedCacheDao.getByValue(CloudTokenCache,
```
edalex-ian, line 126: Why is this using the DAO rather than the Service (`ReplicatedCacheService`)?

Source/Plugins/Core/com.equella.serverbase/src/com/tle/core/replicatedcache/dao/ReplicatedCacheDaoImpl.java
```diff
@@ -63,6 +63,28 @@ public Object doInHibernate(Session session) throws HibernateException {
...
+    return (CachedValue)
+        getHibernateTemplate()
+            .execute(
+                new HibernateCallback() {
+                  @Override
+                  public Object doInHibernate(Session session) throws HibernateException {
+                    Query q =
+                        session.createQuery(
+                            "FROM CachedValue WHERE value = :value"
+                                + " AND cacheId = :cacheId AND institution = :institution");
```
edalex-ian, line 78: I know it's convention for this class (and many others) but I'd definitely like to see less use of HQL and more of Criteria and named queries to try and add some robustness. :disappointed:

edalex-ian (changes requested):
Heya @PenghaiZhang 

Just a suggestion to maintain encapsulation and also a query around DAO usage rather than Service.

BTY.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/oauthserver/OAuthServerAccess.scala
PenghaiZhang: I actually had asked myself the same question. According to the previous code, there seems no need to save the `CachedValue` in a cache. If you look into `ReplicatedCacheService` you will see it only provides a method which returns a cache.

I feel there is some confusion here. `ReplicatedCacheDao` is just a dao which serves entity `CachedValue`, but `ReplicatedCacheService` is not like other Services we have. It doesn't do anything about `CachedValue` but `Cache`.

edalex-ian: hmmm, let's discuss in person tomorrow. From a quick look now it'd seem that `ReplicatedCacheService` manages `Caches` with the default implementation being cluster aware. Seems to then use a cache that is stored locally on a server, but then distributed across the cluster via the DB. (kind of like hibernates L1/L2 caches)

Source/Plugins/Core/com.equella.serverbase/src/com/tle/core/replicatedcache/dao/ReplicatedCacheDaoImpl.java
PenghaiZhang: Not a problem. I have replaced this HQL with Criteria.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/oauthserver/OAuthServerAccess.scala
edalex-ian: Okay, after our discussion I can more importantly see that you have matched the code as was.

The question about the use of the `ReplicatedCacheService` will more so come in when we look at  `OAuthClientService.scala`. As we've noted, in that file is seems to use the scala re-implementation of `ReplicatedCacheService` in the form of `DBCacheBuilder`. So that'll be interesting times.

edalex-ian (approved):
GTG! Thanks @PenghaiZhang

PR #3038, opened 2021-05-26
chore(deps): Remove unnecessary ts-node

edalex-ian:
##### Checklist

- [ ] tests are included

##### Description of change

Not sure why this was there, however based on what it is (a TS REPL) and the fact all tests etc pass without it, I think we can drop it.

### DIFF
--- a/oeq-ts-rest-api/package-lock.json
+++ b/oeq-ts-rest-api/package-lock.json
@@ -791,30 +791,6 @@
         "@sinonjs/commons": "^1.7.0"
       }
     },
-    "@tsconfig/node10": {
-      "version": "1.0.7",
-      "resolved": "https://registry.npmjs.org/@tsconfig/node10/-/node10-1.0.7.tgz",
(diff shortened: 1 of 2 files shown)


ChristianMurphy (approved):
Huh, if I recall correctly at one point there were some ts scripts.
Good catch spotting those have been factored out of the application. :+1:

edalex-ian:
> some ts scripts

Oh interesting, is that what it's commonly used for - having TS based scripts run from your package.json? (As maybe some of the TSDX stuff would've used that.)

PenghaiZhang (approved):

ChristianMurphy:
> Oh interesting, is that what it's commonly used for - having TS based scripts run from your package.json

You got it, its primary use case is being able to call `ts-node sctipt.ts` instead of `node script.js`.

> As maybe some of the TSDX stuff would've used that

That could be :thinking: 
The user guide for tsdx does reference ts-node https://github.com/formium/tsdx/blob/462af2d002987f985695b98400e0344b8f2754b7/templates/basic/README.md#tsdx-user-guide

PR #3039, opened 2021-05-26
chore(ci): Include ESLint checks for oeq-ts-rest-api

edalex-ian:
##### Checklist

- [ ] commit message follows [commit guidelines][]
- [ ] tests are included

##### Description of change

We seem to be missing these checks, and currently their are a couple of issues. ~~So at first I expect this PR to fail, and then we need to make a few minor commits to address the issues. (But I want to confirm it fails first to ensure I've probably added the new task.)~~

A quick run down of the issues being fixed:

- `Search.ts` : Add missing explicit return types - straightforward enough, and of good value;
- `Utils.ts` : Stop using `any` and replace `unknown` for known objects to `Record<string,unknown>` as per TS best practice. Also as a result add some more explicit run-time type validation to ensure we're working on what we think we are. Which ultimately, is why ESLint raises warnings around `any`.
- `schema.test.ts` : Rework conditional `expect` calls to ensure no false positives. (This file could probably be updated further to use async/await, but that'll do for now to pass ESLint and not blur things.)

### DIFF
--- a/oeq-ts-rest-api/src/Utils.ts
+++ b/oeq-ts-rest-api/src/Utils.ts
@@ -16,6 +16,17 @@
  * limitations under the License.
  */
 import { cloneDeep, set } from 'lodash';
+import { is } from 'typescript-is';
+
+/**
+ * The idiomatic modelling for an `object` in Typescript is to use a `Record<string, unknown>`, so
+ * this is a simple type guard to help below where we want to go from an unknown potential object
+ * to a type more suitable for TS contexts.
+ *
+ * @param r a potential `object` / `Record`
+ */
+const isRecord = (r?: unknown): r is Record<string, unknown> =>
+  r !== null && is<Record<string, unknown>>(r);
 
 /**
  * Performs inplace conversion of specified fields with supplied converter.
@@ -31,16 +42,9 @@ const convertFields = <R>(
   recursive: boolean,
   converter: (value: unknown) => R
 ): void => {
-  const entries: [string, unknown][] = Object.entries(input);
-
-  entries.forEach(([field, value]) => {
-    if (value && typeof value === 'object' && recursive) {
-      convertFields(
-        value as Record<string, unknown>,
-        targetFields,
-        recursive,
-        converter
-      );
+  Object.entries(input).forEach(([field, value]) => {
+    if (isRecord(value) && recursive) {
+      convertFields(value, targetFields, recursive, converter);
     } else {
       targetFields
         .filter((targetField) => targetField === field)
@@ -58,9 +62,6 @@ const convertFields = <R>(
  */
 export const convertDateFields = <T>(input: unknown, fields: string[]): T => {
   const cloneRecord = (obj: unknown): Record<string, unknown> => {
-    const isRecord = (r?: unknown): r is Record<string, unknown> =>
-      r !== null && typeof r === 'object';
-
     const cloned = cloneDeep(obj);
     if (isRecord(cloned)) {
       return cloned;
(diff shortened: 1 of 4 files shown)


edalex-ian:
Nice, so that failed as expected. Now to fix the issues.

oeq-ts-rest-api/src/Utils.ts
```diff
@@ -54,11 +57,34 @@ const convertFields = <T, R>(
...
+        : new Date(maybeDateString);
+    } else {
+      return undefined;
+    }
+  };
+
+  const inputClone: Record<string, unknown> = cloneRecord(input);
+  convertFields(inputClone, fields, true, potentialStringToDate);
+
+  return inputClone as T;
```
edalex-ian, line 87: I don't love this, but at least it's explicit where as before it was just implied the `unknown` could be returned as a `T`. :man_shrugging:

oeq-ts-rest-api/src/Utils.ts
```diff
@@ -25,23 +25,26 @@ import { cloneDeep } from 'lodash';
...
-  input: unknown,
+const convertFields = <R>(
+  input: Record<string, unknown>,
   targetFields: string[],
   recursive: boolean,
-  converter: (value: T) => R
+  converter: (value: unknown) => R
 ): void => {
-  const entries: [string, any][] = Object.entries(input as any);
+  const entries: [string, unknown][] = Object.entries(input);
```
ChristianMurphy, line 34: The typing here may be optional, testing on the ts playground it looks like `[string, unknown][]` is inferred https://www.typescriptlang.org/play?#code/MYewdgzgLgBApgDwIYFsAOAbOAuGAlOUAJwBMAeaIgSzAHMAaGAVzAGswQB3MAPhgF4YAbwC+AKDGhIsKHGgCYAeQBGAK0JQAdHDBRqcgBSJUmOAEogA

```suggestion
  const entries = Object.entries(input);
```

oeq-ts-rest-api/src/Utils.ts
ChristianMurphy: For the current approach this is probably fine.
It could still be interesting to explore a lens approach to transforming the data, but that is well outside the scope of this PR.

ChristianMurphy (approved):

oeq-ts-rest-api/src/Utils.ts
edalex-ian: Ah nice. I'll take it for a spin. Thanks.

oeq-ts-rest-api/src/Utils.ts
```diff
@@ -25,23 +25,26 @@ import { cloneDeep } from 'lodash';
...
 
   entries.forEach(([field, value]) => {
-    if (typeof value === 'object' && recursive) {
-      convertFields(value, targetFields, recursive, converter);
+    if (value && typeof value === 'object' && recursive) {
+      convertFields(
+        value as Record<string, unknown>,
```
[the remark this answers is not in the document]
edalex-ian: Hmmm, interesting thought. I'll see if I can get something to work as it doesn't make for the clearest read (checking typeof 'object' and then casting to Record).

PenghaiZhang (approved):
LGTM

oeq-ts-rest-api/src/Utils.ts
edalex-ian: Yes, with that done it did kind reduce readability further down, so I simply merged the lines. All done. :+1:

oeq-ts-rest-api/src/Utils.ts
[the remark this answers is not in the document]
edalex-ian: Yup, all nicely working with `typescript-is` after a slight refactor - which also allowed for some simple doco.

edalex-ian:
Alright, should give a green build - tests passed locally still after last commit - so will merge when green.

ChristianMurphy (approved):

PR #3040, opened 2021-05-26
Bugfix/jsonb not supported in oracle

PenghaiZhang:
#### Checklist

- [ ] tests are included

##### Description of change

Since `jsonb` is a Postgres specific column type, previous testing work found errors when oEQ was installed with SQL Server or Oracle. To fix this issue, a custom hibernate type has been created to support saving JSON strings in databases. This custom type returns type 'nclob' for Oracle, `nvarchar(max)` for SQL Server and `jsonb` for Postgres.

Video for Oracle
https://user-images.githubusercontent.com/47203811/119589398-09542700-be16-11eb-996b-01b4fe000fd1.mp4

Video for Postgres
https://user-images.githubusercontent.com/47203811/119589615-7d8eca80-be16-11eb-937d-ed7ab45e7c64.mp4

Video for SQL Server
https://user-images.githubusercontent.com/47203811/119589751-d2cadc00-be16-11eb-8fd0-a4bf7570fce1.mp4

### DIFF
--- a/Source/Plugins/Core/com.equella.serverbase/src/com/tle/core/hibernate/type/HibernateCustomTypes.java
+++ b/Source/Plugins/Core/com.equella.serverbase/src/com/tle/core/hibernate/type/HibernateCustomTypes.java
@@ -37,8 +37,8 @@ public class HibernateCustomTypes {
       new CustomType(new HibernateCsvType(Types.VARCHAR), new String[] {"csv"});
 
   public static List<BasicType> getCustomTypes(ExtendedDialect dialect) {
-    final CustomType TYPE_JSON_STRING =
-        new CustomType(new JsonStringCustomType(dialect), new String[] {"json_string"});
-    return Arrays.asList(TYPE_JSON_STRING, TYPE_BLANKABLE, TYPE_CSV, TYPE_XSTREAM);
+    final CustomType TYPE_JSON =
+        new CustomType(new JsonStringCustomType(dialect), new String[] {"json"});
+    return Arrays.asList(TYPE_JSON, TYPE_BLANKABLE, TYPE_CSV, TYPE_XSTREAM);
   }
 }
--- a/Source/Plugins/Core/com.equella.serverbase/src/com/tle/core/hibernate/type/HibernateCustomTypes.java
+++ b/Source/Plugins/Core/com.equella.serverbase/src/com/tle/core/hibernate/type/HibernateCustomTypes.java
@@ -18,7 +18,7 @@
 
 package com.tle.core.hibernate.type;
 
-import com.vladmihalcea.hibernate.type.json.JsonBinaryType;
+import com.tle.core.hibernate.ExtendedDialect;
 import java.sql.Types;
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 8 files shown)


Source/Plugins/Core/com.equella.serverbase/src/com/tle/core/hibernate/type/HibernateCustomTypes.java
```diff
@@ -36,9 +36,9 @@
   private static final CustomType TYPE_CSV =
       new CustomType(new HibernateCsvType(Types.VARCHAR), new String[] {"csv"});
 
-  private static final BasicType TYPE_JSONB = new JsonBinaryType(String.class);
-
-  public static List<BasicType> getCustomTypes() {
-    return Arrays.asList(TYPE_JSONB, TYPE_BLANKABLE, TYPE_CSV, TYPE_XSTREAM);
+  public static List<BasicType> getCustomTypes(ExtendedDialect dialect) {
+    final CustomType TYPE_JSON_STRING =
+        new CustomType(new JsonStringCustomType(dialect), new String[] {"json_string"});
```
edalex-ian, line 41: Curious, why have you defined TYPE_JSON_STRING at the method level where as all the others were at the class level?

Aside from that, two other minor items:

1. should it just be TYPE_JSON?; and
2. should the name just be a simple "json" rather than "json_string"?

Doesn't really matter, just seems string is superfluous.

Source/Plugins/Core/com.equella.serverbase/src/com/tle/hibernate/dialect/ExtendedPostgresDialect.java
```diff
@@ -45,6 +45,7 @@ public ExtendedPostgresDialect() {
     super();
     uniqueDelegate = new InPlaceUniqueDelegate(this);
     registerColumnType(Types.BLOB, "bytea");
+    registerColumnType(Types.OTHER, "jsonb");
```
edalex-ian, line 48: I wonder if using `Types.OTHER` will limit us in any way. :thinking:  I wonder if we can use our own custom `Type` code. :thinking:

edalex-ian (commented):
Looks pretty good @PenghaiZhang . You'll see I've got one query about naming, and then one about the use of `OTHER`.

But on a broader scale, when I've done custom types like this in the past it's not uncommon to also do some of the serialisation/deserialisation in the get and set methods. For example, here you might bring in Jackson to go from/to Object/JSON. Should we consider that here?

Source/Plugins/Core/com.equella.serverbase/src/com/tle/core/hibernate/type/HibernateCustomTypes.java
PenghaiZhang: I defined it at the method level because it is dependent on the `dialect` passed in. 

Yeah I have renamed these two.

Source/Plugins/Core/com.equella.serverbase/src/com/tle/hibernate/dialect/ExtendedPostgresDialect.java
PenghaiZhang: In the custom type class, we use `Types.OTHER` as the SQL type for Postgres. So here we tell the dialect to use `jsonb` for columns whose SQL types are `OTHER`.

Source/Plugins/Core/com.equella.serverbase/src/com/tle/core/hibernate/type/HibernateCustomTypes.java
edalex-ian: Aaaaha! :+1:

edalex-ian (approved):
Good stuff @PenghaiZhang . Let's get it merged and take it for another spin. :+1:

PR #3047, opened 2021-05-28
Refactor 'OAuthClientService.scala'

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

This PR have quite a few changes made for 'OAuthClientService.scala`, including

1. Removing the usage of simpleDBA;
2. Replace the use of Scala cache implementation with the Java `ReplicatedCache`.

There is one remaining `DB` monad. Unfortunately, we can remove it until we rework Cloud provider.
 
In my mind, the only feature that gets impacted is Cloud provider, so I have lots of testing around it. But there might be features which I have missed and we need to test.

Tested with IntegTester. Also tested the replicated cache by debugging. 
Screenshot from 2021-05-28 11-40-41


Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/oauthclient/OAuthClientService.scala
```diff
@@ -81,126 +89,118 @@ object OAuthTokenState {
 
 }
 
-object OAuthClientService {
+object OAuthTokenCacheHelper {
```
PenghaiZhang, line 92: This object was a nested object. I think it's good to move it to the top level.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/oauthclient/OAuthClientService.scala
```diff
@@ -81,126 +89,118 @@ object OAuthTokenState {
...
+      .basic(token.clientId, token.clientSecret)
+      .body(
+        OAuthWebConstants.PARAM_GRANT_TYPE -> OAuthWebConstants.GRANT_TYPE_CREDENTIALS
+      )
+      .response(asJson[OAuthTokenResponse])
+      .post(uri"${token.authTokenUrl}")
+    lazy val oauthTokenState = postRequest
+      .send()
+      .map(_.unsafeBody.fold(de => throw de.error, responseToState))
+      .unsafeRunSync()
```
PenghaiZhang, line 118: A bit unsure about `unsafeRunSync` as the doc says ` You should ideally only call this function *once*, at the very end of your program.`

I was trying to avoid `unsafeRunSync` by a for comprehension similar to the previous code.
``` scala
for {
  state <- postRequest.send().map(....)
  _ <- IO(LegacyGuice.replicatedCacheDao.put(..., state))
}
yield state

```

However, I then got errors about `not current data source`.  I guess this is because current data source is saved in a specific thread and the thread is not accessible for the `IO(LegacyGuice.replicatedCacheDao.put)`.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/oauthclient/OAuthClientService.scala
```diff
@@ -81,126 +89,118 @@ object OAuthTokenState {
...
 
-  private val clientTokenCache = DBCacheBuilder.buildCache(
-    OAuthTokenCache,
-    CacheBuilder
-      .newBuilder()
-      .maximumSize(1000)
-      .expireAfterAccess(10, TimeUnit.MINUTES)
-      .build[String, AnyRef]()
-      .asMap())
+  def responseToState(response: OAuthTokenResponse): OAuthTokenState = {
```
PenghaiZhang, line 151: Not changes made for this function.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/oauthclient/OAuthClientService.scala
```diff
@@ -81,126 +89,118 @@ object OAuthTokenState {
...
-                          tokenType: OAuthTokenType.Value): IO[Response[T]] = {
+                          tokenType: OAuthTokenType.Value): Response[T] = {
     val authHeader = tokenType match {
       case OAuthTokenType.EquellaApi =>
         OAuthWebConstants.HEADER_X_AUTHORIZATION -> s"${OAuthWebConstants.AUTHORIZATION_ACCESS_TOKEN}=$token"
       case OAuthTokenType.Bearer =>
         OAuthWebConstants.HEADER_AUTHORIZATION -> s"${OAuthWebConstants.AUTHORIZATION_BEARER} $token"
     }
-    request.headers(authHeader).send[IO]()
+    request.headers(authHeader).send[IO].unsafeRunSync()
```
PenghaiZhang, line 193: `unsafeRunSync` again...

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/oauthclient/OAuthClientService.scala
```diff
@@ -81,126 +89,118 @@ object OAuthTokenState {
...
     val tokenRequest = TokenRequest(authTokenUrl, clientId, clientSecret)
-    for {
-      token    <- tokenForClient(tokenRequest)
-      response <- dbLiftIO.liftIO(requestWithToken(request, token.token, token.tokenType))
-      _        <- if (response.code == StatusCodes.Unauthorized) removeToken(tokenRequest) else ().pure[DB]
-    } yield response
+    val token        = tokenForClient(tokenRequest)
+    val res          = requestWithToken(request, token.token, token.tokenType)
+    if (res.code == StatusCodes.Unauthorized) removeToken(tokenRequest)
+    res.pure[DB]
```
PenghaiZhang, line 204: This is the `Applicative Functor` I asked before. The doc says `pure lifts any value into the Applicative Functor.` 

All I know is calling it can put the value into a monad which is `DB` here. Not sure if this is a good approach. But we do have more  `.pure[DB]` in other files.

Source/Plugins/Core/com.equella.serverbase/src/com/tle/core/replicatedcache/dao/ReplicatedCacheDaoImpl.java
```diff
@@ -122,6 +122,12 @@ public void put(final String cacheId, final String key, final Date ttl, final by
     save(cv);
   }
 
+  @Override
+  @Transactional
+  public void delete(CachedValue entity) {
```
PenghaiZhang, line 127: Got an exception complaining no existing transaction because the overridden `delete`  is annotated by `@Transactional(propagation = Propagation.MANDATORY)` in `GenericDaoImpl`.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/oauthclient/OAuthClientService.scala
edalex-ian: How does all this sit now that you've moved away from using `replicatedCacheDao` directly?

I've a feeling that if sttp is wrapping this in a monad then arguably you'll need the unsafeRunSync still. And yes, although they say only once at the end of the program, its always a bit different in this context and probably more so is at the end of any process/task.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/oauthclient/OAuthClientService.scala
edalex-ian: Maybe the short version helps: https://adit.io/posts/2013-04-17-functors,_applicatives,_and_monads_in_pictures.html

Or maybe the longer wordy version helps: http://learnyouahaskell.com/functors-applicative-functors-and-monoids#applicative-functors

But the short of it is that it's attempting to lift the response into a `DB` to match the return so that it can be used elsewhere as intended. (In a very simplified example, if we wanted to use java `Optional` as our Functor we'd be going with `Optional.of(res)`. Hence 'lifting' `res` into an `Optional`.)

I do wonder how `res.pure[DB]` differs from `dbLiftIO.liftIO(res). :thinking:

Source/Plugins/Core/com.equella.serverbase/src/com/tle/core/replicatedcache/ReplicatedCacheService.java
```diff
@@ -45,12 +46,41 @@
...
 
+    /**
+     * Similar to { @link #put(String, Serializable)} }, but also allow specifying TTL of the value
+     * saved in DB.
+     *
+     * @param key Key of the object
+     * @param value Value of the object
+     * @param dbEntryTTL TTL of the value saved in DB
+     */
+    default void put(@NonNull String key, @NonNull V value, Date dbEntryTTL) {
```
edalex-ian, line 80: Hmmm, I'm unsure about the default implementation simply ignoring the additional parameter. Feels like that could accidentally get very confusing. Would it be better to throw a `UnsupportedOperationException`?

And also:
* Is that name too specific, should it just be `ttl`?
* And why the old `Date` rather than `Instant` or one of the newer Java 8 date/time options?

Source/Plugins/Core/com.equella.serverbase/src/com/tle/core/replicatedcache/dao/ReplicatedCacheDaoImpl.java
edalex-ian: Hmmm, but does that mean transactional context has not been correctly established in the service layer?

Source/Plugins/Core/com.equella.serverbase/src/com/tle/core/replicatedcache/ReplicatedCacheService.java
```diff
@@ -45,12 +46,41 @@
...
   @NonNullByDefault
-  public interface ReplicatedCache<V extends Serializable> {
+  interface ReplicatedCache<V extends Serializable> {
     Optional<V> get(@NonNull String key);
 
     void put(@NonNull String key, @NonNull V value);
 
+    /**
+     * Similar to { @link #put(String, Serializable)} }, but also allow specifying TTL of the value
+     * saved in DB.
```
edalex-ian, line 74: So I'm slight confused why this is needed, as wouldn't that normally be looked after by the cache - I mean, it does have to be specified when the cache is instantiated. Is this in the case where you want to use a value other than the cache default?

I think it would be useful to catch the use case in the comment.

edalex-ian (commented):
Thanks @PenghaiZhang , some responses for you and also some queries. BTY.

Source/Plugins/Core/com.equella.serverbase/src/com/tle/core/replicatedcache/dao/ReplicatedCacheDaoImpl.java
PenghaiZhang: Ah actually, this overridden `delete` does not get used now. I will remove it. But yes, there is no methods in service layer to do a deletion of CachedValue.

Source/Plugins/Core/com.equella.serverbase/src/com/tle/core/replicatedcache/ReplicatedCacheService.java
PenghaiZhang: The reason for `Date` is just because lower levels use `Date`, but I can definitely change the `Date` here to `Instant`.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/oauthclient/OAuthClientService.scala
```diff
@@ -81,9 +80,38 @@ object OAuthTokenState {
...
+
+    lazy val newOAuthTokenState = postRequest
+      .send()
+      .map(_.unsafeBody.fold(de => throw de.error, responseToState))
+      .unsafeRunSync()
+
+    // Save the token in both cache and DB.
+    replicatedCache.put(tokenKey,
+                        newOAuthTokenState,
+                        Date.from(newOAuthTokenState.expires.getOrElse(Instant.now())))
```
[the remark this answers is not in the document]
PenghaiZhang: Ah yes, this is definitely wrong.
What about `getOrElse(Instant.ofEpochMilli(TimeUnit.MINUTES.toMillis(10))) ` ?

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/oauthclient/OAuthClientService.scala
```diff
@@ -93,114 +121,38 @@ object OAuthClientService {
...
-
-  private def tokenStillValid(token: OAuthTokenState): Boolean =
-    token.expires.isEmpty || token.expires.exists(_.isAfter(Instant.now))
-
-  def removeToken(tokenRequest: TokenRequest): DB[Unit] = {
-    OAuthTokenCache.removeFromDB(tokenRequest) *>
-      clientTokenCache.invalidate(tokenRequest).flatMap(dbLiftIO.liftIO)
+  def removeToken(): Unit = {
+    // Not only invalidate the cache but also remove the DB entries.
+    replicatedCache.invalidate()
```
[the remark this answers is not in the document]
PenghaiZhang: Yes it should only delete one token. I will fix this up.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/oauthclient/OAuthClientService.scala
PenghaiZhang: THanks for the links! The shorted version helps more :smile:

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/oauthclient/OAuthClientService.scala
PenghaiZhang: Yes, found an explanation

##
since the result.unsafeRunSync() is called at last , for-comprehension of IO monad execution is performed at “end of the world” — which is the desired and logical way of execution
##

edalex-ian (approved):
GTG @PenghaiZhang 

Thanks. :pray:

Issue #2707, opened 2021-01-28
MIME types can be configured such that they cause an error message in the new search UI


SammyIsConfused:
**Describe the bug**
If a MIME type entry is configured with a default viewer of File Viewer, but the Enabled checkbox for File Viewer is unchecked, then when an attachment of that MIME type turns up in a search on the new UI, an error snackbar appears at the bottom of the page labelled "Failed to get attachment viewer details: Missing viewer details for default viewer with id: "file"".

**To Reproduce**

1. As an admin, go to Settings -> General -> MIME Types.
2. Edit any MIME type entry or create a new one.
3. Scroll down to the Viewers section. Make the default radio button select File Viewer, and make sure the corresponding "Enabled" checkbox is unchecked. Save the MIME type.
4. If none exist, make a contribution that contains a file attachment of that MIME type.
5. Go to the search page and search for that item (it's likely already at the top of the page if you just added it)
6. Notice the error that displays at the bottom of the screen.

**Expected behavior**
There are a couple ways this could be fixed. We could make it so that if a MIME type is set to default then that option automatically checks Enabled and can't be unchecked without setting a new default. 

Or, we could handle this silently, such that if this configuration occurs it just uses a default viewer without displaying an error to the user, since most users wouldn't have MIME editing privileges and wouldn't be able to do anything about this.

**Screenshots**
![image](https://user-images.githubusercontent.com/24543345/106099106-4244fd80-618e-11eb-8550-4c6c6c5229da.png)
![image](https://user-images.githubusercontent.com/24543345/106099265-80422180-618e-11eb-93c7-cfafa37a6f79.png)

**Platform:**

- OpenEquella Version: 2020.2 onwards
- OS: Agnostic
- Browser: Agnostic

snehanaik22:
Done Testing

PR #3064, opened 2021-06-02
Bugfix/default viewer disabled

PenghaiZhang:
#2707 

- [ ] tests are included

##### Description of change

When attachment default viewer is not enabled, New Search UI complains that viewer detail is missing. To fix this issue, we need to do three things: 
1. fix existing data by a data migration;
2. fix imported data;
3. prevent users from saving a MIME type which does not enable its default viewer.

So a new function was added to the `MimeTypeService` to build the correct list of enabled viewers, and then this function is used in the new migration `EnableDefaultViewerMigration` and `MimeEntryConverter`. Lastly, UI side changes were made in `MimeDefaultViewerSection`.

Examples:

1. `video/x-flv`
Before migration: 
image
after:
image

2. `image/jpeg`
Before migration:
image
after:
image

Show an error message when save without enabling default viewer.
image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/resources/lang/i18n-resource-centre.properties
+++ b/Source/Plugins/Core/com.equella.core/resources/lang/i18n-resource-centre.properties
@@ -82,6 +82,7 @@
 /com.tle.core.entity.services.migration.v20202.facetedsearch.classification=Create a new table for faceted search classification
 /com.tle.core.entity.services.migration.v20202.removelastknownuserconstraint=Remove last known user composite constraint (username and institution ID)
 /com.tle.core.entity.services.migration.v20202.indexing.errored=Add column to attachment table to allow indexer to skip individual attachments that have failed.
+/com.tle.core.entity.services.migration.v20211.enable.default.viewer=Enable default viewer
 /com.tle.core.entity.services.query.contains={0} is {1}
 /com.tle.core.entity.services.query.date.after={0} after {1}
 /com.tle.core.entity.services.query.date.before={0} before {1}
--- a/Source/Plugins/Core/com.equella.core/resources/lang/i18n-resource-centre.properties
+++ b/Source/Plugins/Core/com.equella.core/resources/lang/i18n-resource-centre.properties
@@ -2747,6 +2747,7 @@ mimetypes.migration.title=Update default MIME Type icons
 mimetypes.settings.description=Display a list of MIME types, where the properties can be edited, including viewer defaults
 mimetypes.settings.title=MIME types
 mimetypes.title=MIME type editor
+mimetypes.default.viewer.not.enabled=Default viewer is not enabled
 missingprivileges=You do not have the required privilege to access this page\: {0}
 moddialog.accepted=Moderators already accepted
 moddialog.and=and
--- a/Source/Plugins/Core/com.equella.core/resources/lang/i18n-resource-centre.properties
+++ b/Source/Plugins/Core/com.equella.core/resources/lang/i18n-resource-centre.properties
@@ -82,7 +82,7 @@
 /com.tle.core.entity.services.migration.v20202.facetedsearch.classification=Create a new table for faceted search classification
 /com.tle.core.entity.services.migration.v20202.removelastknownuserconstraint=Remove last known user composite constraint (username and institution ID)
 /com.tle.core.entity.services.migration.v20202.indexing.errored=Add column to attachment table to allow indexer to skip individual attachments that have failed.
-/com.tle.core.entity.services.migration.v20211.enable.default.viewer=Enable default viewer
+/com.tle.core.entity.services.migration.v20211.enable.default.viewer=Ensure configured default viewers are also enabled
 /com.tle.core.entity.services.query.contains={0} is {1}
 /com.tle.core.entity.services.query.date.after={0} after {1}
 /com.tle.core.entity.services.query.date.before={0} before {1}
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/core/institution/migration/v20211/EnableDefaultViewerMigration.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/core/institution/migration/v20211/EnableDefaultViewerMigration.java
@@ -46,6 +46,11 @@ import org.hibernate.annotations.AccessType;
 import org.hibernate.annotations.MapKeyType;
 import org.hibernate.annotations.Type;
 
+/**
+ * Some existing MIME types do not have their default viewers enabled. This results in error
+ * messages showing in the New Search UI. So we use this migration to ensure the default viewer is
+ * in list of enabled viewers.
+ */
 @Bind
 @Singleton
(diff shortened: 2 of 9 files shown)


Source/Plugins/Core/com.equella.core/resources/lang/i18n-resource-centre.properties
```diff
@@ -82,6 +82,7 @@
 /com.tle.core.entity.services.migration.v20202.facetedsearch.classification=Create a new table for faceted search classification
 /com.tle.core.entity.services.migration.v20202.removelastknownuserconstraint=Remove last known user composite constraint (username and institution ID)
 /com.tle.core.entity.services.migration.v20202.indexing.errored=Add column to attachment table to allow indexer to skip individual attachments that have failed.
+/com.tle.core.entity.services.migration.v20211.enable.default.viewer=Enable default viewer
```
edalex-ian, line 85:
```suggestion
/com.tle.core.entity.services.migration.v20211.enable.default.viewer=Ensure configured default viewers are also enabled
```

Source/Plugins/Core/com.equella.core/src/com/tle/core/institution/migration/v20211/EnableDefaultViewerMigration.java
```diff
@@ -0,0 +1,103 @@
...
+import javax.persistence.Lob;
+import javax.persistence.MapKeyColumn;
+import org.hibernate.Session;
+import org.hibernate.annotations.AccessType;
+import org.hibernate.annotations.MapKeyType;
+import org.hibernate.annotations.Type;
+
+@Bind
+@Singleton
+public class EnableDefaultViewerMigration extends AbstractHibernateDataMigration {
```
edalex-ian, line 51: Be good to add JavaDoc to capture details about what this migration is for/doing.

Source/Plugins/Core/com.equella.core/src/com/tle/core/institution/migration/v20211/EnableDefaultViewerMigration.java
```diff
@@ -0,0 +1,103 @@
...
+  protected void executeDataMigration(
+      HibernateMigrationHelper helper, MigrationResult result, Session session) throws Exception {
+    final List<FakeMimeEntry> entries = session.createQuery("FROM MimeEntry").list();
+    for (FakeMimeEntry entry : entries) {
+      Map<String, String> attributes = entry.attributes;
+      String enabledViewers = mimeTypeService.getEnabledViewerList(attributes);
+      if (!Check.isEmpty(enabledViewers)) {
+        attributes.put(MimeTypeConstants.KEY_ENABLED_VIEWERS, enabledViewers);
+        session.update(entry);
+        session.flush();
```
edalex-ian, line 65: I'm missing something here, but this reads:

1. For each entry
2. Read in the attributes
3. From the attributes (from the entry) get the list of enabled viewers
4. If there are any enabled viewers
5. then write back the same values to the entry

So, the end result is no change. :confused: 

Now I'm going to take a punt and say you're relying on some kind of data correction in `getEnabledViewerList` and so really you should document that here in a comment.

Source/Plugins/Core/com.equella.core/src/com/tle/core/mimetypes/MimeTypeService.java
```diff
@@ -87,6 +88,15 @@
   @Nullable
   String getMimeEntryForAttachment(Attachment attachment);
 
+  /**
+   * Return a list of enabled viewers. If the default viewer does not exist in the list, add it to
+   * the list. However, if the default viewer is "file", it will not be included.
```
edalex-ian, line 93: > However, if the default viewer is "file", it will not be included.

Why not?

Often, the Why in a comment is the most important thing. It's that knowledge that the comment preserves when we come back to this and try and understand what is going on.

Source/Plugins/Core/com.equella.core/src/com/tle/core/mimetypes/MimeTypeService.java
```diff
@@ -87,6 +88,15 @@
...
+   *
+   * @param attributes Configured attributes of MIME type
+   * @return A string representing a list of enabled viewers
```
edalex-ian, line 96: What format is that string?

Source/Plugins/Core/com.equella.core/src/com/tle/core/mimetypes/MimeTypeServiceImpl.java
```diff
@@ -393,45 +386,6 @@ public void setListAttribute(MimeEntry entry, String key, @Nullable Collection<?
...
-    // });
-    //
-    // List<TextExtracterExtension> extracters = new
-    // ArrayList<TextExtracterExtension>();
-    // for( Extension e : extensions )
-    // {
-    // extracters.add(textExtracterTracker.getBeanByExtension(e));
-    // }
-    // return extracters;
-
```
edalex-ian, line 434: Thank you! :+1:

edalex-ian (changes requested):
Good stuff @PenghaiZhang , all feedback is just around doco and language string.

Source/Plugins/Core/com.equella.core/src/com/tle/core/mimetypes/MimeTypeService.java
PenghaiZhang: I think it's array.

Source/Plugins/Core/com.equella.core/src/com/tle/core/mimetypes/MimeTypeService.java
PenghaiZhang: The reason is `file` is added to the list by `MimeDefaultViewerSection, depending on some conditions IIRC. Also the MIME type REST endpoint does this, too.

Source/Plugins/Core/com.equella.core/src/com/tle/core/mimetypes/MimeTypeService.java
```diff
@@ -90,10 +90,11 @@
 
   /**
    * Return a list of enabled viewers. If the default viewer does not exist in the list, add it to
-   * the list. However, if the default viewer is "file", it will not be included.
+   * the list. However, if the default viewer is "file", it will not be included because "file" is
+   * added to the list dynamically by {@link com.tle.web.mimetypes.section.MimeDefaultViewerSection}
    *
    * @param attributes Configured attributes of MIME type
-   * @return A string representing a list of enabled viewers
+   * @return A string representing a list of enabled viewers in the format of array.
```
edalex-ian, line 97: That's still unclear really, best to either give an example or specify the standard format being used e.g.  "[..] formatted as a JSON array of strings".

edalex-ian (approved):
GTG @PenghaiZhang , but be good if you could just tweak that comment on format a bit more.

edalex-ian (approved):

Issue #3021, opened 2021-05-21
Oracle versions prior to 12c Release 2 cannot upgrade to/install openEQUELLA past 2020.2


SammyIsConfused:
**Describe the bug**
When upgrading to or installing openEQUELLA 2020.2 or above, with Oracle Database 12c Release 1 (12.1.0.2) or previous, migration will fail with a "Identifier too long" error. 

This is because prior to Oracle Database 12c Release 2 there was a character limit for identifiers of 30 characters, and an index brought in in 2020.2 migration "CreateFacetedSearchClassificationTable" was longer than this. The index in question was an institution_id index on FACETED_SEARCH_CLASSIFICATION table - facetSearchClassificationInstitutionIndex  which is longer than 30 characters.

**To Reproduce**

1. Install or upgrade openEQUELLA to 2020.2 or above, using an Oracle Database 12c Release 1 or earlier.
2. Open the server administration page and note that the server requires data migration. Run the migration.
3. Note it fails with this error message: 

![image](https://user-images.githubusercontent.com/24543345/119084385-07b8e680-ba45-11eb-9745-cb87f068919f.png)

**Expected behavior**
The index should not have such a long name, so that the migration will pass and openEQUELLA will be able to be used. 
Alternately, the minimum Oracle version for openEQUELLA should be increased to 12.2.0.1.

**Stacktrace**
```
Failed when doing migration database, got errors
Error running SQL: 'create index facetSearchClassificationInstitutionIndex on EQUELLA.faceted_search_classification (institution_id)'
Stacktrace:
ORA-00927 (Identifier too long) when doing database migration.
javax.persistence.PersistenceException: org.hibernate.exception.SQLGrammarException: could not execute statement
at org.hibernate.internal.ExceptionConverterImpl.convert(ExceptionConverterImpl.java:154)
at org.hibernate.internal.ExceptionConverterImpl.convert(ExceptionConverterImpl.java:181)
at org.hibernate.query.internal.AbstractProducedQuery.executeUpdate(AbstractProducedQuery.java:1621)
at com.tle.core.migration.AbstractHibernateMigration.executeSqlStatements(AbstractHibernateMigration.java:69)
at com.tle.core.migration.AbstractHibernateMigration$1.run(AbstractHibernateMigration.java:60)
at com.tle.core.migration.AbstractHibernateMigration.runInTransaction(AbstractHibernateMigration.java:86)
[...]

PR #3076, opened 2021-06-04
Reduce the length of the facetSearchClassificationInstitutionIndex name

SammyIsConfused:
##### Checklist

##### Description of change
#3021 

This index is longer than 30 characters, which causes problems for Oracle Database 12c Release 1 or earlier. 
When Oracle Database 12c Release 2 was released they increased the character limit to 128. 
This seems to be the only place this index is referred to.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/core/facetedsearch/bean/FacetedSearchClassification.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/core/facetedsearch/bean/FacetedSearchClassification.java
@@ -41,7 +41,7 @@ public class FacetedSearchClassification {
 
   @JoinColumn(nullable = false)
   @ManyToOne(fetch = FetchType.LAZY)
-  @Index(name = "facetSearchClassificationInstitutionIndex")
+  @Index(name = "FSCInstitutionIndex")
   @XStreamOmitField
   private Institution institution;
 


Source/Plugins/Core/com.equella.core/src/com/tle/core/facetedsearch/bean/FacetedSearchClassification.java
```diff
@@ -41,7 +41,7 @@
 
   @JoinColumn(nullable = false)
   @ManyToOne(fetch = FetchType.LAZY)
-  @Index(name = "facetSearchClassificationInstitutionIndex")
+  @Index(name = "FSCInstitutionIndex")
```
PenghaiZhang, line 44: I think If I am using 2020.2 and I have this long name index created, then I upgrade to 2021.2,  the index will not get recreated with this short name.

PenghaiZhang (approved):
LGTM Sam.

ChristianMurphy (approved):

edalex-ian (commented):
Heya @SammyIsConfused 

Could you please confirm if upgrades across Postgres, Oracle and MS SQL all work with this.

SammyIsConfused:
Just tested upgrading with this - upgrading from 2020.1 to the head of this branch works on:
- Oracle 12.1
- Oracle 12.2
- PostgreS 9.6
- MS SQL 2019 15.0.4123.1

It also works for 2020.2 to the head of this branch works too, for Oracle 12.2, PostgreS and MS SQL. I noted that as @PenghaiZhang said it doesn't rename the long index if it already exists as there is no migration, which is fine, it all still works as expected. Hibernate validation doesn't trip us up at all.

SammyIsConfused:
Still worth testing: 
- Upgrade to 2020.2 on Oracle 12.1 (trigger index too long error)
- Apply the .sql file fix
- Upgrade to this branch's head

edalex-ian:
So @SammyIsConfused , where are things at with this?

Source/Plugins/Core/com.equella.core/src/com/tle/core/facetedsearch/bean/FacetedSearchClassification.java
edalex-ian: Sam seemed to confirm this all still worked fine.

PR #3078, opened 2021-06-04
EBP for the Kaltura dialog

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

Some EBP work to make the Kaltura dialog look nicer.

image

image

image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
+++ b/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
@@ -4262,13 +4262,73 @@ a.modal-control.modal-save {
   }
 }
 
-.urlHandler {
+.urlHandler,
+.kalturaHandler {
   .modal-content .modal-content-background .modal-content-inner .float-left {
     @include editAttachmentDialogStyles;
   }
 
   .modal-content-inner {
     flex-direction: row;
+  }
+}
+
+.kalturaHandler {
+  @include buttonShadowBorder;
+
+  .choice.odd {
+    background-color: white;
+  }
+
+  .choice.even {
+    background-color: #f3f1eb;
+  }
+
+  img#kaltura-logo {
+    position: absolute;
+    bottom: 25px;
+    right: 35px;
+    zoom: 1;
+    filter: alpha(opacity=35);
+    opacity: 0.35;
+  }
+
+  .choice-list div {
+    position: relative;
+    padding: 8px 10px;
+  }
+
+  .choice-list div {
+    padding-bottom: 10px;
+  }
+
+  div.choice-list {
+    border: 1px solid #8f8b7e;
+  }
+
+  div.choice-list .choice {
+    cursor: pointer;
+  }
+
+  div.choice-list .choice:hover {
+    background-color: #ffffcc !important;
+  }
+
+  div.choice-list .choice.selected {
+    background-color: #80d2ee !important;
+  }
+
+  .choice h4 {
+    font-size: 12px;
+    padding-bottom: 4px;
+  }
+
+  #kaltura-query button {
+    background-color: $primaryColor;
+    color: set-text-color($primaryColor);
+    padding-left: $singleMargin;
+    padding-right: $singleMargin;
+    margin-left: $singleMargin;
   }
 }
 
--- a/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
+++ b/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
@@ -4288,18 +4288,12 @@ a.modal-control.modal-save {
     position: absolute;
     bottom: 25px;
     right: 35px;
-    zoom: 1;
-    filter: alpha(opacity=35);
     opacity: 0.35;
   }
 
   .choice-list div {
     position: relative;
-    padding: 8px 10px;
-  }
-
-  .choice-list div {
-    padding-bottom: 10px;
+    padding: $doubleMargin;
   }
 
   div.choice-list {
@@ -4311,11 +4305,11 @@ a.modal-control.modal-save {
   }
 
   div.choice-list .choice:hover {
-    background-color: #ffffcc !important;
+    background-color: #ffffcc;
   }
 
   div.choice-list .choice.selected {
-    background-color: #80d2ee !important;
+    background-color: #80d2ee;
   }
 
   .choice h4 {
--- a/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
+++ b/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
@@ -4314,7 +4314,7 @@ a.modal-control.modal-save {
 
   .choice h4 {
     font-size: 12px;
-    padding-bottom: 4px;
+    padding-bottom: $minimalMargin;
   }
 
   #kaltura-query button {


Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -4272,6 +4273,65 @@ a.modal-control.modal-save {
...
+  div.choice-list .choice {
+    cursor: pointer;
+  }
+
+  div.choice-list .choice:hover {
+    background-color: #ffffcc !important;
+  }
+
+  div.choice-list .choice.selected {
+    background-color: #80d2ee !important;
```
SammyIsConfused, line 4318: These !important flags - where are you overriding? Is it somewhere in legacy.scss? If so, we might be able to fix up the rule causing this. 
If its somewhere in equella, but not in legacy.scss, you should be able to `import thefileinquestion.css` at the top of legacy.scss. 
If its external, then yes this is probably the only way.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -4272,6 +4273,65 @@ a.modal-control.modal-save {
...
 
+.kalturaHandler {
+  @include buttonShadowBorder;
+
+  .choice.odd {
+    background-color: white;
+  }
+
+  .choice.even {
+    background-color: #f3f1eb;
```
SammyIsConfused, line 4284: Do these need to refer to the theme at all?

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -4272,6 +4273,65 @@ a.modal-control.modal-save {
...
+  }
+
+  img#kaltura-logo {
+    position: absolute;
+    bottom: 25px;
+    right: 35px;
+    zoom: 1;
+    filter: alpha(opacity=35);
+    opacity: 0.35;
```
SammyIsConfused, line 4293: These lines appear to do the same thing. Does one need removing?

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -4272,6 +4273,65 @@ a.modal-control.modal-save {
...
+  }
+
+  .choice-list div {
+    position: relative;
+    padding: 8px 10px;
```
SammyIsConfused, line 4298: Any time we refer to 8px, we can use $singleMargin.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -4272,6 +4273,65 @@ a.modal-control.modal-save {
...
+  }
+
+  .choice-list div {
+    padding-bottom: 10px;
```
SammyIsConfused, line 4302: Interesting that you need 10px, which is close to our standard $singleMargin. Does it look okay with 2px less?

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -4272,6 +4273,65 @@ a.modal-control.modal-save {
...
+  div.choice-list .choice:hover {
+    background-color: #ffffcc !important;
+  }
+
+  div.choice-list .choice.selected {
+    background-color: #80d2ee !important;
+  }
+
+  .choice h4 {
+    font-size: 12px;
```
SammyIsConfused, line 4322: There might be a mixin at the top of the file that achieves what we need here in the same way as native MUI, Worth checking, so we aren't adding text that is styled inconsistently with others.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -4272,6 +4273,65 @@ a.modal-control.modal-save {
...
+
+  .choice.even {
+    background-color: #f3f1eb;
+  }
+
+  img#kaltura-logo {
+    position: absolute;
+    bottom: 25px;
+    right: 35px;
+    zoom: 1;
```
SammyIsConfused, line 4291: isn't zoom: 1 the default? Is it needed?

SammyIsConfused (commented):
Thanks Penghai! Just a few suggestions :+1:

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
PenghaiZhang: Yeah looks like we can remove these `!important`.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
PenghaiZhang: Not quite sure.  Should this go with primary color or secondary color ?

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
PenghaiZhang: Yes, the `filter` is an ancient thing.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
PenghaiZhang: Good idea! And I think `doubleMargin` looks nicer!
image

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
PenghaiZhang: I just copied this from the Kaltura codebase.  I think this rule is redundant. I will remove it.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
PenghaiZhang: Looks like it is! I have removed it. Cheers.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
PenghaiZhang: Since this rule applies to a `h4`, I thought the mixin of `typography-headline-4` could be the one to use. But it turned out this mixin provides very different font styles. The only one that looks similar to what we need is `typography-body-1`, but I have to add `font-weight: bold;`.

You can see the font size is bigger.
image

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
PenghaiZhang: We will have different user experience in Old UI if we change this.

PenghaiZhang: According to the slack discussions, let's keep this as it is now.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
PenghaiZhang: According to the slack discussions, let's keep this as it is now

PenghaiZhang:
image

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -4272,6 +4273,59 @@ a.modal-control.modal-save {
...
+    background-color: #ffffcc;
+  }
+
+  div.choice-list .choice.selected {
+    background-color: #80d2ee;
+  }
+
+  .choice h4 {
+    font-size: 12px;
+    padding-bottom: 4px;
```
SammyIsConfused, line 4317:
```suggestion
    padding-bottom: $minimalMargin;
```

SammyIsConfused (changes requested):
Thanks Penghai! Just one last tweak to do and then you're good to go.

SammyIsConfused:
Also of note - remember in future to write your commit messages such that they fit in the sentence:
"If applied, this commit will {blank}."
Such as
Improve styles for Kaltura wizard control dialog.

For more info see https://chris.beams.io/posts/git-commit/ (this is also in our CONTRIBUTING.md)

SammyIsConfused (approved):
Looks good to me! Thanks for fixing those up @PenghaiZhang . Good to go.

edalex-ian (approved):
LGTM. :+1:

PR #3091, opened 2021-06-10
Feature/navigate attachments in gallery

PenghaiZhang:
##### Checklist

- [x] tests are included

##### Description of change

This PR adds the support for viewing previous/next attachments with Lightbox in Gallery mode. The function used to build the handler for `onNext` and `onPrevious` was focused on `Attachment` so it did not support `GalleryEntry` very well. So I refactored this function to support both.

During implementation, Sam found a bug where if you are on the first/last image and you press the left/right arrow button you will see a blank Lightbox. The fix is also added in this PR.

https://user-images.githubusercontent.com/47203811/121481993-052c2a00-ca10-11eb-82f1-182278a90282.mp4

### DIFF
--- a/react-front-end/tsrc/components/Lightbox.tsx
+++ b/react-front-end/tsrc/components/Lightbox.tsx
@@ -33,13 +33,7 @@ import OpenInNewIcon from "@material-ui/icons/OpenInNew";
 import { pipe } from "fp-ts/function";
 import * as O from "fp-ts/Option";
 import * as React from "react";
-import {
-  ReactNode,
-  SyntheticEvent,
-  useCallback,
-  useEffect,
-  useState,
-} from "react";
+import { ReactNode, SyntheticEvent, useEffect, useState } from "react";
 import { Literal, match, Unknown } from "runtypes";
 import {
   CustomMimeTypes,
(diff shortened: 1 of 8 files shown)


react-front-end/tsrc/components/Lightbox.tsx
```diff
@@ -140,22 +134,17 @@ const Lightbox = ({ open, onClose, config }: LightboxProps) => {
...
   const { src, title, mimeType, onPrevious, onNext } = lightBoxConfig;
 
-  const handleOnPrevious = useCallback(() => {
-    setContent(undefined);
-    onPrevious && setLightBoxConfig(onPrevious());
-  }, [onPrevious]);
-
-  const handleOnNext = useCallback(() => {
+  const handleNav = (getLightboxConfig: () => LightboxConfig) => {
     setContent(undefined);
```
PenghaiZhang, line 138: Here is the cause for the bug. We set `content` to undefined before we get the new lightbox config. The fix is that we do nothing when the handlers (onNext / onPrevious) provided are undefined.

react-front-end/tsrc/modules/ViewerModule.ts
```diff
@@ -187,39 +201,50 @@ export const getViewerDefinitionForAttachment = (
 };
 
 /**
- * Build a function to handler navigation between Lightbox attachments.
- * @param lightboxAttachments All attachments that can be viewed in Lightbox and their Viewer definitions.
- * @param attachmentIndex Index of the attachment to be viewed
+ * Build a function to handler navigation between Lightbox entries.
```
edalex-ian, line 204:
```suggestion
 * Build a function to handle navigation between Lightbox entries.
```

react-front-end/tsrc/search/components/GallerySearchResult.tsx
```diff
@@ -110,16 +126,31 @@ const GallerySearchResult = ({ items }: GallerySearchResultProps) => {
...
+                true
+              ),
+              onPrevious: buildLightboxNavigationHandler(
+                lightboxEntries,
+                initialLightboxEntryIndex - 1,
+                true
+              ),
             },
           });
+        };
```
edalex-ian, line 153: I think this change - maybe the one before - has pushed `mapItemsToTiles()` over the edge of readability. I think perhaps some of the nested functions should be pushed out - such as `buildOnClickHandler()` - and we probably also probably want to minimise use of external variables to help - such as using `lightboxEntries`.

Refactor needed.

edalex-ian (changes requested):
Looks good @PenghaiZhang , but I think we need one refactor before proceeding.

edalex-ian (approved):
Good stuff @PenghaiZhang . Merge on :+1:

PR #3101, opened 2021-06-15
Update @tinymce/tinymce-react to 3.12.6

SammyIsConfused:
##### Checklist

- [x] tests are included (tests already existed and didn't need to change)

##### Description of change
This came from #2435  - since I made a code change I'm making a new PR.
This branch failed because the Editor became completely non-interactible, typing would do nothing. 
This is because the `htmlInput` prop we were using plugs into the Editor's value parameter. In previous versions, 
this would work as an _initial_ value - every edit would run the `onStateChange` callback but `htmlInput` didn't need to change. 

Now, this works as a standard controlled component and `htmlInput` _is_ the value - meaning if it doesn't update, TinyMCE won't either. We were already holding state for this in this.state.current.notice, for `preventNavigation` checks - it just needs feeding back in.

As `this.state.current.notice` is initialized at the same time as `this.state.db.notice`, we don't need to change any of the initializing code. 

### DIFF
--- a/react-front-end/package-lock.json
+++ b/react-front-end/package-lock.json
@@ -3535,7 +3535,8 @@
           "dev": true,
           "optional": true,
           "requires": {
-            "bindings": "^1.5.0"
+            "bindings": "^1.5.0",
+            "nan": "^2.12.1"
           }
         },
         "glob-parent": {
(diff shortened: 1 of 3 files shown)


edalex-ian (approved):

PenghaiZhang (approved):
LGTM.

PR #3132, opened 2021-06-23
feat: support copying resource embed code from Lightbox

PenghaiZhang:
##### Checklist

- [x] tests are included

##### Description of change

This PR adds the support for copying resource embed code from Lightbox. I also reordered language strings in this PR.

image

image

image


react-front-end/tsrc/components/Lightbox.tsx
```diff
@@ -119,19 +118,28 @@ const {
...
-    close: labelClose,
-    openInNewWindow: labelOpenInNewWindow,
-  } = languageStrings.common.action;
-  const {
-    unsupportedContent: labelUnsupportedContent,
-    youTubeVideoMissingId,
-  } = languageStrings.lightboxComponent;
 
-  const [content, setContent] = useState<ReactNode | undefined>();
+  const [content, setContent] = useState<ReactElement | undefined>();
```
PenghaiZhang, line 138: Since type of the content we build is `JSX.Element`,  `ReactNode`  is too broad. So I changed it to `ReactElement`.

react-front-end/tsrc/components/Lightbox.tsx
```diff
@@ -250,6 +250,19 @@ const Lightbox = ({ open, onClose, config }: LightboxProps) => {
     onClose();
   };
 
+  // Generate a HTML string from current Lightbox content.
+  // Also use DOMParser to help remove unneeded attributes such as 'class'.
+  const generateEmbedCode = (content: ReactElement): string => {
+    const unneededAttributes = ["class"];
```
PenghaiZhang, line 256: I don't think we want to include `class` as part of the embed code. But if we want, this function can be simplified to 
```
 return renderToStaticMarkup(content);
```

react-front-end/tsrc/search/components/EmbedCodeDialog.tsx
```diff
@@ -0,0 +1,72 @@
...
+ * Provide a Dialog which allows users to copy embed code of a Lightbox viewable resource(e.g. image && video).
+ */
+export const EmbedCodeDialog = ({
+  open,
+  closeDialog,
+  embedCode,
+}: EmbedCodeDialogProps) => (
+  <Dialog open={open} fullWidth>
+    <DialogTitle>{embedCodeLabel}</DialogTitle>
+    <DialogContent>{embedCode}</DialogContent>
```
ChristianMurphy, line 58: Would it make sense from an HTML semantics perspective to wrap this in a `<code>` tag?
```suggestion
    <DialogContent><code>{embedCode}</code></DialogContent>
```

react-front-end/tsrc/components/Lightbox.tsx
```diff
@@ -250,6 +250,19 @@ const Lightbox = ({ open, onClose, config }: LightboxProps) => {
...
+    const doc = domParser.parseFromString(
+      renderToStaticMarkup(content),
+      "text/xml"
+    );
```
ChristianMurphy, line 260: I'm cautious about including `renderToStaticMarkup` which is part of `react-dom/server` on the clientside.
Would it be possible to render the component using something like
```jsx
<div ref={staticRenderRef} style={{display: 'none'}}>{content}</div>
```
 and get the inner HTML of the `staticRenderRef`?

react-front-end/tsrc/util/langstrings.ts
```diff
@@ -191,13 +256,64 @@ export const languageStrings = {
...
+        title: "Cloud provider disclaimer",
+      },
+    },
+    deletecloudprovider: {
+      title: "Are you sure you want to delete cloud provider - '%s'?",
+      message: "It will be permanently deleted.",
+    },
+    refreshed: "Completed refresh",
+  },
+  dateRangeSelector: {
```
ChristianMurphy, line 283: it looks like there is a mix of camelcase and all lowercase, would it make sense to use just one or the other?
Personally I find the camelcase names a bit easier to read, but either could work.

ChristianMurphy (commented):
Thanks @PenghaiZhang!
Just a few questions and considerations.

react-front-end/tsrc/util/langstrings.ts
edalex-ian: Yup, we're slowly moving towards all being camelCase - so at least all new ones are. I think what we're seeing here is @PenghaiZhang has gone through and sorted these more (thanks @PenghaiZhang ) and as a result highlighting some of the old 'all lowercase' varieties.

react-front-end/tsrc/search/components/EmbedCodeDialog.tsx
PenghaiZhang: Yeah this makes sense! Thank you!

react-front-end/tsrc/components/Lightbox.tsx
PenghaiZhang: I have tried this approach and got an error. The result of `innerHTML` does not have a closing tag for `img`. For example, it gives me `` and then the DOM parser makes errors. 
`

PenghaiZhang: If I change to HTML parser, I don't have errors but the result is wrapped in a bare-bones `<html><head></head><body></body></html>`.

PenghaiZhang: Do you have any suggestion @ChristianMurphy ?

ChristianMurphy: > I don't have errors but the result is wrapped in a bare-bones `<html><head></head><body></body></html>`

Right, dom parser will try to generate a full HTML document, the fragment can be accessed from `doc.body.childNodes[0]`

PenghaiZhang: Thanks! I have pushed some changes.

react-front-end/tsrc/components/Lightbox.tsx
ChristianMurphy: Probably fine to filter it out.
Down the road, if the list grows, I could potentially see inverting this.
Listing the attributes to keep, rather than those to remove.
But with just a single attr at the moment it seems fine.

react-front-end/tsrc/search/components/EmbedCodeDialog.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+export interface EmbedCodeDialogProps {
+  /**
+   * `true` to open the dialog.
+   */
+  open: boolean;
+  /**
+   * Fired when the dialog is closed.
+   */
+  closeDialog: () => void;
```
ChristianMurphy, line 36: Potentially `onCloseDialog` to clarity its intent as an event listener, rather than an action to close.

ChristianMurphy (approved):

react-front-end/tsrc/search/components/EmbedCodeDialog.tsx
```diff
@@ -0,0 +1,74 @@
...
+  embedCode: string;
+}
+
+const {
+  label: embedCodeLabel,
+  copy: copyEmbedCode,
+} = languageStrings.embedCode;
+
+/**
+ * Provide a Dialog which allows users to copy embed code of a Lightbox viewable resource(e.g. image && video).
```
edalex-ian, line 49: There is nothing in here specific to Lightbox. I think this should be moved to `tsrc/components` and the reference here to `Lightbox` removed. I can see in the future this being used in other contexts.

react-front-end/tsrc/components/Lightbox.tsx
```diff
@@ -250,6 +257,24 @@ const Lightbox = ({ open, onClose, config }: LightboxProps) => {
     onClose();
   };
 
+  // Generate a HTML string from current Lightbox content.
```
edalex-ian, line 260: Interesting idea @PenghaiZhang . I like it. I wonder if we'll get tripped up at some point (like if we need to add SSO tokens or the like), but for now it's great. :+1:

edalex-ian (changes requested):
Looks good @PenghaiZhang . Just need to relocate the new dialog to keep it nice and generic. No need to tie it down into Search. ;)

After that, should be GTG.

edalex-ian (approved):
Looks good @PenghaiZhang , and thanks for the additional cosmetic tweaks.

PR #3142, opened 2021-06-25
feat: preview Youtube videos on standard list mode

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

Changes of this PR enable previewing of YouTube videos from the Search results attachment list in the same way that local videos can currently be previewed from the same place (Lightbox). 

The approach is to update the MIME type of Youtube attachment and what viewer to be used.

https://user-images.githubusercontent.com/47203811/123376446-d5635180-d5cd-11eb-9c22-845eeaaa74cd.mp4


react-front-end/tsrc/search/components/SearchResultAttachmentsList.tsx
```diff
@@ -142,11 +146,14 @@ export const SearchResultAttachmentsList = ({
...
-    const getViewerID = async (broken: boolean, mimeType: string) => {
-      let viewerDetails: OEQ.MimeType.MimeTypeViewerDetail | undefined;
-      if (broken) {
-        return undefined;
+    const getViewerID = async (
+      attachmentType: string,
+      mimeType: string
+    ): Promise<OEQ.MimeType.ViewerId | undefined> => {
+      if (attachmentType === ATYPE_YOUTUBE) {
+        return "htmlFiveViewer";
```
PenghaiZhang, line 154: Not quite sure if we should `htmlFiveViewer` or `file` here, but `htmlFiveViewer` makes more sense to me.

react-front-end/tsrc/search/components/SearchResultAttachmentsList.tsx
```diff
@@ -161,11 +168,15 @@ export const SearchResultAttachmentsList = ({
...
-            const viewerId = mimeType
-              ? await getViewerID(brokenAttachment, mimeType)
-              : undefined;
+          async (originalAttachment) => {
+            const attachment = updateAttachmentForCustomInfo(
+              originalAttachment
+            );
+            const { attachmentType, mimeType, brokenAttachment } = attachment;
+            const viewerId =
+              mimeType && !brokenAttachment
```
PenghaiZhang, line 177: No need to pass `brokenAttachment` to `getViewerID` - just return `undefined` for broken attachments.

react-front-end/tsrc/modules/YouTubeModule.ts
```diff
@@ -57,3 +60,23 @@ export const buildViewUrl = (videoId: string): string =>
...
+  const { externalId } = links;
+  if (externalId) {
+    return {
+      ...attachment,
+      mimeType: CustomMimeTypes.YOUTUBE,
+      links: { ...links, view: buildViewUrl(externalId) },
+    };
+  }
+
+  throw new Error("Missing Youtube Video Attachment external ID.");
```
edalex-ian, line 81:
```suggestion
  throw new Error("Missing YouTube Video Attachment external ID.");
```

edalex-ian (commented):
Hmmm, I've made a few comments here on trivial things, but then I realised we have some rather fundamental issues. Although this change does the job, I fear it's built on top of a foundation that is going to be costly to maintain.

A few areas that drew my attention:

1. We have a rather meandering useEffect of over 100 lines;
2. Inside that affect we have an async IIFE which is over 70 lines long - and I resorted to an IDE to determine where it starts/ends and is called;
3. The function `getViewerId` should probably be over in the `ViewerModule` and just take an `Attachment`;
4. And then the question comes in is there a better way to transform the attachments to be fit for purpose in building the attachment and viewer definitions rather; and
5. How does this overlap with what's in the gallery (and referencing back to the above point, can we have a clearer pipeline like we have there).

I think it'd be worthwhile to discuss and figure out what direction to take.

react-front-end/tsrc/modules/ViewerModule.ts
```diff
@@ -114,7 +119,7 @@ export const determineViewer = (
   if (broken) {
     return simpleLinkView;
   }
-  if (attachmentType !== "file" && attachmentType !== "custom/resource") {
+  if (![ATYPE_FILE, ATYPE_RESOURCE, ATYPE_YOUTUBE].includes(attachmentType)) {
```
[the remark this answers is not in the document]
PenghaiZhang: I am cool with this, but Intellij keeps suggesting me that `[...].includes(x)===false` can be simplified to what I have done.

react-front-end/tsrc/modules/ViewerModule.ts
```diff
@@ -248,3 +263,184 @@ export const buildLightboxNavigationHandler = (
...
+ * will be caught and converted to a rejected promise.
+ *
+ * @param getViewerDetails Function called to retrieve Viewer ID.
+ */
+const getViewerId = async (
+  getViewerDetails: () => Promise<OEQ.MimeType.MimeTypeViewerDetail>
+): Promise<E.Either<string, OEQ.MimeType.ViewerId>> => {
+  // Since the task passed into 'TE.tryCatch' should never throw an error and 'getViewerDetail' could throw an error,
+  // we use a try-catch block to ensure a rejected promise is returned when an error occurs.
+  const task = () => {
```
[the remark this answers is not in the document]
PenghaiZhang: Hi @ChristianMurphy , do you mean we refactor this to 
```
  const task = async () => {
    try {
      return await getViewerDetails();
    } catch (error) {
      return (
        `${languageStrings.searchpage.searchResult.errors.getAttachmentViewerDetailsFailure}: ${error.message}`
      )
    }
  };

```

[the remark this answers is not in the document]
PenghaiZhang: Oh I see. I should throw an error with the string rather than `return`.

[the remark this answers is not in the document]
ChristianMurphy: right, something like

```ts
  const task = async () => {
    return getViewerDetails();
  };
```

or

```ts
  const task = async () => {
    try {
      return getViewerDetails();
    } catch (error) {
      throw new Error(`${languageStrings.searchpage.searchResult.errors.getAttachmentViewerDetailsFailure}: ${error.message}`);
    }
  };
```

ChristianMurphy (approved):

react-front-end/tsrc/modules/ViewerModule.ts
```diff
@@ -248,3 +263,184 @@ export const buildLightboxNavigationHandler = (
...
+              attachment,
+              viewerId?.right
+            ),
+          });
+    })
+  );
+  const lefts: string[] = A.lefts(either);
+  const rights: AttachmentAndViewerDefinition[] = A.rights(either);
+
+  return A.isEmpty(lefts) ? E.right(rights) : E.left(lefts.join());
```
edalex-ian, line 340: As per the comment on the return type, there'll be no need to do the `join`...

react-front-end/tsrc/modules/ViewerModule.ts
```diff
@@ -248,3 +263,184 @@ export const buildLightboxNavigationHandler = (
...
+
+/**
+ * Returns a list of provided attachments and their viewer configurations.
+ *
+ * @param attachments Attachments provided to find their viewer configurations.
+ * @param uuid UUID of the Item which the attachments belong to.
+ * @param version Version of the Item which the attachments belong to.
+ * @param getViewerDetails Function called to retrieve viewer detail for each attachment.
+ */
+export const buildViewerConfigForAttachment = async (
```
edalex-ian, line 418: Shouldn't this be plural?

```suggestion
export const buildViewerConfigForAttachments = async (
```

react-front-end/tsrc/modules/ViewerModule.ts
```diff
@@ -248,3 +263,184 @@ export const buildLightboxNavigationHandler = (
...
+  attachments: OEQ.Search.Attachment[],
+  uuid: string,
+  version: number,
+  getViewerDetails: (
+    mimeType: string
+  ) => Promise<OEQ.MimeType.MimeTypeViewerDetail>
+): Promise<AttachmentAndViewerConfig[]> => {
+  const attachmentsAndViewerDefinitions = await buildAttachmentsAndViewerDefinitions(
```
edalex-ian, line 426: For readability, could I please request that you explicitly type this? I.e.:

```suggestion
  const attachmentsAndViewerDefinitions: Either<string, AttachmentAndViewerDefinition> = await buildAttachmentsAndViewerDefinitions(
```

react-front-end/tsrc/modules/YouTubeModule.ts
```diff
@@ -57,3 +60,23 @@ export const buildViewUrl = (videoId: string): string =>
  */
 export const extractVideoId = (viewUrl: string): string | null =>
   new URL(viewUrl).searchParams.get("v");
+
+/**
+ * Update YouTube Video Attachment with Custom MIME type and View URL.
+ * @param attachment An Youtube Video Attachment
```
edalex-ian, line 66:
```suggestion
 *
 * @param attachment An Youtube Video Attachment
```

edalex-ian (changes requested):
Great work on tidying up the Uber Effect.

Now as discussed, we just want to tidy-up that pipeline.

react-front-end/tsrc/modules/ViewerModule.ts
```diff
@@ -248,3 +263,184 @@ export const buildLightboxNavigationHandler = (
...
+            attachment,
+            viewerDefinition: getViewerDefinitionForAttachment(
+              uuid,
+              version,
+              attachment,
+              viewerId?.right
+            ),
+          });
+    })
+  );
```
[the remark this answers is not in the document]
PenghaiZhang: I also modified `AttachmentAndViewerConfig`

PenghaiZhang:
Ah dear I did not install the new version of prettier and then lots of format issues come up. I will fix this now.

react-front-end/tsrc/modules/ViewerModule.ts
```diff
@@ -248,3 +248,190 @@ export const buildLightboxNavigationHandler = (
...
+  attachments: OEQ.Search.Attachment[],
+  uuid: string,
+  version: number,
+  getViewerDetails: (
+    mimeType: string
+  ) => Promise<OEQ.MimeType.MimeTypeViewerDetail>
+): Promise<E.Either<string[], AttachmentAndViewerDefinition[]>> => {
+  const either: E.Either<string, AttachmentAndViewerDefinition>[] =
+    await Promise.all(
+      attachments.map(async (originalAttachment) => {
```
edalex-ian, line 272: Did you consider doing `pipe(attachments, A.map ... )`? or perhaps to accommodate broken attachments `pipe(attachments, A.partition ... )`?

Doing so you might have been able to minimise the hybrid approach and possibly skip the need for any `return` statements - vs. the two here.

react-front-end/tsrc/modules/ViewerModule.ts
```diff
@@ -248,3 +248,190 @@ export const buildLightboxNavigationHandler = (
...
+ * Build Lightbox entries based on a list of attachments and viewer definitions
+ *
+ * @param attachmentsAndViewerDefinitions A list of attachments and their viewer definitions.
+ */
+export const buildLightboxEntries = (
+  attachmentsAndViewerDefinitions: AttachmentAndViewerDefinition[]
+): LightboxEntry[] =>
+  pipe(
+    attachmentsAndViewerDefinitions,
+    A.map(({ attachment: { id, description, mimeType }, viewerDefinition }) =>
```
edalex-ian, line 327: Might be possible here to utilise `flow`:

```suggestion
export const buildLightboxEntries = (
  attachmentsAndViewerDefinitions: AttachmentAndViewerDefinition[]
): LightboxEntry[] =
  flow(
    A.map(({ attachment: { id, description, mimeType }, viewerDefinition }) =>
```

react-front-end/tsrc/modules/ViewerModule.ts
```diff
@@ -248,3 +248,190 @@ export const buildLightboxNavigationHandler = (
...
+      })
+    );
+
+  const lefts = A.lefts(either);
+  const rights = A.rights(either);
+  return A.isEmpty(lefts) ? E.right(rights) : E.left(lefts);
+};
+
+/**
+ * Build Lightbox entries based on a list of attachments and viewer definitions
```
edalex-ian, line 318: Might be value in mentioning that it drops any non-lightbox entries.

react-front-end/tsrc/modules/ViewerModule.ts
```diff
@@ -248,3 +248,190 @@ export const buildLightboxNavigationHandler = (
...
+                    },
+                  },
+                }
+              : {
+                  attachment,
+                  viewerConfig: {
+                    viewerType: viewer,
+                    url: viewUrl,
+                  },
+                }
```
edalex-ian, line 392: This section has some readability issues. Would it be possible to create a builder function for each of the two object types? Then maybe you'd have something like:

```typescript
viewer === "lightbox"
  ? buildLightboxViewerConfig(attachment, viewer, viewUrl, lightboxEntries, initialLightboxEntryIndex)
  : buildBasicViewerConfig(attachment, viewer, viewUrl)

**Review by edalex-ian — CHANGES_REQUESTED** [review]
Fair effort here @PenghaiZhang ! Thank you! :+1: :pray: :star: 

I do have some final comments for some last tweaks. And after that it should be fine to merge.

File: react-front-end/tsrc/components/ItemAttachmentLink.tsx [hunk 15]
```diff
@@ -118,6 +113,15 @@ const ItemAttachmentLink = ({
...
 
+  // Only broken attachments do not have viewer configuration so return a 'Typography'.
+  if (!viewerConfig) {
+    return (
+      <Typography aria-label={`${attachmentLink} ${description}`}>
+        {description}
+      </Typography>
+    );
+  }
+
```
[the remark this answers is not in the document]
PenghaiZhang: We also added one more function `pfTernaryTypeGuard`  to fit where the ternary is used as a type guard .

react-front-end/tsrc/modules/ViewerModule.ts
PenghaiZhang: Can do this. But the type of viewer will be `lightbox` and `link` in each function rather than `Viewer`.

react-front-end/__tests__/tsrc/modules/ViewerModule.test.ts
```diff
@@ -17,9 +17,62 @@
...
+    id: "44528005-fb39-4461-bac7-12cd33ce4330",
+    description: "Australian Kelpie Pet Profile | Bondi Vet",
+    preview: false,
+    mimeType: "image/gif",
+    brokenAttachment: false,
+    links: {
+      view: "http://localhost:8080/ian/items/40e879db-393b-4256-bfe2-9a78771d6937/1/?attachment.uuid=44528005-fb39-4461-bac7-12cd33ce4330",
+      thumbnail:
+        "http://localhost:8080/ian/thumbs/40e879db-393b-4256-bfe2-9a78771d6937/1/44528005-fb39-4461-bac7-12cd33ce4330",
+    },
```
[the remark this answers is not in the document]
PenghaiZhang: As discusses, they where copied from some other mock data.

react-front-end/tsrc/modules/ViewerModule.ts
```diff
@@ -248,3 +248,190 @@ export const buildLightboxNavigationHandler = (
...
+              viewerId
+            ),
+          }))
+        )();
+      })
+    );
+
+  const lefts = A.lefts(either);
+  const rights = A.rights(either);
+  return A.isEmpty(lefts) ? E.right(rights) : E.left(lefts);
```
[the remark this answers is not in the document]
PenghaiZhang: How about this ? 

```
pipe(either, A.separate, ({ left, right }) =>
    pipe(
      left,
      A.isEmpty,
      B.match(
        () => E.left(left),
        () => E.right(right)
      )
    )
  )
```

react-front-end/tsrc/modules/ViewerModule.ts
PenghaiZhang: Refactored by `partitionMapWithIndex`.

react-front-end/tsrc/util/pointfree.ts
```diff
@@ -0,0 +1,63 @@
...
+  ) =>
+  (a: A): B =>
+    pipe(a, E.fromPredicate(predicate, identity), E.match(onLeft, onRight));
+
+/**
+ * Similar to {@link pfTernary}, but the predicate is a type guard.
+ */
+export const pfTernaryTypeGuard =
+  <A, B, C>(
+    predicate: (a: A | B) => a is A,
```
edalex-ian, line 54:
```suggestion
    guard: (a: A | B) => a is A,
```

react-front-end/tsrc/modules/ViewerModule.ts
```diff
@@ -248,3 +248,243 @@ export const buildLightboxNavigationHandler = (
...
+      O.map(([_, viewerUrl]) => ({
+        src: viewerUrl,
+        title: description,
+        mimeType: mimeType ?? "",
+        id,
+      }))
+    )
+  ),
+  A.filter(O.isSome),
+  A.map((entry) => entry.value)
```
edalex-ian, line 377: I thought I wrote this comment already, but I can't find it. So if you get this twice...

Would this be more idiomatic as:

```suggestion
  A.map(O.getOrElse(absurd))
```

Requires:

```typescript
import {absurd} from "fp-ts/function";
```

edalex-ian (approved):
Just a few minor considerations, then go for merge when happy.

A lot of efforts here @PenghaiZhang . Thanks for sticking with it! :+1:

edalex-ian (approved):

Issue #2934, opened 2021-04-29
New Video Gallery: YouTube preview pane in video lightbox defaults and reverts to small - have to refresh to fix


cathfitz:
**Describe the bug**
Currently, when you first go into Video mode and click on a YouTube video, the preview pane is very small. Doing a Ctrl/F5 makes it revert to the normal size. But if you do something in between (like open a Summary page) it reverts back to the small size.

**To Reproduce**

1. Go to Search page
2. Go to video mode
3. Click on a YouTube video
4. Note that it's small
5. Close the preview pane
6. Refresh (Ctrl/F5)
7. Reopen the preview pane for the YouTube video - note it's the proper size.
8. Close preview pane then open the Resource summary page for a video. 
9. Use the back button or click the Search menu
10. Preview another YouTube video - it's reverted back to being small again.

**Expected behavior**
The video preview should always show in the larger box

**Screenshots**
Small
![image](https://user-images.githubusercontent.com/32081370/116506656-0b06f980-a901-11eb-8070-b35388912194.png)
Correct size
![image](https://user-images.githubusercontent.com/32081370/116506680-14906180-a901-11eb-8a9e-80874b49be51.png)

**Stacktrace**
If applicable, a stacktrace to help explain your problem.

**Platform:**

- OpenEquella Version: [e.g. 2021.1]
- OS: [e.g. Windows 10, MacOS 10.13]
- Browser [e.g. Chrome, Firefox]

**Additional context**

snehanaik22:
Tested.

snehanaik22:
closed by mistake.

snehanaik22:
Tested.

PR #3147, opened 2021-06-30
Exclude YouTube preview player from legacy.scss Reset rule

SammyIsConfused:
Make sure the reset rule in legacy.scss doesn't apply to the youtube preview iframe,
as width:auto and height:auto cause the preview to render incorrectly.

##### Checklist

##### Description of change
#2934 
The small YouTube preview panel was caused by the "Reset rule" at the top of legacy.scss. This reset rule is designed to unset any rules that apply to common tags and replace them with an openEQUELLA default value.
```css
html,
body,
div,
span,
applet,
iframe,
h1,
h2,
h3,
h4,
h5,
h6,
p,
blockquote,
pre,
a,
abbr,
acronym,
address,
big,
cite,
code,
del,
dfn,
em,
font,
ins,
kbd,
q,
s,
samp,
small,
strike,
strong,
sub,
sup,
tt,
var,
b,
u,
i,
center,
dl,
dt,
dd,
ol,
ul,
li,
fieldset,
form,
label,
legend,
input,
textarea,
table,
caption,
tbody,
tfoot,
thead,
tr,
th,
td,
select {
  width: auto;
  height: auto;
  margin: 0;
  padding: 0;
  border: 0;
  outline: 0;
  font-size: 100%;
  vertical-align: baseline;
  background: transparent;
  font-family: "Roboto";
}
```
 
However - this applied to iframe, causing the YouTube preview player to be styled incorrectly.

Whenever legacy.css is present in the browser at the time of a YouTube preview, it will render incorrectly like this:
image

To avoid this, I have added a className to the YoutubeEmbed iframe `skip-css-reset`, and then I check for this in the CSS Reset rule.
```css
...
iframe:not(.skip-css-reset),
...
```

This fixes the issue. 

image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
+++ b/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
@@ -334,7 +334,6 @@ body,
 div,
 span,
 applet,
-iframe:not[title="YouTube video player"],
 h1,
 h2,
 h3,
@@ -403,6 +402,19 @@ select {
   background: transparent;
   font-family: "Roboto";
 }
+iframe:not[title="YouTube video player"]{
+  width: auto;
+  height: auto;
+  margin: 0;
+  padding: 0;
+  border: 0;
+  outline: 0;
+  font-size: 100%;
+  vertical-align: baseline;
+  background: transparent;
+  font-family: "Roboto";
+}
+
 ul {
   list-style: none;
 }
--- a/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
+++ b/Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
@@ -328,7 +328,7 @@ Animations
   }
 }
 
-//need to explicitly exclude youtube preview iframe
+// need to explicitly exclude youtube preview iframe
 html,
 body,
 div,
@@ -354,6 +354,7 @@ del,
 dfn,
 em,
 font,
+iframe:not(.skip-css-reset),
 ins,
 kbd,
 q,
@@ -391,18 +392,6 @@ tr,
 th,
 td,
 select {
-  width: auto;
-  height: auto;
-  margin: 0;
-  padding: 0;
-  border: 0;
-  outline: 0;
-  font-size: 100%;
-  vertical-align: baseline;
-  background: transparent;
-  font-family: "Roboto";
-}
-iframe:not[title="YouTube video player"] {
   width: auto;
   height: auto;
   margin: 0;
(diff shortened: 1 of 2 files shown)


Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -328,12 +328,12 @@ Animations
   }
 }
 
+//need to explicitly exclude youtube preview iframe
```
ChristianMurphy, line 331:
```suggestion
// need to explicitly exclude youtube preview iframe
```

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -402,6 +402,19 @@ select {
...
+  width: auto;
+  height: auto;
+  margin: 0;
+  padding: 0;
+  border: 0;
+  outline: 0;
+  font-size: 100%;
+  vertical-align: baseline;
+  background: transparent;
+  font-family: "Roboto";
```
ChristianMurphy, line 415: :thinking: unless I'm missing something this matches the rules applied on 394-403?
Would it make sense to add the selector to that group?

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
```diff
@@ -402,6 +402,19 @@ select {
   background: transparent;
   font-family: "Roboto";
 }
+iframe:not[title="YouTube video player"] {
```
ChristianMurphy, line 405: Hmmm, this works, but feels potentially brittle. (depending on how often language strings change)
Would it make sense to add a css class to `iframe`, something like `.skip-css-reset`? Which would be more stable.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
SammyIsConfused: Yup, it was having problems allowing a :not([title='Youtube video player"]) within a list of selectors. It works fine with :not(.skip-css-reset). Simplified to account for that.

Source/Plugins/Core/com.equella.core/resources/web/sass/legacy.scss
SammyIsConfused: Great call - this way if YouTube restyle their video player this will still work. Added.

SammyIsConfused:
Alright cool, that's fixed up and ready for re-review :+1:

edalex-ian (approved):

ChristianMurphy (approved):

PR #3157, opened 2021-07-01
feat: save wild card mode in local storage

PenghaiZhang:
##### Checklist

- [x] tests are included

##### Description of change

This PR adds the feature of saving wild card mode in browser local storage.

Screenshot from 2021-07-01 13-13-14


react-front-end/tsrc/util/BrowserStorageUtil.ts
```diff
@@ -0,0 +1,37 @@
...
+  return value ? JSON.parse(value) : undefined;
+};
+
+/**
+ * Save data to browser local storage.
+ * @param key Key of the data.
+ * @param value Value of the data.
+ */
+export const setDataToLocalStorage = <T>(key: string, value: T) =>
+  localStorage.setItem(key, JSON.stringify(value));
```
SammyIsConfused, line 37: JSON.stringify can throw a TypeError on a couple circumstances - 
    Throws a TypeError ("cyclic object value") exception when a circular reference is found.
    Throws a TypeError ("BigInt value can't be serialized in JSON") when trying to stringify a BigInt value.
Since value isn't constrained to type, this function could cause that error to be thrown if value was a BigInt. 
Note - I'm not necessarily saying we need to handle that, just that it could potentially throw an error here.

react-front-end/tsrc/util/BrowserStorageUtil.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Retrieve data from browser local storage by a key. Return undefined if
+ * no data is found.
+ *
+ * @param key Key of the data to be retrieved.
+ */
+export const getDataFromLocalStorage = <T>(key: string): T | undefined => {
+  const value = localStorage.getItem(key);
+  return value ? JSON.parse(value) : undefined;
```
SammyIsConfused, line 28: JSON.parse can throw a SyntaxError if the value is not in an expected format.

SammyIsConfused (commented):
Looks good Penghai!
Could be worth adding this to some tests, and I'm unsure whether or not we need to handle potential errors thrown by the JSON functions.

PenghaiZhang:
> Looks good Penghai!
> Could be worth adding this to some tests, and I'm unsure whether or not we need to handle potential errors thrown by the JSON functions.

Thanks Sam! I have added tests.

react-front-end/__tests__/tsrc/search/SearchPage.test.tsx
```diff
@@ -1139,3 +1140,31 @@ describe("Hide Gallery", () => {
     }
   );
 });
+describe("Interact with Browser local storage", () => {
+  it("saves wildcard mode in local storage", async () => {
```
SammyIsConfused, line 1144:
```suggestion
describe("Wildcard mode persistence", () => {
  it("saves wildcard mode value in local browser storage", async () => {
```

react-front-end/__tests__/tsrc/search/SearchPage.test.tsx
```diff
@@ -1139,3 +1140,31 @@ describe("Hide Gallery", () => {
...
+    const wildcardModeSwitch = getByText(
+      languageStrings.searchpage.wildcardSearch
+    );
+    await act(async () => {
+      await userEvent.click(wildcardModeSwitch);
+    });
+    expect(mockSetDataToLocalStorage).toHaveBeenCalledTimes(1);
+  });
+
+  it("retrieves wildcard mode from local storage", async () => {
```
SammyIsConfused, line 1159:
```suggestion
  it("retrieves wildcard mode value from local storage", async () => {
```

react-front-end/tsrc/search/SearchPage.tsx
```diff
@@ -495,9 +500,32 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
...
     );
 
-  const handleWildcardModeChanged = (wildcardMode: boolean) =>
+  const handleWildcardModeChanged = (wildcardMode: boolean) => {
     // `wildcardMode` is a presentation concept, in the lower levels its inverse is the value for `rawMode`.
     search({ ...searchPageOptions, rawMode: !wildcardMode });
+    try {
+      setDataToLocalStorage(WILD_CARD_MODE, JSON.stringify(wildcardMode));
+    } catch (error) {
+      console.error(
```
SammyIsConfused, line 509: I like that we are catching these and logging them out - nice work.

react-front-end/tsrc/search/SearchPageHelper.ts
```diff
@@ -378,3 +379,5 @@ export const getPartialSearchOptions = (
   options: SearchOptions,
   fields: SearchOptionsFields[]
 ) => pick(options, fields);
+
+export const WILD_CARD_MODE = `wildcardMode_${AppConfig.baseUrl}`;
```
SammyIsConfused, line 383: wildcard is one word, needs no underscore. Also - it might be worth having a better identifier as just calling it WILDCARD_MODE is somewhat ambiguous. WILDCARD_MODE_STORAGE_KEY or something along those lines.

SammyIsConfused (approved):
Looks great Penghai! Thanks for clearing those up. Just a couple suggestions, but you should be mostly good to go.

react-front-end/__tests__/tsrc/search/SearchPage.test.tsx
PenghaiZhang: We might have more data that need to be saved in local storage, but I am happy to rename the describe block. `local storage` is a term so I guess it should be `brower  local storage` ?

react-front-end/tsrc/search/SearchPage.tsx
PenghaiZhang: Yeah I think this error should not stop users from using the new search UI. An console error should be enough.

react-front-end/tsrc/search/SearchPage.tsx
```diff
@@ -495,9 +500,35 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
...
+        `${searchStrings.wildcardMode.errors.save}: ${error.message}`
+      );
+    }
+  };
+
+  const getWildcardMode = (): boolean => {
+    const wildcard = getDataFromLocalStorage(WILDCARD_MODE_STORAGE_KEY);
+    if (wildcard) {
+      try {
+        return JSON.parse(wildcard);
```
ChristianMurphy, line 522: Would it make sense to use runtypes or typescript-is here to assert that the parsed value is a boolean? :thinking: 
`JSON.parse` can return a lot of possible values.

ChristianMurphy (approved):

react-front-end/tsrc/search/SearchPage.tsx
PenghaiZhang: Good idea! I use runtypes to do this.

SammyIsConfused (approved):

react-front-end/tsrc/search/SearchPage.tsx
```diff
@@ -495,9 +501,39 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
...
+        }
+        return parsedValue;
+      }
+    } catch (error) {
+      console.error(
+        `${searchStrings.wildcardMode.errors.retrieve}: ${error.message}`
+      );
+    }
+
+    return !searchPageOptions.rawMode;
```
edalex-ian, line 535: hmmm. :thinking: 

hmmmmmmm... :thinking: 

This oddity caught my attention. We have potential for a mismatch here of what we're displaying in the search results vs what the toggle is displaying. I'm thinking maybe the logic should more so be:

* Wildcard toggle display state is _only_ based on `searchPageOptions` - as that is kept in sync with the search results (due to the reducer)
* We should indeed store the value of `rawMode` on each change, but it should use the same sequencing of how we store the SearchPageOptions into history. (i.e. do it as part of the `state.status === "searching"` effect just after we do the `history.replace`)
* Reading the persisted value should then be part of the process of setting the initial state of `SearchPageOptions`. It's kind of interesting how that is currently done, it should probably be within the `state.status === "initialising"` effect and I'd suggest that's where this initial read is done. To do so may require 'fixing' that setting of it the initial value of `SearchPageOptions` so that it is in the effect and then if there is nothing in 'history' but there is a value in LocalStorage then it's combined with the `defaultSearchPageHistory.searchPageOptions`, otherwise just use plain old `defaultSearchPageHistory.searchPageOptions`.

react-front-end/tsrc/util/BrowserStorageUtil.ts
```diff
@@ -0,0 +1,35 @@
...
+ * no data is found.
+ *
+ * @param key Key of the data to be retrieved.
+ */
+export const getDataFromLocalStorage = (key: string): string | null =>
+  localStorage.getItem(key);
+
+/**
+ * Save data to browser local storage.
+ * @param key Key of the data.
```
edalex-ian, line 31:
```suggestion
 *
 * @param key Key of the data.
```

react-front-end/tsrc/util/BrowserStorageUtil.ts
```diff
@@ -0,0 +1,35 @@
...
+ */
+const localStorage = window.localStorage;
+
+/**
+ * Retrieve data from browser local storage by a key. Return undefined if
+ * no data is found.
+ *
+ * @param key Key of the data to be retrieved.
+ */
+export const getDataFromLocalStorage = (key: string): string | null =>
```
edalex-ian, line 26: Feels more idiomatic to go with:

```suggestion
export const readFromLocalStorage = (key: string): string | null =>
```

react-front-end/tsrc/util/BrowserStorageUtil.ts
```diff
@@ -0,0 +1,35 @@
...
+  localStorage.getItem(key);
+
+/**
+ * Save data to browser local storage.
+ * @param key Key of the data.
+ * @param value Value of the data.
+ */
+export const setDataToLocalStorage = (key: string, value: string) =>
```
edalex-ian, line 34: Feels more idiomatic to go with:

```suggestion
export const saveToLocalStorage = (key: string, value: string) =>
```

react-front-end/tsrc/search/SearchPageHelper.ts
```diff
@@ -378,3 +379,5 @@ export const getPartialSearchOptions = (
   options: SearchOptions,
   fields: SearchOptionsFields[]
 ) => pick(options, fields);
+
+export const WILDCARD_MODE_STORAGE_KEY = `wildcardMode_${AppConfig.baseUrl}`;
```
edalex-ian, line 383: I think it could be useful to flip these around such as:

```suggestion
export const WILDCARD_MODE_STORAGE_KEY = `${AppConfig.baseUrl}_wildcardMode`;
```

That way if have more, then we can easily delete for an institution. But, not to phased. :shrug:

edalex-ian (changes requested):
Thanks @PenghaiZhang , this is an interesting one as we continue to add feature on top of feature for the SearchPage.

Main comment on the code you'll see is around changing the logic of when and at what point to store, read and use the value in LocalStorage.

My second main feedback is around the `BrowserStorageUtil.ts` file. Best to place that here I guess as it's at the file level.

**`BrowserStorageUtil.ts`**

I like the idea of abstracting out this storage, but as it stands this file only seems to offer that it's closed over `window.localStorage`. Could I suggest that we expand this further - which possibly then also pushes it more so to `BrowserStorageModule.ts`.

* Have it handle the prefixing of keys with an institution specific identifier;
* Remove all the logic and try/catch logic for reading and storing the values from `SearchPage.tsx` and encapsulate it here in one place
* Consider making the functions object aware - don't just return a string and receive a string
  * In doing so, maybe then the validation of the read in data shapes can be centralised here too
* Have these functions do the serialisation and deserialisation - and hey, why not play with some more fp-ts via the [`Json.ts` Module](https://gcanti.github.io/fp-ts/modules/Json.ts.html) which can then feed into the above point about error handling etc.

react-front-end/__tests__/tsrc/modules/BrowserStorageModule.test.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  it("returns undefined when data format doesn't match", () => {
+    mockGetItem.mockReturnValueOnce('{"a": 1}');
+    expect(readDataFromLocalStorage(KEY, booleanValidator)).toBeUndefined();
+    expect(mockConsoleError).toHaveBeenCalledTimes(1);
+  });
+
+  it("supports reading boolean data", () => {
+    mockGetItem.mockReturnValueOnce("true");
+    expect(readDataFromLocalStorage("test", booleanValidator)).toBe(true);
```
edalex-ian, line 51:
```suggestion
    expect(readDataFromLocalStorage(KEY, booleanValidator)).toBe(true);
```

react-front-end/tsrc/modules/BrowserStorageModule.ts
```diff
@@ -0,0 +1,82 @@
...
+import * as O from "fp-ts/Option";
+import { AppConfig } from "../AppConfig";
+
+const localStorage = window.localStorage;
+
+const buildStorageKey = (key: string) => `${AppConfig.baseUrl}_${key}`;
+
+/**
+ * Retrieve data from browser local storage by a key. Return undefined if
+ * no data is found.
```
edalex-ian, line 31:
```suggestion
 * no data is found, cannot be parsed or fails validation.
```

react-front-end/tsrc/modules/BrowserStorageModule.ts
```diff
@@ -0,0 +1,82 @@
...
+export const readDataFromLocalStorage = <T>(
+  key: string,
+  validator: (value: unknown) => value is T
+): T | undefined =>
+  pipe(
+    localStorage.getItem(buildStorageKey(key)),
+    O.fromNullable,
+    O.fold(
+      () => undefined,
+      flow(
```
edalex-ian, line 45: I think the idomatic way to do the interop is to use the provided interop function `toUndefined`. In which case your pipe would more so be:

```typescript
pipe(
  localStorage.getItem(...),
  O.fromNullable,
  O.map(
    flow(
      ...
    )
  ),
  O.toUndefined
);
```

react-front-end/tsrc/modules/BrowserStorageModule.ts
```diff
@@ -0,0 +1,82 @@
...
+        J.parse,
+        E.mapLeft((error) => `Failed to parse data due to ${error}`),
+        E.filterOrElse(
+          validator,
+          () => "Data format mismatch with data read from storage"
+        ),
+        E.mapLeft(console.error),
+        E.getOrElseW(() => undefined)
```
edalex-ian, line 53: With the above suggestion, perhaps this line should just be:

```suggestion
        O.fromEither)
```

Mind you, in doing that, you'd then want to use `O.chain` rather than `O.map`.

But at least with these two comments we only go back to the wild west of `undefined` and `null` when we exit the functional area.

react-front-end/tsrc/modules/BrowserStorageModule.ts
```diff
@@ -0,0 +1,82 @@
...
+    )
+  );
+
+/**
+ * Save data to browser local storage.
+ *
+ * @param key Key of the data.
+ * @param value Value of the data.
+ */
+export const saveDataToLocalStorage = (key: string, value: unknown) =>
```
edalex-ian, line 64: Please add an explicit return type.

(I should find the ESLint rule used over in the kaltura module to check for this and enable it on this repo. :thinking: )

react-front-end/tsrc/modules/BrowserStorageModule.ts
```diff
@@ -0,0 +1,82 @@
...
+        `Failed to stringify the provided value to JSON format: ${error}`
+    ),
+    E.fold(console.error, (right) =>
+      localStorage.setItem(buildStorageKey(key), right)
+    )
+  );
+
+/**
+ * Clear data from browser local storage.
+ * @param key Key of the data.
```
edalex-ian, line 79:
```suggestion

 * @param key Key of the data.
```

react-front-end/tsrc/modules/BrowserStorageModule.ts
```diff
@@ -0,0 +1,82 @@
...
+ */
+export const clearDataFromLocalStorage = (key: string) =>
```
edalex-ian, line 81: Explicit return type please.

It's good to make sure we have explicit types on exported / public functions.

react-front-end/tsrc/search/SearchPage.tsx
```diff
@@ -420,6 +419,8 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
               ...history.location,
               state: { searchPageOptions: state.options, filterExpansion },
             });
+            // Save the value of wildcard mode to LocalStorage.
+            saveDataToLocalStorage(RAW_MODE_STORAGE_KEY, state.options.rawMode);
```
edalex-ian, line 423: Like you've done for `getRawModeFromStorage`, this should probably be encapsulated in a `writeRawModeToStorage`.

react-front-end/tsrc/search/SearchPage.tsx
```diff
@@ -556,6 +527,7 @@ const SearchPage = ({ updateTemplate }: TemplateUpdateProps) => {
       externalMimeTypes: isSelectionSessionOpen()
         ? searchPageOptions.externalMimeTypes
         : undefined,
+      rawMode: searchPageOptions.rawMode,
```
edalex-ian, line 530: I think this warrants a comment:

```suggestion
      // As per requirements for persistence of rawMode, it is _not_ reset for New Searches
      rawMode: searchPageOptions.rawMode,
```

react-front-end/tsrc/legacycontent/LegacyContent.tsx
```diff
@@ -244,6 +246,11 @@ export const LegacyContent = React.memo(function LegacyContent({
   ) {
     submitRequest(toRelativeUrl(formAction || pathname), submitValues)
       .then((content) => {
+        // Clear raw mode saved in local storage after a login request is resolved.
+        if (pathname.indexOf("logon.do") > 0) {
+          clearDataFromLocalStorage(RAW_MODE_STORAGE_KEY);
```
edalex-ian, line 251: Maybe encapsulate this in `SearchPageHelper` as `deleteRawModeFromStorage()`.

edalex-ian (changes requested):
That's looking really tidy and well structured @PenghaiZhang .

Just a few minor tweaks remain and then should be good to go.

react-front-end/tsrc/modules/BrowserStorageModule.ts
PenghaiZhang: Ah ha I forgot `O.toUndefined`. Thanks !

react-front-end/tsrc/modules/BrowserStorageModule.ts
PenghaiZhang: Sure :smile:   Perhaps I thought the type is void so ...

edalex-ian (approved):
Very tidy and robust @PenghaiZhang . Great work! :+1: :tada:

PR #3209, opened 2021-07-14
feat: add support for DRM to REST module

PenghaiZhang:
##### Checklist

- [x] tests are included

##### Description of change

This PR adds the support for DRM to REST module, including two functions talking to server and some type definitions.

### DIFF
--- a/oeq-ts-rest-api/src/Drm.ts
+++ b/oeq-ts-rest-api/src/Drm.ts
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 import { is } from 'typescript-is';
-import { GET, POST } from './AxiosInstance';
+import { GET, POST_void } from './AxiosInstance';
 
 export interface DrmParties {
   /**  Server side language string for DRM party. */
@@ -85,13 +85,9 @@ export const listDrmTerms = (
  * @param apiBasePath Base URI to the oEQ institution and API
  * @param uuid UUID of an Item
  * @param version Version of an Item
- * @return `true` to indicate terms have been accepted.
  */
 export const acceptDrmTerms = (
   apiBasePath: string,
   uuid: string,
   version: number
-): Promise<true> =>
-  POST(buildPath(apiBasePath, uuid, version), (data): data is true =>
-    is<true>(data)
-  );
+): Promise<void> => POST_void(buildPath(apiBasePath, uuid, version));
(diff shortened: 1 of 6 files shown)


oeq-ts-rest-api/src/Drm.ts
```diff
@@ -0,0 +1,92 @@
...
+    is<ItemDrmDetails>(data)
+  );
+
+/**
+ * Accept an Item's DRM terms.
+ *
+ * @param apiBasePath Base URI to the oEQ institution and API
+ * @param uuid UUID of an Item
+ * @param version Version of an Item
+ */
```
edalex-ian, line 84: Be good to add a `@return` specification here, as what is this random `number` it's returning.

Should this API perhaps just return true/false? (even on the server side?)

oeq-ts-rest-api/src/Drm.ts
```diff
@@ -0,0 +1,92 @@
...
+  additionalPermission?: string;
+  /** Text describing that the use of Item is limited to education sector. */
+  educationSector?: string;
+  /** Text describing parties related to the Item. */
+  parties?: DrmParties;
+  /** Other terms and conditions applied to the Item. */
+  customTerms?: DrmCustomTerms;
+}
+
+export interface ItemDrmDetails {
```
edalex-ian, line 48: It would be useful for this various interfaces to have some JSDoc at the interface level.

edalex-ian (commented):
Looks good thanks @PenghaiZhang .

One question though (and I probably should've picked it up in the other PR), we discussed the idea that the Accept terms endpoint should probably _not_ return the internal DB id for the stored acceptance. Probably only needs at most a true/false, or may be fine to just rely on the return code from the server - i.e. if you get a 200 you know all is good.

Thoughts?

oeq-ts-rest-api/src/Drm.ts
PenghaiZhang: I think let's just rely on the return code because it seems confusing to see a 200 response which however returns false.

PenghaiZhang: But this make the validator looks a bit strange .

```
(data): data is void =>
    is<void>(data)
```
We know if the response code is 200, it must be `true` . So maybe let's return true.

edalex-ian: Sorry, I think I've confused you. My thinking was simply the server need not return any value at all. We'd simply rely on it succeeding (200) or failing (non-200).

Doing that, then even a validator is redundant.

edalex-ian (approved):
GTG :+1:

PR #3220, opened 2021-07-26
feat: Return Kaltura externalId in search results

edalex-ian:
##### Checklist

- [ ] tests are included

##### Description of change

This enables for previewing Kaltura media in search results and galleries. And by adding this to `AttachmentBean` also supports work in the future with Summary Pages etc.

Integrating with this will also be a PR over at https://github.com/openequella/openEQUELLA-Kaltura

### DIFF
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
@@ -41,16 +41,17 @@ import com.tle.web.api.item.equella.interfaces.beans.{
 }
 import com.tle.web.api.item.interfaces.beans.AttachmentBean
 import com.tle.web.api.search.model.{
+  DrmStatus,
   SearchParam,
   SearchResultAttachment,
-  SearchResultItem,
-  DrmStatus
+  SearchResultItem
 }
 import com.tle.web.controls.resource.ResourceAttachmentBean
 import com.tle.web.controls.youtube.YoutubeAttachmentBean
+
 import java.time.format.DateTimeParseException
 import java.time.{LocalDate, LocalDateTime, LocalTime, ZoneId}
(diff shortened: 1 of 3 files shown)


Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
```diff
@@ -435,11 +436,15 @@ object SearchHelper {
     *         if suitable for the attachment type.
     */
   def buildAttachmentLinks(attachment: AttachmentBean): java.util.Map[String, String] = {
-    val externalIdKey = "externalId"
-    val links         = getLinksFromBean(attachment).asScala
+    val addExternalId = (links: Map[String, String], externalId: Optional[String]) =>
+      if (externalId.isPresent) links ++ Map("externalId" -> externalId.get) else links
+
+    val links = getLinksFromBean(attachment).asScala.toMap
```
PenghaiZhang, line 442: `getLinksFromBean` actually returns a Java map so there is no need to call `toMap` again.
`getLinksFromBean` is missing its return type. What about adding it  ?

PenghaiZhang (approved):
LGTM.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
edalex-ian: This was to ensure I explicitly have a Scala immutable Map. Basically what we have here is:

- `getLinksFromBean` -> Java Map
- `...asScala` -> Scala **mutable** Map
- `...toMap` -> Scala **immutable** Map

It's kind of odd, as the default in Scala land (as per good FP practice) is to use the immutable datatypes. However I _guess_ because we want interop with Java when we to `asScala` we end up with a mutable instance. :man_shrugging:

PR #3240, opened 2021-07-29
Feature/show DRM violations in dialog

PenghaiZhang:
##### Checklist

- [x] tests are included

##### Description of change

This PR adds the support for listing DRM violations for unauthorised Items and attachments. The main change is to update the logic of DRM check in `DrmHelper`. This change results in function `createDrmDialog` becoming an async function, and therefore the `useEffect` which uses this function gets slightly changed as well. 

I tried to implement a custom hook `useAsyncEffect` but stopped after a couple hours experiments as I thought the next DRM ticket is more important.

https://user-images.githubusercontent.com/47203811/127578666-fe879e2a-1ea5-4687-9571-f3371a170598.mp4

### DIFF
--- a/react-front-end/__tests__/tsrc/search/components/GallerySearchResult.test.tsx
+++ b/react-front-end/__tests__/tsrc/search/components/GallerySearchResult.test.tsx
@@ -19,14 +19,19 @@ import "@testing-library/jest-dom/extend-expect";
 import { queryByText, render, screen, waitFor } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { createMemoryHistory } from "history";
+import { act } from "react-dom/test-utils";
 import { Router } from "react-router-dom";
 import * as React from "react";
-import { drmTerms } from "../../../../__mocks__/Drm.mock";
+import { DRM_VIOLATION, drmTerms } from "../../../../__mocks__/Drm.mock";
 import * as DrmModule from "../../../../tsrc/modules/DrmModule";
 import { GallerySearchResultItem } from "../../../../tsrc/modules/GallerySearchModule";
 import GallerySearchResult from "../../../../tsrc/search/components/GallerySearchResult";
 import { languageStrings } from "../../../../tsrc/util/langstrings";
-import { buildItems, galleryDrmItem } from "./GallerySearchResultHelpers";
+import {
+  buildItems,
+  galleryDrmItem,
+  galleryDrmUnauthorisedItem,
+} from "./GallerySearchResultHelpers";
 import * as ReactRouterDom from "react-router-dom";
 
 const {
@@ -59,17 +64,21 @@ describe("<GallerySearchResult />", () => {
       </Router>
     ); // 16 entries in total.
 
-  it("displays the lightbox when the image is clicked on", () => {
+  it("displays the lightbox when the image is clicked on", async () => {
     const { getAllByLabelText, queryAllByLabelText } = renderGallery();
-    userEvent.click(getAllByLabelText(ariaLabel)[0]);
+    await act(async () => {
+      await userEvent.click(getAllByLabelText(ariaLabel)[0]);
+    });
 
     // Then they see the lightbox
     expect(queryAllByLabelText(openInNewWindow)[0]).toBeInTheDocument();
   });
 
-  it("navigates to the images item when the information icon is clicked on", () => {
+  it("navigates to the images item when the information icon is clicked on", async () => {
     const { getAllByLabelText } = renderGallery();
-    userEvent.click(getAllByLabelText(viewItem)[0]);
+    await act(async () => {
+      await userEvent.click(getAllByLabelText(viewItem)[0]);
+    });
 
     expect(mockUseHistoryPush).toHaveBeenCalled();
     expect(mockUseHistoryPush.mock.calls[0][0].match("/items/")).toBeTruthy();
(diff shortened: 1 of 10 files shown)


react-front-end/tsrc/drm/DrmHelper.tsx
```diff
@@ -66,8 +71,40 @@ export const createDrmDialog = (
...
+    );
   } else {
     drmProtectedHandler();
     return;
   }
 };
+
+// Call the API to list DRM violations in a TaskEither where Left is why fail to
+// list violations and Right is the retrieved violation.
+const listDrmViolationsInTask = async (
```
PenghaiZhang, line 92: This function was in the module but I moved it to the helper.

react-front-end/__tests__/tsrc/search/components/GallerySearchResult.test.tsx
```diff
@@ -120,5 +134,26 @@ describe("<GallerySearchResult />", () => {
         });
       }
     );
+
+    it.each([
+      ["item", languageStrings.searchpage.gallerySearchResult.viewItem],
+      ["attachment", languageStrings.searchpage.gallerySearchResult.ariaLabel],
```
edalex-ian, line 140: The `languageStrings` have been de-structured up the top, so:

```suggestion
      ["item", viewItem],
      ["attachment", ariaLabel],
```

That `ariaLabel` though keeps grabbing my attention. At somepoint we may want to refactor that with a more useful name - or maybe just in the de-structured context it loses meaning.

react-front-end/__tests__/tsrc/search/components/GallerySearchResult.test.tsx
```diff
@@ -120,5 +134,26 @@ describe("<GallerySearchResult />", () => {
...
+    ])(
+      "shows a dialog to list DRM violations for %s",
+      async (_: string, galleryEntryLabel: string) => {
+        const { getByLabelText } = await renderGallery([
+          galleryDrmUnauthorisedItem,
+        ]);
+        await act(async () => {
+          await userEvent.click(getByLabelText(galleryEntryLabel));
+        });
```
edalex-ian, line 149: There's a lot of boilerplate here. I wonder if we need a general (available for all tests) function like:

```typescript
const asyncClick = (element: Element, init?: MouseEventInit, {skipHover, clickCount}?: clickOptions): void =>
  act(async () => {
    await userEvent.click(element);
  });
```

And then we'd just be able to have:

```suggestion
        await asyncClick(getByLabelText(galleryEntryLabel));
```

I wonder too if `asyncClick` is the best name. Maybe `clickThenWait` as really this is all about clicking a button and then waiting for React to do it's state updates etc. Hmmm... :thinking:

react-front-end/__tests__/tsrc/search/components/SearchResult.test.tsx
```diff
@@ -537,5 +540,23 @@ describe("<SearchResult/>", () => {
         });
       }
     );
+
+    it.each([
+      ["item", "DRM Item"],
+      ["attachment", "DRM Attachment"],
```
edalex-ian, line 546: The string literals of `"DRM Item"` and `"DRM Attachment"` are now in three places and ideally should all be linking to the same thing. Perhaps over in the mock we need to set a constant, and then use that to tie them altogether.

react-front-end/tsrc/drm/DrmHelper.tsx
```diff
@@ -66,8 +71,40 @@ export const createDrmDialog = (
...
+    TE.map<OEQ.Drm.DrmViolation, string>(({ violation }) => violation)
+  )();
+
+  return pipe(
+    violation,
+    E.getOrElse(
+      (error) => `Failed to list DRM violations due to ${error.message}`
+    )
+  );
+};
```
edalex-ian, line 110: Does this need to be broken up, or can it just be:

```suggestion
): Promise<string> =>
  await pipe(
    TE.tryCatch<Error, OEQ.Drm.DrmViolation>(
      () => listDrmViolations(uuid, version),
      (error) => new Error(`${error}`)
    ),
    TE.map<OEQ.Drm.DrmViolation, string>(({ violation }) => violation)
    TE.getOrElse(
      (error) => `Failed to list DRM violations due to ${error.message}`
    )
  )();
```

Are are the issues with the async/await?

react-front-end/tsrc/drm/DrmHelper.tsx
```diff
@@ -66,8 +71,40 @@ export const createDrmDialog = (
         open={!!drmProtectedHandler}
       />
     );
+  } else if (!isAuthorised) {
```
edalex-ian, line 74: I wonder if we should more so go with a 'deny all' default for the sake of robustness. To demonstrate with pseudo code:

```
if isAuthorised
  if termsAccepted
    call handler
  else
    show DrmAcceptanceDialog
else
  show Violations dialog
```

edalex-ian (changes requested):
Nice @PenghaiZhang , this has come together rather cleanly.

Just a couple of minor requests, and we probably need to check the wording on the dialog with @cathfitz

react-front-end/__tests__/tsrc/search/components/GallerySearchResult.test.tsx
PenghaiZhang: Perhaps rename to `viewAttachment` ?

react-front-end/__tests__/tsrc/search/components/GallerySearchResult.test.tsx
PenghaiZhang: As discussed, we leave this for now and try to do it maybe in next tech debt.

Issue #1160, opened 2019-08-08
"Item not found" when clicking fullscreen on DRM-enabled attachment in new UI


SammyIsConfused:
**Describe the bug**
In 2019.1, when using the new UI, an incorrect "Item not found" error displays when clicking "Fullscreen" on a DRM-protected attachment.
`DRMTest` highlighted this issue. 

**To Reproduce**
1. Run `DRMTest` in the new UI.

OR

1. Go to the Resource Summary page of an item with DRM on the attachments, using an account that requires agreement of this DRM before viewing.
2. Click "Fullscreen".
3. See error.

**Expected behavior**
The DRM accept/reject dialog should pop up, and on accept it should show the attachment fullscreen.

**Screenshots**
![image](https://user-images.githubusercontent.com/24543345/62670291-12240080-b9d6-11e9-8fb5-43490398352b.png)

**Platform:**

- OpenEquella Version: 2019.1
- OS: Ubuntu 19.04
- Browser: Chrome, Firefox

SammyIsConfused:
Reopening, as this came up in support (discovered and reported by Carl, not by a client.)

snehanaik22:
Tested.

PR #3250, opened 2021-08-03
Bugfix/show DRM acceptance page before viewing content in full screen

PenghaiZhang:
#1160 

##### Checklist

- [ ] tests are included

##### Description of change

This bug was firstly caused by `LegacyContentApi` failing to parse the URL, which resulted in Item version being -1.

After above issue was fixed, the next issue was the page stayed in Summary page rather than going to the DRM acceptance page. After lots of investigations, I found the key thing is to make sure `RootItemFileModel` has a proper file name which can be extracted from the URL and is used to determine what viewers to be used in the full screen mode.

After fixing the second issue, there was a third issue. Once DRM terms had been accepted, although the page showing contents was opened, the page looked very bad because CSS styles applied to this page were never included in `Legacy.scss`.

Once the third issue was fixed, this bug was completed fixed. So I reverted workaround added in DRMTest.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/web/viewitem/treeviewer/AbstractTreeViewerSection.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/web/viewitem/treeviewer/AbstractTreeViewerSection.java
@@ -63,6 +63,7 @@ import com.tle.web.sections.standard.Link;
 import com.tle.web.sections.standard.annotations.Component;
 import com.tle.web.template.Decorations;
 import com.tle.web.template.Decorations.FullScreen;
+import com.tle.web.template.RenderNewTemplate;
 import com.tle.web.viewable.ViewableItem;
 import com.tle.web.viewitem.section.RootItemFileSection;
 import com.tle.web.viewitem.treeviewer.js.TreeLibrary;
@@ -113,6 +114,8 @@ public abstract class AbstractTreeViewerSection<M extends AbstractTreeViewerMode
   @Component
   @PlugKey(value = "navbar.last", icon = Icon.JUMP_TO_LAST)
   private Button last;
+
+  public static final String VIEW_FROM_NEW_UI = "view_from_new_ui";
 
   protected interface NodeUrlGenerator {
     @Nullable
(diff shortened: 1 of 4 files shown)


edalex-ian:
> Once the third issue was fixed, this bug was completed fixed.

Gah! Fair effort @PenghaiZhang 

Ye olde can o' worms.

Source/Plugins/Core/com.equella.core/src/com/tle/web/viewitem/treeviewer/AbstractTreeViewerSection.java
```diff
@@ -179,6 +182,12 @@ public SectionResult view(RenderContext info, ViewItemResource resource) {
     model.setHideTree(oneAttachment);
     model.setHideNavControls(oneAttachment);
     model.setDefinition(getTreeDef(info, tree, new HashSet<String>()));
+
+    // If this section is used in new UI, set a flag to tell LegacyContentApi NOT to use
+    // 'Legacy.css'.
```
edalex-ian, line 187: If the intention is to "tell LegacyContentApi" then shouldn't the constant for this flag reside over there? And maybe a name like `DISABLE_LEGACY_CSS`?

This will allow the concept to be nicely encapsulated over where it's used. That in turn will mean over in the check in `LegacyContentApi` there'll be no need to store the value in an intermediate variable (just make the call in the if/else; and you won't need to explicitly check a flag unique to this Section - so if we need similar for other sections in the future it'll be simple.

edalex-ian (commented):
Looks good @PenghaiZhang , and great effort documenting the regular expression. :+1: 

Just one query around management of the flag to disable Legacy.css.

Source/Plugins/Core/com.equella.core/src/com/tle/web/viewitem/treeviewer/AbstractTreeViewerSection.java
PenghaiZhang: Good idea! I have refactored this.

edalex-ian (approved):

PR #3257, opened 2021-08-04
fix(rest/search2): Use description from Attachment

edalex-ian:
##### Checklist

- [ ] tests are included

##### Description of change

Rather than use the description from AttachmentBean, use the one from
Attachment as it supports any tweaks done in CustomAttachment instances
(such as Kaltura) to provide a string more suitable.

In turn, the resultant strings better align with those seen in the
LegacyUI.

(I also tweaked the broken attachment check to try and ensure consistency
with convention. No logical change though.)

### DIFF
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
@@ -20,8 +20,8 @@ package com.tle.web.api.search
 
 import com.dytech.edge.exceptions.{BadRequestException, DRMException}
 import com.tle.beans.entity.DynaCollection
-import com.tle.beans.item.attachments.{Attachment, CustomAttachment, FileAttachment}
-import com.tle.beans.item.{Comment, ItemId, ItemIdKey}
+import com.tle.beans.item.attachments.{CustomAttachment, FileAttachment}
+import com.tle.beans.item.{Comment, ItemId, ItemIdKey, ItemKey}
 import com.tle.common.Check
 import com.tle.common.beans.exception.NotFoundException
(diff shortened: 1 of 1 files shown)


Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala
```diff
@@ -463,4 +461,17 @@ object SearchHelper {
...
+    * the value more commonly seen in the LegacyUI. And specifically uses any tweaks done for
+    * Custom Attachments - such as with Kaltura where the Kaltura Media `title` is pushed into
+    * the `description` rather than using the optional (and multi-line) Kaltura Media `description`.
+    *
+    * @param itemKey the details of the item the attachment belongs to
+    * @param attachmentUuid the UUID of the attachment
+    * @return the description for the attachment if available
+    */
+  def getAttachmentDescription(itemKey: ItemKey, attachmentUuid: String): Option[String] =
+    Option(LegacyGuice.itemService.getAttachmentForUuid(itemKey, attachmentUuid).getDescription)
```
PenghaiZhang, line 476: `getAttachmentForUuid` can throw `AttachmentNotFoundException`. Do we want to handle this ?

edalex-ian: I was thinking not, because if that occurs then we have some corrupted data or state. I'd assume this should be handled further up.

edalex-ian: Gah, but now after deliberately triggering an exception, I see we have nothing in there. One exception brings down the who search query.

PenghaiZhang (approved):
LGTM.

PR #3262, opened 2021-08-05
chore: add legacy peer dependencies

ChristianMurphy:
suggested in https://github.com/renovatebot/renovate/discussions/9204
this allows working around https://github.com/peternewnham/react-html-parser/pull/80 and https://github.com/gvaldambrini/storybook-router/pull/66 by using npm 6 style peer dependencies, this can be removed when upstream peer conflicts have been resolved

##### Checklist

- [x] tests are included

##### Description of change

Address https://github.com/openequella/openEQUELLA/pull/3259#issuecomment-893042580


react-front-end/.npmrc
```diff
@@ -0,0 +1 @@
+legacy-peer-deps=true
```
ChristianMurphy, line 1: alternatively

```suggestion
force=true
```

could also work, neither is ideal, as these should be patched upstream

edalex-ian (commented):
Thanks @ChristianMurphy .

I can see how this would hopefully fix the issue, but arguably it then introduces a new issue - i.e. we'd have to hopefully make sure we remove this in the future.

So that got me wondering, how far in the future?

Well, from looking at the two mentioned repositories it almost looks like both are dead. Indeed, the contributor for the router module has not even done anything on GitHub in general for 2+ years, and the other has not had any substantive updates since 2017!

I wonder if maybe the issue is we need to replace those dependencies with something that is maintained... Or less desirable, fork them and maintain them....

Another thing I've observed in this PR is that some of the direct dependencies in `package.json` are not correct in `package-lock.json` - e.g. in `package.json` we have `fp-ts` at 2.11.1 and yet `package-lock.json` still has it at `2.10.5`. I wonder how much benefit we might get just be updating `package-lock.json`. :thinking:

ChristianMurphy:
> I wonder if maybe the issue is we need to replace those dependencies with something that is maintained... Or less desirable, fork them and maintain them

Agreed, moving to libraries which are still active is the better solution.
I'm fine closing this out.

> Another thing I've observed in this PR is that some of the direct dependencies in package.json are not correct in package-lock.json - e.g. in package.json we have fp-ts at 2.11.1 and yet package-lock.json still has it at 2.10.5.

In this PR and in the `react-front-end` folder?
I'm seeing fp-ts 2.11.1 line up between the package.json and package-lock.json. :thinking: 

>  wonder how much benefit we might get just be updating package-lock.json. :thinking: 

It's been a while since we last requested a lock file maintenance PR from https://github.com/openequella/openEQUELLA/issues/1532, getting another lockfile update would probably be beneficial.

edalex-ian:
Strange, I thought there had been some more recent ones. But none the less, we now have #3265 . It's been a bit bumpy, but will see how it goes with GHA and then look closer.

PenghaiZhang:
I have not found any alternative for `storybook-router`, but we have to use it ? Can't we just use `react-router` such as `addDecorator(story => <Router>{story()}</Router>);`

edalex-ian:
> I have not found any alternative for `storybook-router`, but do we have to use it ? Can't we just use `react-router` such as `addDecorator(story => <Router>{story()}</Router>);`

Hey! If we can do that and get rid of yet another dependency I'd be pretty happy.

Have you had a chance to try?

PR #3273, opened 2021-08-09
bugfix: don't include Bootstrap.css when open legacy DRM dialog in new UI

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

In new UI, the legacy DRM acceptance dialog is usually NOT open by `LegacyContent.tsx`, which means additional CSS files will be added to the whole page once the dialog is opened. One of those additional CSS files is `Bootstrap.css` which can break new UI components like buttons. So this PR adds a fix which allows the DRM dialog to include this CSS file only in Old UI.

Before: 
image

After:
image

Before: 
image

After:
image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/LegacyContentApi.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/LegacyContentApi.scala
@@ -203,6 +203,7 @@ object LegacyContentController extends AbstractSectionsController with SectionFi
   val RedirectedAttr = "REDIRECTED"
 
   val DISABLE_LEGACY_CSS = "DISABLE_LEGACY_CSS"
+  val SKIP_BOOTSTRAP     = "skip_bootstrap";
 
   override protected def getTreeForPath(path: String): SectionTree =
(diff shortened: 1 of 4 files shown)


edalex-ian (commented):
I like that this is the same pattern as #3250 . And with that in mind, I do wonder whether the new attribute `SKIP_BOOTSTRAP` should perhaps reside over in `LegacyContentController` along side `DISABLE_LEGACY_CSS`. What do you think?

PenghaiZhang:
> I like that this is the same pattern as #3250 . And with that in mind, I do wonder whether the new attribute `SKIP_BOOTSTRAP` should perhaps reside over in `LegacyContentController` along side `DISABLE_LEGACY_CSS`. What do you think?

The reason I did not put this attirbute in `LegacyContentController` is because the request to open the DRM dialog does not hit the endpoint of LegacyContentApi. I do agree this is the same pattern as #3250  so it makes sense to move the attribute to `LegacyContentController`. Maybe in the future we would have LegacyContentApi requests that would use this attribute.

PenghaiZhang:
I have moved this constant to LegacyContentApi alongside with `DISABLE_LEGACY_CSS`.

edalex-ian (approved):

PR #3344, opened 2021-08-26
Feature/wizard definition endpoint

PenghaiZhang:
##### Checklist

- [x] tests are included

##### Description of change

This PR adds a REST endpoint for retrieving Advanced search wizard definition. Several data models are created to represent controls that wizard controls. 

The initial idea was to use inheritance - create a basic control to provide common fields and then those having special fields just extend the basic one. I tried `trait` and `abstract class` both of them can produce lots of duplication. 

So I went to composition which works well, but I am not so sure if this is the best choice.

Screenshot from 2021-08-26 15-10-52

I have attached a full json file.
[controls.zip](https://github.com/openequella/openEQUELLA/files/7051791/controls.zip)

### DIFF
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/settings/AdvancedSearchResource.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/settings/AdvancedSearchResource.scala
@@ -18,13 +18,17 @@
 
 package com.tle.web.api.settings
 
+import com.dytech.edge.wizard.beans.control._
+import com.tle.beans.entity.PowerSearch
 import com.tle.legacy.LegacyGuice
+import com.tle.web.api.ApiErrorResponse.resourceNotFound
 import com.tle.web.api.entity.BaseEntitySummary
+import com.tle.web.api.wizard._
 import io.swagger.annotations.{Api, ApiOperation}
+import org.slf4j.LoggerFactory
 import javax.ws.rs.core.Response
-import javax.ws.rs.{GET, Path, Produces}
-
-import scala.collection.JavaConverters.collectionAsScalaIterableConverter
+import javax.ws.rs.{GET, Path, PathParam, Produces}
+import scala.collection.JavaConverters._
 
 /**
   * API for managing Advanced Searches (internally - and historically - known as Power Searches).
@@ -33,6 +37,7 @@ import scala.collection.JavaConverters.collectionAsScalaIterableConverter
 @Produces(value = Array("application/json"))
 @Api(value = "Settings")
 class AdvancedSearchResource {
+  private val LOGGER             = LoggerFactory.getLogger(classOf[AdvancedSearchResource])
   private val powerSearchService = LegacyGuice.powerSearchService
 
   @GET
@@ -52,4 +57,48 @@ class AdvancedSearchResource {
           .asScala
           .map(be => BaseEntitySummary(be)))
       .build()
+
+  @GET
+  @Path("{uuid}")
+  @ApiOperation(
+    value = "Get Advanced Search definition",
+    notes = "This endpoint is used to retrieve Wizard definition of an Advanced Search by UUID.",
+    response = classOf[WizardBasicControl],
+    responseContainer = "list"
+  )
+  def getAdvancedSearchWizardDefinition(@PathParam("uuid") uuid: String): Response = {
+    Option(powerSearchService.getByUuid(uuid)) match {
+      case Some(ps) =>
+        Response
+          .ok()
+          .entity(getWizardDefinition(ps))
+          .build()
+      case None => resourceNotFound(s"Failed to find Advanced search for ID: ${uuid}")
+    }
+  }
+
+  private def getWizardDefinition(ps: PowerSearch): List[WizardControlDefinition] =
+    ps.getWizard.getControls.asScala.map {
+      // These controls do not have special fields
+      case c @ (_: ListBox | _: CheckBoxGroup | _: RadioGroup | _: ShuffleBox | _: Html) =>
+        WizardBasicControl(c)
+      case c: Calendar => WizardCalendarControl(WizardBasicControl(c), c.isRange)
+      case c: ShuffleList =>
+        WizardShuffleListControl(WizardBasicControl(c),
+                                 c.isTokenise,
+                                 c.isForceUnique,
+                                 c.isCheckDuplication)
+      case c: EditBox =>
+        WizardEditBoxControl(WizardBasicControl(c),
+                             c.isAllowLinks,
+                             c.isNumber,
+                             c.isAllowMultiLang,
+                             c.isForceUnique,
+                             c.isCheckDuplication)
+      case c: CustomControl =>
+        WizardCustomControl(WizardBasicControl(c), c.getAttributes.asScala.toMap)
+      case _ =>
+        LOGGER.error("Unknown Wizard Control type")
+        UnknownWizardControl()
+    }.toList
 }
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/settings/AdvancedSearchResource.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/settings/AdvancedSearchResource.scala
@@ -68,7 +68,8 @@ class AdvancedSearchResource {
   @Path("{uuid}")
   @ApiOperation(
     value = "Get Advanced Search definition",
-    notes = "This endpoint is used to retrieve Wizard definition of an Advanced Search by UUID.",
+    notes =
+      "This endpoint is used to retrieve an Advanced Search's name, description, Collections and Wizard definition by UUID.",
     response = classOf[AdvancedSearch],
   )
   def getAdvancedSearchWizardDefinition(@PathParam("uuid") uuid: String): Response = {
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/settings/AdvancedSearchResource.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/settings/AdvancedSearchResource.scala
@@ -18,15 +18,13 @@
 
 package com.tle.web.api.settings
 
-import com.dytech.edge.wizard.beans.control._
-import com.tle.beans.entity.PowerSearch
 import com.tle.legacy.LegacyGuice
 import com.tle.web.api.ApiErrorResponse.resourceNotFound
 import com.tle.web.api.entity.BaseEntitySummary
 import com.tle.web.api.language.LanguageStringHelper.getStringFromCurrentLocale
+import com.tle.web.api.wizard.WizardControlHelper.wizardControlConverter
 import com.tle.web.api.wizard._
 import io.swagger.annotations.{Api, ApiModelProperty, ApiOperation}
-import org.slf4j.LoggerFactory
 import javax.ws.rs.core.Response
 import javax.ws.rs.{GET, Path, PathParam, Produces}
 import scala.collection.JavaConverters._
@@ -46,7 +44,6 @@ case class AdvancedSearch(name: Option[String],
 @Produces(value = Array("application/json"))
 @Api(value = "Settings")
 class AdvancedSearchResource {
-  private val LOGGER             = LoggerFactory.getLogger(classOf[AdvancedSearchResource])
   private val powerSearchService = LegacyGuice.powerSearchService
 
   @GET
@@ -77,7 +74,7 @@ class AdvancedSearchResource {
   def getAdvancedSearchWizardDefinition(@PathParam("uuid") uuid: String): Response = {
     Option(powerSearchService.getByUuid(uuid)) match {
       case Some(ps) =>
-        val controls    = getWizardDefinition(ps)
+        val controls    = wizardControlConverter(ps.getWizard.getControls.asScala.toList)
         val collections = ps.getItemdefs.asScala.map(c => c.getUuid).toList
         val name        = getStringFromCurrentLocale(ps.getName)
         val description = getStringFromCurrentLocale(ps.getDescription)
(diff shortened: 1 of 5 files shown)


edalex-ian:
Another pondering @PenghaiZhang ... Seeing this is `GET /settings/advancedsearch/{uuid}` are there any other values about the Advanced Search we should also be returning? For example, should we return the name, description and collection(s) configured to search against?

And although the search endpoint has the odd idea of specifying the UUID of the advanced search to get the list of collections to use in the search, we simply returned the list of collections as part of this endpoint then it would further help us construct the query client side with existing logic. :thinking:

PenghaiZhang:
> Another pondering @PenghaiZhang ... Seeing this is `GET /settings/advancedsearch/{uuid}` are there any other values about the Advanced Search we should also be returning? For example, should we return the name, description and collection(s) configured to search against?
> 
> And although the search endpoint has the odd idea of specifying the UUID of the advanced search to get the list of collections to use in the search, we simply returned the list of collections as part of this endpoint then it would further help us construct the query client side with existing logic.

Ah I totally forgot that an advanced search can be configured to work for certain collections. Yes, I agree that we should also return the list of configured collections rather than specifying the UUID of the advanced search in search2.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/wizard/WizardBasicControl.scala
```diff
@@ -0,0 +1,151 @@
...
+  * @param include #todo
+  * @param size1 #todo
+  * @param size2 #todo
+  * @param customName The controls' customised name.
+  * @param title Title of the control.
+  * @param description Description of the control.
+  * @param visibilityScript Script running on Client to control the visibility of the control.
+  * @param targetNodes Schema nodes that the control targets to.
+  * @param options Options available for selection.
+  * @param powerSearchFriendlyName #todo
```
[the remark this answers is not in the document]
edalex-ian: On some controls if you go down the bottom of their settings you'll see an 'Advanced Search Options' section which then has a text field for 'Friendly Name'. I think - you'll need to test - this is used when showing the summary of criteria instead of simply showing the XML node name.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/wizard/WizardBasicControl.scala
```diff
@@ -0,0 +1,169 @@
...
+      controlType = control.getClassType
+    )
+  }
+}
+
+/**
+  * Data structure for Calendar control.
+  *
+  * @param basicControl The basic control providing common fields.
+  * @param isRange Whether to support a date range.
```
edalex-ian, line 105: Probably need more. Something like: If false, should only display a single calendar control and supply a single date value, however if true, two controls should be displayed with the values from both representing a range.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/wizard/WizardBasicControl.scala
```diff
@@ -0,0 +1,169 @@
...
+case class WizardCalendarControl(@JsonUnwrapped
+                                 basicControl: WizardBasicControl,
+                                 isRange: Boolean)
+    extends WizardControlDefinition
+
+/**
+  * Data structure for ShuffleList control.
+  *
+  * @param basicControl The basic control providing common fields.
+  * @param isTokenise Whether to tokenise the value.
```
edalex-ian, line 116: Tokenise how?

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/wizard/WizardBasicControl.scala
```diff
@@ -0,0 +1,169 @@
...
+    isCheckDuplication: Boolean
+) extends WizardControlDefinition
+
+/**
+  * Data structure for EditBox control.
+  *
+  * @param basicControl The basic control providing common fields.
+  * @param isAllowLinks Whether to allow links.
+  * @param isNumber Whether to use numbers only.
+  * @param isAllowMultiLang Whether to allow multiple lines.
```
edalex-ian, line 134: Curious, the comment talks about multiple lines, but the variable name seems to talk about multiple languages (Lang). What's going on here?

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/wizard/WizardBasicControl.scala
```diff
@@ -0,0 +1,169 @@
...
+  * @param isForceUnique Whether each value must be unique.
+  * @param isCheckDuplication Whether to check duplicated values.
```
edalex-ian, line 136: These seem to be a common theme, should then be put into another case class to compose with? e.g. ControlUniqueConstraints or something...

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/wizard/WizardBasicControl.scala
```diff
@@ -0,0 +1,169 @@
...
+/**
+  * Data structure for Custom Wizard control.
+  *
+  * @param basicControl The basic control providing common fields.
+  * @param attributes Custom attributes
+  */
+case class WizardCustomControl(
+    @JsonUnwrapped
+    basicControl: WizardBasicControl,
+    attributes: Map[Object, Object]
```
edalex-ian, line 157: This `Object, Object` concerns me. But I've no idea what this WizardCustomControl is about - the description in the comment is unhelpful. Could you please update comment and let me know? Then I can figure out this very generic map.

autotest/OldTests/src/test/java/io/github/openequella/rest/AdvancedSearchApiTest.java
```diff
@@ -35,6 +37,21 @@ public void testRetrieveAdvancedSearches() throws Exception {
         searches.size());
   }
 
+  @Test(description = "list all Wizard controls of an Sdvanced search")
```
edalex-ian, line 40:
```suggestion
  @Test(description = "list all Wizard controls of an Advanced search")
```

autotest/OldTests/src/test/java/io/github/openequella/rest/AdvancedSearchApiTest.java
```diff
@@ -35,6 +37,21 @@ public void testRetrieveAdvancedSearches() throws Exception {
...
+  public void testRetrieveWizardDefinition() throws IOException {
```
edalex-ian, line 41: Needs a bit more work. Checks that it gets the right number of controls back, but they could arguably all be 'unknown' controls....

edalex-ian (changes requested):
Looks almost there @PenghaiZhang . Great work. :+1: 

Just needs:

* Minor refactoring;
* Tidy up of some comments - and clarification; and
* Expand the test a little.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/settings/AdvancedSearchResource.scala
```diff
@@ -52,4 +57,48 @@ class AdvancedSearchResource {
...
+                             c.isAllowLinks,
+                             c.isNumber,
+                             c.isAllowMultiLang,
+                             c.isForceUnique,
+                             c.isCheckDuplication)
+      case c: CustomControl =>
+        WizardCustomControl(WizardBasicControl(c), c.getAttributes.asScala.toMap)
+      case _ =>
+        LOGGER.error("Unknown Wizard Control type")
+        UnknownWizardControl()
```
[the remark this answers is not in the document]
PenghaiZhang: I added another object named `WizardControlHelper` to provide this function.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/wizard/WizardBasicControl.scala
```diff
@@ -0,0 +1,169 @@
...
+package com.tle.web.api.wizard
+
+import com.dytech.edge.wizard.TargetNode
+import com.dytech.edge.wizard.beans.control.{WizardControl, WizardControlItem}
+import com.fasterxml.jackson.annotation.JsonUnwrapped
+import com.tle.beans.entity.LanguageBundle
+import com.tle.common.i18n.{CurrentLocale, LangUtils}
+import scala.collection.JavaConverters._
+
+/** Data structure for Wizard Control option
```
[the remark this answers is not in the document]
PenghaiZhang: It's for a Option' type wizard control. I will update the doco.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/wizard/WizardBasicControl.scala
```diff
@@ -0,0 +1,169 @@
...
+
+// This trait is used to support the union type of different Wizard control types.
+// We can probably remove it trait once we can use Scala 3.
+sealed trait WizardControlDefinition
+
+/**
+  * Data structure for key information of 'com.dytech.edge.wizard.beans.control.WizardControl'.
+  *
+  * @param mandatory Whether the control must have a value.
+  * @param reload Whether to reload all controls after this control has different selection. This may affect how scripting works.
```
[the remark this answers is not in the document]
PenghaiZhang: Great!

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/wizard/WizardBasicControl.scala
PenghaiZhang: I have added this - `If true, an '*' must be appended to the schema node in the Lucene query.`

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/wizard/WizardBasicControl.scala
PenghaiZhang: :sweat:  I might misread the property.  I will update the comment. 

Actually, these is no usage of this boolean property and I have not found any configuration for it in admin console.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/wizard/WizardBasicControl.scala
PenghaiZhang: This class is all about `CustomControl` which is a subtype of `WizardControl` such as `Term Selector`. The most important property of a CustomControl is `attributes` which is `Map[Object, Object]`.

PenghaiZhang: The property `attributes` is a map where keys and values represent configurations that are specific to a certain type of CustomControl.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/wizard/WizardBasicControl.scala
PenghaiZhang: Not very common though :smile:  Only in two types. If put it into another case class, then we need to use `@JsonUnwrapped` again.

edalex-ian: True, but once you have the same in more than once place you normally want a good reason for the repetition - which I'm not sure avoiding an annotation is a great one.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/wizard/WizardBasicControl.scala
edalex-ian: Okay, that additional information in your first comment here would be great to help flesh out the code comment describing the case class. As for the map, if it is for configurations is the key seriously different types of Object and not just `String`?

PenghaiZhang: Not sure why the map was designed to use `Object` as key, but I can confirm that all the keys are `String`. So on Scala side I will use String.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/wizard/WizardBasicControl.scala
```diff
@@ -0,0 +1,151 @@
...
+  * @param include #todo
+  * @param size1 #todo
+  * @param size2 #todo
```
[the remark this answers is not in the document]
PenghaiZhang: So should we filter out controls whose `include` is false on server then  ?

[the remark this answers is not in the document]
PenghaiZhang: Yes I know what `size1` and `size2` mean now! Thank you!

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/wizard/WizardBasicControl.scala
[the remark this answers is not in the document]
PenghaiZhang: You are 100% correct! I have tested.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/wizard/WizardBasicControl.scala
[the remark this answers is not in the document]
edalex-ian: > So should we filter out controls whose `include` is false on server then ?

No, I don't think so. Again, it's REST endpoint, leave it to the client.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/wizard/WizardBasicControl.scala
```diff
@@ -0,0 +1,169 @@
...
+  * @param customName The controls' customised name.
```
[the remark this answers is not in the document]
PenghaiZhang: After some debugging, I think this property is used in `admin console` rather than the page.
Screenshot from 2021-08-31 10-43-49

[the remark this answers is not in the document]
edalex-ian: Aha! Good to describe that, because in the future if we move this to the Web then we'd most likely be using this endpoint. :wink:

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/wizard/WizardBasicControl.scala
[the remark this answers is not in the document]
PenghaiZhang: Actually, `include` is also used in admin console - whether the control is ticked or not.

Screenshot from 2021-08-31 11-20-39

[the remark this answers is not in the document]
PenghaiZhang: But even though the control in not selected in Admin Console, it is still displayed in the page.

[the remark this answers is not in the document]
edalex-ian: Yeah, I played with that ticket too but couldn't get it to do anything useful. :thinking: :man_shrugging:

edalex-ian (approved):
LGTM! Merge on! :tada:

PR #3372, opened 2021-09-03
bugfix: do not update SearchResult state when it is unmounted

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

Due to the support for DRM, we need to update state to show or hide the DRM
dialog. However, when a non-DRM item is clicked, the page just navigates to
Summary page, which unmounts the component. In this case, we should skip the
state update.


edalex-ian (approved):
LGTM! :+1:

PR #3393, opened 2021-09-14
feat: don't show Collection Selector in Advanced search page

PenghaiZhang:
##### Checklist

- [x] tests are included

##### Description of change

This PR adds the feature of not showing Collection Selector and Classification panel in Advanced search mode. 

Also the REST API call to list Classifications will not be executed. Should not need to list Collections as well, but due to the existing implementation this will require more changes.

https://user-images.githubusercontent.com/47203811/133177963-7685b440-e692-44eb-8707-ed476a0a1ae7.mp4

### DIFF
--- a/react-front-end/__tests__/tsrc/search/SearchPageTestHelper.tsx
+++ b/react-front-end/__tests__/tsrc/search/SearchPageTestHelper.tsx
@@ -325,3 +325,25 @@ export const queryCollectionSelector = (container: HTMLElement) =>
  */
 export const queryStatusSelector = (container: HTMLElement) =>
   queryRefineSearchComponent(container, "StatusSelector");
+
+/**
+ * Helper function to assist in finding the Classification panel.
+ * @param container The root container where <ClassificationsPanel/> exists
+ */
+export const queryClassificationPanel = (
+  container: Element
+): HTMLDivElement | null =>
+  container.querySelector<HTMLDivElement>("#classification-panel");
+
+/**
+ * Similar to queryClassificationPanel but throws an error if ClassificationsPanel is not found.
+ * @param container The root container where <ClassificationsPanel/> exists
+ */
+export const getClassificationPanel = (container: Element): HTMLDivElement => {
+  const classificationPanel = queryClassificationPanel(container);
+  if (!classificationPanel) {
+    throw new Error("Unable to find Classification panel");
+  }
+
+  return classificationPanel;
+};
--- a/react-front-end/tsrc/search/SearchPage.tsx
+++ b/react-front-end/tsrc/search/SearchPage.tsx
@@ -157,6 +157,7 @@ const SearchPage = ({ updateTemplate, advancedSearchId }: SearchPageProps) => {
   // If an Advanced Search ID has been provided, use that otherwise check to see if one
   // was passed in by the Router.
   advancedSearchId = advancedSearchId ?? advancedSearchIdParam;
+  const isInAdvancedSearchMode = advancedSearchId !== undefined;
 
   const [state, dispatch] = useReducer(reducer, { status: "initialising" });
   const defaultSearchPageHistory: SearchPageHistoryState = {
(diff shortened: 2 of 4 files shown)


react-front-end/__tests__/tsrc/search/SearchPageTestHelper.tsx
```diff
@@ -325,3 +325,25 @@ export const queryCollectionSelector = (container: HTMLElement) =>
  */
 export const queryStatusSelector = (container: HTMLElement) =>
   queryRefineSearchComponent(container, "StatusSelector");
+
+/**
+ * Helper function to assist in finding the Classification panel.
+ * @param container The root container where <ClassificationsPanel/> exists
+ */
+export const queryClassificationPanel = (
```
edalex-ian, line 333: Wow! I'm amazed that we don't already have these functions. It'd appear we don't currently have a test to cover the display of the Classifications panel on the Search page under normal usage. :disappointed:

react-front-end/tsrc/search/SearchPage.tsx
```diff
@@ -338,26 +339,27 @@ const SearchPage = ({ updateTemplate, advancedSearchId }: SearchPageProps) => {
...
-              return listClassifications;
-            default:
-              throw new TypeError(
-                "Unexpected `displayMode` for determining classifications listing function"
-              );
-          }
-        }
-      );
+      ) => Promise<Classification[]> = isInAdvancedSearchMode
+        ? emptyClassifications
```
edalex-ian, line 348: Hiding this logic down in the get function could lead to confusion. I think we should make it explicit in the below calls. That may involve rather than using the simple `Promise.all` the code need to make one explicit call and then (if `advancedSearchId` is not set) make the second one. (Probably best done async/await style.)

edalex-ian (changes requested):
I think we need a few changes here:

1. Let's not add a new `isInAdvancedSearchMode` - just stick with the current check of the `advancedSearchId` as is already in place;
2. Make the skipping of retrieving classifications more explicit (see comment); and
3. Modify `<SidePanel>` such that `classificationsPanelProps` is optional and when not present all logic related to the classifications is skipped - which probably just means changing on line 47 `classificationsPanelProps.classifications.some` to `classificationsPanelProps?.classifications.some`. Then use `advancedSearchId` to optionally set that value of SidePanel in SearchPage.

react-front-end/tsrc/search/SearchPage.tsx
```diff
@@ -707,7 +716,7 @@ const SearchPage = ({ updateTemplate, advancedSearchId }: SearchPageProps) => {
           value={searchPageOptions.collections}
         />
       ),
-      disabled: false,
+      disabled: isInAdvancedSearchMode,
```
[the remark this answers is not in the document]
PenghaiZhang: oh dear I forgot that Collection Selector is hidden in this case :face_exhaling:

react-front-end/tsrc/search/SearchPage.tsx
```diff
@@ -824,11 +842,15 @@ const SearchPage = ({ updateTemplate, advancedSearchId }: SearchPageProps) => {
...
-        }}
+        classificationsPanelProps={
+          advancedSearchId !== undefined
+            ? undefined
+            : {
+                classifications: getClassifications(),
+                onSelectedCategoriesChange: handleSelectedCategoriesChange,
+                selectedCategories: searchPageOptions.selectedCategories,
+              }
+        }
```
edalex-ian, line 853: How about:

```suggestion
        classificationsPanelProps={
          // When in advanced search mode, hide classifications panel
          !!advancedSearchId
            ? {
                  classifications: getClassifications(),
                  onSelectedCategoriesChange: handleSelectedCategoriesChange,
                  selectedCategories: searchPageOptions.selectedCategories,
                }
              : undefined
        }
```

react-front-end/tsrc/search/SearchPage.tsx
```diff
@@ -707,7 +725,7 @@ const SearchPage = ({ updateTemplate, advancedSearchId }: SearchPageProps) => {
           value={searchPageOptions.collections}
         />
       ),
-      disabled: false,
+      disabled: advancedSearchId !== undefined,
```
edalex-ian, line 728: How about:

```suggestion
      disabled: !!advancedSearchId,
```

react-front-end/tsrc/search/SearchPage.tsx
```diff
@@ -360,33 +360,51 @@ const SearchPage = ({ updateTemplate, advancedSearchId }: SearchPageProps) => {
...
-        )
-        .catch(searchPageErrorHandler);
+      (async () => {
+        try {
+          const searchResult: SearchPageSearchResult = await doSearch(
+            state.options
+          );
+          // Do not list classifications in Advanced search mode.
+          const classifications: Classification[] =
+            advancedSearchId === undefined
```
edalex-ian, line 370: Based on the comments below, I wonder if for consistency this should be:

```suggestion
            !!advancedSearchId
```

But up to you, as I also don't really mind this way. I'm just mentioning it.

react-front-end/tsrc/search/components/SidePanel.tsx
```diff
@@ -30,7 +30,7 @@ interface SidePanelProps {
   /**
    * Props passed to Classifications Panel
    */
-  classificationsPanelProps: CategorySelectorProps;
+  classificationsPanelProps?: CategorySelectorProps;
```
edalex-ian, line 33: Maybe tweak the comment to say, "Props passed to Classifications Panel, if undefined the panel is not displayed"

edalex-ian (commented):
Thanks for the updates @PenghaiZhang . I think that really helps readability and comprehension.

I do however just a have a few further suggestions on the logic checks based on what is already in place though. (And one which is really optional though.)

react-front-end/tsrc/search/SearchPage.tsx
PenghaiZhang: I personally like `!!`, but I remember there were discussion around the confusion caused by `!!`. Excellent I will go with `!!`

react-front-end/tsrc/search/SearchPage.tsx
PenghaiZhang: I think here should be single exclamation mark ?

react-front-end/tsrc/search/SearchPage.tsx
PenghaiZhang: I think here should be single `!`.

react-front-end/__tests__/tsrc/search/SearchPageTestHelper.tsx
PenghaiZhang: Do you want me to add some tests in this PR ?

edalex-ian: Nah, wasn't in the scope. I was just surprised. No action required.

edalex-ian (approved):
GTG @PenghaiZhang - well, other than it seems your other PR you merged conflicts. Merge when sorted. :wink:

PR #3399, opened 2021-09-14
chore(front-end): facilitate leave in debug logging

edalex-ian:
##### Checklist

- [ ] tests are included

##### Description of change

As the front-end is about to become even more complex would be good to be able to add some basic logging which we can leave in place.

Maybe in the future we can build some common functions, but I don't think we need to bring in one of the various JS logging libraries at this point. Just keeping it simple and using what the browser gives us.

As we start doing this more, then we'll see what kind of pattern we really want to support, and add a group of functions to match.

### DIFF
--- a/react-front-end/tsrc/mainui/index.tsx
+++ b/react-front-end/tsrc/mainui/index.tsx
@@ -27,6 +27,17 @@ export type EntryPage = "mainDiv" | "searchPage" | "settingsPage";
 const App = React.lazy(() => import("./App"));
 
 export default function main(entry: EntryPage) {
+  if (process.env.NODE_ENV === "production") {
+    const nop = () => {};
+    console = {
+      ...console,
+      debug: nop,
+      log: nop,
+      time: nop,
+      timeEnd: nop,
+    };
+  }
+
   initStrings();
   ReactDOM.render(
     <React.Suspense fallback={<>loading</>}>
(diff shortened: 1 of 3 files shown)


react-front-end/tsrc/mainui/index.tsx
```diff
@@ -27,6 +27,17 @@ export type EntryPage = "mainDiv" | "searchPage" | "settingsPage";
...
+  if (process.env.NODE_ENV === "production") {
+    const nop = () => {};
+    console = {
+      ...console,
+      debug: nop,
+      log: nop,
+      time: nop,
+      timeEnd: nop,
+    };
+  }
```
edalex-ian, line 39: This is arguably the real part of this PR. The calls later are just to get some bits in there and make sure all is well.

edalex-ian:
The two failed GHA jobs are just from flaky tests - MyResourcesTest and AdvancedSearchPageTest

react-front-end/tsrc/mainui/index.tsx
PenghaiZhang: Do you want to disable the eslint rule `no-global-assign` here ?

PenghaiZhang (approved):

react-front-end/tsrc/mainui/index.tsx
edalex-ian: Oh, that's odd. I thought I ran the checks... Oh hold on, that might have been for the visibility sensor.

Yup, will add that for sure. Thanks. :+1:

edalex-ian:
All sorted, thanks @PenghaiZhang . Will merge on green. :wink:

PR #3412, opened 2021-09-16
Feature/retrieve advanced search definition

PenghaiZhang:
##### Checklist

- [x] tests are included

##### Description of change

This PR adds the support for retrieving Advanced search definition in Search page and passing the Wizard controls to component `AdvancedSearchPanel`. 

In Advanced search mode, the retrieval is started when Search page is in initialising state. Once that is done, update `SearchPageOptions` to use the Collections configured for Advanced search.

Screenshot from 2021-09-16 14-32-19


edalex-ian:
The new tests for the Advanced Search Panel displaying logic are failing @PenghaiZhang , and seeing they're pretty critical to this I better wait until you fix to review. :wink:

react-front-end/tsrc/search/SearchPage.tsx
```diff
@@ -197,6 +200,9 @@ const SearchPage = ({ updateTemplate, advancedSearchId }: SearchPageProps) => {
   const [advancedSearches, setAdvancedSearches] = useState<
     OEQ.Common.BaseEntitySummary[]
   >([]);
+  const [wizardDefinition, setWizardDefinition] = useState<
```
edalex-ian, line 203: Just a quick one while you fix the tests though... Would make sense to go with a more specific name here:

```suggestion
  const [advancedSearchDefinition, setAdvancedSearchDefinition] = useState<
```

Mind you though, we've quickly accumulated a bit of state relating to advanced search mode, should we merge it - maybe even another reducer I wonder.

Bits that come to mind:

- this definition
- show/hide the panel
- maybe the id
- and soon the values or query...

hmmm, what do you think?

oeq-ts-rest-api/src/index.ts
```diff
@@ -35,3 +35,4 @@ export * as Security from './Security';
 export * as Settings from './Settings';
 export * as UserQuery from './UserQuery';
 export * as Utils from './Utils';
+export * as WizardControl from './wizard/WizardControl';
```
edalex-ian, line 38: Also, just another quick one, I missed in the code review that there was a new subfolder created for the wizard types. You'll note there are no others, and it is just using a flat directory structure at this point. Any chance we could just move the two files to the top level?

If we're lucky, it may be fine to keep this whole module flat. You'll also note there are already files in there which are purely shape definitions - Common and Acl.

PenghaiZhang: Yeah I will update the structure.

react-front-end/tsrc/search/SearchPage.tsx
PenghaiZhang: > The new tests for the Advanced Search Panel displaying logic are failing @PenghaiZhang , and seeing they're pretty critical to this I better wait until you fix to review.

Ah I forgot to mock the new function. No problems. Fixed.

PenghaiZhang: At the moment we have two. How about we merge them into a object first ?

edalex-ian: Sounds good, we can start there and see how it goes when we add the others.

But keeping in mind that one of the key purposes of a reducer is to manage multiple pieces of related state. And where this was already jumping out at me was when you were checking for both show advanced search panel and whether there was a wizard definition.

PenghaiZhang: I have implemented a reducer for this.

react-front-end/__tests__/tsrc/modules/AdvancedSearchModule.test.ts
```diff
@@ -0,0 +1,76 @@
...
+describe("getAdvancedSearchByUuid", () => {
+  it("retrieves Advanced search definition by UUID", async () => {
+    const uuid = "e82207be-a9f2-442a-a17f-5c834d5b36cc";
+    const definition = await getAdvancedSearchByUuid(uuid);
+
+    expect(OEQ.AdvancedSearch.getAdvancedSearchByUuid).toHaveBeenCalledTimes(1);
+    expect(mockGetAdvancedSearchByUuid.mock.calls[0][1]).toBe(uuid);
+    expect(definition.controls).toHaveLength(2);
+  });
+});
```
edalex-ian, line 59: Hmmmm, is this actually testing anything? Both those functions really have no logic, just wiring to call the REST module and then you've mocked that out. So unless I'm missing something, you're just testing that you've mocked the functions correctly. :thinking:

react-front-end/tsrc/search/SearchPage.tsx
```diff
@@ -230,6 +238,21 @@ const SearchPage = ({ updateTemplate, advancedSearchId }: SearchPageProps) => {
     [dispatch]
   );
 
+  const showAdvSearchPanel = useCallback(
+    (showPanel: boolean) => {
+      if (searchPageModeState.useMode !== "advSearch") {
+        throw new Error("Control Advanced search panel in wrong state");
```
edalex-ian, line 244:
```suggestion
        throw new Error("Request to toggle Advanced Search Panel when _not_ in Advanced Search mode. Request ignored.");
```

react-front-end/tsrc/search/SearchPage.tsx
```diff
@@ -230,6 +238,21 @@ const SearchPage = ({ updateTemplate, advancedSearchId }: SearchPageProps) => {
...
+      }
+
+      searchPageModeDispatch({
+        type: "useAdvSearch",
+        selectedAdvSearch: searchPageModeState.definition,
+        showPanel,
+      });
```
edalex-ian, line 251: As discussed, rather than simply using the action of `useAdvSearch` to even change the state of the panel display, how about adding three more actions to cover the various usages: `toggleAdvSearchPanel`, `hideAdvSearchPanel` and `showAdvSearchPanel`. In the reducer, they could then use the previous `state` to just tweak the `showPanel` value as required.

Then you'd update the various callbacks below and they'd be nice and clean like `searchPageModeDispatch({type: "toggleAdvSearchPanel"})`.

react-front-end/tsrc/search/SearchPageModeReducer.ts
```diff
@@ -0,0 +1,51 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import * as OEQ from "@openequella/rest-api-client";
+
+export type State =
+  | {
+      useMode: "normal";
```
edalex-ian, line 22: Nit: How about just `mode`?

oeq-ts-rest-api/src/index.ts
```diff
@@ -35,3 +35,4 @@ export * as Security from './Security';
 export * as Settings from './Settings';
 export * as UserQuery from './UserQuery';
 export * as Utils from './Utils';
+export * as WizardControl from './WizardControl';
```
edalex-ian, line 38: While here, could you also:

```suggestion
export * as WizardCommonTypes from './WizardCommonTypes';
export * as WizardControl from './WizardControl';
```

As I've needed to do that in my branch, so can I assure you it is needed. :wink:

edalex-ian (changes requested):
Looks great @PenghaiZhang . Love the move to the reducer, has really tidied things up.

Just two main things:

1. Are those tests I've highlighted actually needed/doing anything? and
2. Let's add a bit more to the reducer as discussed.

react-front-end/__tests__/tsrc/modules/AdvancedSearchModule.test.ts
PenghaiZhang: I think we have similar tests which are meant to ensure the front end talks to rest module properly. :thinking:

PenghaiZhang: I have tests these redundant tests.

react-front-end/tsrc/search/SearchPage.tsx
```diff
@@ -971,19 +956,21 @@ const SearchPage = ({ updateTemplate, advancedSearchId }: SearchPageProps) => {
...
+                  searchPageModeState.mode === "advSearch"
                     ? {
                         onClick: () =>
-                          showAdvSearchPanel(
-                            !searchPageModeState.isAdvSearchPanelOpen
-                          ),
+                          searchPageModeDispatch({
+                            type: "toggleAdvSearchPanel",
+                            showPanel:
+                              !searchPageModeState.isAdvSearchPanelOpen,
```
edalex-ian, line 965: This should not be required, just keep the logic (even though it's simple) in the reducer - see other comment.

react-front-end/tsrc/search/SearchPageModeReducer.ts
```diff
@@ -32,20 +32,45 @@ export type Action =
...
+  action: "toggle" | "hide",
+  showPanel: boolean
+) => {
+  if (state.mode !== "advSearch") {
+    throw new Error(
+      `Request to ${action} Advanced Search Panel when _not_ in Advanced Search mode. Request ignored.`
+    );
+  }
+  return { ...state, isAdvSearchPanelOpen: showPanel };
+};
```
edalex-ian, line 58: This function should really just have all the logic based on the `action` - and then it better aligns with the name. For example:

```
const toggleOrHidePanel = (
  state: State,
  action: "toggle" | "hide"
) => {
  if (state.mode !== "advSearch") {
    throw new Error(
      `Request to ${action} Advanced Search Panel when _not_ in Advanced Search mode. Request ignored.`
    );
  }
  return { ...state, isAdvSearchPanelOpen: (action === "toggle" ? !state.isAdvSearchPanelOpen : false) };
};
```

Notes:

- I've dropped the third arg; and
- Used the `action` to determine what `isAdvSearchPanelOpen` should be.

react-front-end/tsrc/search/SearchPageModeReducer.ts
```diff
@@ -32,20 +32,45 @@ export type Action =
       type: "useNormal";
     }
   | {
-      type: "useAdvSearch";
+      type: "showAdvSearchPanel";
       selectedAdvSearch: OEQ.AdvancedSearch.AdvancedSearchDefinition;
+    }
+  | {
+      type: "toggleAdvSearchPanel";
       showPanel: boolean;
```
edalex-ian, line 40: Not required - the action is `toggleAdvSearchPanel`, so the reducer can do that logic based on previous state.

edalex-ian (changes requested):
Almost there @PenghaiZhang , just need one more tweak to the reducer to allow it to look after the logic of toggling etc.

react-front-end/tsrc/search/SearchPageModeReducer.ts
PenghaiZhang: Good idea!

edalex-ian (approved):
GTG! :tada:

PR #3420, opened 2021-09-21
Feature/advanced search form (part 1)

edalex-ian:
##### Checklist

- [x] tests are included

##### Description of change

The is the first part / foundation of the Advanced Search Form. It provides:

- 3 new components (`WizardEditBox`, `WizardLabel` and `WizardUnsupported`)
- `WizardHelper` with the logic for rendering a wizard definition
- Storybook stories for the new components
- Unit tests for the `WizardHelper`

Next step/part is to then integrate this with the new `AdvancedSearchForm`. And following on from that, is to implement and support all the wizard control types (and then visibility scripting).

I've not included screenshots, as I've already walked @PenghaiZhang through the new components.

And, I expect this will be further tweaked as we get into the integration and later and more components. But hopefully it provides a reasonable foundation for us.

### DIFF
--- a/react-front-end/tsrc/components/wizard/WizardEditBox.tsx
+++ b/react-front-end/tsrc/components/wizard/WizardEditBox.tsx
@@ -1,0 +1,83 @@
+/*
+ * Licensed to The Apereo Foundation under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * The Apereo Foundation licenses this file to you under the Apache License,
+ * Version 2.0, (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { OutlinedInput } from "@material-ui/core";
+import * as React from "react";
+import { WizardLabel } from "./WizardLabel";
+
+export interface WizardEditBoxProps {
+  /**
+   * DOM id
+   */
+  id?: string;
+  /**
+   * The label to display for the control.
+   */
+  label?: string;
+  /**
+   * A description to display alongside the control to assist users.
+   */
+  description?: string;
+  /**
+   * Indicate that this control is 'mandatory' to the user.
+   */
+  mandatory: boolean;
+  /**
+   * If greater than 1 creates a multi-line edit box with the specified number of rows.
+   */
+  rows: number;
+  /**
+   * The current value of the control.
+   */
+  value?: string;
+  /**
+   * On change handler.
+   */
+  onChange: (_: string) => void;
+}
+
+/**
+ * Basic wizard 'EditBox' control.
+ */
+export const WizardEditBox = ({
+  id,
+  label,
+  description,
+  mandatory,
+  rows,
+  value,
+  onChange,
+}: WizardEditBoxProps): JSX.Element => (
+  <>
+    <WizardLabel
+      mandatory={mandatory}
+      label={label}
+      description={description}
+      labelFor={id}
+    />
+    <OutlinedInput
+      id={id}
+      multiline={rows > 1}
+      rows={rows}
+      value={value}
+      onChange={({ target }: React.ChangeEvent<HTMLInputElement>) =>
+        onChange(target.value)
+      }
+      aria-required={mandatory}
+    />
+  </>
+);
--- a/react-front-end/tsrc/components/wizard/WizardHelper.tsx
+++ b/react-front-end/tsrc/components/wizard/WizardHelper.tsx
@@ -1,0 +1,268 @@
+/*
+ * Licensed to The Apereo Foundation under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * The Apereo Foundation licenses this file to you under the Apache License,
+ * Version 2.0, (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import * as OEQ from "@openequella/rest-api-client";
+import * as A from "fp-ts/Array";
+import * as E from "fp-ts/Either";
+import { Eq, struct } from "fp-ts/Eq";
+import { absurd, flow, pipe } from "fp-ts/function";
+import * as N from "fp-ts/number";
+import * as O from "fp-ts/Option";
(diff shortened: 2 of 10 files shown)


react-front-end/tsrc/components/wizard/WizardHelper.tsx
```diff
@@ -0,0 +1,268 @@
...
+ * @param refinement method to validate head element.
+ */
+const isHeadType =
+  <T,>(refinement: Refinement<unknown, T>) =>
+  (xs: unknown[]): boolean =>
+    pipe(
+      xs,
+      A.head,
+      O.map(refinement),
+      O.getOrElse((): boolean => false)
```
edalex-ian, line 72: Ah this would've been more concise (save function definition) and would've had given the typing grief if I had've used the fp-ts `function` util of `constFalse`. Something like:

```suggestion
      O.getOrElse(constFalse)
```

react-front-end/tsrc/components/wizard/WizardHelper.tsx
```diff
@@ -0,0 +1,268 @@
...
+  type: c.controlType,
+});
+
+/**
+ * For a control which just needs a singular value, validate and split out the value from the
+ * unionised ControlValue.
+ *
+ * @param value a potential string value
+ */
+const getStringControlValue = (value?: ControlValue): string | undefined => {
```
PenghaiZhang, line 124: Why `value` can be optional in this function? 
Is the idea similar to this `value ? getStringControlValue(value) : undefined` ?

react-front-end/tsrc/components/wizard/WizardHelper.tsx
```diff
@@ -0,0 +1,268 @@
...
+        E.fromPredicate(
+          isStringArray,
+          () => new TypeError("Expected string[] but got something else!")
+        ),
+        E.map(flow(A.head, O.toUndefined))
+      )
+    )
+  );
+
+  if (E.isLeft(potentialValue)) {
```
PenghaiZhang, line 140: Perhaps you can do a `E.foldW` after `E.map` ?

react-front-end/tsrc/components/wizard/WizardHelper.tsx
```diff
@@ -0,0 +1,268 @@
...
+    case "html":
+    case "listbox":
+    case "radiogroup":
+    case "shufflebox":
+    case "shufflelist":
+    case "termselector":
+    case "userselector":
+      return <WizardUnsupported />;
+    default:
+      return absurd(controlType);
```
PenghaiZhang, line 187: My understanding of `absurd` is to use it to handle something impossible like `never`. But, if something is impossible then why it needs handling.. ?

PenghaiZhang (commented):
Thanks Ian for this good foundation ! 

I have some queries.

react-front-end/tsrc/components/wizard/WizardEditBox.tsx
```diff
@@ -0,0 +1,83 @@
...
+    />
+    <OutlinedInput
+      id={id}
+      multiline={rows > 1}
+      rows={rows}
+      value={value}
+      onChange={({ target }: React.ChangeEvent<HTMLInputElement>) =>
+        onChange(target.value)
+      }
+      aria-required={mandatory}
```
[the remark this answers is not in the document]
edalex-ian: You raise a good point, but... Gah! After adding support for mandatory, and although you _can_ configure it in Advanced Search wizards, the legacy UI doesn't appear to do anything more than a visual indicator. :disappointed: 


https://user-images.githubusercontent.com/43919233/134270085-87d52c70-5f67-41e1-8920-6f9aab9bd7bb.mp4

So I'm going to leave it for now and we can revisit when we need to support that.

react-front-end/tsrc/components/wizard/WizardLabel.tsx
```diff
@@ -0,0 +1,62 @@
...
+  label,
+  description,
+  mandatory = false,
+  labelFor,
+}: WizardLabelProps): JSX.Element => (
+  <label htmlFor={labelFor}>
+    {label && (
+      <Typography variant="h6" gutterBottom>
+        {label}
+        {mandatory ? " *" : ""}
```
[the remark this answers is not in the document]
edalex-ian: Yes, but different components handle it different ways and not strictly inline with Material Design. This gives us full control - allowing us to reach consistency (especially when we do some of our custom controls - shuffle boxes, date selectors, etc).

react-front-end/tsrc/components/wizard/WizardHelper.tsx
```diff
@@ -0,0 +1,268 @@
...
+/**
+ * Identifies a Wizard 'field' and specifies its value.
+ */
+export interface FieldValue {
+  target: ControlTarget;
+  value: ControlValue;
+}
+
+/**
+ * Check the type of the head of an array. Returns `false` if empty.
```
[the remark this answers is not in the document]
edalex-ian: Oh yes, good pick-up. I'll rework this!

react-front-end/tsrc/components/wizard/WizardHelper.tsx
```diff
@@ -0,0 +1,268 @@
...
+      O.map(refinement),
+      O.getOrElse((): boolean => false)
+    );
+
+/**
+ * Used to check if the `ControlValue` is of the string[] variety.
+ * (Not a general purpose array util!)
+ */
+const isStringArray = (xs: ControlValue): xs is string[] =>
+  pipe(xs as unknown[], isHeadType<string>(S.isString));
```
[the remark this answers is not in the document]
edalex-ian: Yes this is needed. It serves two purposes:

1. It makes an `Array` available to the fp-ts `Array` functions used (in what is now `isHeadType`); and
2. I kind of like it in the end as it highlights we don't really know which incarnation of the ControlValue union this is.

react-front-end/tsrc/components/wizard/WizardHelper.tsx
edalex-ian: Are you saying you can `throw` from within an `E.foldW`?

react-front-end/tsrc/components/wizard/WizardHelper.tsx
edalex-ian: Yup, it is slightly odd now you call it out. But it is encapsulating the logic you mention there, as this function was lifted out of the initial context of where it was used.

Hmmmm, I'm somewhat inclined to leave it to keep where it will be used DRY, but I might be able to come up with something better. I'll have a play. :wink:

react-front-end/tsrc/components/wizard/WizardHelper.tsx
edalex-ian: This forces the switch statement to be an exhaustive switch statement - thereby triggering a compile time error (rather than runtime) if we're missing a possible case.

An example of when this might happen, say we suddenly go and add another `controlType` but forget to come and add it here. When we go to build, a compile time error will trigger to highlight we need to come and fix this. (Like exhaustive pattern matching in Scala.)

Make sense?

react-front-end/tsrc/components/wizard/WizardHelper.tsx
edalex-ian: Ah! So it does. Handy. :wink:

react-front-end/tsrc/components/wizard/WizardHelper.tsx
edalex-ian: I've come up with something. Obviously I could've just done with the ternary again and again as we add all the controls, but I've codified it instead.

edalex-ian:
Thanks @PenghaiZhang . All feedback addressed.

PenghaiZhang (approved):
LGTM!

edalex-ian:
I'll merge now, as all the key working bits of CI are green.

PR #3429, opened 2021-09-30
Feature/wizard check box group

PenghaiZhang:
##### Checklist

- [x] tests are included

##### Description of change

This PR adds a new component for the Wizard CheckBox Group. Thanks to the foundation work done previously, it wasn't hard to create the component. Maybe a couple things worth to note: 

1. I did not use `MUI Grid` because it seems not quite fit to this case. For example, if 8 options are displayed in 2 rows and each row can display up to 5 options, then the each grid item in the first row should take 20% width, but the question is how to represent 20% width by the number of grids (from 1 to 12) ? 

2.The `values` passed to this component should be `currentValues` controlled by  the state of `AdvancedSearchPanel`, nor the one provided by `SearchPage`. If we don't do this, the result is only the lastly selected option will be added to `currentValues`.

For the test, I have made some significant changes due to the labels of a CheckBox Group. 

1. Update how to trigger an event because the control's label (title) is not so important and what matters is its options' labels. 

2. Update how to check whether values are saved once search is triggered.

3. I also dropped some fp-ts code, particularly around the `Map` which subsequently introduced `Ord`. I feel it made the test much more complicated. 

https://user-images.githubusercontent.com/47203811/135372522-f9c0ace7-b083-45bd-85be-346d2eb78cd6.mp4

### DIFF
--- a/react-front-end/tsrc/components/wizard/WizardHelper.tsx
+++ b/react-front-end/tsrc/components/wizard/WizardHelper.tsx
@@ -26,6 +26,7 @@ import * as O from "fp-ts/Option";
 import { Refinement } from "fp-ts/Refinement";
 import * as S from "fp-ts/string";
 import * as React from "react";
+import { WizardCheckBoxGroup } from "./WizardCheckBoxGroup";
 import { WizardEditBox } from "./WizardEditBox";
 import { WizardUnsupported } from "./WizardUnsupported";
 
@@ -83,6 +84,12 @@ const isHeadType =
 const isStringArray = (xs: ControlValue): xs is NEA.NonEmptyArray<string> =>
   pipe(xs as unknown[], isHeadType<string>(S.isString));
 
+/**
+ * Typically used to check if the `ControlValue` of an Option type control (e.g. CheckBox Group) is a non-empty array.
+ * If you also want to confirm if the value is `string`, use `isStringArray`.
+ */
+const isControlValueNonEmpty = (xs: ControlValue): boolean => xs.length > 0;
+
 const eqControlTarget: Eq<ControlTarget> = struct({
   schemaNode: A.getEq(S.Eq),
   type: S.Eq,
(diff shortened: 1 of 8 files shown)


react-front-end/tsrc/components/wizard/WizardOptionGroup.tsx
```diff
@@ -0,0 +1,103 @@
...
+  return (
+    <>
+      <WizardLabel
+        mandatory={mandatory}
+        label={label}
+        description={description}
+      />
+      {range(0, rowNumber).map((rowIndex) => (
+        <div className={classes.optionRow} key={`${id}-${rowIndex}`}>
+          {getOptionsForRow(rowIndex).map((option, optionIndex) => (
```
edalex-ian, line 91: How about using lodash's [chunk](https://lodash.com/docs/4.17.15#chunk)? You can then remove the `getOptionsForRow` function and have:

```suggestion
      {_.chunk(options, columnNumber).map((row, rowIndex) => (
        <div className={classes.optionRow} key={`${id}-${rowIndex}`}>
          row.map((option, optionIndex) => (
```

Aside: This came about from reviewing the complexity in `getOptionsForRow`. The call to `filter` would unnecessarily traverse the whole array for each row, and I was trying to find the JS equivalent of what you see in some other modern languages where you can use the subscript operator (`[]`) to specify a range of values to grab. However, really you'd probably just have to use `Array.slice`. But in the end `_.chunk` is really exactly what you're doing here - chunking through and processing.

react-front-end/tsrc/components/wizard/WizardHelper.tsx
```diff
@@ -150,25 +188,46 @@ const controlFactory = (
   const ifAvailable = <T,>(
     value: ControlValue | undefined,
     getter: (_: ControlValue) => T
-  ): T | undefined => pipe(value, O.fromNullable, O.map(getter), O.toUndefined);
+  ): T | undefined =>
+    pipe(
+      value,
+      O.fromNullable,
+      O.filter(isControlValueNonEmpty),
```
edalex-ian, line 195: It's a shame that `ControlValue` is slightly awkward being a union of two array types, as otherwise we could simplify this whole thing with just `NEA.fromArray`. But alas, `ControlValue` is what it is - just hopefully we need that complexity.

edalex-ian (changes requested):
Love where you've taken this @PenghaiZhang :+1: 

Few bits of minor feedback, plus although I love your move simplify stuff and remove fp-ts Map, Ord etc, I think you may need to bring some of it back - see comment.

react-front-end/tsrc/components/wizard/WizardOptionGroup.tsx
PenghaiZhang: `chunk` is really a good idea! I did not know lodash provides this function. Thank you!!

react-front-end/tsrc/components/wizard/WizardHelper.tsx
PenghaiZhang: Yeah this is indeed awkward. Let's see if we really need to support `number[]` later.

PR #3555, opened 2021-11-03
refactor: update how to build the default value of Calendar

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

The testing work found that if the default value of Calendar is configured as `today` or `today plus xx days`, the default value is not displayed properly in the advanced search panel.  The reason is when using `today`, the value is actually a number which represents how many days to be added to today.

So the fix is to check if the value is numeric, if yes then plus it to the day of today, otherwise use the value.

Please note that the Date Range Selector will show a warning if the default day is a future day.

image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/wizard/WizardBasicControl.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/wizard/WizardBasicControl.scala
@@ -25,6 +25,8 @@ import com.tle.common.i18n.LangUtils
 import com.tle.common.taxonomy.SelectionRestriction
 import com.tle.common.taxonomy.wizard.TermSelectorControl.TermStorageFormat
 import com.tle.web.api.language.LanguageStringHelper.getStringFromCurrentLocale
+import java.time.LocalDate
+import java.time.temporal.ChronoUnit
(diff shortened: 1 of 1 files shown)


Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/wizard/WizardBasicControl.scala
```diff
@@ -91,11 +93,21 @@ case class WizardBasicControl(
 
 object WizardBasicControl {
   def getDefaultValues(options: List[WizardControlItem], controlType: String): List[String] = {
+    def processValue(item: WizardControlItem): String = {
+      val value = item.getValue
+      controlType match {
+        case Calendar.CLASS if value.matches("^[0-9]+$") =>
```
edalex-ian, line 99: Rather than pulling in a regex, how about something like `Try(s.toInt).toOption` and checking for `Some`. That way you have symmetry in validation and then later use.

edalex-ian (approved):
Looks good @PenghaiZhang , but just one thing to consider and then feel free to merge. :wink:

edalex-ian (approved):
Nice refactor with Try there @PenghaiZhang . :+1:

Issue #3558, opened 2021-11-03
Items with EQUELLA resource attachments that have been set to latest live version breaks the new Search U/I.


cathfitz:
Tested in 2021.1.1, found by a client, but could affect other institutions.

**Overview:**

If you have an institution with multiple collections, one of which has an attachment control to allow the selection of a second collection's resources as an attachment.  If you have your integration settings that allow the user to select the latest live version (0) and attach that, then when you perform a search, you'll get a "Request failed with code 500" message and no search results.

**Example:**

- I've contributed one resource into the "Remote Items Collection".
- I've contributed two resources into the "Collection with Remote Items".
- Both link to the single item in the "Remote Items Collection":

Item with just a local attachment and a EQUELLA Resource Attachment fixed at version 2 

Item with just a local attachment and a EQUELLA Resource Attachment fixed at version 0 (latest live version). Archive this item and the error goes away. 

1. Login as sysadmin/equella
2. Perform a regular (basic) search (no collections selected).  You'll see the error.
3. Refine the search to just the "Remote Items Collection" and you'll see one resource.
4. Refine the search to just the "Collection with Remote Items" and you'll see the error.

Go to Manage Resources and archive the item:

Item with just a local attachment and a EQUELLA Resource Attachment fixed at version 0 (latest live version). 

Then go back and perform a basic search.  You'll see two resources.  Refining by each collection also returns results.

The stack trace says:

```
2021-11-03 12:03:31,140 context="00360c14-6509-4fd0-8f85-dde60684999d" level="ERROR" category="RestEasyExceptionMapper" message="REST API error"
java.lang.Error: Real version numbers are used in ItemService
        at com.tle.core.item.service.impl.ItemServiceImpl.ensureItemKey(ItemServiceImpl.java:253)
        at com.tle.core.item.service.impl.ItemServiceImpl.getUnsecureIfExists(ItemServiceImpl.java:270)
        at com.tle.web.api.search.SearchHelper$.getBrokenAttachmentStatusForResourceAttachment(SearchHelper.scala:340)
        at com.tle.web.api.search.SearchHelper$.recurseBrokenAttachmentCheck(SearchHelper.scala:361)
        at com.tle.web.api.search.SearchHelper$.$anonfun$convertToAttachment$3(SearchHelper.scala:284)
        at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:286)
[...]

PR #3563, opened 2021-11-04
bugfix: use real version to check if custom attachment is broken

PenghaiZhang:
#3558 

##### Checklist

- [ ] tests are included

##### Description of change

If a Custom Attachment links to the latest version of an Item,  it uses 0 to represent the latest version. However, when we check whether the attachment is broken, we need to use the real latest version number rather than 0. 

#3558


edalex-ian (changes requested):
All good, except...

Really should have a test around this kind of thing. Any chance when this functionality was added there was more than one customAttachment and we could configure one to point to 'latest'/version 0?

PenghaiZhang:
> All good, except...
> 
> Really should have a test around this kind of thing. Any chance when this functionality was added there was more than one customAttachment and we could configure one to point to 'latest'/version 0?

I have updated the testing data. Without the fix, new search page shows 500 error. And with the fix it works fine. :smile:

edalex-ian (approved):
LGTM! :wink:

PR #3575, opened 2021-11-10
Component/add clear button to advanced search

edalex-yinzi:
##### Checklist

- [x] tests are included

##### Description of change

Add clear button in advanced search.
Click Clear button will clear filters and update search results.

Also, Ian fix a bug in `getOptionValues`.

https://user-images.githubusercontent.com/92769668/141209281-12ee23fe-8375-40ee-8f81-2a97a5613ab1.mp4


edalex-ian:
That test failure is all fine, just on a known flakey test - `EquellaConnectorTest`.

react-front-end/__tests__/tsrc/search/AdvancedSearch.test.tsx
```diff
@@ -285,11 +300,67 @@ describe("Rendering of wizard", () => {
...
+          ),
+          O.fromNullable,
+          O.getOrElse(() => new Map())
+        )
+      )
+    );
+
+    const isNotEmptyMap = (m: Map<unknown, unknown>): boolean => {
+      return !M.isEmpty(m);
+    };
```
edalex-ian, line 336: Not required (see following comment). But also, be aware this could have been done as:

```suggestion
    const isNotEmptyMap = (m: Map<unknown, unknown>): boolean => !M.isEmpty(m);
```

react-front-end/__tests__/tsrc/search/AdvancedSearch.test.tsx
```diff
@@ -285,11 +300,67 @@ describe("Rendering of wizard", () => {
...
+        )
     );
 
+    // Make sure the current values match the default
     expect(defaultLabelsAndValues).toEqual(mockedLabelsAndValues);
+
+    // Clear all input value
+    await clickClearButton(container);
+
+    const afterClearLabelsAndValues: WizardControlLabelValue[] = pipe(
```
edalex-ian, line 319: Great to see you using fp-ts already. Well done.

Just something to keep in mind though, it is important to aim for consistent style. So unless you really needed to use fp-ts here, you'd really want to use the pattern used elsewhere in this function - .e.g `buildMockedControls.map()`.

And further keeping that in mind, you now have multiple implementation of the same code (more or less), so to follow the [DRY principal](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), you'd want to functionalise it and in this and the above places use the new function. Perhaps in the new function you could do it with fp-ts, but up to you.

react-front-end/__tests__/tsrc/search/AdvancedSearch.test.tsx
```diff
@@ -285,11 +300,67 @@ describe("Rendering of wizard", () => {
...
+
+    const isNotEmptyMap = (m: Map<unknown, unknown>): boolean => {
+      return !M.isEmpty(m);
+    };
+
+    // filter all empty value
+    const filteredLabelsAndValues = pipe(
+      afterClearLabelsAndValues,
+      // filter {}
+      A.filter(isNotEmptyMap),
```
edalex-ian, line 342: Only recently did I finally find how better to do these negated cases with fp-ts.

First:

```typescript
import { not } from "fp-ts/Predicate";
```

Then you can simply do this line as:

```suggestion
      A.filter(not(M.isEmpty)),
```

react-front-end/__tests__/tsrc/search/AdvancedSearch.test.tsx
```diff
@@ -285,11 +300,67 @@ describe("Rendering of wizard", () => {
...
+      // filter "" and []
+      A.filter(
+        flow(
+          M.filter((mapValue) => !A.isEmpty(mapValue as string[])),
+          isNotEmptyMap
+        )
+      ),
```
edalex-ian, line 349: With the above in mind, this becomes:

```suggestion
      A.filter(
        flow(
          M.filter(not(A.isEmpty)),
          not(M.isEmpty)
        )
      ),
```

react-front-end/__tests__/tsrc/search/AdvancedSearch.test.tsx
```diff
@@ -285,11 +300,67 @@ describe("Rendering of wizard", () => {
...
+      A.filter(
+        flow(
+          M.filter((m) =>
+            flow(
+              A.filter((arrayItem) => arrayItem !== ""),
+              A.isNonEmpty
+            )(m as string[])
+          ),
+          isNotEmptyMap
+        )
```
edalex-ian, line 360: And let's continue the theme:

```suggestion
        flow(
          M.filter(flow(
              A.filter(not(S.isEmpty)),
              A.isNonEmpty
            )
          ),
          not(M.isEmpty)
        )
```

Not also here I've dropped your type coercion (i.e. `m as string[]`). You don't want to be doing that at all, most of the type you can solve this by explicitly specifying the types for the generics. So if you take the above approach but then have type issues, look at the generic type specifiers rather than adding the coercion back in.

react-front-end/__tests__/tsrc/search/AdvancedSearch.test.tsx
```diff
@@ -285,11 +300,67 @@ describe("Rendering of wizard", () => {
...
+      )
+    );
```
edalex-ian, line 362: And now at the end of the pipe let's revisit the `not(M.isEmpty)`. Rather than doing it as the first step, and then continuing to do it in each of the intermediate steps. You could do it just once right at the end here.

react-front-end/tsrc/search/components/AdvancedSearchPanel.tsx
```diff
@@ -140,6 +158,13 @@ export const AdvancedSearchPanel = ({
         >
           {languageStrings.common.action.search}
         </Button>
+        <Button
+          id={`${idPrefix}-clearBtn`}
+          onClick={() => onClear()}
```
edalex-ian, line 163: Redundant arrow operator. Keep in mind we have first class function support, so:

```suggestion
          onClick={onClear}
```

react-front-end/tsrc/search/components/AdvancedSearchPanel.tsx
```diff
@@ -65,6 +65,7 @@ export const AdvancedSearchPanel = ({
 }: AdvancedSearchPanelProps) => {
   const [currentValues, setCurrentValues] =
     useState<WizardHelper.FieldValueMap>(values);
+  const [isTriggerClearSearch, setIsTriggerClearSearch] = useState(false);
```
edalex-ian, line 68: Not required. Only `useState` and `useEffect` if you have to - as they quickly amp up the complexity.

Instead, you'd simply want to factor out the function currently on the search button, so to something like:

```typescript
const doSearch = () => onSubmit(currentValues)
```

Then update the search button to use that, and in your `onClear` function rather than calling `setIsTriggerClearSearch()` call `doSearch()`.

react-front-end/tsrc/search/components/AdvancedSearchPanel.tsx
```diff
@@ -93,6 +94,23 @@ export const AdvancedSearchPanel = ({
...
   );
 
+  /**
+   * Handler for when user click clear button
+   * set state to an empty map
+   * update search results
+   */
+  const onClear = () => {
+    setCurrentValues(new Map());
+    setIsTriggerClearSearch(true);
```
edalex-ian, line 104: See above comment.

react-front-end/tsrc/search/components/AdvancedSearchPanel.tsx
```diff
@@ -93,6 +94,23 @@ export const AdvancedSearchPanel = ({
...
+  };
+
+  useEffect((): void => {
+    if (isTriggerClearSearch) {
+      onSubmit(currentValues);
+      setIsTriggerClearSearch(false);
+    }
+  }, [currentValues, isTriggerClearSearch, onSubmit]);
+
```
edalex-ian, line 113: As per previous comments, not required.

edalex-ian (changes requested):
Great start there @edalex-yinzi , and good to see you taking fp-ts for a spin already.

As you can see, a few bits of feedback to look at, main ones being:

- Tiding up your pipe in the test; and
- Reworking how you're trigger a search after the clear button is clicked.

react-front-end/__tests__/tsrc/search/AdvancedSearch.test.tsx
edalex-yinzi: Still got errors by explicitly specifying the types for generics. I tried to use ternary to avoid the error. Not sure whether it's a good solution? See the latest commit.

react-front-end/__tests__/tsrc/search/AdvancedSearch.test.tsx
edalex-ian: Ummm, although you marked this as resolved you don't appear to have taken any action on it. Please action.

react-front-end/__tests__/tsrc/search/AdvancedSearch.test.tsx
```diff
@@ -285,11 +301,67 @@ describe("Rendering of wizard", () => {
...
+        flow(
+          M.filter((m) =>
+            typeof m !== "string"
+              ? flow(
+                  A.filter((arrayItem) => arrayItem !== ""),
+                  A.isNonEmpty
+                )(m)
+              : false
+          ),
+          not(M.isEmpty)
```
edalex-ian, line 360: As mentioned, rather than repeating yourself, just do this once as a final pass - i.e. Add another `A.filter` which is just doing a `not(M.isEmpty)`.

react-front-end/__tests__/tsrc/search/AdvancedSearch.test.tsx
```diff
@@ -285,11 +301,67 @@ describe("Rendering of wizard", () => {
...
+    const filteredLabelsAndValues = pipe(
+      afterClearLabelsAndValues,
+      // filter "" and []
+      A.filter(
+        flow(
+          M.filter((mapValue) =>
+            typeof mapValue === "string"
+              ? mapValue !== ""
+              : not(A.isEmpty)(mapValue)
+          ),
```
edalex-ian, line 345: You'll probably find that you have a look at our `pfTernary` and then either `isStringArray` or `S.isString` you'll be able to improve the readability of this. (e.g. `M.filter(pfTernary(S.isString,not(S.isEmpty),not(A.isEmpty))` - but your YMMV depending on the types etc - and hence why `isStringArray` may be a better fit.

react-front-end/tsrc/search/components/AdvancedSearchPanel.tsx
```diff
@@ -93,6 +93,24 @@ export const AdvancedSearchPanel = ({
     [currentValues, setCurrentValues]
   );
 
+  /**
+   * do search based on input value
+   *
+   * @param value WizardHelper.FieldValueMap, default value is `currentValues`
+   */
+  const doSearch = (value = currentValues) => onSubmit(currentValues);
```
edalex-ian, line 101: Do you really need the default value here? Ideally we want `doSearch` to only ever use what's in state. :thinking:

edalex-ian (changes requested):
Ah okay, I see the union type of `string | string[]` may be causing your havoc, but see what you can do with the suggestions.

But you're almost there. Grab me tomorrow if there's time.

react-front-end/__tests__/tsrc/search/AdvancedSearch.test.tsx
edalex-yinzi: Aha,  I didn't notice the duplicated code in the previous test. But now I know what you mean.

react-front-end/__tests__/tsrc/search/AdvancedSearch.test.tsx
edalex-yinzi: Since there is a filter wrapper around this, I need to change the logic of this function to achieve this improvement.

react-front-end/tsrc/search/components/AdvancedSearchPanel.tsx
edalex-yinzi: There is a mistake here, onSubmit should use `values` as param. And the reason I left a param here is we can call doSearch immediately after `setCurrentValues`  by passing an empty map rather than the default value `currentValues` which might not update immediately?

react-front-end/__tests__/tsrc/search/AdvancedSearch.test.tsx
edalex-yinzi: Work out!

react-front-end/tsrc/search/components/AdvancedSearchPanel.tsx
```diff
@@ -93,6 +93,24 @@ export const AdvancedSearchPanel = ({
     [currentValues, setCurrentValues]
   );
 
+  /**
+   * do search based on input value
+   *
+   * @param value WizardHelper.FieldValueMap, default value is `currentValues`
+   */
+  const doSearch = (value = currentValues) => onSubmit(value);
```
PenghaiZhang, line 101: I am not sure if you need this and `onClear`. You can update the props to have `onClear` and then in SearchPage you can just do `onClear={() => handleSubmitAdvancedSearch(new Map())}`  

Please note `handleSubmitAdvancedSearch` is a new function in my PR.

react-front-end/__tests__/tsrc/search/AdvancedSearch.test.tsx
```diff
@@ -285,11 +303,60 @@ describe("Rendering of wizard", () => {
...
+          )
+        )
+      ),
+      // convert ["", ""] to {}
+      A.map(
+        M.filter(
+          pfTernaryTypeGuard<string[], string, boolean>(
+            isStringArrayValues,
+            flow(
+              A.filter((arrayItem) => arrayItem !== ""),
```
edalex-ian, line 349: Can that just be:

```suggestion
              A.filter(not(S.isEmpty)),
```

react-front-end/__tests__/tsrc/search/AdvancedSearch.test.tsx
```diff
@@ -285,11 +303,60 @@ describe("Rendering of wizard", () => {
...
+              A.isNonEmpty
+            ),
+            () => false
```
edalex-ian, line 352: Have a look at things like `constTrue`, `constFalse` which provide nice little helpers for these cases.

react-front-end/__tests__/tsrc/search/AdvancedSearch.test.tsx
```diff
@@ -285,11 +303,60 @@ describe("Rendering of wizard", () => {
...
+
+    // Make sure the current values match the default
+    expect(getCurrentLabelsAndValues()).toEqual(mockedLabelsAndValues);
+
+    // Clear all input value
+    await clickClearButton(container);
+
+    // get current labels and values and
+    // filter all empty value
+    const filteredLabelsAndValues = pipe(
```
edalex-ian, line 324: This now being rather long, it makes sense to pull it out into a function over in `AdvancedSearchTestHelper.ts` so  as to maintain readability of this test.

Perhaps a function like: `const filterEmptyValues = (labelsAndValues: WizardControlLabelValue[]): WizardControlLabelValue[] => flow(...)`.

Then you can just call that direct in your expect statement below. (e.g. `expect(pipe(getCurrentLabelsAndValues, filterEmptyValues)).toEqual([])`).

edalex-ian (changes requested):
Very close now. Last tweaks for test, and then @PenghaiZhang 's suggestion for the clear prop for the button, and you'll be done.

edalex-ian:
Heya @edalex-yinzi 

To action that last part of feedback from @PenghaiZhang , could you please merge `component/advanced_search` into your branch and then proceed?

Thanks,

edalex-ian (approved):
LGTM. Let's await a green build and then merge. :+1:

edalex-ian:
Cool, just failed on ol' flaky `AdvancedScriptControlTests`.

Merging.

PR #3580, opened 2021-11-15
refactor: use KeywordAnalyzer to support the phrases in Lucene query

PenghaiZhang:
##### Checklist

- [x] tests are included

##### Description of change

This is the fifth PR to support the Advanced search criteria. Previous testing work found that if a value is phrase, the parsed Lucene query is not quite right. So this PR brings in the fix and updates the test as well.

The fix includes changes on both server and client. On server, we use `KeywordAnalyzer` instead of `TleAnalyzer` to when the raw query is parsed. on Client, we add `""` for each term and phrase unless the control supports tokenisation or the control is a Calendar.

I also noticed that the query built from `ShuffleList` still has one issue so I fixed that in this PR.


edalex-ian (approved):
LGTM, let's merge as is on green.

If time was permitting (which it's not) I'd suggest `_values` be called something like `quotedValues` and only quote strings which include whitespace.

But, it will be fine as is, and we need it for tomorrow. So let's proceed as is.

edalex-ian:
Let's merge while we go and fix the test.

PR #3582, opened 2021-11-15
fix(test): remove flakiness of MyResourcesTest - testAddMyResource

edalex-yinzi:
##### Checklist

- [ ] tests are included

##### Description of change

Trying to fix this issue, but failed to reproduce in local env. 

image

But based on this [reference](https://stackoverflow.com/questions/38596456/how-to-over-staleelementreferenceexception-when-working-with-pagefactory)  I add a waiter before clicking the finish button. Not sure if it works or not.

### DIFF
--- a/autotest/OldTests/src/test/java/io/github/openequella/rest/DrmApiTest.java
+++ b/autotest/OldTests/src/test/java/io/github/openequella/rest/DrmApiTest.java
@@ -9,7 +9,6 @@ import java.io.IOException;
 import org.apache.commons.httpclient.methods.GetMethod;
 import org.apache.commons.httpclient.methods.PostMethod;
 import org.codehaus.jackson.JsonNode;
-import org.junit.After;
 import org.testng.annotations.Test;
 
 @TestInstitution("fiveo")
@@ -48,7 +47,6 @@ public class DrmApiTest extends AbstractRestApiTest {
     assertEquals(acceptDrm(ITEM_UUID, ITEM_VERSION), 400);
   }
 
-  @After
   @Test(description = "Fail to accept DRM terms due to access denied", priority = 10)
   public void accessDenied() throws IOException {
     logout();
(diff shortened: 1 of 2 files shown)


edalex-yinzi:
Seems the usage is wrong....

edalex-yinzi:
Also, I found an interesting situation. It seems every failed test was come after by a test call `accessDenied`.

image

edalex-yinzi:
I reproduce the same error by adding a command `logout` which is also used in `accessDenied`. So,  I guess this test will influence the login session globally and cause failure to another test that is running at the same time.

edalex-yinzi:
Looks good so far, I will move this change to #3584 and see how it works.

autotest/OldTests/src/test/java/io/github/openequella/rest/DrmApiTest.java
```diff
@@ -47,6 +48,7 @@ public void alreadyAccepted() throws IOException {
     assertEquals(acceptDrm(ITEM_UUID, ITEM_VERSION), 400);
   }
 
+  @After
```
edalex-ian, line 51: As discussed, this should be removed.

autotest/Tests/src/main/java/com/tle/webtests/pageobject/selection/SelectionStatusPage.java
```diff
@@ -22,6 +24,7 @@ protected WebElement findLoadedElement() {
   }
 
   public SelectionCheckoutPage finishSelections() {
+    ExpectWaiter.waiter(ExpectedConditions.visibilityOf(finishedButton), this).get();
```
edalex-ian, line 27: Keep in mind that the main symptom being seeing is that `finishedButton` has become **stale**. As a result, even if you wait here, when you go to click below that's still no good - the original object is still stale.

Perhaps rather than relying on the class level button property:

- Change the class level property to a `By`; then
- In this method add a `WebElement finishButton = ...`; and
- Consider using something like `com.tle.webtests.pageobject.AbstractPage#waitForElement(org.openqa.selenium.By)`

edalex-ian (changes requested):
Good start @edalex-yinzi . I like where you've put the waiter as it aligns with the stale element exception we're seeing. However I think it still needs a bit more consideration.

(And let's not worry about the `@After`.)

edalex-ian:
Ran cleanly other than flaking on known flaky - AdvancedScriptControlTests.

Doing rerun 2, and if we get 3 all happy we'll merge.

edalex-ian (approved):

edalex-ian:
Run 2 was all green. Starting third and final. If green, will merge. :wink:

edalex-ian:
Alright, another pass. Merging now.

PR #3589, opened 2021-11-18
feat(front-end): add User Selector wizard component

edalex-ian:
##### Checklist

- [x] tests are included

##### Description of change

Here is the User Selector Wizard control. It re-uses existing functionality to search for users, however further updates are still to come to properly support the level of multi-user selection capable with the legacy user selector (as that common control will need to be further developer). Additionally, there is not yet support for filtering to specific groups etc.

It does include some optimisation to minimise server calls to resolve full user details, however there slightly more optimisation could be done but things were getting out of hand. That said though, it should be fine without those but if needs be adding it wont be too involved. (I'll highlight what this is in comments on the code.)

Up next I'll integrate this into the wizard/advanced search, and then loop back around after that and add filtering support.

https://user-images.githubusercontent.com/43919233/142357582-6c59d9d0-6f81-41c6-b637-eb1eed387746.mp4

### DIFF
--- a/react-front-end/__tests__/tsrc/components/wizard/WizardUserSelector.test.tsx
+++ b/react-front-end/__tests__/tsrc/components/wizard/WizardUserSelector.test.tsx
@@ -1,0 +1,42 @@
+/*
+ * Licensed to The Apereo Foundation under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * The Apereo Foundation licenses this file to you under the Apache License,
+ * Version 2.0, (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import "@testing-library/jest-dom/extend-expect";
+import { render, waitFor } from "@testing-library/react";
+import * as React from "react";
+import * as UserSearchMock from "../../../../__mocks__/UserSearch.mock";
+import { WizardUserSelector } from "../../../../tsrc/components/wizard/WizardUserSelector";
+
+describe("<WizardUserSelector/>", () => {
+  it("displays a specified set of users", async () => {
+    const testUser = UserSearchMock.users[1];
+    const { queryByText } = render(
+      <WizardUserSelector
+        groupsFilter={new Set()}
+        multiple
+        onChange={jest.fn()}
+        users={new Set([testUser.id])}
+        mandatory={false}
+        resolveUsersProvider={UserSearchMock.resolveUsersProvider}
+      />
+    );
+
+    await waitFor(() =>
+      expect(queryByText(testUser.username)).toBeInTheDocument()
+    );
+  });
+});
--- a/react-front-end/__tests__/tsrc/modules/UserModule.test.ts
+++ b/react-front-end/__tests__/tsrc/modules/UserModule.test.ts
@@ -1,0 +1,149 @@
+/*
+ * Licensed to The Apereo Foundation under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * The Apereo Foundation licenses this file to you under the Apache License,
+ * Version 2.0, (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import * as OEQ from "@openequella/rest-api-client";
+import { pipe } from "fp-ts/function";
+import * as ORD from "fp-ts/Ord";
+import * as RA from "fp-ts/ReadonlyArray";
+import * as RR from "fp-ts/ReadonlyRecord";
+import * as RSET from "fp-ts/ReadonlySet";
+import { last } from "fp-ts/Semigroup";
+import * as S from "fp-ts/string";
+import {
+  resolveUsersProvider,
+  users,
+} from "../../../__mocks__/UserSearch.mock";
+import {
+  resolveUsersCached,
+  UserCache,
+} from "../../../tsrc/modules/UserModule";
+import { expectRight } from "../FpTsMatchers";
+
+describe("resolveUsersCached", () => {
+  // Note that this function retrieves unknown users via `resolveUsersProvider` mock which just
+  // looks up IDs within the `users` mock from `UserSearch.mock`.
+  const retrieveTestUser = async (
+    ids: ReadonlySet<string>,
+    cache: UserCache,
+    setCache: (_: UserCache) => void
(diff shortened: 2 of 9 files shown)


react-front-end/__tests__/tsrc/components/wizard/WizardUserSelector.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import "@testing-library/jest-dom/extend-expect";
+import { render, waitFor } from "@testing-library/react";
+import * as React from "react";
+import * as UserSearchMock from "../../../../__mocks__/UserSearch.mock";
+import { WizardUserSelector } from "../../../../tsrc/components/wizard/WizardUserSelector";
+
+describe("<WizardUserSelector/>", () => {
```
edalex-ian, line 24: Only a very simple test case due to time limitations. More coverage will be provided by the WizardHelper, AdvancedSearch tests.

react-front-end/__tests__/tsrc/modules/UserModule.test.ts
```diff
@@ -0,0 +1,149 @@
...
+  resolveUsersProvider,
+  users,
+} from "../../../__mocks__/UserSearch.mock";
+import {
+  resolveUsersCached,
+  UserCache,
+} from "../../../tsrc/modules/UserModule";
+import { expectRight } from "../FpTsMatchers";
+
+describe("resolveUsersCached", () => {
```
edalex-ian, line 36: I did though spend more time on tests here due tot he complexity, and I wanted to make sure all was well with some of the fp-ts usage.

react-front-end/tsrc/components/SelectUserDialog.tsx
```diff
@@ -0,0 +1,84 @@
...
+  /** Handler for when dialog closes. */
+  onClose: (selection?: OEQ.UserQuery.UserDetails) => void;
+  /** Function which will provide the list of users for UserSearch. */
+  userListProvider?: (query?: string) => Promise<OEQ.UserQuery.UserDetails[]>;
+}
+
+/**
+ * Simple dialog to prompt user to search and select a user to use in the owner filter.
+ */
+export const SelectUserDialog = ({
```
edalex-ian, line 43: This is an 'extract' refactor from OwnerSelector. No changes, just making it stand-alone so that it could be re-used.

react-front-end/tsrc/components/wizard/WizardUserSelector.tsx
```diff
@@ -0,0 +1,300 @@
...
+
+    pipe(
+      selection,
+      O.fromNullable,
+      O.map((u) =>
+        multiple
+          ? pipe(fullUsers, RSET.insert(eqUserById)(u))
+          : RSET.singleton(u)
+      ),
+      O.map(callOnChange)
```
edalex-ian, line 233: The other optimisation is to here also update `fullUsers` with the details of any additional new users. But for now we just call straight to onChange and rely on the two way binding to sort it out. BUT, that means an additional call is made (potentially - depending on cache state) to the server to get the new user's details.

After all the other caching, and the increasing level of complexity in here it felt unwarranted - especially with time constraints. However, if we fine performance is lacking then we can either a. add a spinner when list the users; and/or b. do the above mentioned optimisation.

react-front-end/tsrc/components/wizard/WizardUserSelector.tsx
```diff
@@ -0,0 +1,300 @@
...
+  /**
+   * What do do when the click the clear button.
+   */
+  onClear: () => void;
+}
+
+/**
+ * Display component for display any errors in the list of users.
+ */
+const ListItemError = ({ error, onClear }: ListItemErrorProps): JSX.Element => (
```
edalex-ian, line 88: For readability I chose to split out the two main ListItem types into local components to this file. It also obviously helps when reviewing the component tree in the browser.

react-front-end/tsrc/modules/UserModule.ts
```diff
@@ -67,3 +97,83 @@ export const findUserById = async (
...
+ *
+ * @param cache A cache update from previous calls, or initially simply `{}`
+ * @param setCache A function which will be called if any updates where made which should trigger
+ *                 an updated version of the cache. Typically this will occur if additional user
+ *                 details had to be fetched. However if no additional details were required, then
+ *                 no update to cache will be made and this function wont be called.
+ * @param resolver a function (such as `resolveUsers`) which can make a call to the server to fetch
+ *                 the details of any unknown user ids.
+ */
+export const resolveUsersCached =
```
edalex-ian, line 115: This is the core of the optimisation to minimise calls to the server. Unfortunately couldn't just rely on a simple memoize due to wanting to incrementally build a cache and not just base on input/output.

`cache` and `setCache` were left (I guess) pluggable to allow for them to be controlled independently as needed. Plus, it kept the implementation more FP focused.

edalex-ian:
When you return from leave @PenghaiZhang , could you please review?

react-front-end/tsrc/components/wizard/WizardUserSelector.tsx
```diff
@@ -0,0 +1,300 @@
...
+  userCache = newCache;
+};
+
+interface ListItemErrorProps {
+  /**
+   * The error to display.
+   */
+  error: string;
+  /**
+   * What do do when the click the clear button.
```
PenghaiZhang, line 80:
```suggestion
   * What do do when click the clear button.
```

react-front-end/tsrc/components/wizard/WizardUserSelector.tsx
```diff
@@ -0,0 +1,300 @@
...
+  useEffect(() => {
+    if (RSET.getEq(S.Eq).equals(users, pipe(fullUsers, userIds))) {
+      // Already in sync, no action required.
+      return;
+    }
+
+    console.debug("useEffect(users)", users);
+
+    // Update `fullUsers` but only if there's a change
+    const updateFullUsers: (updatedList: typeof fullUsers) => void = flow(
```
PenghaiZhang, line 197: May I ask why use `typeof` ? Is that not equal to `ReadonlySet<OEQ.UserQuery.UserDetails>` ?

react-front-end/tsrc/components/wizard/WizardUserSelector.tsx
```diff
@@ -0,0 +1,300 @@
...
+
+    const callOnChange = (
+      updatedFullUsers: ReadonlySet<OEQ.UserQuery.UserDetails>
+    ) => pipe(updatedFullUsers, userIds, onChange);
+
+    pipe(
+      selection,
+      O.fromNullable,
+      O.map((u) =>
+        multiple
```
PenghaiZhang, line 229: Seeing `multiple` here, I guess the way to select multiple users is to open the dialog multiple times and select one user each time. Is that correct ?

PenghaiZhang (approved):
Thanks Ian! LGTM. I only have a few queries.

react-front-end/tsrc/components/wizard/WizardUserSelector.tsx
edalex-ian: Correct, but this function is intended to specifically work on `fullUsers`, so it seemed a way to make sure it is coupled and if they typing of `fullUsers` change then this will break at compilation time.

This kind of pattern could be abused, but I think here it's safe and clean.

react-front-end/tsrc/components/wizard/WizardUserSelector.tsx
edalex-ian: Oh, what a mess, I see there's even another typo to clean up. I'll address. :wink:

react-front-end/tsrc/components/wizard/WizardUserSelector.tsx
edalex-ian: That's correct. In the future - when/if time permits - we'll add a feature to the user browser control to allow multiple selections within it. (Looks like that may need to wait until a future version though.)

edalex-ian:
Thanks @PenghaiZhang . I'll wait for another green build and then merge. :wink:

PR #3590, opened 2021-11-19
feat(rest-api): add support for the theme endpoint to oeq-ts-rest-api

edalex-yinzi:
##### Checklist

- [x] tests are included

##### Description of change

Create api with tests for Theme module.

API: 
  - getThemeSettings,
  - updateThemeSettings,
  - getThemeLogo,
  - updateThemeLogo,

Because I am still struggling with tests for theme logo, only contains tests for:
 - getThemeSettings
 - updateThemeSettings

### DIFF
--- a/oeq-ts-rest-api/src/Theme.ts
+++ b/oeq-ts-rest-api/src/Theme.ts
@@ -62,15 +62,12 @@ export const updateThemeSettings = (
   );
 
 /**
- * Retrieve theme logo
+ * Retrieve theme logo url
  *
  * @param apiBasePath Base URI to the oEQ institution and API
  */
-export const getThemeLogo = (apiBasePath: string): Promise<string> =>
-  GET<string>(
-    apiBasePath + THEME_ROOT_PATH + '/newLogo.png',
-    (data): data is any => is<any>(data)
-  );
+export const getThemeLogoUrl = (apiBasePath: string): string =>
+  apiBasePath + THEME_ROOT_PATH + '/newLogo.png';
 
 /**
  * Delete theme logo, reset logo to default
@@ -84,7 +81,7 @@ export const deleteLogo = (apiBasePath: string): Promise<void> =>
  * Update theme logo.
  *
  * @param apiBasePath Base URI to the oEQ institution and API
- * @param updatedSettings New theme Settings
+ * @param file New logo file
  */
 export const updateThemeLogo = (
   apiBasePath: string,
--- a/oeq-ts-rest-api/src/Theme.ts
+++ b/oeq-ts-rest-api/src/Theme.ts
@@ -1,0 +1,94 @@
+/*
+ * Licensed to The Apereo Foundation under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * The Apereo Foundation licenses this file to you under the Apache License,
+ * Version 2.0, (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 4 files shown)


edalex-yinzi:
testAddMyResource failed

oeq-ts-rest-api/src/Theme.ts
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Retrieve theme logo
+ *
+ * @param apiBasePath Base URI to the oEQ institution and API
+ */
+export const getThemeLogo = (apiBasePath: string): Promise<string> =>
+  GET<string>(
+    apiBasePath + THEME_ROOT_PATH + '/newLogo.png',
+    (data): data is any => is<any>(data)
```
edalex-ian, line 72: As discussed, don't use `is` (typescript-is) here, instead you need to manually write a type guard.

oeq-ts-rest-api/test/Theme.test.ts
```diff
@@ -0,0 +1,45 @@
...
+import { ThemeSettings } from '../src/Theme';
+
+beforeAll(() => OEQ.Auth.login(TC.API_PATH, TC.USERNAME, TC.PASSWORD));
+
+afterAll(() => OEQ.Auth.logout(TC.API_PATH, true));
+
+describe('Theme settings', () => {
+  let settingsAtStart: ThemeSettings;
+
+  beforeAll(async () => {
```
edalex-ian, line 29: Be a good idea to add an `afterAll()` to return them to `settingsAtStart`.

oeq-ts-rest-api/tsconfig.json
```diff
@@ -2,7 +2,10 @@
   "include": ["src", "types"],
   "compilerOptions": {
     "module": "ES6",
-    "lib": ["ES6"],
+    "lib": [
+      "ES6",
+      "dom"
```
edalex-ian, line 7: Do we really need this?

I gather you've added for `File`, but the oEQ server has no notion of a `dom` `File`. So perhaps we need something else, as that concern would only be in the front-end's Theme Module.

edalex-ian (approved):
Looks pretty good @edalex-yinzi , just need to:

- Write you own type guard instead of using `is`;
- Tidy up the modification of the settings after test; and
- Revisit the use of `File` and remove `dom` from the libraries.

edalex-ian (changes requested):
Whoops, that was meant to be a 'request changes'. Sorry. :wink:

oeq-ts-rest-api/src/Theme.ts
edalex-yinzi: Yes, just wonder do we still need this API, if we are not going to test the reset of logo API.

oeq-ts-rest-api/test/Theme.test.ts
edalex-yinzi: Do you mean reset the value of the setting in afterAll?

edalex-ian: Yup. :+1:

oeq-ts-rest-api/src/Theme.ts
edalex-ian: Could drop it, and just add something like:

```typescript
export const getThemeLogoUrl = (apiBasePath: string): string =>
  apiBasePath + THEME_ROOT_PATH + '/newLogo.png';
```

oeq-ts-rest-api/tsconfig.json
edalex-yinzi: Use blob instead?

edalex-ian: Yeah, if that works and matches the server API.

edalex-yinzi: After testing, `Buffer` might be a good type.

edalex-ian: Just make sure whatever you choose works in Browsers.

edalex-yinzi:
I might not implement any test for theme logo in this branch because it needs more consideration. 
Also, I keep the `File` type in theme logo API in order to use it in browsers.

What do you think? @edalex-ian

edalex-ian (approved):
Sure thing. It's not super critical, and it's not leaving it worse than you found it, so let's proceed.

edalex-ian:
Heh, flaked on `MyResourcesTest`. Good to merge. :wink:

PR #3598, opened 2021-11-25
Add support to the Wizard User Selector for filtering by group

edalex-ian:
##### Checklist

- [x] tests are included

##### Description of change

This adds support for the feature of the Wizard User Selector control where one or more groups can be configured to limit the users returned in search results - and thereby the users available for selection.

Points of note:

- Refactoring was done to create a new `UserModule.mock.ts` from what was in `UserSearch.mock.ts` to better align the emergent naming standard for our mocks - which was highlighted when creating the new `GoupModule.mock.ts`. This resulted in a lot of minor updates to `*.test.ts*` and `*.stories.tsx` files.
- In the interest of time, no additional tests (other than a minor one) have been explicitly added for this, but manual testing and testing via the new storybook test was undertaken.
- Main effort is focused in `UserSearch.tsx` with support from `GroupModule.ts` (with most of the code really just being around the display of the notice - and tooltip - for when a group filter is active)
- There is a change in `UserModule.ts` which changes which endpoint is used to do searching for users - which is also used by the `SearchPage` Owner Selector (as it uses the `SelectUserDialog` which uses `UserSearch`)
- Speaking of which, the component tree updated to support this is `WizardUserSelector` -> `SelectUserDialog` -> `UserSearch` - these were updated to support the passing of props related to group filtering for the search, and a provider function for resolving group UUIDs for display.

That's the gist of it.

Following this, I will need one more PR to wire it into the new search code.

https://user-images.githubusercontent.com/43919233/143394267-468137e0-fb7d-477d-9eb1-dc2f1a3fb448.mp4

### DIFF
--- a/react-front-end/tsrc/components/UserSearch.tsx
+++ b/react-front-end/tsrc/components/UserSearch.tsx
@@ -32,12 +32,19 @@ import ErrorOutline from "@material-ui/icons/ErrorOutline";
 import InfoIcon from "@material-ui/icons/Info";
 import SearchIcon from "@material-ui/icons/Search";
 import * as OEQ from "@openequella/rest-api-client";
-import { pipe } from "fp-ts/function";
+import { flow, pipe } from "fp-ts/function";
+import * as O from "fp-ts/Option";
+import * as ORD from "fp-ts/Ord";
+import { not } from "fp-ts/Predicate";
 import * as RA from "fp-ts/ReadonlyArray";
 import * as RSET from "fp-ts/ReadonlySet";
+import * as S from "fp-ts/string";
+import * as T from "fp-ts/Task";
+import * as TE from "fp-ts/TaskEither";
 import * as React from "react";
-import { KeyboardEvent, useState } from "react";
+import { KeyboardEvent, useEffect, useState } from "react";
 import { sprintf } from "sprintf-js";
+import * as GroupModule from "../modules/GroupModule";
 import * as UserModule from "../modules/UserModule";
 import { languageStrings } from "../util/langstrings";
 import { OrdAsIs } from "../util/Ord";
@@ -63,6 +70,13 @@ export interface UserSearchProps {
     query?: string,
     groupFilter?: ReadonlySet<string>
   ) => Promise<OEQ.UserQuery.UserDetails[]>;
+  /**
+   * Function which will resolve group IDs to full group details so that the group names can be
+   * used for display.
+   */
+  resolveGroupsProvider?: (
+    ids: ReadonlyArray<string>
+  ) => Promise<OEQ.UserQuery.GroupDetails[]>;
 }
 
 /**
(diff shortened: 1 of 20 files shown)


react-front-end/tsrc/components/UserSearch.tsx
```diff
@@ -54,27 +87,54 @@ const UserSearch = ({
...
+  }, [groupFilter, resolveGroupsProvider]);
+
   // Simple helper function to assist with providing useful id's for testing and theming.
   const genId = (suffix?: string) =>
     (id ? `${id}-` : "") + "UserSearch" + (suffix ? `-${suffix}` : "");
 
   const handleOnSearch = () => {
     setShowSpinner(true);
-    userListProvider(query)
+    userListProvider(query, groupFilter)
```
edalex-ian, line 137: Other than this one change, all the other (almost 200 lines) of changes in this file are to support the display of the filter active notice and list out the names of the groups in the filter. :roll_eyes:

react-front-end/tsrc/modules/UserModule.ts
```diff
@@ -51,16 +51,16 @@ export const userIds: (
...
+ * @param groupFilter A list of group UUIDs to filter the search by
  */
-export const listUsers = (
-  query?: string
+export const listUsers = async (
+  query?: string,
+  groupFilter?: ReadonlySet<string>
 ): Promise<OEQ.UserQuery.UserDetails[]> =>
-  OEQ.UserQuery.search(API_BASE_URL, {
+  await OEQ.UserQuery.filtered(API_BASE_URL, {
```
edalex-ian, line 60: Note the change of API being used here. Although a different endpoint, arguably a more targetted and suitable one - even if only recently added (in the previous PR).

react-front-end/tsrc/util/langstrings.ts
```diff
@@ -710,6 +710,8 @@ export const languageStrings = {
   },
   userSearchComponent: {
     failedToFindUsersMessage: "Unable to find any users matching '%s'",
+    filterActiveNotice: "Results will be filtered.",
+    filteredByPrelude: "Your search results will be filtered by these groups:",
```
edalex-ian, line 714: IIRC, these match the Legacy UI strings.

react-front-end/tsrc/modules/UserModule.ts
```diff
@@ -51,16 +51,16 @@ export const userIds: (
...
     q: query,
-    users: true,
-    groups: false,
-    roles: false,
-  }).then((result: OEQ.UserQuery.SearchResult) => result.users);
+    byGroups: groupFilter ? Array.from<string>(groupFilter) : undefined,
```
PenghaiZhang, line 62: Do you think `pfTernaryTypeGuard` would be a good choice here ?

PenghaiZhang (approved):
LGTM Ian!

react-front-end/tsrc/modules/UserModule.ts
edalex-ian: Unfortunately not, two main reasons:

- Bit verbose to do the simple initial check for `groupFilter` - be probably better to just go with `O.fromNullable` and finish with `O.toUndefined`
- Need to provide the typing for `Array.from` however the expression `Array.from<string>` ends up thinking it's a value rather than a function.

So in the end, feels like this is the most straightforward.

PR #3609, opened 2021-11-30
feat: integrate WizardSimpleTermSelector with Adv search panel.

PenghaiZhang:
##### Checklist

- [x] tests are included

##### Description of change

This PR integrates the `WizardSimpleTermSelector` into the Advanced search panel. Due to time limit I only updated `WizardHelper.test`.

https://user-images.githubusercontent.com/47203811/144141588-037d119b-e394-47b6-821b-20bfb5db5c3e.mp4

### DIFF
--- a/react-front-end/tsrc/components/wizard/WizardHelper.tsx
+++ b/react-front-end/tsrc/components/wizard/WizardHelper.tsx
@@ -29,7 +29,6 @@ import { first } from "fp-ts/Semigroup";
 import * as SET from "fp-ts/Set";
 import * as S from "fp-ts/string";
 import * as React from "react";
-import { searchTaxonomyTerms } from "../../modules/TaxonomyModule";
 import { OrdAsIs } from "../../util/Ord";
 import { WizardCalendar } from "./WizardCalendar";
 import { WizardCheckBoxGroup } from "./WizardCheckBoxGroup";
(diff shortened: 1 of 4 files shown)


react-front-end/tsrc/components/wizard/WizardHelper.tsx
```diff
@@ -387,7 +389,26 @@ const controlFactory = (
...
+      return (
+        <WizardSimpleTermSelector
+          {...commonProps}
+          values={valueAsStringSet()}
+          onSelect={flow(RSET.toSet, onChangeForStringSet)}
+          isAllowMultiple={isAllowMultiple}
+          selectedTaxonomy={selectedTaxonomy}
+          selectionRestriction={selectionRestriction}
+          termStorageFormat={termStorageFormat}
+          termProvider={searchTaxonomyTerms}
```
edalex-ian, line 409: Ah, I must've missed this in the other review. The pattern we have here is that for these providers we provide the default implementation at the usage point. For example, have a look at this line in the WizardUserSelector: https://github.com/PenghaiZhang/openEQUELLA/blob/feature/term-selector-simple/react-front-end/tsrc/components/wizard/WizardUserSelector.tsx#L190

That way we only need to specify the 'providers' when we want to do something special - tests, story book, etc.

react-front-end/tsrc/components/wizard/WizardHelper.tsx
```diff
@@ -387,7 +389,26 @@ const controlFactory = (
         )
       );
     case "termselector":
-      return <WizardUnsupported id={id} />;
+      const {
+        isAllowMultiple,
+        selectedTaxonomy,
+        selectionRestriction,
+        termStorageFormat,
+      } = control as OEQ.WizardControl.WizardTermSelectorControl;
```
edalex-ian, line 397: You'll note up above for `userselector` (and somewhere else) I've taken the approach of using `pipe` for this and de-structuring in an arrow function. The reason for all this is that these variables are actually placed in the same scope as all the other case statements.

I'd suggest that seeing you have so many, the same should be done here.

(Something I've not tried, is actually creating a block (i.e. with curly braces) for the content of the case statement like we do in other languages. That may then force block scope. However, I think prettier may also deem it unnecessary and remove them on us. :cry: )

edalex-ian (commented):
Good stuff @PenghaiZhang , just two tweaks requested.

And I support skipping the test at this point. We're a bit behind so unfortunately something has to give - and like often the case, that's tests.

react-front-end/tsrc/components/wizard/WizardHelper.tsx
PenghaiZhang: good idea!!

react-front-end/tsrc/components/wizard/WizardHelper.tsx
PenghaiZhang: I have confirmed that we can create a block to limit the variable scope.

edalex-ian (approved):
LGTM! :+1:

PR #3635, opened 2021-12-07
feat: With visibility scripting, clear values of any hidden fields

edalex-ian:
##### Checklist

- [x] tests are included

##### Description of change

To match Legacy UI workings, when a field is hidden due to visibility scripting, then it's value should be cleared. This change adds that.

Also includes minor fix to AdvancedSearchPanel.stories... Not sure how long that's been broken....

https://user-images.githubusercontent.com/43919233/144963888-4c22cfb2-9ede-4154-b67d-ed526e948118.mp4

### DIFF
--- a/react-front-end/tsrc/components/wizard/WizardHelper.tsx
+++ b/react-front-end/tsrc/components/wizard/WizardHelper.tsx
@@ -35,6 +35,7 @@ import * as RA from "fp-ts/ReadonlyArray";
 import * as RSET from "fp-ts/ReadonlySet";
 import { Refinement } from "fp-ts/Refinement";
 import { first } from "fp-ts/Semigroup";
+import * as SEP from "fp-ts/Separated";
 import * as SET from "fp-ts/Set";
 import * as S from "fp-ts/string";
 import * as React from "react";
@@ -527,14 +528,16 @@ export const buildVisibilityScriptContext = (
  *               no value, then it should not be in the collection.
  * @param onChange The high level callback to use to update `values` - this will be wrapped so that
  *                 Wizard components only have to worry about calling a typical simple callback
- *                 with their updated value.
+ *                 with their updated value. Takes an array so as to support bulk-updating of values
+ *                 for controls which may be hidden due to visibility scripting. (But could also be
+ *                 used for other optimisations.)
  * @param visibilityScriptContext The `ScriptContext` to be used when evaluating any visibility
  *                                scripts defined in the `controls`.
  */
 export const render = (
   controls: OEQ.WizardControl.WizardControl[],
   values: FieldValueMap,
-  onChange: (update: FieldValue) => void,
+  onChange: (updates: FieldValue[]) => void,
   visibilityScriptContext: ScriptContext
 ): JSX.Element[] => {
   const buildOnChangeHandler = (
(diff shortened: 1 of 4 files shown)


react-front-end/tsrc/components/wizard/WizardHelper.tsx
```diff
@@ -561,6 +564,46 @@ export const render = (
...
+  ) => O.Option<void> = flow(
+    RA.filter(OEQ.WizardControl.isWizardBasicControl),
+    // Only call the onChange if field currently has a value - or we'll end up in an infinite render
+    RA.filter(hasValue),
+    // Only call the onChange if there are changes - or we'll end up in an infinite render loop
+    O.fromPredicate(RA.isNonEmpty),
+    O.map(
+      flow(
+        RA.map((c) => ({ target: buildControlTarget(c), value: [] })),
+        RA.toArray,
```
PenghaiZhang, line 586: I think here is exactly what I did in my PR - converting an `ReadOnlyArray`  to a normal array. Otherwise, we will get an error complaining not being able to assign a ReadOnlyArray<T> to `T[]`

react-front-end/tsrc/components/wizard/WizardHelper.tsx
```diff
@@ -581,20 +624,14 @@ export const render = (
...
-        c,
-        buildOnChangeHandler(c),
-        values,
-        retrieveControlsValue(c)
-      )
-    ),
-    RA.toArray
+    RA.partition(isVisible),
+    // Clear values of all hidden controls - with a bulk call to onChange
+    SEP.mapLeft(clearControls),
```
PenghaiZhang, line 631: I don't want to make things more complicated, but what would happen if a hidden control has a default value ?

PenghaiZhang (approved):
LGTM @edalex-ian . I just have one query.

And good to see the use of `separated`!

react-front-end/tsrc/components/wizard/WizardHelper.tsx
edalex-ian: Yup, exactly what I was mentioning in Slack. It gets nuked and the default value is never seen again.

I tried to do some comparisons to the Legacy UI for this but it was getting a bit confusing - as it's a bit broken in a few ways. So I'm currently thinking to get this all in and working, and then we can review the UX as a group and decide what the rules are for default values.

react-front-end/tsrc/components/wizard/WizardHelper.tsx
edalex-ian: Yeah, that was why I was querying yours (as it was also followed by `A.flatten` so I wasn't sure if that was why.

react-front-end/tsrc/components/wizard/WizardHelper.tsx
edalex-ian: I wonder if we may end up having to pass the default values down into `render` and have it use them as needed. :thinking:

edalex-ian:
Thanks @PenghaiZhang , if a green build comes along I'll merge this and then tee up a meeting with the team to discuss how to handle default values.

react-front-end/tsrc/components/wizard/WizardHelper.tsx
edalex-ian: Ahhh, I see they're already passed in with the `controls` so at least that part should be relatively straight forward. Just need to consider if we need to modify/remove the other code that already pulls in the default values...

PR #3636, opened 2021-12-07
fix: additional whitespace in user selector

edalex-yinzi:
##### Checklist

- [ ] tests are included

##### Description of change
Fix additional whitespace in user selector by adding two classes to the root div for wizards in AdvSearchPanel.

Solution found in [here](https://stackoverflow.com/questions/51445959/unexpected-empty-space-using-flexbox).

https://user-images.githubusercontent.com/92769668/144966006-e633327d-04bf-4014-88e9-c61d14f91900.mp4

### DIFF
--- a/react-front-end/tsrc/search/components/AdvancedSearchPanel.tsx
+++ b/react-front-end/tsrc/search/components/AdvancedSearchPanel.tsx
@@ -130,7 +130,7 @@ export const AdvancedSearchPanel = ({
             currentValues,
             onChangeHandler
           ).map((e) => (
-            <Grid key={e.props.id} item>
+            <Grid key={e.props.id} item container direction="column">
               {e}
             </Grid>
           ))}
--- a/react-front-end/tsrc/search/components/AdvancedSearchPanel.tsx
+++ b/react-front-end/tsrc/search/components/AdvancedSearchPanel.tsx
@@ -130,7 +130,7 @@ export const AdvancedSearchPanel = ({
             currentValues,
             onChangeHandler
           ).map((e) => (
-            <Grid key={e.props.id} item container direction="column">
+            <Grid key={e.props.id} container direction="column">
               {e}
             </Grid>
           ))}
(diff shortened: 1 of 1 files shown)


react-front-end/tsrc/search/components/AdvancedSearchPanel.tsx
```diff
@@ -130,7 +130,7 @@ export const AdvancedSearchPanel = ({
             currentValues,
             onChangeHandler
           ).map((e) => (
-            <Grid key={e.props.id} item>
+            <Grid key={e.props.id} item container direction="column">
```
edalex-ian, line 133: Hmmmmmmm.... Not sure on this one @edalex-yinzi , you normally have either `item` or `container`... Not both. :thinking:

PenghaiZhang: I think a nested container is OK. Here is an [example](https://github.com/mui-org/material-ui/pull/12313/files)

react-front-end/tsrc/search/components/AdvancedSearchPanel.tsx
```diff
@@ -130,7 +130,7 @@ export const AdvancedSearchPanel = ({
             currentValues,
             onChangeHandler
           ).map((e) => (
-            <Grid key={e.props.id} item>
+            <Grid key={e.props.id} container direction="column">
```
PenghaiZhang, line 133: It should really be an `item` at least.

edalex-yinzi: And we do need its margin style anyway, I will add it back.

PenghaiZhang: If you make this `item` a nested `container`, whatever inside it should be wrapped by an `item`. 
I am not quite what styles you need, but if they are quite simple, perhaps an inline style is sufficient ??

edalex-ian (approved):
Makes sense to say by default each WizardControl is provided with 100% width, so, sure. :wink: :+1:

PenghaiZhang (approved):

PR #3637, opened 2021-12-08
feat(frontend): add error handling for wizard controls

edalex-ian:
##### Checklist

- [ ] tests are included

##### Description of change

Some Wizard Controls are a bit smarter than others (say, with async
calls) and as a result are prone to have errors. Therefore, we need a
standard way for them to report their errors.

For this, I've created a React Context that sits in WizardHelper so that
 any wizard rendered controls can be wrapped in that context and can
 rely on it be present. Currently, this is only in the Advanced Search
 Panel, but in the future it would be in the components for contribution
  wizards.

Tests: Covered mostly be existing tests, plus manual testing to ensure when network connectivity was broken that the standard search page error toast was displayed.

### DIFF
--- a/react-front-end/tsrc/components/wizard/WizardUserSelector.tsx
+++ b/react-front-end/tsrc/components/wizard/WizardUserSelector.tsx
@@ -26,12 +26,11 @@ import {
 } from "@material-ui/core";
 import AccountCircleIcon from "@material-ui/icons/AccountCircle";
 import AddIcon from "@material-ui/icons/Add";
-import ClearIcon from "@material-ui/icons/Clear";
 import DeleteIcon from "@material-ui/icons/Delete";
-import ErrorIcon from "@material-ui/icons/Error";
 import ListIcon from "@material-ui/icons/List";
 import * as OEQ from "@openequella/rest-api-client";
-import { flow, pipe } from "fp-ts/function";
+import * as E from "fp-ts/Either";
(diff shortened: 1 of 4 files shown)


PenghaiZhang (approved):
Looks very good!

react-front-end/tsrc/components/wizard/WizardUserSelector.tsx
```diff
@@ -218,16 +192,31 @@ export const WizardUserSelector = ({
...
       // of `users` and `onChange`
       resolveUsersCached(userCache, updateUserCache, resolveUsersProvider),
-      TE.match(setError, updateFullUsers)
+      // Report any errors
+      TE.mapLeft(flow(E.toError, handleError)),
+      // But either way ensure we've got something to show
+      TE.match(
+        () =>
+          pipe(
+            users,
```
PenghaiZhang, line 201: Why do we show all the selected users in this case ? The previous implementation shows an error icon and a clear icon.

edalex-ian: Two reasons:

1. (primary) I ended up in an infinite render loop due to the updates triggered from the toast error - where the guard at the top of this effect then always passes; and
2. Although we don't have the ability to present pretty information, we can still search with what's there so might as well just display something - and that then helps with one, where the other fix would be to add (keep around) some kind of flag to indicate an error state and use that in the guard.

PR #3640, opened 2021-12-10
fix: display attachment actions only in editable mode

edalex-yinzi:
##### Checklist

- [ ] tests are included

##### Description of change
Now attachment actions only display after the user click the edit button.

So before we were misled by the local variable `editable` and props `editable`. :smile: 
It can actually be done in the front end.

https://user-images.githubusercontent.com/92769668/145497673-7911dbc4-1244-4442-8222-4b21031b670f.mp4

 

### DIFF
--- a/react-front-end/tsrc/fileuploader/InlineFileUploader.tsx
+++ b/react-front-end/tsrc/fileuploader/InlineFileUploader.tsx
@@ -295,8 +295,20 @@ export const InlineFileUploader = ({
    * Build three text buttons for UploadedFile or one icon button for UploadingFile.
    */
   const buildActions = (file: UploadedFile | UploadingFile): UploadAction[] => {
-    if (isUploadedFile(file)) {
-      if (!editable) return [];
+    const uploadComplete = isUploadedFile(file);
+
+    if (!uploadComplete) {
+      return [
+        {
+          onClick: () => onCancel(file.localId),
+          text: strings.cancel,
+          icon: <CancelIcon />,
+        },
+      ];
+    } else if (!editable) {
+      // No further actions available
+      return [];
+    } else {
       const basicAction = [
         {
           onClick: () => onDelete(file),
@@ -304,7 +316,7 @@ export const InlineFileUploader = ({
         },
       ];
       const { id, editable: fileEditable } = file.fileEntry;
-      return fileEditable
+      const actions = fileEditable
         ? [
             {
               onClick: () => onEdit(id),
@@ -317,14 +329,8 @@ export const InlineFileUploader = ({
             ...basicAction,
           ]
         : basicAction;
+      return actions;
     }
-    return [
-      {
-        onClick: () => onCancel(file.localId),
-        text: strings.cancel,
-        icon: <CancelIcon />,
-      },
-    ];
   };
 
   // Build an Icon button for adding resources. In Old UI, this is achieved by legacy CSS styles.
--- a/react-front-end/tsrc/fileuploader/InlineFileUploader.tsx
+++ b/react-front-end/tsrc/fileuploader/InlineFileUploader.tsx
@@ -296,14 +296,15 @@ export const InlineFileUploader = ({
    */
   const buildActions = (file: UploadedFile | UploadingFile): UploadAction[] => {
     if (isUploadedFile(file)) {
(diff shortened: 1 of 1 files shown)


react-front-end/tsrc/fileuploader/InlineFileUploader.tsx
```diff
@@ -296,14 +296,15 @@ export const InlineFileUploader = ({
    */
   const buildActions = (file: UploadedFile | UploadingFile): UploadAction[] => {
     if (isUploadedFile(file)) {
+      if (!editable) return [];
```
edalex-ian, line 299: Good practice in C based syntax languages (so JS, TS, Java) to use curly braces even for simple statements like this to protect against accidental future editing causing modified logic errors.

```suggestion
      if (!editable) {
        return [];
      }
```

react-front-end/tsrc/fileuploader/InlineFileUploader.tsx
```diff
@@ -296,14 +296,15 @@ export const InlineFileUploader = ({
    */
   const buildActions = (file: UploadedFile | UploadingFile): UploadAction[] => {
```
edalex-ian, line 297: Gah, this function's become a bit of a meandering mess due to it's nested `if`s combined with ternaries for return values.

Could you please rework as part of this?

Something along the lines of:

```typescript
const buildActions = (file: UploadedFile | UploadingFile): UploadAction[] => {
  const uploadComplete = isUploadedFile(file);

  if(!uploadComplete) {
    return [
      {
        onClick: () => onCancel(file.localId),
        text: strings.cancel,
        icon: <CancelIcon />,
      },
    ];
  }
  else if(!editable) {
    // No further actions available
    return [];
  }
  else {
    const actions = ...

    return actions;
  }
}
```

Mind you, I still don't love that heavily imperative approach, but it kind of fits with what is there. That said though, if you come up with nice readable declarative approach (perhaps with some judicious use of fp-ts) then that'd be a bonus. :wink:

edalex-ian (changes requested):
Good working finding the fix @edalex-yinzi . Now just need to tidy things a bit. :wink: 

And are there any existing supporting tests for this? I can't remember...

react-front-end/tsrc/fileuploader/InlineFileUploader.tsx
```diff
@@ -296,14 +296,15 @@ export const InlineFileUploader = ({
...
       const basicAction = [
         {
           onClick: () => onDelete(file),
           text: strings.delete,
         },
       ];
-      const { id, editable } = file.fileEntry;
-      return editable
+      const { id, editable: fileEditable } = file.fileEntry;
+      return fileEditable
```
PenghaiZhang, line 307: Looks like the issue is caused by variable shadowing.

edalex-yinzi:
We have some tests that test the drop and drag function and attachments list for this.

react-front-end/tsrc/fileuploader/InlineFileUploader.tsx
edalex-yinzi: I refactored the func based on your demo and will come to see if we can use fp-ts in here later.

edalex-yinzi:
I refactored the func based on @edalex-ian 's demo and will come back to see if we can improve it by using fp-ts later.

edalex-ian (approved):

PenghaiZhang (approved):

edalex-yinzi:
@edalex-ian @PenghaiZhang 
Tried to extract each action and use fp-ts, but `actions` are based on the `file` type so it's hard to get rid of the type error.
image

Also get more error after adding getOrElse func.

I think we can merge it.

edalex-ian:
Heya @edalex-yinzi 

Thanks for the heads up about the target branch, I've tweaked it.

As for the code change (screenshot), that's not quite what we're after. So let's stick with what you've got for now. :wink:

PenghaiZhang:
Sorry I am late. At your first line of the pipe, I think you should use `O.fromPredicate` rather than `O.some`. And then in the `O.map` I am not sure where the `match` comes from

edalex-yinzi:
@PenghaiZhang from fp-ts boolean

PR #3644, opened 2021-12-12
bugfix: show Wizard control that uses an empty string as visibility scripting

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

Do not execute the visibility script is it's just an empty string. As a result, the Wizard control is visible.


edalex-ian (changes requested):
Good stuff @PenghaiZhang . :+1: 

But, be worthwhile (seeing we know this is a case in the wild) to add a test case for this over in WizardHelper.test.

edalex-ian (approved):
Thanks @PenghaiZhang :+1:

PR #3650, opened 2021-12-13
chore(deps): stop tika parsers transitive log4j dep

edalex-ian:
##### Checklist

- [ ] tests are included

##### Description of change

When we upgraded to Tika 2 (#3296) it turns out that the microsoft parser in
standard parsers brings in log4j including an SLF4J binding. This causes
some of our logging to then fail, as it uses SLF4J and picks up the
Log4J2 binding where we only configure Log4J 1.2.x.

This was only new in the version we're currently working on and have not
yet released - 2021.2.

I tested this manually to ensure that Microsoft Office files (tested with docx and doc) are still nicely indexed with searches for words in the content returning the items and highlighting the search result was in the attachments. This means that tika is still successfully extracting text from Microsoft files.


PenghaiZhang (approved):
GTG.

Issue #2873, opened 2021-04-12
New UI IMS Package display shows nested template, hides info about the error


SammyIsConfused:
**Describe the bug**
When a full page error occurs within an IMS viewer, it shows the template including the menu bar within the viewer - which can lead to nested equella windows.

**To Reproduce**

1. Go to an IMS attachment with a missing file referred to in the imsmanifest.xml and enter the viewer.
2. Click on the missing file from within the viewer.
3. Note that the 404 error page that displays contains the menu bar.
4. Note that the error page just reads 404 Not Found - no info about the request that failed.

**Expected behavior**
The error should not contain the menu bar. Also - 404 errors should display information about the request that failed - just as it does in the old UI.

**Screenshots**
![image](https://user-images.githubusercontent.com/24543345/114340647-631ccc80-9b9b-11eb-9130-429103f0f389.png)

In the old UI, this is the error. 

![image](https://user-images.githubusercontent.com/24543345/114340757-98c1b580-9b9b-11eb-8daf-79640c9ba4d8.png)

**Stacktrace**
If applicable, a stacktrace to help explain your problem.

**Platform:**

- OpenEquella Version: 2021.1+
- OS: Agnostic
- Browser: Agnostic

SammyIsConfused:
@cathfitz Carl has informed me that a client has asked about this issue again - should definitely go into 2021.2 and not later.

PR #3744, opened 2022-01-12
fix: hide menu bar for error page in IMS viewer

edalex-yinzi:
##### Checklist

- [ ] tests are included

##### Description of change

This is a draft PR to address #2873

There are two cases in old UI if there is a missing file:

https://user-images.githubusercontent.com/92769668/149067406-a5ddd62c-233c-4531-ae5d-4dbfee38e529.mp4

An index in the catalogue points to a missing file:
image

An internal link in IMS viewer points to a missing file:
image

But in New UI both cases will display the menu:
image

In old UI it actually uses param `temp.hn=true` to hide the menu (`RenderTemplate.java:394`).
And 404 page will directly return after the request `IntroToDBS_summary.html?temp.hn=true&temp.hb=true`.

But in new UI, `IntroToDBS_summary.html?temp.hn=true&temp.hb=true` will only return 404 code with some useless HTML content, and trigger a new request to the `LeagcyContent content/submit API`. 

If the URL points to a normal Legacy content, the server will find corresponding content from old UI (`LeagcyContentApi.scala:347`) and generate a `fullscreenMode` variable and pass it to React component (`LeagcyContentApi.scala:576`, `LeagcyContent.tsx:272`).

But if the URL is points to a missing file, the legacy content API will direct return a 404 response (`LeagcyContentApi.scala:346`).

And this 404 response will trigger a catch statement (`LegacyContent.tsx:292`) and finally execute a props method passed from `IndexPage.tsx:154`
image
image

So, currently, I just change the code in the frontend to make new UI has the same behaviour as old UI.

The solution idea is quite simple: if the error response contains the params 'temp.hn=true' let `fullscreenMode` props equal to "YES" otherwise keep undefined.

Here is the screenshot:
image

The next step is to hide the menu for the internal link. 

### DIFF
--- a/react-front-end/tsrc/legacycontent/LegacyContent.tsx
+++ b/react-front-end/tsrc/legacycontent/LegacyContent.tsx
@@ -16,11 +16,9 @@
  * limitations under the License.
  */
 import { CircularProgress, Grid } from "@material-ui/core";
-import Axios, { AxiosResponse } from "axios";
-import { pipe } from "fp-ts/function";
+import Axios from "axios";
 import { isEqual } from "lodash";
 import * as React from "react";
-import * as O from "fp-ts/Option";
 import { useContext } from "react";
 import { v4 } from "uuid";
 import {
@@ -120,18 +118,6 @@ interface LegacyContentSubmission {
    * Payload of the submission.
    */
   payload?: StateData;
-}
-
-/**
- * temp.hn and temp.hb are two possible params contains in the request and response
- * and are used in old UI to hide navigation (menu bar) and banner.
- * In new UI, for normal legacy content we have a template props `fullscreenMode`
- * received from server to hide menu bar, but not the error page.
- * Thus, it is used to set `fullscreenMode` for error page.
- */
-interface AxiosResponseConfigData {
-  "temp.hn"?: [string];
-  "temp.hb"?: [string];
 }
 
 export type SubmitResponse =
@@ -258,6 +244,24 @@ export const LegacyContent = React.memo(function LegacyContent({
     }
   }
 
+  /**
+   * temp.hn and temp.hb are two possible params contains in the request and response
+   * and are used in old UI to hide navigation (menu bar) and banner.
+   * In new UI, for normal legacy content we have a template props `fullscreenMode`
+   * received from server to hide menu bar, but not the error page.
+   * Thus, it is used to set `fullscreenMode` for error page.
+   */
+  function preUpdateFullscreenMode(submitValues: StateData) {
+    const isFullscreenMode = submitValues["temp.hn"]
+      ? submitValues["temp.hn"][0] === "true"
+      : false;
+
+    updateTemplate((tp) => ({
+      ...tp,
+      fullscreenMode: isFullscreenMode ? "YES" : tp.fullscreenMode,
+    }));
+  }
+
   function submitCurrentForm(
     fullScreen: boolean,
     scrollTop: boolean,
@@ -283,6 +287,8 @@ export const LegacyContent = React.memo(function LegacyContent({
       };
     }
 
+    preUpdateFullscreenMode(submitValues);
+
     submitRequest(toRelativeUrl(formAction || pathname), submitValues)
       .then((content) => {
         // Clear raw mode saved in local storage after a login request is resolved.
@@ -304,28 +310,11 @@ export const LegacyContent = React.memo(function LegacyContent({
         }
       })
       .catch((error) => {
-        pipe(
-          O.fromNullable(error.response),
-          O.map((e: AxiosResponse) => {
-            const configData: AxiosResponseConfigData = JSON.parse(
-              e.config.data
-            );
-
-            const isFullscreenMode = configData["temp.hn"]
-              ? configData["temp.hn"][0] === "true"
-              : false;
-
-            updateTemplate((tp) => ({
-              ...tp,
-              fullscreenMode: isFullscreenMode ? "YES" : undefined,
-            }));
-            return fromAxiosResponse(e);
-          }),
-          O.getOrElseW<ErrorResponse>(() => generateFromError(error)),
-          (e) => {
-            handleError(fullScreen, e);
-          }
-        );
+        const errorResponse: ErrorResponse =
+          error.response !== undefined
+            ? fromAxiosResponse(error.response)
+            : generateFromError(error);
+        handleError(fullScreen, errorResponse);
       })
       .finally(() => {
         if (formAction) {
--- a/react-front-end/tsrc/legacycontent/LegacyContent.tsx
+++ b/react-front-end/tsrc/legacycontent/LegacyContent.tsx
@@ -17,7 +17,7 @@
  */
 import { CircularProgress, Grid } from "@material-ui/core";
 import Axios from "axios";
-import { constFalse, pipe } from "fp-ts/function";
+import { pipe } from "fp-ts/function";
 import { isEqual } from "lodash";
 import * as O from "fp-ts/Option";
 import * as React from "react";
@@ -31,7 +31,11 @@ import {
 import { LEGACY_CSS_URL } from "../AppConfig";
 import { AppRenderErrorContext } from "../mainui/App";
 import { BaseOEQRouteComponentProps } from "../mainui/routes";
-import { templateDefaults, templatePropsForLegacy } from "../mainui/Template";
+import {
+  FullscreenMode,
+  templateDefaults,
+  templatePropsForLegacy,
+} from "../mainui/Template";
 import { deleteRawModeFromStorage } from "../search/SearchPageHelper";
 import { LegacyContentRenderer } from "./LegacyContentRenderer";
 import { getEqPageForm, legacyFormId } from "./LegacyForm";
@@ -253,19 +257,20 @@ export const LegacyContent = React.memo(function LegacyContent({
    * received from server to hide menu bar, but not the error page.
    * Thus, it is used to set `fullscreenMode` for error page.
    */
-  function preUpdateFullscreenMode(submitValues: StateData) {
+  const preUpdateFullscreenMode = (submitValues: StateData) => {
     pipe(
       O.fromNullable(submitValues["temp.hn"]),
-      O.map(([firstValue]) => firstValue === "true"),
-      O.getOrElse(constFalse),
-      (isFullscreenMode) => {
+      O.chain<string[], FullscreenMode>(([firstValue]) =>
+        firstValue === "true" ? O.some("YES") : O.none
+      ),
+      O.map((value) =>
         updateTemplate((tp) => ({
           ...tp,
-          fullscreenMode: isFullscreenMode ? "YES" : tp.fullscreenMode,
-        }));
-      }
+          fullscreenMode: value,
+        }))
+      )
     );
-  }
+  };
 
   function submitCurrentForm(
     fullScreen: boolean,
--- a/react-front-end/tsrc/legacycontent/LegacyContent.tsx
+++ b/react-front-end/tsrc/legacycontent/LegacyContent.tsx
@@ -17,7 +17,9 @@
  */
 import { CircularProgress, Grid } from "@material-ui/core";
 import Axios from "axios";
+import { constFalse, pipe } from "fp-ts/function";
 import { isEqual } from "lodash";
+import * as O from "fp-ts/Option";
 import * as React from "react";
 import { useContext } from "react";
 import { v4 } from "uuid";
@@ -252,14 +254,17 @@ export const LegacyContent = React.memo(function LegacyContent({
    * Thus, it is used to set `fullscreenMode` for error page.
    */
   function preUpdateFullscreenMode(submitValues: StateData) {
-    const isFullscreenMode = submitValues["temp.hn"]
-      ? submitValues["temp.hn"][0] === "true"
-      : false;
-
-    updateTemplate((tp) => ({
-      ...tp,
-      fullscreenMode: isFullscreenMode ? "YES" : tp.fullscreenMode,
-    }));
+    pipe(
+      O.fromNullable(submitValues["temp.hn"]),
+      O.map(([firstValue]) => firstValue === "true"),
+      O.getOrElse(constFalse),
+      (isFullscreenMode) => {
+        updateTemplate((tp) => ({
+          ...tp,
+          fullscreenMode: isFullscreenMode ? "YES" : tp.fullscreenMode,
+        }));
+      }
+    );
   }
 
   function submitCurrentForm(
--- a/react-front-end/tsrc/legacycontent/LegacyContent.tsx
+++ b/react-front-end/tsrc/legacycontent/LegacyContent.tsx
@@ -16,9 +16,11 @@
  * limitations under the License.
  */
 import { CircularProgress, Grid } from "@material-ui/core";
-import Axios from "axios";
+import Axios, { AxiosResponse } from "axios";
+import { pipe } from "fp-ts/function";
 import { isEqual } from "lodash";
 import * as React from "react";
+import * as O from "fp-ts/Option";
 import { useContext } from "react";
 import { v4 } from "uuid";
 import {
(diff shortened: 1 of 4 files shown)


edalex-ian (changes requested):
Heya Yinzi,

Great work finding out what is going on and find a possible fix.

A few things, what you have here would appear to directly address what's in #2873 and achieve parity with the Legacy UI. With that in mind, this is all that is required for now. (To address the other issue looks like a more fundamental fix and so we'd want that in another ticket.)

Next, as far as the fix goes a few things need attention:

1. First, needs a bit of work cosmetically - notice how the paper element looses any padding next to the IMS navigator - need bring that back in; 
2. This logic really shouldn't be down in the depths of Error.ts and tied to processing of Axios responses.  It needs to come back up higher and be (most likely) in legacy content. Perhaps even this needs to be stored in state somewhere because obviously the IMS viewer has already been forced to full screen mode at some point - so it seems odd that in the code processing the response, we're first checking the request again, and then modifying the state; lastly
3. If that kind of code was to remain, a bit more rigour should be placed around the processing of `response.config.data` - such as overlaying with an interface and I wonder why JSON.parse is being used for what wouldn't be JSON data (maybe)... Perhaps doco needed to as to what is going on...

edalex-yinzi:
> This logic really shouldn't be down in the depths of Error.ts and tied to processing of Axios responses. It needs to come back up higher and be (most likely) in legacy content. Perhaps even this needs to be stored in state somewhere because obviously the IMS viewer has already been forced to full screen mode at some point - so it seems odd that in the code processing the response, we're first checking the request again, and then modifying the state; lastly

In my understanding, the IMS viewer is actually in old UI, so it already left SPA, thus IMS viewer's fullscreenMode is not related to frontend state.

edalex-yinzi:
> If that kind of code was to remain, a bit more rigour should be placed around the processing of response.config.data - such as overlaying with an interface and I wonder why JSON.parse is being used for what wouldn't be JSON data (maybe)... Perhaps doco needed to as to what is going on...

The data: 
image

react-front-end/tsrc/mainui/ErrorPage.tsx
```diff
@@ -24,6 +24,8 @@ const useStyles = makeStyles((t) => ({
     display: "flex",
     justifyContent: "center",
     marginTop: t.spacing(8),
+    marginLeft: t.spacing(3),
+    marginRight: t.spacing(3),
```
edalex-yinzi, line 28: In the old UI the padding is 27px.

edalex-yinzi:
image

edalex-ian:
> > This logic really shouldn't be down in the depths of Error.ts and tied to processing of Axios responses. It needs to come back up higher and be (most likely) in legacy content. Perhaps even this needs to be stored in state somewhere because obviously the IMS viewer has already been forced to full screen mode at some point - so it seems odd that in the code processing the response, we're first checking the request again, and then modifying the state; lastly
> 
> Thanks Ian, In my understanding, the IMS viewer is actually in old UI, so it already left SPA, thus IMS viewer's fullscreenMode is not related to frontend state.

Yes, but.... Keep in mind we maintain various bits of state about whether we're viewing Legacy UI etc. For example, the control of whether we're in full screen.

edalex-yinzi:
Force push for correcting wrong typing in commit msg....

edalex-yinzi:
Gah! testBulkMetadataEdit failed.

react-front-end/tsrc/legacycontent/LegacyContent.tsx
```diff
@@ -290,11 +304,28 @@ export const LegacyContent = React.memo(function LegacyContent({
         }
       })
       .catch((error) => {
-        const errorResponse: ErrorResponse =
-          error.response !== undefined
-            ? fromAxiosResponse(error.response)
-            : generateFromError(error);
-        handleError(fullScreen, errorResponse);
+        pipe(
+          O.fromNullable(error.response),
```
PenghaiZhang, line 308: The type of `error` is `any`. Do we need the type guard `isAxiosError` here ?

react-front-end/tsrc/legacycontent/LegacyContent.tsx
```diff
@@ -290,11 +304,28 @@ export const LegacyContent = React.memo(function LegacyContent({
...
+          O.map((e: AxiosResponse) => {
+            const configData: AxiosResponseConfigData = JSON.parse(
```
PenghaiZhang, line 310: Do you want to try `fp-ts/Json` and then use `J.parse` which returns a `Either` ?

react-front-end/tsrc/legacycontent/LegacyContent.tsx
```diff
@@ -290,11 +304,28 @@ export const LegacyContent = React.memo(function LegacyContent({
...
       })
       .catch((error) => {
-        const errorResponse: ErrorResponse =
-          error.response !== undefined
-            ? fromAxiosResponse(error.response)
-            : generateFromError(error);
-        handleError(fullScreen, errorResponse);
+        pipe(
+          O.fromNullable(error.response),
+          O.map((e: AxiosResponse) => {
```
PenghaiZhang, line 309:
```suggestion
          O.map((response: AxiosResponse) => {
```

react-front-end/tsrc/legacycontent/LegacyContent.tsx
```diff
@@ -290,11 +304,28 @@ export const LegacyContent = React.memo(function LegacyContent({
...
+            const configData: AxiosResponseConfigData = JSON.parse(
+              e.config.data
+            );
+
+            const isFullscreenMode = configData["temp.hn"]
+              ? configData["temp.hn"][0] === "true"
```
PenghaiZhang, line 315: I have tried this code but all I got was an empty object from `error.response.config.data`. Perhaps I missed something ?

react-front-end/tsrc/legacycontent/LegacyContent.tsx
```diff
@@ -290,11 +304,28 @@ export const LegacyContent = React.memo(function LegacyContent({
...
+              : false;
+
+            updateTemplate((tp) => ({
+              ...tp,
+              fullscreenMode: isFullscreenMode ? "YES" : undefined,
+            }));
+            return fromAxiosResponse(e);
+          }),
+          O.getOrElseW<ErrorResponse>(() => generateFromError(error)),
```
PenghaiZhang, line 324: Is not `getOrElse` enough ?

react-front-end/tsrc/legacycontent/LegacyContent.tsx
```diff
@@ -290,11 +304,28 @@ export const LegacyContent = React.memo(function LegacyContent({
...
+          (e) => {
+            handleError(fullScreen, e);
```
PenghaiZhang, line 326: Do not need the curly braces.

PenghaiZhang (commented):
Thanks Yinzi. I have added some feedback for you to think.

react-front-end/tsrc/legacycontent/LegacyContent.tsx
edalex-ian, line 309: I was looking at the anonymous function you have inline here (and, specifically how it impacts readability) and was thinking you need to either create a function for it, or perhaps it'd be clearer if you simply do it more declaratively with a pipe rather then the switch to the imperative approach.

It'd be (really roughly) something like:

```typescript
O.map((response: AxiosResponse) => {
  updateTemplate((tp) => ({
    ...tp,
    fullscreenMode: pipe(
      response.config.data,
      J.parse,
      O.fromEither,
      O.chain(({temp:{hn}}) =>
        // Here you now need to still do a few things:
        // 1. Make sure it's not undefined
        // 2. Make sure it's a non-empty array
        // 3. Get the first element (A.head - which may also fulfil the above step)
        // 4. Then if `true` you can return "YES"
      ),
      O.toUndefined
    )
  );

  return fromAxiosResponse(e);
})
```

It is a bit odd though to have the two disjointed operations (update template, and then the actual map operation - `fromAxiosResponse`) in this O.map but maybe that's good enough here.

react-front-end/tsrc/mainui/IndexPage.tsx
```diff
@@ -153,7 +153,6 @@ export default function IndexPage() {
 
   const errorCallback = React.useCallback((err: ErrorResponse) => {
     errorShowing.current = true;
-    setTemplateProps((p) => ({ ...p, fullscreenMode: undefined }));
```
edalex-ian, line 156: Wonder if that causes problems anywhere else. :thinking:

edalex-ian (changes requested):
Thanks for the updates @edalex-yinzi . @PenghaiZhang has added some good comments, so I've added to those.

I've also got one thing for you to query... Rather than reaching around and trying to figure out what was used in the previous Axios call (i.e. `config.data`) you are now executing this from within `submitForm` so you're very close to where the values are known before they're submitted. So couldn't you actually execute the `updateTemplate` call up before you even call `submitRequest`? Or, should it perhaps be done _in_ `submitRequest`? May simplify how you process the additional params - possibly no JSON parsing required.

react-front-end/tsrc/legacycontent/LegacyContent.tsx
```diff
@@ -120,6 +122,18 @@ interface LegacyContentSubmission {
...
+ * temp.hn and temp.hb are two possible params contains in the request and response
+ * and are used in old UI to hide navigation (menu bar) and banner.
+ * In new UI, for normal legacy content we have a template props `fullscreenMode`
+ * received from server to hide menu bar, but not the error page.
+ * Thus, it is used to set `fullscreenMode` for error page.
+ */
+interface AxiosResponseConfigData {
+  "temp.hn"?: [string];
+  "temp.hb"?: [string];
+}
```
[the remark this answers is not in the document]
edalex-yinzi: The original data is quite odd: {"temp.hn": ["true"], "temp.hb":["true"]}

react-front-end/tsrc/mainui/ErrorPage.tsx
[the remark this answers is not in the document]
edalex-yinzi: Should be OK there is not much difference between appearance.

react-front-end/tsrc/legacycontent/LegacyContent.tsx
edalex-yinzi: The previous code is judging error is whether `Error` and `AxiosError`. 
But I wonder because the code block is actually in catch statement from an Axios request, is that means the error must be AxiosError type?

PenghaiZhang: No it didn't. It just checked whether `error.response` is undefined or not.

react-front-end/tsrc/legacycontent/LegacyContent.tsx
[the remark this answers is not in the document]
PenghaiZhang: I'ts hard to guess what `hn` and `hb` mean. Please add some JSDoc.

react-front-end/tsrc/mainui/IndexPage.tsx
edalex-yinzi: Not sure but the initial value is "No", and I also tried deleting this line before I made any other changes, nothing happened.

react-front-end/tsrc/legacycontent/LegacyContent.tsx
edalex-yinzi: It depends on whether the request url contains the params or not.

react-front-end/tsrc/legacycontent/LegacyContent.tsx
edalex-yinzi: Yes, it uses this attribute to judge and call different function `fromError` and `fromAxiosError`

edalex-ian (approved):
Nice, much simpler. I think it's good to go, what do you think @PenghaiZhang ?

react-front-end/tsrc/legacycontent/LegacyContent.tsx
```diff
@@ -244,6 +244,24 @@ export const LegacyContent = React.memo(function LegacyContent({
...
+   * In new UI, for normal legacy content we have a template props `fullscreenMode`
+   * received from server to hide menu bar, but not the error page.
+   * Thus, it is used to set `fullscreenMode` for error page.
+   */
+  function preUpdateFullscreenMode(submitValues: StateData) {
+    const isFullscreenMode = submitValues["temp.hn"]
+      ? submitValues["temp.hn"][0] === "true"
+      : false;
+
+    updateTemplate((tp) => ({
```
PenghaiZhang, line 259: `updateTemplate` is called regardless of whether `"temp.hn"` is there or not.  I am not sure if we need this ?

react-front-end/tsrc/legacycontent/LegacyContent.tsx
```diff
@@ -244,6 +244,24 @@ export const LegacyContent = React.memo(function LegacyContent({
...
 
+  /**
+   * temp.hn and temp.hb are two possible params contains in the request and response
+   * and are used in old UI to hide navigation (menu bar) and banner.
+   * In new UI, for normal legacy content we have a template props `fullscreenMode`
+   * received from server to hide menu bar, but not the error page.
+   * Thus, it is used to set `fullscreenMode` for error page.
+   */
+  function preUpdateFullscreenMode(submitValues: StateData) {
+    const isFullscreenMode = submitValues["temp.hn"]
```
PenghaiZhang, line 255: If you like 'fp-ts' you may have a look this
```
    pipe(
      O.fromNullable(submitValues["temp.hn"]),
      O.map(([firstValue]) => firstValue === "true"),
      O.getOrElse(constFalse),
      (isFullscreenMode) => {
        updateTemplate((tp) => ({
          ...tp,
          fullscreenMode: isFullscreenMode ? "YES" : tp.fullscreenMode,
        }));
      }
    )
```

react-front-end/tsrc/legacycontent/LegacyContent.tsx
```diff
@@ -244,6 +246,27 @@ export const LegacyContent = React.memo(function LegacyContent({
...
 
+  /**
+   * temp.hn and temp.hb are two possible params contains in the request and response
+   * and are used in old UI to hide navigation (menu bar) and banner.
+   * In new UI, for normal legacy content we have a template props `fullscreenMode`
+   * received from server to hide menu bar, but not the error page.
+   * Thus, it is used to set `fullscreenMode` for error page.
+   */
+  function preUpdateFullscreenMode(submitValues: StateData) {
+    pipe(
```
edalex-ian, line 257: Can be nice and simple like:

```suggestion
  const preUpdateFullscreenMode = (submitValues: StateData) =>
    pipe(
```

react-front-end/tsrc/legacycontent/LegacyContent.tsx
```diff
@@ -244,6 +246,27 @@ export const LegacyContent = React.memo(function LegacyContent({
...
+      O.fromNullable(submitValues["temp.hn"]),
+      O.map(([firstValue]) => firstValue === "true"),
+      O.getOrElse(constFalse),
+      (isFullscreenMode) => {
+        updateTemplate((tp) => ({
```
edalex-ian, line 262: So just keep in mind this is still getting called every-time. Does that need to occur? If it does get called every-time, I'm concerned of unnecessary re-renders and/or screen flickers.

Could you please confirm?

If you see it's best _not_ to call everytime, how about changing the above `O.map` to `O.chain` and simply return `some` if you do need to call it, and then you can have a subsequent `O.map` to do the call to updateTemplate with a hard coded YES.

edalex-ian (changes requested):
Almost there, just a couple of tweaks and really need to make sure there's not an unnecessary call to updateTemplate.

react-front-end/tsrc/legacycontent/LegacyContent.tsx
edalex-yinzi: The props will be updated after submitting the form anyway and currently no screen flickers， but it would absolutely be better not to call everytime. :+1:

edalex-ian (approved):
Excellent, thanks @edalex-yinzi . You do still have some redundant curly braces in `preUpdateFullscreenMode` - so that's something to focus on going forward. But I reckon, let's get it merged.

PR #3756, opened 2022-01-17
Feature/upgrade log2j to v2

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

This is the second PR for the work of consolidating logging layer. And this PR is focused on updating Log4J to the latest version.

Overall,  the changes can be categorised into 5 groups.

1. Update Log4J to the latest version and address SBT dependency conflicts.
2. Drop code that's used to implement the custom daily rolling appender and HTML layout.
3. Rewrite Log4j configuration files in YAML format.
4. Update existing code to use the the new YAML files.
5. Update JPF plugin files to drop useless Log4j relevant configurations.

Testing was done on my local instance to confirm all logs can be properly generated.

1. All the directories in folder `logs` were created.
image

2. Archived resource centre logs were stored in folders representing different days.
image

3. The file size based rolling policy was working.
image


PenghaiZhang:
Failded on `OAIEndpointTest` because the dependency `Oclc Harvester2` still uses Log4j 1.2.

PenghaiZhang:
`Oclc Harvester2` was last updated in 2009!!!

PenghaiZhang:
There is not updates for Oclc Harvester2 at all.

The alternative may be using OAI harvest manager which has a new release for the famous Log4J bug.

But...I can't find OAI harvest manager from maven repo.

[This is the link ](https://github.com/clarin-eric/oai-harvest-manager)

PenghaiZhang:
I am not familiar with the OAI stuff, but I think what we can do is either replacing that OAI lib with another one which is not dependent on the old Log4J or bringing Log4J back only in this project `OldTest`.

Installer/data/server/learningedge-config/learningedge-log4j.yaml
```diff
@@ -0,0 +1,100 @@
...
+#              .....
+#
+
+Configuration:
+  status: warn
+
+  appenders:
+    RollingFile:
+      - name: FILE
+        fileName: "${install.path}/logs/resource-centre/application.html"
```
edalex-ian, line 21: The old templates had the filenames in properties right at the top of the file to make it easy for people to tweak. Perhaps do the same with this YAML format too?

Installer/data/server/learningedge-config/learningedge-log4j.yaml
```diff
@@ -0,0 +1,100 @@
...
+        filePattern: "${install.path}/logs/resource-centre/%d{yyyy-MM-dd}/application-%i.html"
+        immediateFlush: true
+        append: true
+        Filters:
+          ThresholdFilter:
+            level: DEBUG
+          RegexFilter:
+            regex: ".*(HHH90000022|HHH90000014).*"
```
edalex-ian, line 29: Please add a comment defining what this is filtering.

Installer/data/server/manager/all/log4j.yaml
```diff
@@ -0,0 +1,53 @@
...
+  status: warn
+
+  properties:
+    property:
+      - name: filePath
+        value: "../logs/equella-manager"
+      - name: fileName
+        value: "${filePath}/services.html"
+      - name: filePattern
+        value: "${filePath}/%d{yyyy-MM-dd}/services-%i.html"
```
edalex-ian, line 11: Just to be sure, these don't get processed by any templating in the installer do they?

Source/Server/equellaserver/build.sbt
```diff
@@ -116,7 +116,9 @@ libraryDependencies ++= Seq(
   "javax.xml"                 % "jaxrpc"                   % "1.1",
   "jdom"                      % "jdom"                     % "1.1",
   "com.github.equella.jpf"    % "jpf"                      % "1.0.7",
-  "log4j"                     % "log4j"                    % "1.2.17",
+  "org.apache.logging.log4j"  % "log4j"                    % "2.17.1",
+  "org.apache.logging.log4j"  % "log4j-slf4j-impl"         % "2.17.1",
+  "org.apache.logging.log4j"  % "log4j-1.2-api"            % "2.17.1",
```
edalex-ian, line 121: Hmm, shouldn't this just be:

```suggestion
  "org.apache.logging.log4j"  % "log4j-api"            % "2.17.1",
```

And, do we need the API dep? Kind of feels like it should be a transitive for log4j-slf4j-impl. :thinking:

edalex-ian, line 121: Be good to pop `2.17.1` into a `val`. Maybe place that in CommonSettings... Or maybe place the whole dependency specifier in there for easy re-use - I see you also need to specify in the installer. :thinking:

Source/Tools/UpgradeInstallation/build.sbt
```diff
@@ -1,12 +1,15 @@
...
+  "org.apache.logging.log4j" % "log4j-slf4j-impl" % "2.17.1",
+  "org.apache.logging.log4j" % "log4j"            % "2.17.1",
   xstreamDep,
   "commons-configuration" % "commons-configuration" % "1.10",
   "commons-io"            % "commons-io"            % "2.11.0",
-  "commons-lang"          % "commons-lang"          % "2.6"
+  "commons-lang"          % "commons-lang"          % "2.6",
+  // Need these two jackson deps to allow processing log4j yaml config files.
+  "com.fasterxml.jackson.core"       % "jackson-databind"        % "2.11.3",
+  "com.fasterxml.jackson.dataformat" % "jackson-dataformat-yaml" % "2.11.3"
```
edalex-ian, line 12: How about use `jacksonVersion`?

```suggestion
  "com.fasterxml.jackson.core"       % "jackson-databind"        % jacksonVersion,
  "com.fasterxml.jackson.dataformat" % "jackson-dataformat-yaml" % jacksonVersion
```

docker/learningedge-log4j.yaml
```diff
@@ -0,0 +1,103 @@
...
+
+  appenders:
+    Console:
+      name: STDOUT
+      target: SYSTEM_OUT
+      PatternLayout:
+        Pattern: "%d{ABSOLUTE} %-5p [%c{10}] %m%n"
+      Filters:
+        RegexFilter:
+          regex: ".*(HHH90000022|HHH90000014).*"
```
edalex-ian, line 17: As per the other, please add comment.

docker/learningedge-log4j.yaml
```diff
@@ -0,0 +1,103 @@
...
+          onMatch: "DENY"
+          onMismatch: "NEUTRAL"
+        ThresholdFilter:
+          level: "DEBUG"
+          onMatch: "ACCEPT"
+          onMismatch: "DENY"
+
+    RollingFile:
```
edalex-ian, line 25: Please add the RegexFilter for this appender as well.

edalex-ian (changes requested):
Great work @PenghaiZhang . Main feedback is for the dependency declarations add some variables to keep things DRY. :wink:

Installer/data/server/manager/all/log4j.yaml
PenghaiZhang: No, they don't. They are processed by Log4j Lookups.

PenghaiZhang: image

You can see the install leaves these as they are.

Source/Server/equellaserver/build.sbt
PenghaiZhang: Actually, `log4j-1.2-api` is a bridge which allows log4j2 to be used with some old log4j code. 

For example, in `ItemServiceImpl` there is usage of `MDC` which is replaced by `Thread Context Map`. And in 
`AuthenticatedThread` there is usage of `NDC` which is replaced by `Thread Context Stack`. 

I did try to update the code to use the new stuff, but I stopped and then forgot...I will try again.

Source/Tools/UpgradeInstallation/build.sbt
PenghaiZhang: The `jacksonVersion` created in `equellaserver/build.sbt` is not available here. 

`jackson-dataformat-yaml` is only used `UpgradeInstallation` so what about create a val here ?

Source/Server/equellaserver/build.sbt
PenghaiZhang: I have moved these to `CommonSettings`.

Installer/data/server/manager/all/log4j.yaml
PenghaiZhang: After lots of digging, I found `learningedge-log4j.yaml`  does get processed by the installer.

The class is `Interpreter.java`.

Source/Tools/UpgradeInstallation/build.sbt
edalex-ian: Sounds like a plan. :+1:

Installer/data/server/learningedge-config/learningedge-log4j.yaml
PenghaiZhang: As discussed, leave this file as it's now due to the complexity of refactoring how the installer processes templates.

edalex-ian (approved):
GTG! Thanks @PenghaiZhang :bow:

PR #3764, opened 2022-01-19
build(docker): Upgrade base image to openjdk 8u312-jre

edalex-ian:
##### Checklist

- [ ] tests are included

##### Description of change

Main task here was to try and get us to an up to date and known base image - we pinned it a few years back to avoid an issue, but since then have been unsure how to accept renovate pull requests to update. To update the image we also needed to address the deprecation and removal of libav-tools from debian based distros (and this base image is debian).

While doing that I also reviewed and revised some of the docker and simplified the expected filename for the installer.

Undertook manual smoke testing locally, and looks to still function the same (including video previews to ensure I'd done the symbolic links correctly for ffmpeg -> libav).

(I did have an issue getting the Zookeeper part of the docker-compose working even before the upgrade - so at some point we need to loop back around and validate that is still happy.)

This will resolve #2974 

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM openjdk@sha256:0e25c8428a56e32861fe996b528a107933155c98fb2a9998a4a4e9423aad734d as baseequella
+FROM openjdk:8u312-jre as baseequella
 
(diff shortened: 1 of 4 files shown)


PenghaiZhang (approved):
Thanks for updating the Docker REAMDE!  LGTM!

Issue #3769, opened 2022-01-20
New Search UI is broken when a custom resource uses latest version


PenghaiZhang:
This issue is very very similar to #3558 so I suggest to read #3558 first.

The difference between this issue and the previous one is a custom resource attachment links to another **attachments**.

**Expected behavior**
The new Search UI should not break.

**Screenshots**
![image](https://user-images.githubusercontent.com/47203811/150255335-9c606ea4-ef00-47ad-854b-a8047414c938.png)

**Stacktrace**
If applicable, a stacktrace to help explain your problem.

**Platform:**

- OpenEquella Version: [2021.2.1]
- OS: [Ubuntu 21]
- Browser [Chrome]

**Additional context**
N/A

PR #3777, opened 2022-01-21
bugfix: ensure Custom Attachment uses the correct latest version

PenghaiZhang:
#3769 

##### Checklist

- [ ] tests are included

##### Description of change

We have added a fix (#3563)for the bug caused by a custom attachment which points to the latest version of another resource. 

However, that fix only works when the linked resource is an Item. When the resource is an Attachment, the issue can still happen such as when checking the MIME type of the linked attachment. (#3769)

So this PR adds some improvements to make sure whenever a linked resource is used in search2 API, the genuine latest version is used. 

During implementation, I realised that `SearchHelper` has been getting bigger and bigger. So I move some functions to the Scala version `AttachmentResourceService` and some to `AttachmentHelper`.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/service/AttachmentResourceService.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/service/AttachmentResourceService.scala
@@ -1,0 +1,134 @@
+package com.tle.web.api.search.service
+
+import com.tle.beans.item.attachments.{CustomAttachment, FileAttachment}
+import com.tle.beans.item.{ItemId, ItemIdKey, ItemKey}
+import com.tle.legacy.LegacyGuice
+import com.tle.web.api.interfaces.beans.AbstractExtendableBean
+import com.tle.web.api.item.equella.interfaces.beans.{
+  AbstractFileAttachmentBean,
+  FileAttachmentBean
+}
+import com.tle.web.api.item.interfaces.beans.AttachmentBean
+import com.tle.web.controls.resource.ResourceAttachmentBean
+
+/**
+  * Service layer that deals with Attachment related business logic.
+  */
+object AttachmentResourceService {
+
+  /**
+    * Extract the mimetype for AbstractExtendableBean.
+    */
+  def getMimetypeForAttachment[T <: AbstractExtendableBean](bean: T): Option[String] =
+    bean match {
+      case file: AbstractFileAttachmentBean =>
+        Some(LegacyGuice.mimeTypeService.getMimeTypeForFilename(file.getFilename))
+      case resourceAttachmentBean: ResourceAttachmentBean =>
+        Some(
+          LegacyGuice.mimeTypeService.getMimeTypeForResourceAttachmentBean(resourceAttachmentBean))
+      case _ => None
+    }
+
+  /**
+    * If the attachment is a file, then return the path for that attachment.
+    *
+    * @param attachment a potential file attachment
+    * @return the path of the provided file attachment
+    */
+  def getFilePathForAttachment(attachment: AttachmentBean): Option[String] =
+    attachment match {
+      case fileAttachment: FileAttachmentBean => Option(fileAttachment.getFilename)
+      case _                                  => None
+    }
+
+  /**
+    * Use the `description` from the `Attachment` behind the `AttachmentBean` as this provides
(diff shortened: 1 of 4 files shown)


Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/service/AttachmentResourceService.scala
```diff
@@ -0,0 +1,134 @@
...
+  /**
+    * Determines if a given customAttachment is invalid. Required as these attachments can be recursive.
+    * @param customAttachment The attachment to check.
+    * @return If true, this attachment is broken.
+    */
+  def isCustomAttachmentBroken(customAttachment: CustomAttachment): Boolean = {
+    val uuid    = customAttachment.getData("uuid").asInstanceOf[String]
+    val version = customAttachment.getData("version").asInstanceOf[Int]
+
+    val key = new ItemId(uuid, getLatestVersionForCustomAttachment(version, uuid))
```
PenghaiZhang, line 95: Unfortunately,  we need to check the version in here again. This is because what is processed here is an `Attachment` rather than an `AttachmentBean`. Even though we sanitise the bean in earlier stage, that does not make any difference on `Attachment`.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/service/AttachmentResourceService.scala
```diff
@@ -0,0 +1,152 @@
...
+  AbstractFileAttachmentBean,
+  FileAttachmentBean
+}
+import com.tle.web.api.item.interfaces.beans.AttachmentBean
+import com.tle.web.controls.resource.ResourceAttachmentBean
+
+/**
+  * Service layer that deals with Attachment related business logic.
+  */
+object AttachmentResourceService {
```
edalex-ian, line 35: So there's already an `AttachmentResourceService` over in `com.tle.web.viewurl.attachments.AttachmentResourceService` (with the implementation at `com.tle.web.viewurl.attachments.impl.AttachmentResourceServiceImpl`). So that raises two questions:

1. Should this be merged with that?; and
2. Is this in the right package?

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/service/AttachmentResourceService.scala
```diff
@@ -0,0 +1,152 @@
...
+    * the value more commonly seen in the LegacyUI. And specifically uses any tweaks done for
+    * Custom Attachments - such as with Kaltura where the Kaltura Media `title` is pushed into
+    * the `description` rather than using the optional (and multi-line) Kaltura Media `description`.
+    *
+    * @param itemKey the details of the item the attachment belongs to
+    * @param attachmentUuid the UUID of the attachment
+    * @return the description for the attachment if available
+    */
+  def getAttachmentDescription(itemKey: ItemKey, attachmentUuid: String): Option[String] =
+    Option(LegacyGuice.itemService.getAttachmentForUuid(itemKey, attachmentUuid).getDescription)
```
edalex-ian, line 73: This feels more like AttachmentHelper to me.

edalex-ian (changes requested):
It was a bit tricky to review due to the reshuffle, so just to confirm that the key change here is:

- New function - `sanitiseAttachmentBean`; and
- Calling that function in `SearchHelper` on line 293.

And all the other functions stayed as were.

One thing I do wonder, is how will this be for backporting - will the reshuffle cause problems, or should it be pretty straight forward?

But onto specifics, the main feedback is the structure with the new Service. Whether it's in the right package, and how it relates to the existing service.

Other than all that though, I think the fix seems fine. :grin:

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/service/AttachmentResourceService.scala
PenghaiZhang: Were you suggesting that we rewrite these Scala methods in Java ?

edalex-ian:
As discussed, how about:

- rename your AttachmentResourceService to AttachmentService
- and place AttachmentService in the core plugin at `com.tle.core.item.service`

edalex-ian (approved):
Excellent, thanks @PenghaiZhang . LGTM! :+1:

PR #3802, opened 2022-02-03
fix: don't convert LTI custom parameters to string if no parameters provided 

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

New Search UI is broken when an Item has an LTI attachment. This is because the LTI custom params are null 
and calling `customParams.size()` results in NPE. 

So we should properly check if the custom params are provided or not by calling `!Check.isEmpty`.


edalex-ian (approved):
Good ol' NPE. Classic. :+1:

PR #3808, opened 2022-02-09
refactor: don't filter entities by ACL in `PagedResults`

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

Before 2020.2, the value of 'available' in a PagedResult is the full number of available
entities, which is actually not correct because ACL is not taken into account. We fixed
this in 2020.2. However, we found the fix can cause significant performance issue due to
filtering entities by ACL. So we decided to revert the fix and use the full number of
available entities again.


edalex-ian (approved):
LGTM. :+1:

PR #3833, opened 2022-02-11
Feature/rewrite view count dao

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

This PR adds the Service and DAO layers written in Java for Item and Attachment View Count. 

You will see each of `ViewcountItem` and `ViewcountAttachment` has its own DAO but they share one Service. I tried to implement one DAO only, but that seemed to bring in more complexity and did not make any improvement. 

In the Service class, you will also see some duplicated code.

Here is the list of changes:

1. Adding named queries to the entity classes.
2. Adding DAOs and Service.
3. Replacing all the use of `ViewCountJavaDao` and then dropping it.

I have tested locally all affected pages (including advanced scripting) and I can confirm the view counts calculated by the new implementation are correct.


Source/Plugins/Core/com.equella.core/plugin-jpf.xml
```diff
@@ -6001,6 +6001,12 @@
   <extension plugin-id="com.tle.core.hibernate" point-id="domainObjects" id="facetedSearchClassificationObject">
     <parameter id="class" value="com.tle.core.facetedsearch.bean.FacetedSearchClassification" />
   </extension>
+  <extension plugin-id="com.tle.core.hibernate" point-id="domainObjects" id="ViewcountItem">
+    <parameter id="class" value="com.tle.beans.viewcount.ViewcountItem" />
+  </extension>
+  <extension plugin-id="com.tle.core.hibernate" point-id="domainObjects" id="ViewcountAttachment">
+    <parameter id="class" value="com.tle.beans.viewcount.ViewcountAttachment" />
+  </extension>
```
edalex-ian, line 6009: Is this required for every `@Entity`?

Source/Plugins/Core/com.equella.core/src/com/tle/core/item/dao/impl/ItemDaoImpl.java
```diff
@@ -506,7 +507,7 @@ public String getNameForId(long id) {
   @Override
   public void delete(Item entity) {
     itemLockingDao.deleteForItem(entity);
-    ViewCountJavaDao.deleteForItem(entity.getItemId());
+    viewCountService.deleteViewCount(CurrentInstitution.get(), entity.getItemId());
```
edalex-ian, line 510: Should this be here, or possibly higher in an ItemService layer?

Source/Plugins/Core/com.equella.core/src/com/tle/core/item/helper/ItemDetailsHelper.java
```diff
@@ -83,7 +84,7 @@ public void load(PropBagEx xml, Item item) {
     setNode(xml, "/rating/@average", item.getRating());
 
     if (security.hasPrivilege(item, Privilege.VIEW_VIEWCOUNT)) {
-      final int views = ViewCountJavaDao.getSummaryViewCount(item.getItemId());
+      final int views = viewCountService.getItemViewCount(item.getItemId());
```
edalex-ian, line 87: Maybe it's just naming, but is a 'SummaryViewCount' the same as an 'ItemViewCount'?

Source/Plugins/Core/com.equella.core/src/com/tle/core/viewcount/service/ViewCountService.java
```diff
@@ -0,0 +1,100 @@
...
+import java.util.List;
+
+public interface ViewCountService {
+  /**
+   * Get the view count of an Item.
+   *
+   * @param itemKey The Item's unique key.
+   */
+  int getItemViewCount(ItemKey itemKey);
+  /**
```
edalex-ian, line 35: I'll only make this whitespace suggestion here, but seeing they're methods can you please leave a line between the method definition and the Javadoc?

```suggestion

  /**
```

Source/Plugins/Core/com.equella.core/src/com/tle/core/viewcount/service/ViewCountService.java
```diff
@@ -0,0 +1,100 @@
...
+   *
+   * @param col The collection for which to get the total Attachment view count.
+   */
+  int getAttachmentViewCountForCollection(ItemDefinition col);
+  /**
+   * Add or update an instance of {@link ViewcountItem} for an Item.
+   *
+   * @param viewcountItem The instance of ViewcountItem to be updated.
+   */
+  void setItemViewCount(ViewcountItem viewcountItem);
```
edalex-ian, line 71: Being that this is a service layer, should it be more along the lines of:

```suggestion
  void setItemViewCount(ItemKey item, int viewCount);
```

Source/Plugins/Core/com.equella.core/src/com/tle/core/viewcount/service/ViewCountService.java
```diff
@@ -0,0 +1,100 @@
...
+  /**
+   * Add or update an instance of {@link ViewcountAttachment} for an Attachment.
+   *
+   * @param viewcountAttachment The instance of ViewcountAttachment to be updated.
+   */
+  void setAttachmentViewCount(ViewcountAttachment viewcountAttachment);
```
edalex-ian, line 77: Same feedback as per `setItemViewCount` above.

Source/Plugins/Core/com.equella.core/src/com/tle/core/viewcount/service/ViewCountServiceImpl.java
```diff
@@ -0,0 +1,152 @@
...
+    return new ViewcountAttachmentId(
+        CurrentInstitution.get().getDatabaseId(),
+        itemKey.getUuid(),
+        itemKey.getVersion(),
+        attachmentUuid);
+  }
+
+  private Criterion restrictedByInstitution(Institution institution) {
+    return Restrictions.eq("id.inst", institution.getDatabaseId());
+  }
```
edalex-ian, line 62: I'm surprised there isn't already this idea centralised in some kind of common place. :thinking:

Source/Plugins/Core/com.equella.core/src/com/tle/core/viewcount/service/ViewCountService.java
```diff
@@ -0,0 +1,100 @@
...
+   */
+  void incrementItemViewCount(ItemKey itemKey);
+  /**
+   * Increase an Attachment's view count by 1. If the Attachment does not have any view count yet,
+   * initialise the count to 1.
+   *
+   * @param itemKey The Item's unique key.
+   * @param attachmentUuid UUID of the Attachment.
+   */
+  void incrementAttachmentViewCount(ItemKey itemKey, String attachmentUuid);
```
edalex-ian, line 92: :thinking: 

Could be rather useful if these increment functions returned the updated value.

Source/Plugins/Core/com.equella.core/src/com/tle/core/viewcount/service/ViewCountServiceImpl.java
```diff
@@ -0,0 +1,152 @@
...
+
+    setItemViewCount(viewcountItem);
+  }
+
+  @Override
+  public void incrementAttachmentViewCount(ItemKey itemKey, String attachmentUuid) {
+    ViewcountAttachmentId id = buildViewCountAttachmentId(itemKey, attachmentUuid);
+
+    ViewcountAttachment viewcountAttachment =
+        Optional.ofNullable(attachmentViewCountDao.findById(id))
```
edalex-ian, line 98: To solve this duplication, why not have the two viewcount entities share a common abstract base class entity?

edalex-ian (commented):
Looking good @PenghaiZhang , mostly just a few queries and ponderings.

Source/Plugins/Core/com.equella.base/src/com/tle/beans/viewcount/ViewcountAttachment.java
```diff
@@ -22,11 +22,25 @@
...
  * Attachment.
  */
+@NamedQuery(
+    name = "getAttachmentViewCountForCollection",
+    query =
+        "SELECT sum(vca.count) FROM ViewcountAttachment vca "
+            + "inner join com.tle.beans.item.attachments.Attachment a on vca.id.attachment = a.uuid "
+            + "inner join com.tle.beans.item.Item i on a.item.id = i.id "
+            + "inner join com.tle.beans.entity.BaseEntity be on be.id = i.itemDefinition.id "
+            + "WHERE be.id= :collectionId")
```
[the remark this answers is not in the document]
PenghaiZhang: The previous implementation did not limit it by institution, but I can add the limit.

Source/Plugins/Core/com.equella.core/plugin-jpf.xml
PenghaiZhang: Yes, I think so. Without this, there will be an HibernateSystemException complaining "Unable to locate persister".

Source/Plugins/Core/com.equella.core/src/com/tle/core/item/dao/impl/ItemDaoImpl.java
PenghaiZhang: I did add this in ItemService at the beginning. Then I found this method is directly called another place (ItemConverter) so I changed to here.

PenghaiZhang: An alternative is to use the two view count DAOs here rather than ViewCountService.

Source/Plugins/Core/com.equella.core/src/com/tle/core/item/helper/ItemDetailsHelper.java
PenghaiZhang: Yes. They are the same thing. I think `ItemView` is more consistent.

Source/Plugins/Core/com.equella.core/src/com/tle/core/item/dao/impl/ItemDaoImpl.java
edalex-ian: Possibly the ItemConverter then should also be using the ItemService - it is odd it is using the Dao's directly and having to deal with the transactions, flushing and also the ItemLockingDao (which one would argue is a logic concern and so should be encapsulated in a service layer.

But alas, we don't have scope to also rewrite the ItemConverter. :cry: 

So for now maybe it is best this uses the two DAO's directly in this context. :disappointed:

Source/Plugins/Core/com.equella.core/src/com/tle/core/viewcount/service/ViewCountService.java
PenghaiZhang: I have updated this.

Source/Plugins/Core/com.equella.core/src/com/tle/core/viewcount/service/ViewCountServiceImpl.java
PenghaiZhang: There is a `enumerateAll` which can return a list of entities limited by current institution. However, to use it, the entity must have a field whose type is `Institution` instead of `long`.

edalex-ian: Ah well, not to worry. Next time. :wink:

Source/Plugins/Core/com.equella.core/src/com/tle/core/viewcount/service/ViewCountService.java
PenghaiZhang: Ah I can see the issue I used to have.  With the signature of `(ItemKey item, int viewCount)`, we need to create a new instance of `ViewcountItem`, and then calling `dao.save` will trigger this exception.
`DuplicateKeyException: A different object with the same identifier value was already associated with the session`

PenghaiZhang: The reason is `setAttachmentViewCount` is called within `incrementAttachmentViewCount` which already fined the object by ID.

PenghaiZhang: Let's have some discussions for this.

Source/Plugins/Core/com.equella.core/src/com/tle/core/viewcount/service/ViewCountServiceImpl.java
PenghaiZhang: In the end I found this does not bring in much benefit...

PenghaiZhang:
Sorry. Put a wrong commit message so did a force-push.

edalex-ian (approved):
Looks good thanks @PenghaiZhang :+1:

PR #3837, opened 2022-02-15
chore(deps): partial revert of ansi-regex PR #3814

edalex-ian:
##### Checklist

- [ ] tests are included

##### Description of change

Since merging #3814 we've been having GHA build issues. This should address them.

It's unclear how this really got into the lock file in this fashion, however without builds fail in pristine environments (like openjdk
docker containers and GHA) - but still work on CodeBuild for some (ominous) reason.

I did go down the path of installing the ansi-regex package explicitly (no at v6.0.1) and that worked. However, without understanding the tangled web of why this is there, I thought best to simply revert what we needed from #3814 and get things working again.

Also noted (as a red-herring) there is a slight bug in the GHA YAML specification meaning we're not using the version of Node/NPM for the functional tests - just what ever the default is.

(I did test the build of the autotest/IntegTester/ps in a bare openjdk image and all worked. So hopefully the rest now will.)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -22,8 +22,6 @@ jobs:
   build_and_check:
     runs-on: ubuntu-latest
     steps:
-      # Using v1 (rather than v2) through-out this workflow due to issue:
-      # https://github.com/actions/checkout/issues/237
       - uses: actions/checkout@v2
 
       - name: NPM Cache
@@ -244,6 +242,8 @@ jobs:
           distribution: adopt
           java-version: 8
 
+      - uses: actions/checkout@v2
+
       - name: Read .nvmrc
         run: echo ::set-output name=NVMRC::$(cat .nvmrc)
         id: nvm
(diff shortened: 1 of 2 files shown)


edalex-ian:
So it appears to have solved the issue, but now we're getting the odd failures on LoginNotice which started once we moved to ubuntu-latest.

I propose this should merge and then we need to look closely at LoginNotice tests.

PenghaiZhang (approved):
LGTM!

.github/workflows/ci.yaml
```diff
@@ -244,6 +242,8 @@ jobs:
           distribution: adopt
           java-version: 8
 
+      - uses: actions/checkout@v2
```
PenghaiZhang, line 245: I can see examples from `setup-node` where the use of `checkout@v2` must be declared before `setup-node`. But I am not quite sure why is that. Perhaps `setup-node` is dependent on `checkout@v2`  ?

edalex-ian: Typically you want to 'setup NodeJS' based on your `.nvmrc` file which is stored in your repository. So to be able to gain access to that file, first you need to get it. You do that by a checkout of your repository. So the steps are basically:

1. Get your .nvmrc - i.e. `git clone` /  `checkout@v2`;
2. Read the .nvmrc in so that it can be used by `setup-node`; then
3. execute `setup-node`.

PenghaiZhang: Oh yes I forgot we need to access nvmrc first! Thanks Ian!

PR #3838, opened 2022-02-16
fix: correctly set base url in error page under ssl env

edalex-yinzi:
##### Checklist

- [ ] tests are included

##### Description of change
Get schema based on both request and URI.

### DIFF
--- a/Source/Plugins/Core/com.equella.serverbase/src/com/tle/core/services/impl/UrlServiceImpl.java
+++ b/Source/Plugins/Core/com.equella.serverbase/src/com/tle/core/services/impl/UrlServiceImpl.java
@@ -105,8 +105,7 @@ public class UrlServiceImpl implements UrlService {
    * In order to correctly set up schema for an internal forward request, such as a redirect request
    * to error.do, this func will overwrite the schema based on the request and base institution url.
    *
-   * @param request HttpServletRequest
-   * @return UriBuilder
+   * @return UriBuilder with new schema
    */
   private UriBuilder getUriBuilderForRequest(HttpServletRequest request) {
     URI uri = getBaseInstitutionURI();
--- a/Source/Plugins/Core/com.equella.serverbase/src/com/tle/core/services/impl/UrlServiceImpl.java
+++ b/Source/Plugins/Core/com.equella.serverbase/src/com/tle/core/services/impl/UrlServiceImpl.java
@@ -101,10 +101,13 @@ public class UrlServiceImpl implements UrlService {
     return qbuilder;
   }
 
-  // In order to correctly set up schema for an internal forward request,
-  // such as a redirect request to error.do,
-  // this func will overwrite the schema
-  // based on the request and base institution url.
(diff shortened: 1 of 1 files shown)


Source/Plugins/Core/com.equella.serverbase/src/com/tle/core/services/impl/UrlServiceImpl.java
```diff
@@ -101,21 +101,28 @@ public QueryBuilder getQueryBuilderForRequest(HttpServletRequest request) {
     return qbuilder;
   }
 
-  @Override
-  public URI getUriForRequest(HttpServletRequest request, String query) {
+  // In order to correctly set up schema for an internal forward request,
+  // such as a redirect request to error.do,
+  // this func will overwrite the schema
+  // based on the request and base institution url.
```
edalex-ian, line 107: Good job adding the doco. Seeing you've gone to the effort, might as well put it as JavaDoc so that the IDE will display it when someone queries the function.

edalex-ian (changes requested):
Good stuff @edalex-yinzi and great to see the refactor to keep it DRY.

Just one tweak for the comment and then we should be good to go when we get a (mostly) green build. :wink:

Source/Plugins/Core/com.equella.serverbase/src/com/tle/core/services/impl/UrlServiceImpl.java
```diff
@@ -101,21 +101,31 @@ public QueryBuilder getQueryBuilderForRequest(HttpServletRequest request) {
...
   }
 
-  @Override
-  public URI getUriForRequest(HttpServletRequest request, String query) {
+  /**
+   * In order to correctly set up schema for an internal forward request, such as a redirect request
+   * to error.do, this func will overwrite the schema based on the request and base institution url.
+   *
+   * @param request HttpServletRequest
+   * @return UriBuilder
```
edalex-ian, line 109: Don't just leave these as the template place holders, add some _useful_ descriptions.

edalex-ian (commented):
Almost there, just make sure you do the JavaDoc completely.

PenghaiZhang (approved):

edalex-yinzi:
Thanks, @edalex-ian and @PenghaiZhang. Just one question if I keep @param what's the appropriate description for it? I can't figure out a proper des.

edalex-ian:
Well first, keep in mind that both the JavaDoc generator and the IDE can determine the type for you, so the type information like the below is redundant:

```
@param request HttpServletRequest
```

But that also doesn't mean it's good to translate a code type into plain english like:

```
@param request The request object from a HTTP Servlet
```

No, you want to detail any kind of contractual expectations or how it will be used:

```
@param request The source which will be used to determine if the request was over a secure protocol
               from which the scheme in the builder will be derived.
```

Have a look at some of the standard Java JavaDoc to get more of an idea. :wink:

PR #3854, opened 2022-02-18
feat: rewrite the migration for table `entities` in Java

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

This PR adds the migration for table `entities` implemented in Java. However, the entity class is still written in Scala.
And because this migration was the last one written in Scala, after the re-implementation I also dropped those Scala code related to migration.

I tested locally by installing a fresh instance and upgrading from an older version. I also tested with Cloud Provider to confirm all the data is correctly stored in the table.


Source/Plugins/Core/com.equella.core/plugin-jpf.xml
```diff
@@ -1685,6 +1685,11 @@
     <parameter id="bean" value="bean:com.tle.core.institution.migration.v66.NewAuditLogColumn" />
     <parameter id="date" value="2018-06-01" />
   </extension>
+  <extension plugin-id="com.tle.core.migration" point-id="migration" id="NewEntityTable">
+    <parameter id="id" value="NewEntityTable" />
+    <parameter id="bean" value="bean:com.tle.core.institution.migration.v20191.NewEntityTable" />
```
edalex-ian, line 1690: Oh, did this only come in in 2019.1?

(For some reason I thought it was 2018.2 or earlier.)

edalex-ian (approved):
Nice work @PenghaiZhang ! And our first Scala `@Entity` to boot. :+1:

Source/Plugins/Core/com.equella.core/plugin-jpf.xml
PenghaiZhang: Yes, according to the previous implementation `entity` came in 2019.1.

PR #3858, opened 2022-02-21
feat: implement the converter for Entity in Java

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

This PR adds the implementation of `Entity` converter written in Java. 

Here is the list of changes:
1.  Adding a named query in `Entity.scala`.
2.  Creating Service and DAO layers for `Entity`.
3.  Creating the converter to support import/export of `Entity`.
4.  Dropping the previous implementation.

I noticed the format of JSON files created from previous implementation is much better. I wonder how important is the format.

new
image

old
image

I tested locally by adding a Cloud Provider and then importing, exporting and deleting the institution.


Source/Plugins/Core/com.equella.core/src/com/tle/core/newentity/service/EntityService.java
```diff
@@ -0,0 +1,70 @@
...
+   * @param description Description of the Entity.
+   * @param descriptionStrings Description of the Entity including locale in Json format (e.g.
+   *     {"en-GB": "desc"}).
+   * @param created The time when the Entity was created.
+   * @param modified The time when the Entity was lastly modified.
+   * @param owner Owner of the Entity.
+   * @param data Custom data of the Entity in Json format.
+   * @param institution Institution which the Entity belongs to.
+   */
+  void setEntity(
```
edalex-ian, line 52: A more common name for this kind of thing would be `createOrUpdate`. Using `set` here makes it sound like it's meant to be some kind of 'setter'.

Source/Plugins/Core/com.equella.core/src/com/tle/core/newentity/service/EntityService.java
```diff
@@ -0,0 +1,70 @@
...
+      String name,
+      String nameStrings,
+      String description,
+      String descriptionStrings,
+      Instant created,
+      Instant modified,
+      String owner,
+      String data,
+      Institution institution);
+  /**
```
edalex-ian, line 64:
```suggestion

  /**
```

Source/Plugins/Core/com.equella.core/src/com/tle/core/newentity/service/EntityServiceImpl.java
```diff
@@ -0,0 +1,77 @@
...
+      String nameStrings,
+      String description,
+      String descriptionStrings,
+      Instant created,
+      Instant modified,
+      String owner,
+      String data,
+      Institution institution) {
+    EntityID id = new EntityID();
+    id.uuid_$eq(uuid);
```
edalex-ian, line 55: What's this $eq? Is it the iterop to Scala?

Should we see if we can make it align with normal Java usage?

edalex-ian (changes requested):
Thanks @PenghaiZhang .

Needs a few tweaks as discussed, especially the way the JSON is being generated/handled. And feel free to convert that service to Scala and then the 'setter' becomes a non-issue.

Also, we dicussed the odd naming in the Entity - please convert to camel case.

Plus the few other minor comments.

PenghaiZhang:
After changing the type of `data` to `JsonNode` the output is correct and looks better.

image

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/newentity/service/EntityService.scala
```diff
@@ -0,0 +1,52 @@
...
+  def createOrUpdate(uuid: String,
+                     typeId: String,
+                     name: String,
+                     nameStrings: String,
+                     description: String,
+                     descriptionStrings: String,
+                     created: Instant,
+                     modified: Instant,
+                     owner: String,
+                     data: String,
```
edalex-ian, line 43: Seeing this is the service layer, why don't we have it accept a JsonNode rather than a String and look after the handling? (Same as the other two Json fields.)

edalex-ian (approved):
Looks good @PenghaiZhang (nice with it all in Scala), but I have one new query. (See comment.)

Either way, you can probably merge. :wink:

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/newentity/service/EntityService.scala
PenghaiZhang: I believe the other two Json fields refer to  `nameStrings` and `descriptionStrings`  ?

PenghaiZhang: Let me merge with what we have now. I guess in the next PR for Cloud Provider there is a chance I would revisit here and change something.

PR #3892, opened 2022-02-24
Feature/rewrite cloud provider

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

In order to remove the use of simpleDBA in Cloud provider, we firstly need to drop the `CloudProviderDB`. At the beginning, this class name made me think it worked like a DAO, but I was wrong. This one is more like a Service which provided supports like registering a Cloud providers and was mostly used in the Cloud provider REST APIs. So I created a new class named `CloudProviderRegistrationService` which uses the `EntityService` implemented in previous PR.

Major changes:
1. more methods in `EntityService` and `EntityDAO`  and named query in `Entity`.
2. The new implementation in the new Service which required some changes in `EntityValidation` as well.
3. Due to the class `Entity` being a normal class and the types of its fields are either primitive types or String, an extra handling for converting string to `CloudProviderData` is required
4. Use the Service in `CloudProviderApi`.
5. Drop unused methods in `CloudProviderDB`. 

I tried to minimise changes, especially logical changes. But some of the previous implementation did not look quite right to me and did not work seamlessly with the new implementation. So I had to change some logical (but all minor). 

There is only one previous implementation related to cloud provider registration left. It is `refreshCloudProvider`. The main reason is the use of `CloudProviderService.serviceRequest` which obviously has simpleDBA code. Refactoring this one will cause a lot of more changes in different files. So I decided to do this in next PR when I will focus on refactoring `CloudProviderService`.

I have tested initialising a Cloud provider registration, listing all Cloud providers, adding a new one and deleting an existing one via the page. I can't find any UI that allows me to edit a Cloud provider so I tested this in the Swagger API page.

### DIFF
--- a/Source/Plugins/Core/com.equella.base/scalasrc/com/tle/beans/newentity/Entity.scala
+++ b/Source/Plugins/Core/com.equella.base/scalasrc/com/tle/beans/newentity/Entity.scala
@@ -18,6 +18,9 @@
 
 package com.tle.beans.newentity
 
+import com.tle.common.institution.CurrentInstitution
+import com.tle.common.usermanagement.user.CurrentUser
+import java.util.UUID
 import java.time.Instant
 import javax.persistence.{Column, Embeddable, EmbeddedId, Index, Lob, Table}
 import org.hibernate.annotations.{AttributeAccessor, NamedQuery, Type}
@@ -27,6 +30,13 @@ class EntityID extends Serializable {
   @Column(length = 36)
   var uuid: String  = _
   var inst_id: Long = _
+
+  override def equals(obj: Any): Boolean = obj match {
+    case that: EntityID => that.uuid == that.uuid && this.inst_id == that.inst_id
+    case _              => false
+  }
+
+  override def hashCode(): Int = inst_id.hashCode + uuid.hashCode
 }
 
 @javax.persistence.Entity
@@ -36,6 +46,8 @@ class EntityID extends Serializable {
 })
 @NamedQuery(name = "deleteAllEntityByInst",
             query = "DELETE FROM Entity WHERE id.inst_id = :institutionId")
+@NamedQuery(name = "getAllByType",
+            query = "from Entity WHERE id.inst_id = :institutionId and typeid = :typeId")
 class Entity {
   @EmbeddedId
   var id: EntityID = _
@@ -55,4 +67,56 @@ class Entity {
   var modified: Instant = _
   @Type(`type` = "json")
   var data: String = _
+}
+
+object EntityID {
+  def apply(uuid: String, inst_id: Long): EntityID = {
+    val id = new EntityID
+    id.uuid = uuid
+    id.inst_id = inst_id
+    id
+  }
+}
+
+object Entity {
+  def apply(uuid: String,
+            inst_id: Long,
+            typeId: String,
+            name: String,
+            nameStrings: String,
+            description: String,
+            descriptionStrings: String,
+            owner: String,
+            created: Instant,
+            modified: Instant,
+            data: String): Entity = {
+    val entity = new Entity
+    entity.id = EntityID(uuid, inst_id)
+    entity.typeid = typeId
+    entity.name = name
+    entity.nameStrings = nameStrings
+    entity.description = description
+    entity.descriptionStrings = descriptionStrings
+    entity.owner = owner
+    entity.created = created
+    entity.modified = modified
+    entity.data = data
+
+    entity
+  }
+
+  def blankEntity(typeId: String): Entity =
+    Entity(
+      uuid = UUID.randomUUID().toString,
+      inst_id = CurrentInstitution.get().getDatabaseId,
+      typeId = typeId,
+      name = "",
+      nameStrings = "",
+      description = "",
+      descriptionStrings = "",
+      owner = CurrentUser.getDetails.getUniqueID,
+      created = Instant.now(),
+      modified = Instant.now(),
+      data = ""
+    )
 }
--- a/Source/Plugins/Core/com.equella.base/scalasrc/com/tle/beans/newentity/Entity.scala
+++ b/Source/Plugins/Core/com.equella.base/scalasrc/com/tle/beans/newentity/Entity.scala
@@ -36,7 +36,7 @@ class EntityID extends Serializable {
     case _              => false
   }
 
-  override def hashCode: Int = Objects.hash(inst_id, uuid)
+  override def hashCode: Int = Objects.hash(inst_id: java.lang.Long, uuid: java.lang.String)
 }
 
 @javax.persistence.Entity
--- a/Source/Plugins/Core/com.equella.base/scalasrc/com/tle/beans/newentity/Entity.scala
+++ b/Source/Plugins/Core/com.equella.base/scalasrc/com/tle/beans/newentity/Entity.scala
@@ -20,7 +20,7 @@ package com.tle.beans.newentity
 
 import com.tle.common.institution.CurrentInstitution
 import com.tle.common.usermanagement.user.CurrentUser
-import java.util.UUID
+import java.util.{Objects, UUID}
 import java.time.Instant
 import javax.persistence.{Column, Embeddable, EmbeddedId, Index, Lob, Table}
 import org.hibernate.annotations.{AttributeAccessor, NamedQuery, Type}
@@ -32,11 +32,11 @@ class EntityID extends Serializable {
   var inst_id: Long = _
 
   override def equals(obj: Any): Boolean = obj match {
-    case that: EntityID => that.uuid == that.uuid && this.inst_id == that.inst_id
+    case that: EntityID => this.uuid == that.uuid && this.inst_id == that.inst_id
     case _              => false
   }
 
-  override def hashCode(): Int = inst_id.hashCode + uuid.hashCode
+  override def hashCode: Int = Objects.hash(inst_id, uuid)
 }
 
 @javax.persistence.Entity
--- a/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/cloudproviders/CloudProviderDB.scala
+++ b/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/cloudproviders/CloudProviderDB.scala
@@ -20,21 +20,16 @@ package com.tle.core.cloudproviders
 
 import java.util.concurrent.TimeUnit
 import java.util.{Locale, UUID}
-
-import cats.data.Validated.{Invalid, Valid}
 import cats.data.{OptionT, ValidatedNec}
-import cats.effect.{IO, LiftIO}
-import cats.syntax.applicative._
+import cats.effect.IO
 import cats.syntax.apply._
-import cats.syntax.validated._
 import com.softwaremill.sttp._
 import com.softwaremill.sttp.circe._
 import com.tle.core.db._
 import com.tle.core.db.dao.{EntityDB, EntityDBExt}
 import com.tle.core.db.tables.OEQEntity
-import com.tle.core.validation.{EntityValidation, OEQEntityEdits}
+import com.tle.core.validation.{EntityStdEdits, EntityValidation}
 import com.tle.legacy.LegacyGuice
-import com.tle.web.DebugSettings
 import fs2.Stream
 import io.circe.generic.semiauto._
 import io.doolse.simpledba.Iso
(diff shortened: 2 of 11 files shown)


Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/cloudproviders/CloudProviderRegistrationService.scala
```diff
@@ -0,0 +1,204 @@
...
+
+  /**
+    * Get all the registered Cloud providers.
+    *
+    * @return ValidatedNec where left is errors accumulated during the retrieval and right is a list
+    *         of Cloud provider details.
+    */
+  def getAllProviders: ValidatedNec[EntityValidation, List[CloudProviderDetails]] = {
+    entityService
+      .getAllByType(typeId)
```
PenghaiZhang, line 188: The previous implementation used `fs2 scream` to handle reading all the entities. I think we discussed this in previous PR and if needed we can use pagination to optimise here.

Source/Plugins/Core/com.equella.base/scalasrc/com/tle/beans/newentity/Entity.scala
```diff
@@ -27,6 +30,13 @@ class EntityID extends Serializable {
   @Column(length = 36)
   var uuid: String  = _
   var inst_id: Long = _
+
+  override def equals(obj: Any): Boolean = obj match {
+    case that: EntityID => that.uuid == that.uuid && this.inst_id == that.inst_id
+    case _              => false
+  }
+
+  override def hashCode(): Int = inst_id.hashCode + uuid.hashCode
```
edalex-ian, line 39: You'd probably get a more reliable hash with:

```suggestion
  override def hashCode(): Int = Objects.hash(inst_id, uuid)
```

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/cloudproviders/CloudProviderDB.scala
```diff
@@ -103,66 +98,27 @@ object CloudProviderDB {
...
-      if (!tokenCache.get(regToken).isPresent)
-        EntityValidation("token", "invalid").invalidNec
-      else {
-        tokenCache.invalidate(regToken)
-        ().validNec
-      }
-    })
+      EntityValidation
+        .standardValidation(EntityStdEdits(name = reg.name, description = reg.description), locale)
+        .map { _ =>
```
edalex-ian, line 108: hmmm, ignore the value passed into a `map` seems odd - if anything says this is _not_ a map operation... What's happening here?

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/cloudproviders/CloudProviderRegistrationService.scala
```diff
@@ -0,0 +1,204 @@
...
+    entity.nameStrings = edits.nameStrings.asJson.noSpaces
+    entity.description = edits.description.getOrElse("")
+    entity.descriptionStrings = edits.descriptionStrings.map(_.asJson.noSpaces).orNull
+    entityService.createOrUpdate(entity)
+
+    toInstance(entity, data)
+  }
+
+  /**
+    * Check whether a registration token is valid. If yes, invalidate the token.
```
edalex-ian, line 91: This is so confusing. One would use `validToken` to invalidate an existing valid token. Huh?

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/cloudproviders/CloudProviderRegistrationService.scala
```diff
@@ -0,0 +1,204 @@
...
+      tokenCache.invalidate(regToken)
+      ().validNec
+    }
+
+  /**
+    * Create a new token for Cloud provider registration.
+    */
+  def createRegistrationToken: String = {
+    val newToken = UUID.randomUUID().toString
+    tokenCache.put(newToken, newToken)
```
edalex-ian, line 109: Is this right, the token is _both_ the key and the value??

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/cloudproviders/CloudProviderRegistrationService.scala
```diff
@@ -0,0 +1,204 @@
...
+  private def save(entity: Entity,
+                   values: (EntityStdEdits, CloudProviderData)): CloudProviderInstance = {
+    val (edits, data) = values
+    entity.data = data.asJson.noSpaces
+    entity.modified = Instant.now
+    entity.name = edits.name
+    entity.nameStrings = edits.nameStrings.asJson.noSpaces
+    entity.description = edits.description.getOrElse("")
+    entity.descriptionStrings = edits.descriptionStrings.map(_.asJson.noSpaces).orNull
+    entityService.createOrUpdate(entity)
```
edalex-ian, line 85: Hmmm, this got lost in my first reading - indeed, I wondered why this was called `save` as I thought it didn't persist anything.

Maybe to make it jump out:

```suggestion

    // Persist the new Entity
    entityService.createOrUpdate(entity)
```

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/cloudproviders/CloudProviderRegistrationService.scala
```diff
@@ -0,0 +1,204 @@
...
+  // this Entity.
+  private def save(entity: Entity,
+                   values: (EntityStdEdits, CloudProviderData)): CloudProviderInstance = {
+    val (edits, data) = values
+    entity.data = data.asJson.noSpaces
+    entity.modified = Instant.now
+    entity.name = edits.name
+    entity.nameStrings = edits.nameStrings.asJson.noSpaces
+    entity.description = edits.description.getOrElse("")
+    entity.descriptionStrings = edits.descriptionStrings.map(_.asJson.noSpaces).orNull
```
edalex-ian, line 84: Do you really want to mutate the passed in Entity?

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/newentity/service/EntityService.scala
```diff
@@ -20,17 +20,29 @@ package com.tle.core.newentity.service
...
     * @param institution The Institution from where to get the list of Entity.
     */
-  def getAllEntity(institution: Institution): java.util.List[Entity]
+  def getAllByInstitution(institution: Institution): java.util.List[Entity]
+
+  /**
+    * Get a list of Entity by Entity type.
+    * @param typeId The type representing an Entity (e.g. cloudprovider)
+    */
+  def getAllByType(typeId: String): java.util.List[Entity]
```
edalex-ian, line 45: Should it limit by institution?

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/newentity/service/EntityService.scala
```diff
@@ -20,17 +20,29 @@ package com.tle.core.newentity.service
 
 import com.tle.beans.Institution
 import com.tle.beans.newentity.Entity
-
 import java.time.Instant
 
 trait EntityService {
 
+  /**
+    * Get an Entity by UUID.
```
edalex-ian, line 28:
```suggestion
    * Get an Entity by UUID limited to those in the current institution.
```

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/cloudprovider/CloudProviderApi.scala
```diff
@@ -44,27 +48,31 @@ class CloudProviderApi {
...
+        .path(SettingsList.CloudProviderListPage)
+        .build()
+        .toString
+
+    registrationService
+      .register(regtoken, registration)
+      .map(CloudProviderRegistrationResponse(_, forwardUrl))
+      .map(Response.ok(_).build)
+      .leftMap(collectErrors)
+      .valueOr(badRequest(_: _*))
```
edalex-ian, line 75: Oooh, nice. I love how that reads. :+1:

edalex-ian (changes requested):
Fair effort @PenghaiZhang :star: 

Mostly queries, and just a couple of minor items. :+1:

Source/Plugins/Core/com.equella.base/scalasrc/com/tle/beans/newentity/Entity.scala
```diff
@@ -27,6 +30,13 @@ class EntityID extends Serializable {
   @Column(length = 36)
   var uuid: String  = _
   var inst_id: Long = _
+
+  override def equals(obj: Any): Boolean = obj match {
+    case that: EntityID => that.uuid == that.uuid && this.inst_id == that.inst_id
```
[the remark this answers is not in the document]
PenghaiZhang: Oops my mistake! Yes it should be `this`.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/newentity/service/EntityService.scala
PenghaiZhang: It's limited by current institution. Do you want to make it a bit more flexible to support different institution ?

PenghaiZhang: Or I can `limited to those in the current institution.` to the docs like below comment.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/cloudproviders/CloudProviderDB.scala
PenghaiZhang: I did not update too much in this file because the intention is to drop it so I just made minor changes to let the complier happy.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/cloudproviders/CloudProviderRegistrationService.scala
PenghaiZhang: I guess the idea is: if the token is no longer in the cache, return a error; if the token is still available, invalidate it so it cannot be reused.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/cloudproviders/CloudProviderRegistrationService.scala
PenghaiZhang: According to how the cache of token is invalidated, I think this is right. 


```
tokenCache.invalidate(regToken)
```

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/cloudproviders/CloudProviderRegistrationService.scala
PenghaiZhang: Ah yes it's to persist an entity. I will add this comment.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/cloudproviders/CloudProviderRegistrationService.scala
edalex-ian: Yeah, but why do the two disparate operations in the one function called validXyz.. Let's discuss.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/cloudproviders/CloudProviderRegistrationService.scala
edalex-ian: It's not so much about a "Scala coding convention" as it is about being cautious about mutating passed in data - as it could cause an unexpected surprise (especially in any future use/maintenance). It's probably mostly because this seems like an innocuous 'save' function, so you wouldn't typically expect it to be modifying inputs. However if it was called something like 'applyValues' then it wouldn't be unexpected.

But so in this context I'd expect `entity` to be unchanged after use. Now you could add a copy, or you could simple use the constructor with named args in this `save` to build up the new one.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/cloudproviders/CloudProviderRegistrationService.scala
edalex-ian: Okay, cool.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/newentity/service/EntityService.scala
edalex-ian: Ah I see, I went into the DAO and had a closer look. Yes, just make it explicit in the docs.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/cloudproviders/CloudProviderRegistrationService.scala
PenghaiZhang: This method now returns the valid token and the invalidation is moved to `register`.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/cloudproviders/CloudProviderRegistrationService.scala
PenghaiZhang: This method is renamed to `applyValues` and only mutates the Entity. The persisting and CP instance building are move to `register` and `editRegistered`.

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/cloudproviders/CloudProviderRegistrationService.scala
```diff
@@ -139,26 +134,29 @@ class CloudProviderRegistrationService {
...
+            iconUrl = registration.iconUrl,
+            vendorId = registration.vendorId,
+            providerAuth = registration.providerAuth,
+            oeqAuth = CloudOAuthCredentials.random(),
+            serviceUrls = registration.serviceUrls,
+            viewers = registration.viewers
+          )
+          val entity = applyValues(Entity.blankEntity(typeId), fields, newData)
+
+          tokenCache.invalidate(token)         // Invalidate the token so it can't be reused.
```
edalex-ian, line 149: Should this only happen after a successful DB update (i.e. after the below `createOrUpdate`)?

edalex-ian (commented):
Looks like it's there @PenghaiZhang . Just one query....

Source/Plugins/Core/com.equella.core/scalasrc/com/tle/core/cloudproviders/CloudProviderRegistrationService.scala
edalex-ian: I understand this is how it was. All good. :+1:

edalex-ian (approved):

PR #3908, opened 2022-03-07
fix: correctly handle error for UserSearch dialog

edalex-yinzi:
##### Checklist

- [X] the [contributor license agreement][] is signed
- [X] commit message follows [commit guidelines][]
- [ ] tests are included
- [X] screenshots are included showing significant UI changes

##### Description of change
Add a callback onError
Correctly show 404 and error message

https://user-images.githubusercontent.com/92769668/156957434-b1b046df-0449-490b-8dcb-6068f36ac948.mp4

### DIFF
--- a/react-front-end/tsrc/components/UserSearch.tsx
+++ b/react-front-end/tsrc/components/UserSearch.tsx
@@ -63,8 +63,6 @@ export interface UserSearchProps {
   listHeight?: number;
   /** Callback triggered when a user entry is clicked on. */
   onSelect: (username: OEQ.UserQuery.UserDetails) => void;
-  /** Callback triggered when an api error is raised, except 404 error */
-  onError?: (error: OEQ.Errors.ApiError) => void;
   /** A list of group UUIDs to filter the users by. */
   groupFilter?: ReadonlySet<string>;
   /** Function which will provide the list of users. */
@@ -89,7 +87,6 @@ const UserSearch = ({
   id,
   listHeight,
   onSelect,
-  onError,
   groupFilter,
   userListProvider = (query?: string, groupFilter?: ReadonlySet<string>) =>
     UserModule.listUsers(query ? `${query}*` : undefined, groupFilter),
@@ -151,7 +148,6 @@ const UserSearch = ({
         setUsers([]);
         if (error.status !== 404) {
           setErrorMessage(error.message);
-          onError?.(error);
         }
       })
       .finally(() => {
--- a/react-front-end/tsrc/components/UserSearch.tsx
+++ b/react-front-end/tsrc/components/UserSearch.tsx
@@ -104,7 +104,7 @@ const UserSearch = ({
     OEQ.Common.UuidString | undefined
   >(undefined);
   const [hasSearched, setHasSearched] = useState<boolean>(false);
-  const [errorMessage, setErrorMessage] = useState<String | null>(null);
+  const [errorMessage, setErrorMessage] = useState<String>();
   const [showSpinner, setShowSpinner] = useState<boolean>(false);
 
   useEffect(() => {
@@ -138,7 +138,7 @@ const UserSearch = ({
 
   const handleOnSearch = () => {
     setShowSpinner(true);
-    setErrorMessage(null);
+    setErrorMessage(undefined);
     userListProvider(query, groupFilter)
       .then((userDetails: OEQ.UserQuery.UserDetails[]) => {
         setUsers(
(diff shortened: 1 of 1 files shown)


react-front-end/tsrc/components/UserSearch.tsx
```diff
@@ -101,6 +104,7 @@ const UserSearch = ({
     OEQ.Common.UuidString | undefined
   >(undefined);
   const [hasSearched, setHasSearched] = useState<boolean>(false);
+  const [errorMessage, setErrorMessage] = useState<String | null>(null);
```
edalex-ian, line 107: Typically we just use `undefined`, so:

```suggestion
  const [errorMessage, setErrorMessage] = useState<String>();
```

And then update the usages accordingly.

react-front-end/tsrc/components/UserSearch.tsx
```diff
@@ -63,6 +63,8 @@ export interface UserSearchProps {
   listHeight?: number;
   /** Callback triggered when a user entry is clicked on. */
   onSelect: (username: OEQ.UserQuery.UserDetails) => void;
+  /** Callback triggered when an api error is raised, except 404 error */
+  onError?: (error: OEQ.Errors.ApiError) => void;
```
edalex-ian, line 67: Why is this needed? I don't see an other files updated to use this. :thinking:

edalex-ian (changes requested):
Good work @edalex-yinzi . One change required, and one query.

BTY

react-front-end/tsrc/components/UserSearch.tsx
edalex-yinzi: The ticket says 

> Fix notes: Keeping in mind that the `<SelectUserDialog>` wraps the standalone `<UserSearch>` component, `UserSearch` should simply take an additional `onError` prop - as opposed to using one of the existing contexts. This will allow it to remain decoupled.

So I added this callback.

react-front-end/tsrc/components/UserSearch.tsx
```diff
@@ -101,6 +104,7 @@ const UserSearch = ({
     OEQ.Common.UuidString | undefined
   >(undefined);
   const [hasSearched, setHasSearched] = useState<boolean>(false);
+  const [errorMessage, setErrorMessage] = useState<String>();
```
PenghaiZhang, line 107: Looks like `errorMessage` is used more like a boolean.

react-front-end/tsrc/components/UserSearch.tsx
```diff
@@ -134,16 +138,26 @@ const UserSearch = ({
...
       .then((userDetails: OEQ.UserQuery.UserDetails[]) => {
-        setHasSearched(true);
         setUsers(
           userDetails.sort((a, b) =>
             a.username.toLowerCase().localeCompare(b.username.toLowerCase())
           )
         );
       })
-      .finally(() => setShowSpinner(false));
+      .catch((error: OEQ.Errors.ApiError) => {
```
PenghaiZhang, line 150: Can `error` be any other type?

And what would happen if you just disable the network and do the search ?

react-front-end/tsrc/components/UserSearch.tsx
```diff
@@ -134,16 +138,26 @@ const UserSearch = ({
...
+        setUsers([]);
+        if (error.status !== 404) {
```
PenghaiZhang, line 152: Why skip 404 may I ask?

react-front-end/tsrc/components/UserSearch.tsx
edalex-yinzi: Yes. I had used two variables one is a boolean hasError and another one is string errorMessage but I think it would be more elegant to combine them together.

react-front-end/tsrc/components/UserSearch.tsx
edalex-yinzi: Because I think 404 can be regarded as a normal response for a search query, it's not a `real` error.

react-front-end/tsrc/components/UserSearch.tsx
edalex-yinzi: The func is from oeq-ts-rest-api module and calls axios so I think error will always wrapped as ApiError.

Disable network:
image

edalex-ian (approved):

edalex-ian:
Before merging will just check if @PenghaiZhang has anything else needing addressing.

PenghaiZhang:
> Before merging will just check if @PenghaiZhang has anything else needing addressing.

I have left some queries which yinzi marked as resolved but I did not see any update. But they are not big deals.

PenghaiZhang (approved):
Had a chat with Yinzi and it's good to go.

PR #3913, opened 2022-03-08
fix: make sure 'SchemaId' and 'MigrationLog' are added in Hibernate entity binding map

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

After Hibernate was updated to version 5, adding a new database schema from the Institution Manager no longer works. 

According to my investigation, this is because Hibernate no longer has the entity bindings for 'SchemaId' and 'MigrationLog' after `TransactionModule` is configured. As a result, the SQL of creating the two tables are just empty strings, which fail the whole process of adding a new database. 

The fix is to add the two entities in the `plugin-jpf.xml` as plugins of `com.tle.core.hibernate`. This will make sure Hibernate has the bindings and the SQL of creating the two tables are correct.

Please grab to chat if it's hard to understand above description.


edalex-ian:
:crossed_fingers:

edalex-ian (approved):

PR #3922, opened 2022-03-09
feat(rest-module): add support for thumbnailDetails

edalex-ian:
##### Checklist

- [x] the [contributor license agreement][] is signed
- [x] commit message follows [commit guidelines][]
- [x] tests are included
- [ ] screenshots are included showing significant UI changes
- [ ] documentation is changed or added

##### Description of change

Adds support for the new `thumbnailDetails` field introduced in #3916 .

[contributor license agreement]: https://www.apereo.org/node/676
[commit guidelines]: https://chris.beams.io/posts/git-commit/

PR #3893, opened 2022-02-25
feat(rest-module): add support for includeAttachments param of search2

edalex-ian:
##### Checklist

- [x] the [contributor license agreement][] is signed
- [x] commit message follows [commit guidelines][]
- [x] tests are included
- [ ] screenshots are included showing significant UI changes
- [ ] documentation is changed or added

##### Description of change

Short and simple, this updates the REST Module (oeq-ts-rest-api) to match the back-end changes for `includeAttachments` and `attachmentCount`.

[contributor license agreement]: https://www.apereo.org/node/676
[commit guidelines]: https://chris.beams.io/posts/git-commit/

PR #3891, opened 2022-02-24
feat(REST): new includeAttachments param for search2 API

edalex-ian:
##### Checklist

- [x] the [contributor license agreement][] is signed
- [x] commit message follows [commit guidelines][]
- [x] tests are included
- [ ] screenshots are included showing significant UI changes
- [ ] documentation is changed or added

##### Description of change

This adds a new param to the search2 endpoint that allows for attachments to be excluded from the result. Doing so can skip the (sometimes) intensive processing required for attachments and thereby speed up responses.

The intention is to use this in the front-end to exclude attachments on search, but then if a user expands the attachments for an item to use the same endpoint but with a `must` of the item `uuid`. To ensure this can be done and still have the UI display how many attachments an item have there is also a new field in the response called `attachmentCount`.

[contributor license agreement]: https://www.apereo.org/node/676
[commit guidelines]: https://chris.beams.io/posts/git-commit/

PR #3868, opened 2022-02-22
refactor(search2): optimise number of attachment retrievals

edalex-ian:
##### Checklist

- [x] the [contributor license agreement][] is signed
- [x] commit message follows [commit guidelines][]
- [x] tests are included
- [ ] screenshots are included showing significant UI changes
- [ ] documentation is changed or added

##### Description of change

Profiling revealed a lot of time was spent in search2 requests in `ifNotBroken` due to both uses calling off to retrieve the same
attachment. Significant performance (both speed and DB calls) improvements were possible if instead we re-use the attachment we retrieve when we first check for broken attachment status rather than throwing it away.

Existing tests provide coverage for this, and I also undertook manual testing.

(While doing this, I noticed that for broken attachments the New UI was showing _no_description - just the exclamation mark. This was due to the change to show the Kaltura description. As a result, I rejigged that a little too - and so now broken attachments are more likely to show a description.)

[contributor license agreement]: https://www.apereo.org/node/676
[commit guidelines]: https://chris.beams.io/posts/git-commit/

PR #3925, opened 2022-03-10
Bugfix/slow search frontend

edalex-ian:
##### Checklist

- [x] tests are included

##### Description of change

This is the (hopefully) last PR in the slow search performance enhancement series consisting of:

- #3868 
- #3891 
- #3893 
- #3916 and
- #3922 

Basically this PR modifies the front-end such that:

- When in list mode, sets the new `includeAttachments` to `false` for the inital request;
- When a user expands the list of attachments, then does an additional search targetting that specific item (with version) via `must` specifiers;
- Displays a `skeleton` in place of the attachment list while waiting; and
- When second search for attachments completes renders all the attachments as usual.

However, this also impacted thumbnails. To address that:

- `OEQThumb` was basically rewritten to use ThumbnailDetails instead;
- `SearchResult` now uses an item's `thumbnailDetails` with `OEQThumb`; and
- Rather than having a separate flag to control whether or not to explicitly use a 'placeholder' thumbnail, that can be controlled by simply not providing `thumbnailDetails`.

Also the code around `OEQThumb` could skip some additional logic which is now be handled by the server - and that extends to not needing to know of the concept of a 'broken' attachment etc.

### DIFF
--- a/react-front-end/__tests__/tsrc/search/components/SearchResult.test.tsx
+++ b/react-front-end/__tests__/tsrc/search/components/SearchResult.test.tsx
@@ -83,13 +83,16 @@ describe("<SearchResult/>", () => {
     theme: Theme = defaultTheme
   ) => {
     const renderResult = render(
-      //This needs to be wrapped inside a BrowserRouter, to prevent an `Invariant failed: You should not use <Link> outside a <Router>` error  because of the <Link/> tag within SearchResult
+      // This needs to be wrapped inside a BrowserRouter, to prevent an
+      // `Invariant failed: You should not use <Link> outside a <Router>`
+      // error  because of the <Link/> tag within SearchResult
       <MuiThemeProvider theme={theme}>
         <BrowserRouter>
           <SearchResult
             key={itemResult.uuid}
             item={itemResult}
             highlights={[]}
+            getItemAttachments={async () => itemResult.attachments!}
           />
         </BrowserRouter>
       </MuiThemeProvider>
@@ -428,7 +431,7 @@ describe("<SearchResult/>", () => {
         mockData.keywordFoundInAttachmentObj
       );
       expect(expandedAttachment.queryByText("image.png")).toBeVisible();
-      expect(collapsedAttachment.queryByText("config.json")).not.toBeVisible();
+      expect(collapsedAttachment.queryByText("config.json")).toBeFalsy(); // i.e. not rendered so not visible
     });
 
     it("should make each attachment draggable", async () => {
--- a/react-front-end/tsrc/components/OEQThumb.tsx
+++ b/react-front-end/tsrc/components/OEQThumb.tsx
@@ -56,7 +56,7 @@ interface ThumbProps {
 
 export interface OEQThumbProps {
   /**
-   * Details to determine what and how the thumbnail should be displayed. Of if none are provided
+   * Details to determine what and how the thumbnail should be displayed. Or if none are provided
    * a simple placeholder icon will be used instead.
    *
    * If `details` includes a `link` then that will have priority over any others and be used to
--- a/react-front-end/tsrc/components/OEQThumb.tsx
+++ b/react-front-end/tsrc/components/OEQThumb.tsx
@@ -15,18 +15,23 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import * as React from "react";
-import { makeStyles } from "@material-ui/core/styles";
 import { Theme } from "@material-ui/core";
-import * as OEQ from "@openequella/rest-api-client";
+import { makeStyles } from "@material-ui/core/styles";
+import DefaultFileIcon from "@material-ui/icons/InsertDriveFile";
+import WebIcon from "@material-ui/icons/Language";
 import LinkIcon from "@material-ui/icons/Link";
 import VideoIcon from "@material-ui/icons/Movie";
 import ImageIcon from "@material-ui/icons/Panorama";
-import DefaultFileIcon from "@material-ui/icons/InsertDriveFile";
-import WebIcon from "@material-ui/icons/Language";
-import Web from "@material-ui/icons/Web";
 import PlaceholderIcon from "@material-ui/icons/TextFields";
+import Web from "@material-ui/icons/Web";
+import * as OEQ from "@openequella/rest-api-client";
+import { flow, pipe } from "fp-ts/function";
+import * as React from "react";
 import { languageStrings } from "../util/langstrings";
+import * as O from "fp-ts/Option";
+import { simpleMatch } from "../util/match";
+import * as S from "fp-ts/string";
+import * as RNEA from "fp-ts/ReadonlyNonEmptyArray";
 
 const useStyles = makeStyles((theme: Theme) => {
   return {
@@ -51,25 +56,20 @@ interface ThumbProps {
 
 export interface OEQThumbProps {
   /**
-   * On object representing an oEQ attachment. If undefined, a placeholder icon is returned
+   * Details to determine what and how the thumbnail should be displayed. Of if none are provided
+   * a simple placeholder icon will be used instead.
+   *
+   * If `details` includes a `link` then that will have priority over any others and be used to
+   * display an image based thumbnail.
    */
-  attachment?: OEQ.Search.Attachment;
-  /**
-   * True indicates that a placeholder icon should be used instead of the 'real' thumbnail
-   */
-  showPlaceholder: boolean;
+  details?: OEQ.Search.ThumbnailDetails;
 }
 
 /**
- * OEQThumb component
- * Takes an OEQ.Search.Attachment object as a parameter.
- * The thumbnail will be served up from oEQ if the attachment has had a thumbnail generated (Image, video, pdf based files, Youtube, Flickr, Google books)
- * Otherwise, an appropriate Icon is returned
+ * Displays a standard thumbnail using either the details provided to display an appropriate icon or
+ * image from the server. Of if no details provided will display a generic placeholder.
  */
-export default function OEQThumb({
-  attachment,
-  showPlaceholder,
-}: OEQThumbProps) {
+export default function OEQThumb({ details }: OEQThumbProps) {
   const classes = useStyles();
   const thumbLabels = languageStrings.searchpage.thumbnails;
   const generalThumbStyles: ThumbProps = {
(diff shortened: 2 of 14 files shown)


react-front-end/tsrc/components/OEQThumb.tsx
```diff
@@ -51,129 +56,131 @@ interface ThumbProps {
...
-      oeqThumb = oeqProvidedThumb;
-      break;
-  }
-  return oeqThumb;
+  return pipe(
+    details,
+    O.fromNullable,
+    O.map(({ attachmentType, mimeType, link }) =>
+      link
+        ? buildServerProvidedThumb(link)
```
PenghaiZhang, line 176: In the previous implementation, whether to use server provided thumbnail is also determined by `hasGenerated`. Looks like it's missing in the new implementation.

react-front-end/tsrc/modules/SearchModule.ts
```diff
@@ -285,6 +297,63 @@ export const searchItems = (
...
+    ),
+    TE.map(NEA.head)
+  );
+
+  const extractAttachments: (
+    item: OEQ.Search.SearchResultItem
+  ) => OEQ.Search.Attachment[] = flow(
+    ({ attachments }) => attachments,
+    O.fromPredicate((x): x is OEQ.Search.Attachment[] => x !== undefined),
+    O.getOrElse(() => [] as OEQ.Search.Attachment[])
```
PenghaiZhang, line 328: How about this ? 
```
O.getOrElse<OEQ.Search.Attachment[]>(() => [])
```

react-front-end/tsrc/modules/SearchModule.ts
```diff
@@ -285,6 +297,63 @@ export const searchItems = (
...
+      () => `Search for item ${uuid}/${version} returned no results`
+    ),
+    TE.map(NEA.head)
+  );
+
+  const extractAttachments: (
+    item: OEQ.Search.SearchResultItem
+  ) => OEQ.Search.Attachment[] = flow(
+    ({ attachments }) => attachments,
+    O.fromPredicate((x): x is OEQ.Search.Attachment[] => x !== undefined),
```
PenghaiZhang, line 327: Why not just using `O.fromNullable`?

edalex-ian:
Failed 'functional_testing' job is on known flakies:

```
[error] Failed: Total 679, Failed 2, Errors 0, Passed 677
[error] Failed tests:
[error] 	com.tle.webtests.test.users.LDAPTest
[error] 	com.tle.webtests.test.contribute.controls.asc.AdvancedScriptControlTests
```

react-front-end/tsrc/modules/SearchModule.ts
```diff
@@ -285,6 +297,63 @@ export const searchItems = (
...
+    TE.chain(extractAndValidateItem),
+    TE.map(extractAttachments),
+    TE.mapLeft(E.toError)
+  )();
+
+  if (E.isLeft(attachmentsMaybe)) {
+    throw attachmentsMaybe.left;
+  }
+
+  return attachmentsMaybe.right;
```
PenghaiZhang, line 354:
```
pipe(
  attachmentsMaybe,
  E.fold(
    (err) => throw err;
    identity,
))
```
Is this similar to above code ?

react-front-end/tsrc/search/components/SearchResult.tsx
```diff
@@ -140,24 +148,25 @@ export const ItemDrmContext = React.createContext<{
...
-    starRatings,
-    bookmarkId: bookmarkDefaultId,
-    isLatestVersion,
     drmStatus: initialDrmStatus = defaultDrmStatus,
+    isLatestVersion,
+    modifiedDate,
+    name,
+    starRatings,
+    status,
+    uuid,
```
PenghaiZhang, line 168: Do you want to move `thumbnailDetails`  to above `uuid` ?

react-front-end/tsrc/search/components/SearchResult.tsx
```diff
@@ -310,7 +319,7 @@ export default function SearchResult({
           >
             {
               /**Custom metadata can contain html tags, we should make sure that is
-          preserved */
+                             preserved */
```
PenghaiZhang, line 322: Why is the big indentation ?

react-front-end/tsrc/search/components/SearchResult.tsx
```diff
@@ -385,10 +394,13 @@ export default function SearchResult({
         alignItems="flex-start"
         divider
         aria-label={searchResultStrings.ariaLabel}
+        data-item-id={uuid}
+        data-item-version={version}
```
PenghaiZhang, line 398: Did you add these two for testing purpose ?

react-front-end/tsrc/search/components/SearchResultAttachmentsList.tsx
```diff
@@ -244,10 +295,32 @@ export const SearchResultAttachmentsList = ({
...
+      </ListItemIcon>
+      <ListItemText color="primary" primary={e.message} />
+    </ListItem>
+  );
+
+  const buildAttachmentList = (): JSX.Element => {
+    const items =
+      attachmentsList.length > 0
+        ? attachmentsList
+        : buildSkeletonList(attachmentCount);
```
PenghaiZhang, line 311: Do you like this ? 

```
  pipe(
    attachmentsList,
    O.fromPredicate(not(A.isEmpty)),
    O.getOrElse(() => buildSkeletonList(attachmentCount))
  )
```

PenghaiZhang (commented):
Thanks Ian! I have left some comments. I also noticed some warnings in `Search.test.ts`.

react-front-end/__tests__/tsrc/search/components/SearchResult.test.tsx
```diff
@@ -428,7 +431,7 @@ describe("<SearchResult/>", () => {
         mockData.keywordFoundInAttachmentObj
       );
       expect(expandedAttachment.queryByText("image.png")).toBeVisible();
-      expect(collapsedAttachment.queryByText("config.json")).not.toBeVisible();
+      expect(collapsedAttachment.queryByText("config.json")).toBeFalsy(); // i.e. not rendered so not visible
```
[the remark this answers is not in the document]
edalex-ian: No, `toBeVisible` is correct. The change here was more so that we used to always build the full attachment list in the accordion and it was simply _hidden_ so the old check was `not.toBeVisible`. However that's no longer useable as now not only is it not visible, it's not even in the DOM.

It would have been nice (for the human readable side) to still be able to do `not.toBeVisible`, but ah well.

So unless you say otherwise, I think I'll leave as is.

react-front-end/tsrc/components/OEQThumb.tsx
edalex-ian: Very good observation @PenghaiZhang , as this is a key change.

The whole check for `hasGenerated` and then the logic needed around it was rather convoluted for the front-end to be able to navigate the various placeholder thumbnails that oEQ provides from Legacy UI days. This came to my attention when reviewing the thumbnail implementation before all this rework and it's why various things on the server side were done as they were - specifically take note of `determineThumbnailLink` @ https://github.com/openequella/openEQUELLA/blob/develop/Source/Plugins/Core/com.equella.core/scalasrc/com/tle/web/api/search/SearchHelper.scala#L355

With those changes, now the client simply has to see if the server provided a link for a thumbnail, if so that'll either be a 'generated' thumbnail (which mind you, `hasGenerated` was only applicable for `file` attachment types) or it will possibly be one which a third party custom attachment type supplies (like youtube, kaltura, etc). And so if a link is provided, simply use that, otherwise setup a generic replacement thumbnail.

react-front-end/tsrc/modules/SearchModule.ts
edalex-ian: Yes, that's what I had at first. But I had typing issues. :cry: 

I'll try again to check if that's still the case.

react-front-end/tsrc/modules/SearchModule.ts
edalex-ian: Because this is a type `Refinement` (due to the type guard), not just a fromNullable.

react-front-end/tsrc/search/components/SearchResult.tsx
edalex-ian: Yeah, prettier for some reason... Not sure if it's due to the missing space to open this JDoc block or... :shrug:

react-front-end/tsrc/search/components/SearchResult.tsx
edalex-ian: Yup, that I did. Was a bit cheeky, but I also felt that keeping them general like this they could be useful for other purposes. :crossed_fingers:

react-front-end/tsrc/modules/SearchModule.ts
PenghaiZhang: The type of `attachments` is `undefined | OEQ.Search.Attachment[]`, with `O.fromNullable` you don't need the type refinement I think :thinking: .

react-front-end/tsrc/modules/SearchModule.ts
PenghaiZhang: Oh I think you may need this as well as `O.fromNullable`  instead of `O.fromPredicate`.

react-front-end/__tests__/tsrc/search/components/SearchResult.test.tsx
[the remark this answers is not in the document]
PenghaiZhang: Oh I see! Yeah leave as it is.

react-front-end/tsrc/search/components/SearchResultAttachmentsList.tsx
edalex-ian: Ah yes, I do like that. Explicitly captures intent.

I think I'm trying to use ternaries too much now to try and be JS/TS friendly and not over-use FP-TS. Heh, tricky to find the balance.

react-front-end/tsrc/search/components/SearchResult.tsx
edalex-ian: Heh, that's very odd that I would add it at the top of the list. I'm confused what's happened here. :zany_face: 

I'll resort that list me thinks. :wink:

react-front-end/tsrc/search/components/SearchResultAttachmentsList.tsx
edalex-ian: hmm, with that in mind I rewrote the function:

```typescript
  const buildAttachmentList = (): JSX.Element => (
    <List disablePadding className={classes.attachmentListItem}>
      {error
        ? buildErrorListItem(error)
        : pipe(
            attachmentsList,
            O.fromPredicate(not(A.isEmpty)),
            O.getOrElse(() => buildSkeletonList(attachmentCount))
          )}
    </List>
  );
```

That way the old `item` only gets evaluated if there is no error. Minor optimisation, but I also like how it keeps the function as a single expression which is still readable thanks to the pipe (as opposed to a nested ternary).

react-front-end/tsrc/modules/SearchModule.ts
edalex-ian: Hmm, indeed. Fixed. :grin:

react-front-end/tsrc/modules/SearchModule.ts
edalex-ian: Nope, still needed:

2022-03-11-114837_1191x589_scrot

react-front-end/tsrc/search/components/SearchResult.tsx
edalex-ian: I've manually made it 'prettier'. :wink:

react-front-end/tsrc/modules/SearchModule.ts
edalex-ian: Ah yes, I was not fond of this and couldn't remember what alternatives we had. I've now got:

```typescript
  return pipe(
    attachmentsMaybe,
    E.getOrElseW((err) => {
      throw err;
    })
  );
```

edalex-ian:
Thanks @PenghaiZhang , all feedback addressed.

For the warnings in `Search.test.ts` you'll see I've tweaked the eslint config for that. And as mentioned, although we have configuration around this in the root `package.json` we actually don't have the plugin active - hence why we get away with it in the `react-frontend` tests.

PenghaiZhang (approved):
LGTM!

PR #3931, opened 2022-03-11
fix: ensure each view of a webpage has an audit log

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

The audit logs for webpage attachments were missing because cache of the webpage
was used and no requests were sent to server. So we need to update the cache control
for webpage attachment so that a request is sent to check whether the cache is still
valid and this request will also trigger an audit log.

I also noticed the audit log event type for webpage viewing is wrong. It should be `CONTENT_VIEWED` rather than `SUMMARY_VIEWED`. So I added a fix for this as well.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/web/viewitem/section/RootItemFileSection.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/web/viewitem/section/RootItemFileSection.java
@@ -228,17 +228,18 @@ public class RootItemFileSection
     RootItemFileModel model = getModel(info);
     ViewItemViewer viewer = model.getActualViewer();
     ViewItemResource resource = model.getResource();
+    IAttachment attachment = resource.getAttachment();
+    ViewAuditEntry vae =
+        viewer == null ? resource.getViewAuditEntry() : viewer.getAuditEntry(info, resource);
     try {
       ViewableItem<?> viewableItem = resource.getViewableItem();
       checkRestrictedResource(info, resource, viewer);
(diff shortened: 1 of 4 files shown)


Source/Plugins/Core/com.equella.core/src/com/tle/web/viewitem/section/RootItemFileSection.java
```diff
@@ -251,10 +252,11 @@ public SectionResult renderHtml(RenderEventContext info) throws IOException {
       }
       ensureOnePrivilege(resource.getPrivileges(), viewer.ensureOnePrivilege());
       if (viewableItem.isItemForReal() && viewableItem.getItemExtensionType() == null) {
-        auditor.audit(
-            info.getRequest(),
-            viewer.getAuditEntry(info, resource),
-            (ViewableItem<Item>) viewableItem);
```
PenghaiZhang, line 257: In the previous implementation, if `viewer` is not null (the null check is at line 237),  it always create a log where the event type is `SUMMARY_VIEWED`. So I added an attachment check here to support `CONTENT_VIEWED`.

edalex-ian (approved):
LGTM. :+1: 

Are there any existing tests around this to consider? (I'm guessing not, but thought I should check...)

PenghaiZhang:
> LGTM. +1
> 
> Are there any existing tests around this to consider? (I'm guessing not, but thought I should check...)

No I don' think so. I am not aware of any test for audit log.

PR #3946, opened 2022-03-16
fix: handle multiple status for api search

edalex-yinzi:
##### Checklist

- [X] the [contributor license agreement][] is signed
- [X] commit message follows [commit guidelines][]
- [x] tests are included

##### Description of change
Change query param status type from string to CsvList.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/web/api/search/SearchResourceImpl.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/web/api/search/SearchResourceImpl.java
@@ -67,6 +67,7 @@ import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Collectors;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 import javax.ws.rs.core.UriInfo;
@@ -100,7 +101,7 @@ public class SearchResourceImpl implements EquellaSearchResource {
       CsvList info,
       String showall,
       String dynaCollectionCompound,
-      String status,
+      CsvList status,
       String modifiedAfter,
       String modifiedBefore,
       String advancedSearch,
@@ -267,7 +268,7 @@ public class SearchResourceImpl implements EquellaSearchResource {
       String modifiedAfter,
       String modifiedBefore,
       String dynaCollectionCompound,
-      String status,
+      CsvList status,
       String owner,
       DefaultSearch search) {
     FreeTextBooleanQuery freetextQuery = null;
@@ -308,7 +309,11 @@ public class SearchResourceImpl implements EquellaSearchResource {
     boolean useStatus = false;
     if (status != null) {
       try {
-        search.setItemStatuses(ItemStatus.valueOf(status.toUpperCase()));
+        List<ItemStatus> statusT =
+            CsvList.asList(status).stream()
+                .map(sta -> ItemStatus.valueOf(sta.toUpperCase()))
+                .collect(Collectors.toList());
+        search.setItemStatuses(statusT);
         useStatus = true;
       } catch (IllegalArgumentException iae) {
         // Invalid status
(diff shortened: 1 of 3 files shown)


Source/Plugins/Core/com.equella.core/src/com/tle/web/api/search/SearchResourceImpl.java
```diff
@@ -267,7 +268,7 @@ private DefaultSearch createSearch(
       String modifiedAfter,
       String modifiedBefore,
       String dynaCollectionCompound,
-      String status,
+      CsvList status,
```
edalex-ian, line 271: You'll note that by the time CsvList for collectionUuids has been passed to this function it's done the conversion to the standard `Collection` type. This is along the lines of the pattern that the first entry point marshalls the data into a standard model which can then readily be used anywhere.

So here, this should be:

```suggestion
      List<String> status,
```

This can be done rather cleanly up in `searchItems` with:

```java
Optional.ofNullable(status).map(CsvList::asList).orElse(null);
```

Source/Plugins/Core/com.equella.core/src/com/tle/web/api/search/SearchResourceImpl.java
```diff
@@ -308,7 +309,11 @@ private DefaultSearch createSearch(
     boolean useStatus = false;
     if (status != null) {
       try {
-        search.setItemStatuses(ItemStatus.valueOf(status.toUpperCase()));
+        List<ItemStatus> statusT =
+            CsvList.asList(status).stream()
+                .map(sta -> ItemStatus.valueOf(sta.toUpperCase()))
```
edalex-ian, line 314: You could keep this as is. However, seeing you're essentially doing another map in here it makes it really clear if you do something like:

```java
List<ItemStatus> itemStatuses = status.stream()
    .map(String::toUpperCase)
    .map(ItemStatus::valueOf)
    .collect(Collectors.toList());
```

Note also I've changed the name of the variable there - not sure `statusT` really conveys much idea of what's happening here.

Source/Plugins/Core/com.equella.core/src/com/tle/web/api/search/SearchResourceImpl.java
```diff
@@ -308,7 +309,11 @@ private DefaultSearch createSearch(
...
+                .collect(Collectors.toList());
+        search.setItemStatuses(statusT);
         useStatus = true;
       } catch (IllegalArgumentException iae) {
         // Invalid status
```
edalex-ian, line 319: BTW, while playing in this error it would be worthwhile to change this to a WARN log entry capturing what invalid status was attempted.

edalex-ian (changes requested):
Thanks @edalex-yinzi . Needs a couple of tweaks, and don't forget to add something to test this over in SearchApiTest.

edalex-ian (approved):
Thanks @edalex-yinzi . LGTM! :+1:

autotest/OldTests/src/test/java/com/tle/webtests/test/webservices/rest/SearchApiTest.java
```diff
@@ -8,6 +8,7 @@
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
+import com.tle.beans.item.ItemStatus;
```
edalex-ian, line 11: Ah whoops! I didn't notice the failed build.

There's no dependencies across from the `autotest` to the core code based - keeps the completely decoupled. It's acceptable then to define your own constants for `LIVE` and `DRAFT`.

Interesting that your IDE let you do this. :thinking:

edalex-ian (changes requested):

autotest/OldTests/src/test/java/com/tle/webtests/test/webservices/rest/SearchApiTest.java
edalex-yinzi: Actually, it popped up a fix that said add core_base dependencies and didn't change any files so I thought it might be ok.

edalex-ian: Aha! Yeah, that changes the IntelliJ project setup - not the build etc.

edalex-ian (approved):

edalex-ian:
Failed known flaky on CodeBuild(`HierarchyTopicTest`), fine to merge.

Issue #1146, opened 2019-08-02
Login with secure token doesn't show permission error


SammyIsConfused:
**Describe the bug**
In the new UI, when attempting to login with a secure token that you do not have permissions for, the "You do not have permissions to use this token" error that is meant to show simply doesn't. Found with the TokenLoginTest

**To Reproduce**
Run TokenLoginTest in the old UI, note that it passes.
Run TokenLoginTest in the new UI, note that it fails.

**Expected behavior**
Both tests should work identically (I.E there should be equivalent error handling in the new UI).

PR #3951, opened 2022-03-17
fix: make sure Login page shows the captured token error

PenghaiZhang:
##### Checklist

- [x] tests are included

##### Description of change

In New UI, if a user logins with an invalid shared secret or with one that the user does not have permission to use, the token error is not showed in the Login page. 

The reason is `TokenUserStateHook` fails to keep the token error in New UI. In Old UI, the error is saved in the attribute of the login request which is a GET request, but when New UI is turned on, the GET request only returns the page which triggers another POST request to get the page content. Obviously, the POST request does not know anything about the token error. 

If above sounds confusing please grab me to explain in more detail. 

I have tried different approaches to fix the issue, and in the end I chose the one in this PR which I believe is the safest fix and have least impacts on others. It's to save the token error in `LogonSection` rather than the request attribute in New UI.

#1146

### DIFF
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/web/core/filter/TokenUserStateHook.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/web/core/filter/TokenUserStateHook.java
@@ -67,10 +67,10 @@ public class TokenUserStateHook implements UserStateHook {
           LOGGER.warn("Error with token:" + token);
 
           // We can save the login token error in the request attribute in Old UI, but we can't do
-          // this
-          // in New UI because New UI sends another POST request to query the page content. So we
-          // should
-          // call 'LogonSection.setLoginTokenError' to save the error.
+          // this in New UI because it sends another POST request to query the page content.
+          // Instead,
+          // we need to store it in the current session so that it can be persisted across the two
+          // requests.
           if (RenderNewTemplate.isNewUIEnabled()) {
             sessionService.setAttribute(WebConstants.KEY_LOGIN_EXCEPTION, ex.getLocalizedMessage());
           } else {
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/web/core/filter/TokenUserStateHook.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/web/core/filter/TokenUserStateHook.java
@@ -24,9 +24,9 @@ import com.tle.common.Check;
 import com.tle.common.usermanagement.user.UserState;
 import com.tle.core.guice.Bind;
 import com.tle.core.services.user.UserService;
+import com.tle.core.services.user.UserSessionService;
 import com.tle.exceptions.TokenException;
 import com.tle.web.core.filter.UserStateResult.Result;
-import com.tle.web.login.LogonSection;
 import com.tle.web.template.RenderNewTemplate;
 import javax.inject.Inject;
 import javax.inject.Singleton;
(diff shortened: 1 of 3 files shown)


edalex-ian (changes requested):
Thanks @PenghaiZhang , but as discussed we can't store it in a static variable as it could impact other user and uses of the Section. BTY.

Source/Plugins/Core/com.equella.core/src/com/tle/web/login/LogonSection.java
```diff
@@ -116,6 +117,18 @@
   @PlugKey("logon.login")
   private Button logonButton;
 
+  private static String loginTokenError;
+
+  /**
+   * Save a login token error captured when login with a shared secret in New UI. This method should
+   * only be used when New UI is turned on.
```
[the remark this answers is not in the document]
PenghaiZhang: This code has been removed.

PenghaiZhang:
image

Source/Plugins/Core/com.equella.core/src/com/tle/web/core/filter/TokenUserStateHook.java
```diff
@@ -62,7 +65,17 @@ public UserStateResult getUserState(HttpServletRequest request, UserState userSt
           return null;
         } catch (TokenException ex) {
           LOGGER.warn("Error with token:" + token);
-          request.setAttribute(WebConstants.KEY_LOGIN_EXCEPTION, ex);
+
+          // We can save the login token error in the request attribute in Old UI, but we can't do
+          // this
+          // in New UI because New UI sends another POST request to query the page content. So we
+          // should
+          // call 'LogonSection.setLoginTokenError' to save the error.
```
edalex-ian, line 73:
```suggestion
          // We can save the login token error in the request attribute in Old UI, but we can't do
          // this in New UI because it sends another POST request to query the page content. Instead
          // we need to store it in the current session so that it can be persisted across the two
          // requests.
```

Source/Plugins/Core/com.equella.core/src/com/tle/web/login/LogonSection.java
```diff
@@ -245,6 +246,14 @@ public SectionResult renderHtml(RenderEventContext context) {
     }
     model.setLoginLinks(loginLinksRenderables);
 
+    // In New UI, if a login token error is captured and saved in Session, update the model
+    // to show the error and then clear the error from Session.
+    String loginTokenError = sessionService.getAttribute(WebConstants.KEY_LOGIN_EXCEPTION);
+    if (RenderNewTemplate.isNewUIEnabled() && loginTokenError != null) {
+      model.setError(loginTokenError);
+      sessionService.removeAttribute(WebConstants.KEY_LOGIN_EXCEPTION);
+    }
```
edalex-ian, line 255: How about:

```java
Optional
  .ofNullable(sessionService.getAttribute(WebConstants.KEY_LOGIN_EXCEPTION))
  .filter(() -> RenderNewTemplate.isNewUIEnabled())
  .ifPresent(err -> {
    model.setError(err);
    sessionService.removeAttribute(WebConstants.KEY_LOGIN_EXCEPTION);
  });
```

edalex-ian (changes requested):
Good improvement thanks @PenghaiZhang .

Comment needs updating, and how about a bit of `Optional` for the logic.

Source/Plugins/Core/com.equella.core/src/com/tle/web/login/LogonSection.java
PenghaiZhang: This looks good ! The only limitation is Java does not allow us  to skip the parameter in `filter` so I will do
`filter((err) -> RenderNewTemplate.isNewUIEnabled())`.

edalex-ian (approved):
Good stuff @PenghaiZhang . Merge away! :tada:

PR #3970, opened 2022-03-23
fix: ensure audit level has a non-null default value

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

Since 2022.1, if the audit level is null it will break thumbnails. So we use 'NONE' as the default value.


edalex-ian (approved):
Excellent, thanks @PenghaiZhang :+1:

PR #3972, opened 2022-03-25
fix: brightspace connector 500 error and automatically close dialog

edalex-yinzi:
##### Checklist

- [ ] tests are included

##### Description of change
Force return old UI to resolve the 500 error.
Provide a custom close function to close the Dialog in the new UI.

This PR fixes two problems.

The first one is that the LMS auth dialog in the new UI will trigger a 500 error. This 500 error happens after the finishedAuth event is triggered. Since the cause of this error is difficult to trace, we force the new UI to be closed in finishedAuth Event.

The second problem was that after fixing the 500 error, the LMS auth dialog didn't close properly in the new UI. In fact, it didn't close properly even in the old UI (it disappeared because the whole brightspace connector page was redirected). The reason is the returned script `$.fancybox.close` is in the iframe, and it does not have a `self.parent` prefix or context, so this script won't close the dialog outside the iframe. And the solution was to add a correct close JS statement to the returned script.

https://user-images.githubusercontent.com/92769668/160061668-39e86656-79f4-4400-9372-0ff4e15cf1ac.mp4

~~The only issue is now the new UI will send a redundant request at the end.~~ fixed
image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/web/connectors/dialog/LMSAuthDialog.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/web/connectors/dialog/LMSAuthDialog.java
@@ -40,7 +40,6 @@ import com.tle.web.sections.events.js.BookmarkAndModify;
 import com.tle.web.sections.jquery.JQuerySelector;
 import com.tle.web.sections.jquery.JQuerySelector.Type;
 import com.tle.web.sections.jquery.JQueryStatement;
-import com.tle.web.sections.js.JSCallable;
 import com.tle.web.sections.js.generic.expression.ScriptExpression;
 import com.tle.web.sections.js.generic.function.CallAndReferenceFunction;
 import com.tle.web.sections.js.generic.function.ParentFrameFunction;
@@ -142,13 +141,8 @@ public class LMSAuthDialog extends AbstractOkayableDialog<LMSAuthDialog.Model> {
     closeBtn.setContextExpr(
         new ScriptExpression(
             "window.parent.document")); // The script is possible in the iframe, for example in
-                                        // BrightSpace.
+    // BrightSpace.
     return new JQueryStatement(closeBtn, new ScriptExpression("click()")); // Click the button
-  }
-
-  @Override
-  public JSCallable getCloseFunction() {
-    return super.getCloseFunction();
   }
 
   @EventHandlerMethod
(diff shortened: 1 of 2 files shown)


PenghaiZhang:
As discussed, this approach may result in the warning of redundant request.

After hours of experimenting, I finally got a better solution. Let's talk on Monday.

edalex-yinzi:
Now the redundant request issue has been fixed.

edalex-yinzi:
Comment updated.

Source/Plugins/Core/com.equella.core/src/com/tle/web/connectors/dialog/LMSAuthDialog.java
```diff
@@ -108,10 +119,41 @@ public void treeFinished(String id, SectionTree tree) {
...
+  public void setIsUseMockCloseDialog(boolean value) {
+    this.isUseMockCloseDialog = value;
+  }
+
+  /**
+   * Js statement for closing the dialog by clicking the close button. Because the script could be
+   * existing inside the iframe so the close function should have window.parent.ducoment prefix.
+   * Otherwise, it won't be able to close the auth dialog where outside the iframe.
+   */
+  private JQueryStatement mockCloseDialog(SectionInfo info) {
```
edalex-ian, line 135: Not sure if this is really a 'mock', as it is actually closing the dialog - albeit kind of manually (seeing it's is calling the `click()` function). The use of mock kind of implies something not really doing something - just kind of pretending to do it.

Or am I not understanding what's going on here?

edalex-ian (commented):
Some comments need tweaking, plus the naming is slightly confusing me. :thinking:

Source/Plugins/Core/com.equella.core/src/com/tle/web/connectors/dialog/LMSAuthDialog.java
PenghaiZhang: No, it's not a mock.

edalex-yinzi: Yes, your understanding is right. How about changing the name to manualCloseDialog? Or do you have a better suggestion?

edalex-ian: Would it be fair to call it an 'external' (to the IFRAME) auth dialog? If so, how about:

- `closeExternalAuthDialog` and
- `isExternalAuthDialog` etc

edalex-ian (approved):
Good stuff @edalex-yinzi , thank you. :+1:

PenghaiZhang (approved):

PR #4016, opened 2022-04-06
fix: escape the reserved word 'count' for Oracle

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

The two View Count tables have columns named `count` which is a reserved word in Oracle. So we need to escape the reserved word.

### DIFF
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/core/institution/migration/v20221/RenameViewCount.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/core/institution/migration/v20221/RenameViewCount.java
@@ -28,6 +28,11 @@ import java.util.List;
 import org.hibernate.Session;
 import org.hibernate.dialect.Oracle9iDialect;
 
+/**
+ * This migration renames the 'count' column in the viewcount_* tables on Oracle. This was required
+ * due to some odd naming which resulted from the original implementation with SimpleDBA. The new
+ * naming aligns with conventional Oracle usage as well as Hibernate defaults.
+ */
 @Bind
 @Singleton
 public class RenameViewCount extends AbstractHibernateSchemaMigration {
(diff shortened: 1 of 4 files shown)


Source/Plugins/Core/com.equella.core/src/com/tle/core/institution/migration/v20221/RenameViewCount.java
```diff
@@ -0,0 +1,73 @@
...
+import com.tle.core.migration.AbstractHibernateSchemaMigration;
+import com.tle.core.migration.MigrationInfo;
+import com.tle.core.migration.MigrationResult;
+import java.util.List;
+import org.hibernate.Session;
+import org.hibernate.dialect.Oracle9iDialect;
+
+@Bind
+@Singleton
+public class RenameViewCount extends AbstractHibernateSchemaMigration {
```
edalex-ian, line 33: Be good to have a comment (class level JavaDoc) describing this migration.

Something like: "This migration renames the 'count' column in the viewcount_* tables on Oracle. This was required due to some odd naming which resulted from the original implementation with SimpleDBA. The new naming aligns with conventional Oracle usage as well as Hibernate defaults."

edalex-ian (approved):
Looks good @PenghaiZhang 

Because this is a little odd/special I think it'd be prudent to add some doco. Could you please address the suggestion, then only wait for a successful `build_and_check` then merge to release?

Source/Plugins/Core/com.equella.core/src/com/tle/core/institution/migration/v20221/RenameViewCount.java
PenghaiZhang: Great. I will add it.

edalex-ian (approved):
Thanks @PenghaiZhang . :+1:

PR #4037, opened 2022-04-14
fix: blackboard 500 error from manage external resources

edalex-yinzi:
##### Checklist

- [ ] tests are included

##### Description of change
Change func `findAllUsages` to throw UnsupportedOperationException.
Add catch statement to set error message.

image
image

### DIFF
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/impl/BlackboardRESTConnectorServiceImpl.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/impl/BlackboardRESTConnectorServiceImpl.java
@@ -411,8 +411,8 @@ public class BlackboardRESTConnectorServiceImpl extends AbstractIntegrationConne
       int count,
       ConnectorRepositoryService.ExternalContentSortType sortType,
       boolean reverseSort)
-      throws LmsUserNotFoundException {
-    return null;
+      throws LmsUserNotFoundException, UnsupportedOperationException {
+    throw new UnsupportedOperationException("Feature not supported by Blackboard Rest");
   }
 
   @Override
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/impl/BlackboardRESTConnectorServiceImpl.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/impl/BlackboardRESTConnectorServiceImpl.java
@@ -54,6 +54,7 @@ import com.tle.core.connectors.service.ConnectorService;
 import com.tle.core.connectors.utils.ConnectorEntityUtils;
 import com.tle.core.encryption.EncryptionService;
 import com.tle.core.guice.Bind;
+import com.tle.core.i18n.CoreStrings;
 import com.tle.core.institution.InstitutionService;
 import com.tle.core.plugins.AbstractPluginService;
 import com.tle.core.replicatedcache.ReplicatedCacheService;
@@ -412,7 +413,7 @@ public class BlackboardRESTConnectorServiceImpl extends AbstractIntegrationConne
       ConnectorRepositoryService.ExternalContentSortType sortType,
       boolean reverseSort)
       throws LmsUserNotFoundException, UnsupportedOperationException {
-    throw new UnsupportedOperationException("Feature not supported by Blackboard Rest");
+    throw new UnsupportedOperationException(CoreStrings.text("blackboardrest.error.notsupported"));
   }
 
   @Override
--- a/Source/Plugins/Core/com.equella.core/src/com/tle/web/connectors/manage/ConnectorManagementResultsSection.java
+++ b/Source/Plugins/Core/com.equella.core/src/com/tle/web/connectors/manage/ConnectorManagementResultsSection.java
@@ -222,8 +222,7 @@ public class ConnectorManagementResultsSection
 
   @Override
   protected ConnectorManagmentSearchResultEvent createResultsEvent(
-      SectionInfo info, ConnectorManagementSearchEvent connectorSearchEvent)
-      throws RuntimeException {
+      SectionInfo info, ConnectorManagementSearchEvent connectorSearchEvent) {
     try {
       Connector connector = connectorSearchEvent.getConnector();
       List<ConnectorContent> allUsage = new ArrayList<ConnectorContent>();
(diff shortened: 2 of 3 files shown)


edalex-ian:
Curious, both GHA new UI functional_testing jobs failed on `io.github.openequella.search.AdvancedSearchPageTest` but was fine on CodeBuild.

Have attempted to re-run them. Let's see what happens.

Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/impl/BlackboardRESTConnectorServiceImpl.java
```diff
@@ -411,8 +411,8 @@ public SearchResults<ConnectorContent> findAllUsages(
       int count,
       ConnectorRepositoryService.ExternalContentSortType sortType,
       boolean reverseSort)
-      throws LmsUserNotFoundException {
-    return null;
+      throws LmsUserNotFoundException, UnsupportedOperationException {
+    throw new UnsupportedOperationException("Feature not supported by Blackboard Rest");
```
edalex-ian, line 415:
```suggestion
    throw new UnsupportedOperationException("Feature not supported by Blackboard REST");
```

Source/Plugins/Core/com.equella.core/src/com/tle/web/connectors/manage/ConnectorManagementResultsSection.java
```diff
@@ -222,7 +222,8 @@ protected Label getSuggestions(
 
   @Override
   protected ConnectorManagmentSearchResultEvent createResultsEvent(
-      SectionInfo info, ConnectorManagementSearchEvent connectorSearchEvent) {
+      SectionInfo info, ConnectorManagementSearchEvent connectorSearchEvent)
+      throws RuntimeException {
```
edalex-ian, line 226: Don't normally need to declare `RuntimeException` as it's an unchecked exception. :thinking:

edalex-ian (changes requested):
Looks good Yinzi. Just needs a wording tweak, and a query for you.

Source/Plugins/Core/com.equella.core/src/com/tle/core/connectors/blackboard/service/impl/BlackboardRESTConnectorServiceImpl.java
edalex-ian: Oh! Actually, that should be a language string.

edalex-ian:
BTW... What happens in something like Canvas (which does not support Find Uses)? We should be doing the same.

edalex-ian:
Excellent, the two GHA fails are different to before, just flaky known ones.

Source/Plugins/Core/com.equella.core/src/com/tle/web/connectors/manage/ConnectorManagementResultsSection.java
```diff
@@ -300,6 +301,13 @@ protected ConnectorManagmentSearchResultEvent createResultsEvent(
 
     } catch (LmsUserNotFoundException e) {
       throw new RuntimeException(e);
+    } catch (UnsupportedOperationException e) {
+      ConnectorManagmentSearchResultEvent erroredSearchResultEvent =
+          new ConnectorManagmentSearchResultEvent(
+              null, null, null, 0, connectorSearchEvent.getConnector());
```
PenghaiZhang, line 307: I have not dug into the code, but I found that if you pass `null`  to the constructor , you might trigger another NPE. For example, in line 168 of  `ConnectorManagementResultsSection`, calling `results.getResults().getResults()` would result in a NPE.

PenghaiZhang: But I am not sure if this kind of code will be executed or not as you have set `error` and `message`.

edalex-yinzi: Thanks, Penghai, after investigating I found the function `processResults` you mentioned will be triggered at line 220 in `AbstractSearchResultSection.java` which has been wrapped by an if statement so it will just skip it if the result is null or errored.

FYI: 220 line `context.processEvent(resultsEvent)` will trigger the `fire` func  in `ConnectorManagementResultsSection` and `fire` will trigger `processResults` 

image

edalex-yinzi:
Thanks Ian,

> BTW... What happens in something like Canvas (which does not support Find Uses)? We should be doing the same.

The interesting thing is `CanvasConnecterService` actually implemented the func `findAllUsages`.

PenghaiZhang (approved):

edalex-ian:
> > BTW... What happens in something like Canvas (which does not support Find Uses)? We should be doing the same.
> 
> The interesting thing is `CanvasConnecterService` actually implemented the func `findAllUsages`.

Gah, that's curious. Our feature matrix says otherwise... But you've drawn my attention to something else:

- In `CanvasConnecterService` although `findAllUsages` is implemented and based on code history has been since before open sourcing (be worth making sure it works), `findUsages` is not and returns an empty array.
- You'll note just above where you've made your fix is the `findUsages` for Bb, and it also is simply returning a `null`. Is that another NPE waiting to happen?
- So, what is the use cases for these two functions? I think you may need to have a look and then you may need to address that second function too.
  - I think the `findUsages` is used when from the Item Summary page you click the 'Find Uses' link on the right hand of the screen (under Details heading).

We could also follow up with @cathfitz on her return.

edalex-yinzi:
> > > BTW... What happens in something like Canvas (which does not support Find Uses)? We should be doing the same.
> > 
> > 
> > The interesting thing is `CanvasConnecterService` actually implemented the func `findAllUsages`.
> 
> Gah, that's curious. Our feature matrix says otherwise... But you've drawn my attention to something else:
> 
> * In `CanvasConnecterService` although `findAllUsages` is implemented and based on code history has been since before open sourcing (be worth making sure it works), `findUsages` is not and returns an empty array.
> * You'll note just above where you've made your fix is the `findUsages` for Bb, and it also is simply returning a `null`. Is that another NPE waiting to happen?
> * So, what is the use cases for these two functions? I think you may need to have a look and then you may need to address that second function too.
>   
>   * I think the `findUsages` is used when from the Item Summary page you click the 'Find Uses' link on the right hand of the screen (under Details heading).
> 
> We could also follow up with @cathfitz on her return.

Thanks for your info:

`findAllUsage` is used when you click the left menu `manage external resources` and choose a connector:
image

`findUsage` is used when you click the 'Find Uses' link in the Item Summary page:
image

and the workflow is in `FindUsesContentSection`: `loadUsage` -> `checkErrors` -> `getUsages` -> `BlackboardRESTConnectorServiceImpl`:`findUsage`.

The server just saves it into the model and has no other process on it.  

image

image

I think null is ok for this use case :
image

edalex-ian (approved):
Excellent, thanks for checking it all. Let's merge and proceed.

PR #4057, opened 2022-04-22
Migrate to JDK11

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

The intention of this PR is to move oEQ to being built with JDK 11 and targetting JDK11 as the minimum required JVM.

Required changes can be grouped into four categories.

* Apache Daemon update
  - Update JSVC to v1.3.3 for Linux
  - Update Procrun to v1.3.3 for Windows
  - Update the configuration for the new version
  -  Create a new upgrade to support above updates
  - Drop the support for installing OEQ on unsupported OSes

* BIRT packages update
   - Update BIRT OSGI runtime library and BIRT Report Framework to v4.9.1. Details of these two dependencies can be found from [this repo](https://github.com/openequella/birt-packages)
   - Update in `build.sbt` to support the new version
   - Code changes for fixing errors caused by the new version and revealed by  [Sample Reports.zip](https://github.com/openequella/openEQUELLA/files/10212588/Sample.Reports.zip)
   - Code changes for the Selenium tests
  
* Build with JDK11 and target JDK11 as the minimum required JVM
  - Code changes for the update
  - GHA workflow configuration update

* Update for using JDK11 in Docker container


edalex-ian:
So close, just the same two on the three different modes:

```
com.tle.webtests.test.reporting.RunReportTest
com.tle.webtests.test.workflow.ModerationProgressTest
```

Could it be a BiRT issue. :grimacing:

PenghaiZhang:
I downloaded the logs and saw a lot more warnings than I did locally, including the slf4j warning.

autotest/Tests/src/main/java/com/tle/webtests/pageobject/viewitem/ModerationTab.java
```diff
@@ -1,92 +1,93 @@
...
+    return isPresent(
+        By.xpath(
+            "//table[@id='cmc_m']/tbody/tr["
+                + quoteXPath(row)
+                + "]/td[2]/ul[@class='moderators']/li/span[text()="
+                + quoteXPath(moderator)
+                + "]"));
+  }
+
+  public long waitTime() throws ParseException {
```
PenghaiZhang, line 46: The real change is in this method.

PenghaiZhang:
According to the logs, Jar `reporting-common` is missing.

PenghaiZhang:
It's finally ready for review.

.github/workflows/ci.yaml
```diff
@@ -172,11 +172,11 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: Set up JDK 1.8
+      - name: Set up JDK 11
         uses: actions/setup-java@v3
         with:
           distribution: adopt
-          java-version: 8
+          java-version: 11
```
PenghaiZhang, line 179: Should we keep 8 here for the testing ?

edalex-ian: What do you mean, I'm not sure I follow because:

1. This job is about building the import/export tool; and
2. After your changes can you still build with JDK 8, or will this now need 11?

edalex-ian: What do you mean, I'm not sure I follow because:

1. This job is about building the import/export tool; and
2. After your changes can you still build with JDK 8, or will this now need 11?

Installer/data/server/manager/unix/equellaserver
```diff
@@ -145,49 +150,22 @@ case "$1" in
...
-
-      case "$OUTPUT" in
-      *Still*)
-          exit 2
-          ;;
-      *)
-          exit 0
-         ;;
-      esac
+      if [ -f "$EQUELLASERVER_PID" ]; then
```
edalex-ian, line 153: This is a very different check to the above - and much weaker at that. Do we really want to leave it as this?

Installer/data/server/manager/unix/equellaserver
```diff
@@ -145,49 +150,22 @@ case "$1" in
...
-
-      case "$OUTPUT" in
-      *Still*)
-          echo "The EQUELLA App Server is running"
-          ;;
-      *)
-          echo "The EQUELLA App Server is stopped"
-         ;;
-      esac
+      if [ -f "$EQUELLASERVER_PID" ]; then
```
edalex-ian, line 160: As above.

Installer/data/server/manager/unix/manager
```diff
@@ -135,51 +140,21 @@ case "$1" in
...
-
-      case "$OUTPUT" in
-      *Still*)
-          exit 2
-          ;;
-      *)
-          exit 0
-         ;;
-      esac
+      if [ -f "$MANAGER_PID" ]; then
```
edalex-ian, line 143: Same as for equellaserver...

Installer/data/server/manager/unix/manager
```diff
@@ -135,51 +140,21 @@ case "$1" in
...
-
-      case "$OUTPUT" in
-      *Still*)
-          echo "The EQUELLA manager is running"
-          ;;
-      *)
-          echo "The EQUELLA manager is stopped"
-         ;;
-      esac
+      if [ -f "$MANAGER_PID" ]; then
```
edalex-ian, line 150: Same as for equellaserver...

Installer/src/com/dytech/edge/installer/application/JavaCallback.java
```diff
@@ -34,26 +34,11 @@ public class JavaCallback implements Callback {
...
-                parent,
-                "Does the directory refer to a Java 8 installation?",
-                "Warning", //$NON-NLS-1$ //$NON-NLS-2$
-                JOptionPane.YES_NO_OPTION,
-                JOptionPane.WARNING_MESSAGE);
-      }
-
-      if (result == JOptionPane.YES_OPTION) {
-        installer.gotoPage(installer.getCurrentPageNumber() + 1);
-      }
```
edalex-ian, line 56: How come you dropped the version check?

Source/Plugins/Applet/com.tle.common.applet/build.sbt
```diff
@@ -1 +1 @@
-(Compile / unmanagedJars) += file(sys.props("java.home")) / "lib/javaws.jar"
```
edalex-ian, line 1: Ah interesting, indeed the applets were one of the things blocking an upgrade from JDK11.

So what does this change mean?

Source/Plugins/Core/com.equella.base/src/com/tle/core/application/impl/PluginServiceImpl.java
```diff
@@ -104,11 +103,12 @@ public Object getBean(PluginDescriptor plugin, String clazzName) {
...
-      while (callerClass == PluginServiceImpl.class
-          || callerClass == AbstractPluginService.class
-          || callerClass == PluginTracker.class) {
-        callerClass = Reflection.getCallerClass(++i);
+      StackTraceElement[] stackTraceElements = Thread.currentThread().getStackTrace();
+      String callerClass = stackTraceElements[i].getClassName();
+      while (callerClass.equals(PluginServiceImpl.class.getName())
+          || callerClass.equals(AbstractPluginService.class.getName())
+          || callerClass.equals(PluginTracker.class.getName())) {
+        callerClass = stackTraceElements[++i].getClassName();
```
edalex-ian, line 111: Did you validate this code still works the same?

Source/Server/adminTool/build.sbt
```diff
@@ -13,8 +13,6 @@ libraryDependencies ++= Seq(
   "org.springframework"    % "spring-context"  % springVersion
 )
 
-(Compile / unmanagedJars) += file(sys.props("java.home")) / "lib/javaws.jar"
```
edalex-ian, line 16: Oh that's odd. I wonder why the Admin Tool (which isn't that the admin console) required javaws - unless this is not Java WebStart like I was thinking.

Does this still all work?

edalex-ian (commented):
Great work @PenghaiZhang ! :tada: Thank you. :pray: 

Couple of queries:

1. With the updates to JSVC I note there are only changes for linux, what about the others? If they're no longer used should we remove them? And, what about windows?
2. Should we add some kind of doco around what you had to do with JSVC? Maybe a README.md next to them?
3. Did you get to try the oEQ Birt Report Designer against it? Do they still communicate okay?

Other than that, I've place some comments in the review with some queries.

Thanks for your huge effort getting it this far. :+1:

.github/workflows/ci.yaml
PenghaiZhang: Sorry, it's not related to this job. 

This is what I meant: 
* In the job of building main artefacts, we use JDK 11.
* In the job of functional tests, we use JRE 8.

And after the changes,  I cannot build with JDK 8 because the flag `--release` is not in JDK 8.

Installer/data/server/manager/unix/equellaserver
PenghaiZhang: When I used an older version of JSVC and did a `jsvc check`, the output has a message like `XXX still running` if the daemon is running.   

However, the latest version of JSVC no longer outputs that kind of message. Instead, it only tells me `XXXX checked successfully` no matter the daemon is working or not. 

I might miss some important configuration, but I could not more useful information and in the end I went this way - checking the PID file exists or not.

PenghaiZhang: This is the command I use.

``
./jsvc -cwd ./ -check -java-home $JAVA_HOME -pidfile manager-daemon.pid -classpath ./manager-rg5505249.jar -outfile /home/penghai/Edalex/forks/2022/manager/out1.txt  com.tle.upgrademanager.Main
``

Installer/src/com/dytech/edge/installer/application/JavaCallback.java
PenghaiZhang: I feel this check is quite useless because we have another check for whether `javac` exists in the supplied path. And this check is bound to Java 8.

Source/Plugins/Core/com.equella.base/src/com/tle/core/application/impl/PluginServiceImpl.java
PenghaiZhang: Yes I can confirm this still works fine.

Source/Server/adminTool/build.sbt
PenghaiZhang: Not so sure why it added `javaws` to unmanagedJars, but it works fine locally. 
BTW, does a web start APP can only be launched from a web browser ? I have not used and seen web start APP for a very long period and my understanding of this technique may be wrong.

Installer/data/server/manager/unix/equellaserver
edalex-ian: Okay, gotcha. Looks like you had a good play.

Shame we don't know the history of these files. I guess it's a modification of the template from Tomcat (https://github.com/apache/tomcat/blob/main/bin/daemon.sh), and `quickstatus` and `status` must be Pearson additions. I spent (too much) time trying to delve through the archives and figure it out, but no real sense made. :man_shrugging: 

I also tried to find any bug around the check option on jsvc not working, but nothing immediately found in their Jira with https://issues.apache.org/jira/issues/?jql=project%20%3D%20DAEMON%20AND%20text%20~%20%22jsvc%20check%22

.github/workflows/ci.yaml
edalex-ian: Ummm, so now I'm no longer sure what you're asking. Maybe we'll need to chat.

Installer/src/com/dytech/edge/installer/application/JavaCallback.java
edalex-ian: Cool. I think I was just checking here, but I'm struggling to remember 25 days back. :thinking: I'll say this is resolved.

Source/Plugins/Applet/com.tle.common.applet/build.sbt
edalex-ian: Eeeck, so this may be a blocker until we provide the new feature for the File Manager Applet replacement - i.e. improved UI, security and auditing for WebDAV.

We probably need to discuss.

Source/Server/adminTool/build.sbt
edalex-ian: Yeah, it could be a hang-over from before we added the Admin Console Package. That was the point of the Admin Console Package so that we no longer had to use the unsupported WebStart and JNLP.

So although you've run the `ClientLauncher` above, you'd really want to run the JAR via the Admin Console Package to test it.

Could you please give that a go and let us know if it works?

PenghaiZhang:
> Great work @PenghaiZhang ! tada Thank you. pray
> 
> Couple of queries:
> 
> 1. With the updates to JSVC I note there are only changes for linux, what about the others? If they're no longer used should we remove them? And, what about windows?
> 2. Should we add some kind of doco around what you had to do with JSVC? Maybe a README.md next to them?
> 3. Did you get to try the oEQ Birt Report Designer against it? Do they still communicate okay?
> 
> Other than that, I've place some comments in the review with some queries.
> 
> Thanks for your huge effort getting it this far. +1

I feel we should create another ticket for JSVC because there might be nothing or lots of to do for other systems. 

And for the oEQ Birt Report Designer, let's have a chat.

Source/Plugins/Applet/com.tle.common.applet/build.sbt
edalex-ian: Wow, so as far as we can tell this doesn't even work as none of the build environments have `javaws.jar` at `${JAVA_HOME}/lib/javaws`. So removing this should in the end be fine.

Marking as resolved.

Source/Server/adminTool/build.sbt
edalex-ian: As per https://github.com/openequella/openEQUELLA/pull/4057#discussion_r890830399 this is obviously not currently used, so can also be removed.

Resolved.

.github/workflows/ci.yaml
edalex-ian: As per our chat the query is: In the functional tests job, don't we want to test that oEQ still runs successfully with JDK 8?

Technically, _yes_. However, the functional tests job runs the functional tests code which needs to be compiled. And so it will now want JDK 11. So we can't have one JDK for the execution of oEQ and another for the execution of the autotests - because they're in the same job.

That's probably fine, as long as we make sure we undertake manual testing. Or as it is, our docker image is still using JDK 8 (at the moment) so it will inherently keep that baseline checked. (Just need to be mindful if we upgrade the Dockerfile to use openjdk:11.

edalex-ian:
Thanks for the README @PenghaiZhang .

Now when we've got time (might be awhile) we just need to:

1. confirm if the Report Designer will still work if only oEQ has updated BiRT deps; and
2. finalise the JSVC updates to only support supported OSes.

PenghaiZhang:
> Thanks for the README @PenghaiZhang .
> 
> Now when we've got time (might be awhile) we just need to:
> 
> 1. confirm if the Report Designer will still work if only oEQ has updated BiRT deps; and
> 2. finalise the JSVC updates to only support supported OSes.

According to the responses from the consultant team I think we will need to update JSVC for MAC and the Prunmgr for Windows and we can delete others.

edalex-ian:
I think it's even simpler than that from what support are saying. Looks like we can narrow it right down to:

- Windows (64-bit only); and
- Linux (64-bit only).

And that should mean rather than the 13 directories within `Installer/data/server/manager/` we can have three:

- all
- linux
- windows

And merge in the `unix`, `unix-config` and `win-config` into their relevant directories. (And yeah, update the binaries in the `windows` directory too.)

PenghaiZhang:
Merging develop to this branch failed the job that builds the artefacts, but the only difference is scala version from 2.13.8 to 2.13.10.

PenghaiZhang:
Did a force push due to one commit messed up locally.

PenghaiZhang:
By 5pm 7th Dec 2022, OEQ is finally happy to run and stop on Windows. 
The last several commits are all about the update of `Procrun`.

image

PenghaiZhang:
I noticed that I have updated `Procrun` to v1.3.3, but `JSVC` still stays at v1.3.0.

Maybe the next commit is to update `JSVC` again.

edalex-ian:
Great job @PenghaiZhang , you're close now. Keep kicking those goals. :+1:

PenghaiZhang:
JSVC is on the latest version (v1.3.3) now and the BIRT packages are also updated to v.4.9.1.

PenghaiZhang:
Locally ran OEQ against Java 11 and did some basic testing. All looks good.

PenghaiZhang:
The upgrade will also drop JVM options not available in Java 11.

PenghaiZhang:
Because folder `jre` no longer exists in Java 11, an update for the path to `jvm.dll` for Windows in the BAT files is required.

edalex-ian:
Heya @PenghaiZhang ,

I think we should update the title and description for this ticket. For example, perhaps a title like "Migrate to JDK11". And then in the description currently there is this line:

> The intention of this PR is to allow using JDK 11 to build openEquella artifacts and the artifacts can still work with JRE 8.

That probably needs to be changed to something like:

> The intention of this PR is to move oEQ to being built with JDK 11 and targetting JDK11 as the minimum required JVM.

What do you think?

PenghaiZhang:
I have updated the whole description.

I am updating the dockerfile to JDK 11.

PenghaiZhang:
Locally span up a clustered OEQ and confirmed the updated Dockerfile works fine.

edalex-ian:
Great work @PenghaiZhang , that sounds like it's ready for the next step. :+1:

PR #4327, opened 2022-09-05
Update Apache axis 2 packages to 1.8.2

SammyIsConfused:
##### Checklist

##### Description of change
resolves #4217 
`HTTPConstants` was moved to a different location in the package from 1.8.x, so changing the import reference was all that was required there. 
`OMLinkedListMetaFactory` can no longer be instantiated directly - you have to use `OMLinkedListMetaFactoryLoader` to load one up. `load(Map<String,Object> properties)` doesn't ever seem to refer to the `properties` variable so it should be safe to use `null` as the input. 


SammyIsConfused:
hmmm, okay so now we are getting deduplicate errors. :thinking: back to the drawing board.

SammyIsConfused:
Added a deduplication strategy for javax.servlet, this allows me to installerZip so fingers crossed it builds correctly this time.

edalex-ian:
Excellent, thank you @SammyIsConfused

PR #4469, opened 2022-11-15
chore: pin node version and npm version

PenghaiZhang:
##### Checklist

- [ ] tests are included

##### Description of change

Pin Node and NPM versions so that Renovate will use the correct versions to build.

The pinned Node version matches to the one specified in `.nvmrc`.

### DIFF
--- a/oeq-ts-rest-api/package.json
+++ b/oeq-ts-rest-api/package.json
@@ -10,7 +10,8 @@
     "src"
   ],
   "engines": {
-    "node": ">=10"
+    "node": "16.18.1",
+    "npm": "8.19.2"
   },
   "scripts": {
     "start": "rollup --config --watch --bundleConfigAsCjs",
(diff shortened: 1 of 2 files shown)


edalex-ian (commented):
Is it sufficient to only do `react-front-end/package.json`? I'd say we'd also at least want `oeq-ts-rest-api/package.json`. The others after that might be optional due to how little is in them and how little they're updated.

PenghaiZhang:
> Is it sufficient to only do `react-front-end/package.json`? I'd say we'd also at least want `oeq-ts-rest-api/package.json`. The others after that might be optional due to how little is in them and how little they're updated.

Ah yeah thanks for reminding this! I have updated `oeq-ts-rest-api`.

edalex-ian (approved):
Looks good. It will be interesting to see if renovate also updates these values for us going forward. :thinking: :crossed_fingers:

edalex-ian:
I wonder if when this merges several of the JS renovate PRs will suddenly go green and merge (after rebasing of course).

PR #4675, opened 2023-03-01
build (github): change order or node dep install

edalex-ian:
##### Checklist

~~- [ ] tests are included~~
~~- [ ] screenshots are included showing significant UI changes~~
~~- [ ] documentation is changed or added~~

##### Description of change

Need to install the root node dependencies before the others, as the new io-ts code-gen relies on ESLint being installed from the root first.

An alternative could be to try and get the oeq-ts-rest-api to install ESLint before it starts the codegen, however this is probably easier (and all is working on the internal Edalex build - due to this order difference).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -59,6 +59,10 @@ jobs:
           distribution: adopt
           java-version: 11
 
+      - name: Install node dependencies (root)
+        run: |
+          npm ci
+
       - name: Install node dependencies (core)
         working-directory: react-front-end
         run: |
@@ -66,10 +70,6 @@ jobs:
 
       - name: Install node dependencies (IntegTester)
         working-directory: autotest/IntegTester/ps
-        run: |
-          npm ci
-
-      - name: Install node dependencies (root)
         run: |
           npm ci
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -138,6 +138,10 @@ jobs:
         with:
           node-version-file: '.nvmrc'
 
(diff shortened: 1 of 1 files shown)


edalex-ian:
After making that second commit, I now wonder if ESLint should be a dependency down in the codegen package.json file. It does after all depend on it, so why not add it there.

Let me know what you think @PenghaiZhang

PenghaiZhang (approved):
LGTM!

PenghaiZhang:
> After making that second commit, I now wonder if ESLint should be a dependency down in the codegen package.json file. It does after all depend on it, so why not add it there.
> 
> Let me know what you think @PenghaiZhang

I think that's OK, but we end up having `eslint` in both `oeq-ts-rest-api` and `io-ts-gen`, which may be confusing.

edalex-ian:
> > After making that second commit, I now wonder if ESLint should be a dependency down in the codegen package.json file. It does after all depend on it, so why not add it there.
> > Let me know what you think @PenghaiZhang
> 
> I think that's OK, but we end up having `eslint` in both `oeq-ts-rest-api` and `io-ts-gen`, which may be confusing.

But it already has some of it's own ESLint setup as the intention is to still split this out to it's own module for adopters to be able to use for integrations etc.

edalex-ian:
Ah, and now I see the issue is more pervasive. I think that might be what we need to do.

PenghaiZhang:
> Ah, and now I see the issue is more pervasive. I think that might be what we need to do.

Agreed. Let's do that.

edalex-ian:
Resolved with #4679

PR #4676, opened 2023-03-02
build(oeq-ts-rest-api): add missing ESLint deps

edalex-ian:
##### Checklist

~~- [ ] tests are included~~
~~- [ ] screenshots are included showing significant UI changes~~
~~- [ ] documentation is changed or added~~

##### Description of change

oeq-ts-rest-api maintains its own ESLint setup as the intention is to split it out to its own module. However, it seemed to be missing a few dependencies and instead was simply using those from the root of the project. This only became obvious now that the codegen uses ESLint, and on fresh CI builds of just this module the codegen would fail with missing ESLint.

This is an alternate to and closes #4675 

(Note: There appears to be a lot of shuffling of the other dependencies. Seemingly they were out of order previously, but simply using `npm install` nicely sorts them for us. So there's not as many changes as it appears.)


PenghaiZhang:
Oh I thought you were going to add these deps in the package.json of io-ts-gen.

edalex-ian:
> Oh I thought you were going to add these deps in the package.json of io-ts-gen.

Yup, that was my thinking, but then I realised there was a problem with the `oeq-ts-rest-api` module setup so I fixed it there and all worked locally - when doing `npm ci` **in** `oeq-ts-rest-api`. HOWEVER as we're now seeing, if the `npm ci` is done in `react-frontend` then it runs the `oeq-ts-rest-api` `prepare` script and THEN gets the `oeq-ts-rest-api` dependencies. :disappointed: 

I guess I could set all that up in `gen-io-ts` however I'd be further duplicating things.

It'd seem the options (without digging further) are:

1. Split the `oeq-ts-rest-api` module so that we can add it as a proper dependency; or
2. Modify the GHA build as was starting to be done over in #4675 - really, probably merge these two PRs into one might solve it (and improve some things)

I'll leave for the moment while I go and work on my main task. Back soon.

image

edalex-ian:
Resolved with #4679

PR #4680, opened 2023-03-31
Hyphen name change

kapil-bunni:
- Replaced OpenEquella with Hyphen.

-  Bulk upload issue: "appending extra characters while decoding using base64 decoder"


edalex-ian:
Hi There,

Thank you for your submission. Just a few questions/observations:

1. Did you mean to also include the name change? We can't be changing the name in the main repository - it'd effect everyone;
2. You'd need to remove that left over debug println statment; and
3. You seem to have completely removed the template for PR requests - we kind of need that to validate the basics of your PR.

Oh, and is this related to an issue ticket, or some other reference?

Thanks again.

PR #4700, opened 2023-05-30
fix: ScalaCheck tests

edalex-yinzi:
##### Checklist

- [ ] tests are included

##### Description of change
In `UniversalControl`,  the exception is raised from `ExpectedConditions.stalenessOf(row)`. 

And the error message says it can't find the node with the given id.

Consider we updated the web driver and the above code will remove the `row` element from dom,
so I think it's possible the element is already been moved before we do `ExpectedConditions.stalenessOf(row)`.

Thus I add a new function to check if the element is still there before calling `ExpectedConditions.stalenessOf(row)`.

### DIFF
--- a/autotest/Tests/src/test/scala/equellatests/sections/wizard/UniversalControl.scala
+++ b/autotest/Tests/src/test/scala/equellatests/sections/wizard/UniversalControl.scala
@@ -1,13 +1,24 @@
 package equellatests.sections.wizard
 
 import equellatests.domain.TestFile
-import org.openqa.selenium.By
+import org.openqa.selenium.{By, StaleElementReferenceException, WebElement}
 import org.openqa.selenium.support.pagefactory.ByChained
 import org.openqa.selenium.support.ui.{ExpectedCondition, ExpectedConditions}
 
 import scala.util.Try
 
 class UniversalControl(val page: WizardPageTab, val ctrlNum: Int) extends WizardControl {
+
+  private def isElementPresent(element: WebElement): Boolean = {
+    try {
+      // try to get text attribute
+      element.getText()
+      // if no exception, the element should still in dom
+      true
+    } catch {
+      case e => false
+    }
+  }
 
   private def actionLinkBy(action: String) =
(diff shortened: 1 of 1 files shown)


autotest/Tests/src/test/scala/equellatests/sections/wizard/UniversalControl.scala
```diff
@@ -1,14 +1,25 @@
...
 import org.openqa.selenium.support.ui.{ExpectedCondition, ExpectedConditions}
 
 import scala.util.Try
 
 class UniversalControl(val page: WizardPageTab, val ctrlNum: Int) extends WizardControl {
 
+  private def isElementPresent(element: WebElement): Boolean = {
+    try {
+      // try to get text attribute
+      element.getText()
```
edalex-ian, line 15: Is `getText` to specific? Could you just use `isDisplayed`?

If so, can this function also just be:

```scala
  private def isElementPresent(element: WebElement): Boolean =
    Try(element.isDisplayed).getOrElse(false)
```

edalex-ian (commented):
Nice find @edalex-yinzi . Just a query on how best to check this, as well as a more idiomatic scala approach.

autotest/Tests/src/test/scala/equellatests/sections/wizard/UniversalControl.scala
edalex-yinzi: Looks fine locally.

edalex-ian (approved):
Thanks @edalex-yinzi , LGTM. Last build I saw ScalaCheck pass, so that was a win. Let's hope we get it again and then we can merge. :crossed_fingers:

edalex-ian:
This definitely fixes the ScalaCheck tests - they've not failed in all the additional runs I've done. But alas, the ScalaCheck failures were masking failures in the legacy UI test suite of the same nature:

image

image

I've re-run this many times to confirm against flakyness. But on seeing the error message, I can see it's essentially the same issue.

I'll merge this PR though as it addresses the stated issue, and I'll raise another ticket to address the newly identified issues.

PR #4712, opened 2023-06-13
fix(selenium test): check the element before stalenessOf

edalex-yinzi:
##### Checklist

- [ ] tests are included

##### Description of change

Same issue with [!4700](https://github.com/openequella/openEQUELLA/pull/4700). The element has been removed before we monitor it.

### DIFF
--- a/autotest/Tests/src/main/java/com/tle/webtests/pageobject/ExpectedConditions2.java
+++ b/autotest/Tests/src/main/java/com/tle/webtests/pageobject/ExpectedConditions2.java
@@ -46,7 +46,7 @@ public class ExpectedConditions2 {
           try {
             realElement.isDisplayed();
             return null;
-          } catch (StaleElementReferenceException se) {
+          } catch (Exception se) {
             checkingStale = false;
           }
         }
--- a/autotest/Tests/src/main/java/com/tle/webtests/pageobject/ExpectedConditions2.java
+++ b/autotest/Tests/src/main/java/com/tle/webtests/pageobject/ExpectedConditions2.java
@@ -12,6 +12,7 @@ import org.openqa.selenium.NoSuchFrameException;
 import org.openqa.selenium.SearchContext;
 import org.openqa.selenium.StaleElementReferenceException;
 import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebDriverException;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.internal.WrapsElement;
 import org.openqa.selenium.support.ui.ExpectedCondition;
@@ -34,10 +35,25 @@ public class ExpectedConditions2 {
     return ExpectedConditions.stalenessOf(element);
   }
 
+  /**
+   * Creates an {@code ExpectedCondition} which attempts to check if the provided {@code element} is
+   * displayed, however if that element has become stale, then will use the provided {@code context}
+   * and {@code locator} to find a target element and check if it is displayed/visible.
+   *
+   * @param element a previously found element which will initially be used to see if the target is
+   *     displayed.
+   * @param context a context to use when {@code element} is stale, to search for target element
+   *     with {@code locator}.
+   * @param locator a {@code By} instance used to find the target element with {@code context} for
+   *     when {@code element} is stale.
+   * @return an {@code ExpectedCondition} to check if the target element is displayed/visible.
+   */
   public static ExpectedCondition<WebElement> updateOfElementLocated(
       WebElement element, final SearchContext context, final By locator) {
     final WebElement realElement = unwrappedElement(element);
     return new ExpectedCondition<WebElement>() {
+      // Provides the means to toggle between first checking if 'element' is stale, and then if it
+      // becomes stale later using 'context' and 'locator'.
       private boolean checkingStale = true;
 
       @Override
@@ -46,7 +62,9 @@ public class ExpectedConditions2 {
           try {
             realElement.isDisplayed();
             return null;
-          } catch (Exception se) {
+          } catch (WebDriverException se) {
+            // Once the element has become stale, then the provided element/realElement is no longer
+            // used and every check will instead now rely on context and locator down below.
             checkingStale = false;
           }
         }
@@ -165,9 +183,7 @@ public class ExpectedConditions2 {
         try {
           realElement.isDisplayed();
           return false;
-        } catch (StaleElementReferenceException ser) {
-          return true;
-        } catch (NoSuchElementException e) {
+        } catch (WebDriverException ser) {
           return true;
         }
       }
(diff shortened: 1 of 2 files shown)


edalex-yinzi:
Will see how pipeline goes.

edalex-ian:
I have rerun the build, I saw three things across the failed functional_testing jobs:

1. The installer failing on configInstall - as per the other ticket you're working on;
2. Flaky jobs failing; and
3. Some tests still failing with no node for id - so the issue you're attempting to fix here.

Let's see how the re-run goes, but you may need to address some more test issues I feel.

edalex-ian:
Heya @edalex-yinzi ,

So yeah, it looks like under New UI, things are failing due to flaky tests. However, legacy UI seems to have repeat failures with a number related to the no node for ID issue. Please have a closer look at `functional_testing (false, 12)`.

image

(No point re-running, will re-run when you make some more fixes.)

autotest/Tests/src/main/java/com/tle/webtests/pageobject/ExpectedConditions2.java
```diff
@@ -46,7 +46,7 @@ public WebElement apply(WebDriver driver) {
           try {
             realElement.isDisplayed();
             return null;
-          } catch (StaleElementReferenceException se) {
+          } catch (Exception se) {
```
edalex-yinzi, line 49: I can see almost all errors are from this line, the current exception we have is `unknown error`.

edalex-ian: Sorry, I'm not sure I follow you here:

> the current exception we have is `unknown error`.

`unknown error` is obviously not a specific `Exception`. I'm guessing not, but did you find what the `Exception` was? (As catching just all `Exception`s is probably not what we really want here.)

edalex-ian: So in trying to confirm the logic of this code, I've gone through and added documentation. Could you please do the same? 

Here's the doco for you:

```java
  /**
   * Creates an {@code ExpectedCondition} which attempts to check if the provided {@code element} is
   * displayed, however if that element has become stale, then will use the provided {@code context}
   * and {@code locator} to find a target element and check if it is displayed/visible.
   * 
   * @param element a previously found element which will initially be used to see if the target is
   *                displayed.
   * @param context a context to use when {@code element} is stale, to search for target element
   *                with {@code locator}.
   * @param locator a {@code By} instance used to find the target element with {@code context} for
   *                when {@code element} is stale.
   * @return an {@code ExpectedCondition} to check if the target element is displayed/visible.
   */
  public static ExpectedCondition<WebElement> updateOfElementLocated(
      WebElement element, final SearchContext context, final By locator) {
    final WebElement realElement = unwrappedElement(element);
    return new ExpectedCondition<WebElement>() {
      // Provides the means to toggle between first checking if 'element' is stale, and then if it
      // becomes stale later using 'context' and 'locator'.
      private boolean checkingStale = true;

      @Override
      public WebElement apply(WebDriver driver) {
        if (checkingStale) {
          try {
            realElement.isDisplayed();
            return null;
          } catch (StaleElementReferenceException se) {
            // Once the element has become stale, then the provided element/realElement is no longer
            // used and every check will instead now rely on context and locator down below.
            checkingStale = false;
          }
        }
        return elementIfVisible(context.findElement(locator));
      }

      @Override
      public String toString() {
        return "updateOfElementLocated " + context + ":" + locator;
      }
    };
  }
```

edalex-ian:
I've rerun the old UI functional_testing again, as there's only one of those that I'm familiar with being flaky (`BulkMetadataEditTest`). So here's the result to compare against after that run.

image

autotest/Tests/src/main/java/com/tle/webtests/pageobject/ExpectedConditions2.java
edalex-yinzi: The exception type is WebDriveException:
image

edalex-yinzi: `StaleElementReferenceException` is a subclass of `WebDriverException`. How about we use `WebDriverException` instead? @edalex-ian

edalex-yinzi: I found this https://bugs.chromium.org/p/chromedriver/issues/detail?id=4440. 

and a fix has been merged recently https://chromium-review.googlesource.com/c/chromium/src/+/4508952 (return `N`).

If the error exception is `NoSuchElement` then it makes more sense we could simply add this exception rather than using `WebDriverException`.

edalex-ian: Well, `WebDriverException` would be an improvement - that's for sure. With just `Exception` you'll also pick-up `RuntimeException`s which include things like `NullPointerException`s.

But really, I think what this is highlight is that our Selenium is so out of date now that it doesn't know how to handle some of the new standard errors coming from the Web Driver. :disappointed: 

In the end, I wonder if a better approach to all this may be to address #3522. (As catching a generic `WebDriverException` when - as per the doco I provided - this function is really only meant to be checking for a stale element is not the same by far.)

edalex-ian: As discussed, I missed your above message about the Chrome Driver bug, so ignore the bit about Selenium - we can continue to save that for another day.

edalex-ian:
Looking good @edalex-yinzi , only failed on known flaky tests:

**functional_testing (true, 12)**

- `com.tle.webtests.test.contribute.controls.attachments.MultipleAttachmentControlTest`

**functional_testing (false, 12)**

- `com.tle.webtests.test.searching.manage.BulkMetadataEditTest`

----

I'll rerun the failed tests just to see how it goes.

edalex-yinzi:
equellaConnector failed again. So I updated the code.

edalex-ian:
You're pretty close @edalex-yinzi ,

However I keep seeing the EquellaConnectorTest fail, and as can be seen in [this job](https://github.com/openequella/openEQUELLA/actions/runs/5274709880/jobs/9562030716) there's a good number of mentions of the 'no node with id' errors.

Ah, and while just about to click 'comment' I see you've done a commit. Let's see what that has/does.

edalex-yinzi:
Amazing. All passed.

edalex-ian:
Great work dude! Looks like now just need to sort the `configInstall` and we'll have stable builds again. Well done.

edalex-ian (approved):

PR #4775, opened 2023-08-11
fix: com.puppycrawl.tools checkstyle 10.x

edalex-yinzi:
##### Checklist

- [ ] tests are included

##### Description of change

Since checkstyle fixed bug `FinalClassCheck should report private classes without constructor` https://checkstyle.sourceforge.io/releasenotes.html

I fix all `final` errors, and I think it might make sense to decrease the threshold number to 520, because the current error number is 519.

1. Add the `final` keyword for those private classes according to the checkstyle report.
2. Also add `final` keyword for some public classes according to the checkstyle report.
3. Update threshold number.

Fix #4762.

### DIFF
--- a/Platform/Infrastructure/com.tle.jpfclasspath/src/com/tle/jpfclasspath/model/JPFPluginModelManager.java
+++ b/Platform/Infrastructure/com.tle.jpfclasspath/src/com/tle/jpfclasspath/model/JPFPluginModelManager.java
@@ -22,7 +22,7 @@ import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.JavaCore;
 import org.eclipse.jdt.core.JavaModelException;
 
-public class JPFPluginModelManager implements IModelProviderListener {
+public final class JPFPluginModelManager implements IModelProviderListener {
   private static JPFPluginModelManager instance;
   private WorkspaceJarModelManager jarManager;
   private WorkspacePluginModelManager projectManager;
--- a/Installer/src/com/dytech/installer/controls/GPassword.java
+++ b/Installer/src/com/dytech/installer/controls/GPassword.java
@@ -93,7 +93,7 @@ public class GPassword extends GuiControl {
     }
   }
 
-  private static class ConfirmPassword {
+  private static final class ConfirmPassword {
     private boolean confirmed = false;
 
     private ConfirmPassword(final JPasswordField fieldToConfirm) {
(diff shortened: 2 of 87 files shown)


edalex-ian:
Thanks @edalex-yinzi . Two things:

1. Would be good to include in your description what you were address etc. Not just rely on people to go and find then read the comments over in #4762 . Could you please update the description.
2. Why have you also marked some public classes as final?

edalex-yinzi:
> Thanks @edalex-yinzi . Two things:
> 
> 1. Would be good to include in your description what you were address etc. Not just rely on people to go and find then read the comments over in [chore(deps): update dependency com.puppycrawl.tools:checkstyle to v10.12.2 #4762](https://github.com/openequella/openEQUELLA/pull/4762) . Could you please update the description.
> 2. Why have you also marked some public classes as final?

No problem.

I fixed some public classes because it is mentioned in the checkstyle report. So I just fix them as well.

edalex-ian:
Thanks @edalex-yinzi , any chance you could tell me what was the failure for the public classes?

edalex-yinzi:
> Thanks @edalex-yinzi , any chance you could tell me what was the failure for the public classes?
The error message is the same :
image

Platform/Infrastructure/com.tle.jpfclasspath/src/com/tle/jpfclasspath/model/JPFPluginModelManager.java
```diff
@@ -22,7 +22,7 @@
 import org.eclipse.jdt.core.JavaCore;
 import org.eclipse.jdt.core.JavaModelException;
 
-public class JPFPluginModelManager implements IModelProviderListener {
+public final class JPFPluginModelManager implements IModelProviderListener {
```
edalex-yinzi, line 25: @edalex-ian 
The error message is the same :
image

edalex-yinzi:
Flaky tests failed.

edalex-ian:
Aha! Details of check can be found in the JavaDoc at https://checkstyle.org/apidocs/com/puppycrawl/tools/checkstyle/checks/design/FinalClassCheck.html

Key here is that the public classes it's marking as 'final' are utility classes. And by nature, you wouldn't normally extend those and nor should you. So this makes sense.

edalex-ian (approved):
Looks good thanks @edalex-yinzi , and thanks for also dropping the `thresholdNumber`. I still hope that one day we'll have time to get that to zero. One day....

Will merge after the retry.

PR #4994, opened 2024-01-08
build: tidy-up memory settings and GHA workflow

edalex-ian:
##### Checklist

- [ ] tests are included

##### Description of change

Originally a PR to investigate the issues with OOM issues with GHA. However, again the magically resovled (just like they magically appeared) after yet another update to the GHA runner image.

However in the process I did find a few oddities with how `.sbtopts` and `.jvmopts` were setup in the project. And the result was that the memory configuration was nothing like what we thought. So I've consolidated that down into a single root `.jvmopts`.

Further, I've updated the GHA workflow not only to reflect that, but also to use the java distro we actually now use - Temurin.

Now we seem to have a significant issue with `AdvancedSearchPageTest` in that it is almost always failing - after 10 attempts I give up. I say _almost_ though, as it is passing 1 in 20 times or so. But this is not unique to these changes, as I've also validated it over on the build for #4993 . So let's merge this in, and then we need to have a closer look at `AdvancedSearchPageTest`.

(This PR should definitely be squashed when merging.)


edalex-ian:
Oh, I've just noticed the Java version is out of wack on `develop`. So this PR will fail. :sob:

edalex-ian:
Mind you, the default collector is G1GC according to this: https://docs.oracle.com/en/java/javase/21/gctuning/available-collectors.html. I think it might have been in 11 too.

edalex-ian:
As expected, still getting memory issues, let's increase the heap.

edalex-ian:
(Seems like pipelines don't run for Draft PRs. :thinking: )

Anyway, will do one last thing: Capture memory logs.

I'll have a look at them, attach here for propsperity. And then let it be - and await (hopefully) this week's update to the GitHub VM images.

edalex-ian:
Okay, so it looks as though the new GitHub runner image from last week has again magically fixed the memory issues. But now the build has failed on some flaky tests.

I'll run again, and see if we have any success. But at least it didn't fail with OOM errors.

If I can get a build to pass - with the tweaks I've made here - then I think we should keep these changes.

edalex-ian:
I had one build fail with a heap memory issue (whereas when there's issues with GitHub runners that's typically a full blown OOM), and I can't get the new UI tests to pass (specifically `AdvancedSearchPageTest`). I'm think that with the new configuration for memory (which I think previously was always ending up at 4GiB) perhaps it's a bit too frugal.

To address this, the last commit increases from 2GiB max heap to 3. Let's see what that does.

edalex-ian:
Hmmm, `io.github.openequella.search.AdvancedSearchPageTest` just keeps failing - and nothing else:

```
FAILED: userSelector
        Search by using the User selector control
org.openqa.selenium.NoSuchElementException: no such element: Unable to locate element: {"method":"xpath","selector":".//span[text()='AutoTest']"}
```

:confused: 

I'll retry two more times (max).

PenghaiZhang (approved):
LGTM!

PR #5084, opened 2024-04-17
fix: OAuthClientEditorTest fails with latest Chrome

phungnguyenptn:
##### Checklist

- [x] tests are included

##### Description of change
- I've added a condition on OAuthClientEditorTest to check WebElement get attribute by id not empty based on the below condition of this test.
- The issue by when we send the empty string into the find element of Selenium Webdriver it will show an error when we send the empty string. But in the latest Chrome maybe Selenium Webdriver doesn't support the latest version.

### DIFF
--- a/autotest/Tests/src/main/java/com/tle/webtests/pageobject/oauth/OAuthClientEditorPage.java
+++ b/autotest/Tests/src/main/java/com/tle/webtests/pageobject/oauth/OAuthClientEditorPage.java
@@ -6,7 +6,7 @@ import com.tle.webtests.pageobject.WaitingPageObject;
 import com.tle.webtests.pageobject.generic.component.EquellaSelect;
 import com.tle.webtests.pageobject.generic.component.SelectUserDialog;
 import java.util.List;
-import java.util.Objects;
+import org.apache.commons.lang3.StringUtils;
 import org.openqa.selenium.By;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.support.FindBy;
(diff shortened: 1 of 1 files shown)


autotest/Tests/src/main/java/com/tle/webtests/pageobject/oauth/OAuthClientEditorPage.java
```diff
@@ -157,7 +158,8 @@ private boolean isControlInvalid(WebElement we, String selector) {
         driver.findElements(By.xpath("//div[@class='control ctrlinvalid']"));
 
     for (WebElement control : invalidControls) {
-      if (isPresent(control, By.id(we.getAttribute("id")))) {
+      if (!Objects.equals(we.getAttribute("id"), "")
```
edalex-ian, line 161: hmmm, I'm confused why you've used `Objects.equals` here. The idiom would be `"".equals(...)`...

But anyway, in oEQ we do have apache StringUtils so you'd just want to do:

```java
if(StringUtils.isNotEmpty(...) && isPresent(...)) { ... }
```

But that said, if you're going to pull out this value, please just do it once:

```java
final String elementId = we.getAttribute("id")

for (WebElement control : invalidControls) {
  if(StringUtils.isNonEmpty(elementId) && isPresent(control, elementId)) {
    ...
  }
  ...
}
```

edalex-ian (changes requested):
Thanks for the PR @phungnguyenptn 

Please tweak the code, and then I'll approve a pipeline run to see how it goes.

autotest/Tests/src/main/java/com/tle/webtests/pageobject/oauth/OAuthClientEditorPage.java
phungnguyenptn: Hi @edalex-ian . I've updated it based on your suggestion. Please help me check it

edalex-ian (approved):
Thanks @phungnguyenptn . Let's see how it runs. :crossed_fingers:

edalex-ian:
Woohoo! Solid run first time.

Considering it used to have a 100% fail rate, I'd say this is good to go! Thanks @phungnguyenptn :tada:

